PR #463, opened 2013-02-14
Klarna update

martynas-wm:
Digest calculation is done in ASCII encoding to get the correct signature.


straightlight:
Klarna is no longer supported on the master branch on anyhow.

PR #578, opened 2013-03-27
Fixes exploit

Bloafer:
This change fixes the exploit found and described here: http://www.exploit-db.com/exploits/24877/


jjPlusPlus:
Dang this guy is a huge douche.

Issue #177, opened 2012-10-10
Multi-column table Indexes causing slowdowns


chrisatomix:
Several of the tables in the database have Primary Keys on two or more columns, eg:

``` sql
DROP TABLE IF EXISTS `oc_product_attribute`;
CREATE TABLE `oc_product_attribute` (
  `product_id` int(11) NOT NULL,
  `attribute_id` int(11) NOT NULL,
  `language_id` int(11) NOT NULL,
  `text` text COLLATE utf8_bin NOT NULL,
  PRIMARY KEY (`product_id`,`attribute_id`,`language_id`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_bin;
```

AND:

``` sql
 PRIMARY KEY (`product_id`,`category_id`)
```

The problem is that any WHERE statements or JOINs that try to use the category_id will not be indexed (this behavior may be exclusive to MySQL, but I'm not sure).
So it behaves as if category_id has no index at all, which means whenever you're filtering on this it causes the SQL to slow down. My commit in #46 didn't take this into account, which is why a store with 15,000 products in multiple categories would run so painfully slow (because there is only an index on product_id, not on category_id).

Any column which is to be used for joining or searching should have an index. For example, any table containing a product_id field should have it indexed. Possibly also language_id, category_id and store_id. On a demo store the difference will be negligible, but on a large store it would cause major issues.

Pros: Improvement in performance on front-end and searches, especially for large stores
Cons: Miniscule performance drop when inserting/updating products, as indexes will need to be updated.

Example updated code:

``` sql
DROP TABLE IF EXISTS `oc_product_to_category`;
CREATE TABLE `oc_product_to_category` (
  `product_id` int(11) NOT NULL,
  `category_id` int(11) NOT NULL,
  PRIMARY KEY (`product_id`,`category_id`),
  KEY `idx_category_id` (`category_id`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_bin;
```

Happy to commit this myself if nobody else has time.

PR #986, opened 2013-11-07
Dramatically boost load performance of products and categories on large stores

chrisatomix:
Dramatically boost load performance of products and categories on large stores (10,000+ products, 500+ nested categories, product filters and options) due to multi-column indexes, as explained in detail here:

http://bloke.org/php/opencart-is-slow-with-many-categories/
https://github.com/opencart/opencart/issues/177

I have tested these changes across 13 different OpenCart installations, there was a substantial reduction in page load time (even with caching enabled). Furthermore, MySQL load was reduced by an average 15-20% (sometimes more).

As explained in the links above, when a MySQL index has multiple columns, only the first column will perform as an indexed column. The remaining columns will appear un-indexed, unless the first column is also filtered during the query.
For example, if the table has an index on product_id,category_id:

Selecting all categories "WHERE product_id = 123" IS indexed, because product_id comes first in the index.

Selecting all categories "WHERE product_id = 123 AND category_id = 7" IS indexed, because both indexed columns are specified.

Selecting all products "WHERE category_id = 7" is NOT indexed, because the first column in the index is 'product_id' and that has not been specified, which makes the second index useless.

This is all backed up and explained very clearly by the MySQL documentation:
http://dev.mysql.com/doc/refman/5.0/en/multiple-column-indexes.html

> MySQL cannot use the index to perform lookups if the columns do not form a leftmost prefix of the index. Suppose that you have the SELECT statements shown here:
> 
> ``` SQL
> SELECT * FROM tbl_name WHERE col1=val1;
> SELECT * FROM tbl_name WHERE col1=val1 AND col2=val2;
> SELECT * FROM tbl_name WHERE col2=val2;
> SELECT * FROM tbl_name WHERE col2=val2 AND col3=val3;
> ```
> 
> If an index exists on (col1, col2, col3), only the first two queries use the index. The third and fourth queries do involve indexed columns, but (col2) and (col2, col3) are not leftmost prefixes of (col1, col2, col3).

There is obviously going to be a minor performance hit when creating new products (as the indexes are updated), however this is so negligible as to be almost unworthy of mention (bulk-adding 10,000 products may take a few seconds longer).

### DIFF
--- a/upload/install/opencart.sql
+++ b/upload/install/opencart.sql
@@ -318,7 +318,8 @@ CREATE TABLE `oc_category` (
   `status` tinyint(1) NOT NULL,
   `date_added` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
   `date_modified` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
-  PRIMARY KEY (`category_id`)
+  PRIMARY KEY (`category_id`),
+  KEY `parent_id` (`parent_id`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;
 
 --
@@ -435,7 +436,8 @@ CREATE TABLE `oc_category_path` (
   `category_id` int(11) NOT NULL,
   `path_id` int(11) NOT NULL,
   `level` int(11) NOT NULL,
-  PRIMARY KEY (`category_id`,`path_id`)
+  PRIMARY KEY (`category_id`,`path_id`),
+  KEY `path_id` (`path_id`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;
 
 --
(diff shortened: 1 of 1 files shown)


dkgawande:
Does that mean the latest version of opencart (2.1.0.1) released on 6th October 2015 includes these indexes and optimisations? I get 404 while accessing the turbo script (https://github.com/chrisatomix/opencart-turbo). Thank you.

shkodskihk:
I see the repository 'opencart-turbo' moved here: https://github.com/lilalaunesau/opencart-turbo
Does the latest version of opencart these optimizations - I do not know.

dkgawande:
@shkodskihk Thank you!

shkodskihk:
I note in addition that the optimization of the database at the source code made in the Russian build of Opencart called 'ocStore'. References:
http://sourceforge.net/projects/ocstore/files/
http://myopencart.com/
(translated to english link: https://translate.google.com/translate?hl=ru&sl=auto&tl=en&u=http%3A%2F%2Fmyopencart.com%2F)

PR #1113, opened 2014-01-01
check for customer negative store credit balance before confirm order 

kabatza:
A quick and effective fix to prevent customers using the multiple browser trick and double-spent store credit resulting in complete orders but negative store balance.

Modification simply checks the customer balance on callback (confirm or update) and if the store-credit is negative it sets the order to pending and adds a red notice in the order history.


danielkerr:
i have already added a fix for this. the order status will be assigned a fraud order status. you can aset that status to be pending review.

PR #1619, opened 2014-07-09
do not show errors suppressed with @

cornernote:


cornernote:
Did this change ever make it into a release?  The 1.5.6.4 doesn't have it.

PR #1855, opened 2014-09-23
Making menu check permissions

edirpedro:
Driving menu programmatically to let it check permissions before show on the list.

**Benefits:** Check access permissions and shows only pages user can see.

**Doubt:** Extension developers can add new pages using vQmod at this momment, but first idea was to use something like WordPress does, it uses an option "order" to let developer choose the position when registering a page, but I don't know if this is really necessary here as you can map using vQmod, the position where your menu needs to appear on the list easely. The list now is fully organized and is ordered by added first.

**Error:** The page `payment/pp_express/search` won't work because there is no way to set access permission to that. Two options here, create another controller to open this page or modify the way rule works to include the action `search` presented on this route.

**Changed:** In the Front-End, menu had IDs for each ones like `#catalog`, I don't think this is a good idea and changed to something that represents the object `#menu-catalog`.

Adding a Menu or Sub-Menu: (Name, Title, Parent Menu, Icon)
~~~
$this->addMenu('catalog',  $this->language->get('text_catalog'), null, 'fa-tags');
~~~

Adding a Page (Route, Title, Parent Menu, URL Params)
~~~
$this->addPage('catalog/category', $this->language->get('text_category'), 'catalog');
~~~


mahomedalid:
I think problem with this approach is that it mix html/view logic on controller, better if at the end controller just push menu array to the view and on the view the .tpl decides how to render menu.

PR #2312, opened 2014-11-21
Fix Javascript for Cart when SEO URls enabled

uksitebuilder:
Now that you have added SEO Keywords for Cart, Checkout, etc

Remove item button no longer refreshes the Cart page (as per the issue I
raised the other day)

This commit fixes the refresh problem

### DIFF
--- a/upload/catalog/view/javascript/common.js
+++ b/upload/catalog/view/javascript/common.js
@@ -182,6 +182,8 @@ var cart = {
 
 				if (getURLVar('route') == 'checkout/cart' || getURLVar('route') == 'checkout/checkout') {
 					location = 'index.php?route=checkout/cart';
+				} else if (window.location.pathname == '/checkout' || window.location.pathname == '/cart') {
+					location = 'cart';
 				} else {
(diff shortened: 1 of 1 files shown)


dennisbuijsgmailcom:
Hi 
I have the same issue as uksitebuilder, also no refresh of the cart after removing the product.
The solution provided is not working for everybody, and not for me, so I think Danielkerr is right about this.

If have also check the javascript code and found a good solution for everybody, a change that is already used in opencart code.

In the javascript file common.js I see at other places the code: 
if (json['redirect']) {
   location = json['redirect'];
}

This code can be used, I think, it can be replaced, the code at line  218: 
if (getURLVar('route') == 'checkout/cart' || getURLVar('route') == 'checkout/checkout') {
   location = 'index.php?route=checkout/cart';
} 
Replace into:  
if (json['redirect']) {
   location = json['redirect'];
}   

In the file catalog/controller/checkout/cart.php add the code before line 470:   
$json['redirect'] = $this->url->link('checkout/cart');
$json['total'] = sprintf($this->language->.......
 
Best regards
Dennis

PR #2545, opened 2015-01-07
Infinite loop

burakcakirel:
In modification file, when using position="after" attribute, if you add same line in search attribute, it creates infinite loop. 

It seems "line_id" variable has been forgotten to decrease in "after" case. I committed with the solution.

Here is the test xml code:
```xml
<modification>
	<name>Test XML</name>
	<code>test_xml</code>
	<version>0.0.1</version>
	<author>burakcakirel</author>

	<file path="admin/controller/catalog/product.php">
		<operation>
			<search><![CDATA[$this->response->setOutput(json_encode($json));]]></search>
			<add position="after"><![CDATA[
			#test line
			$this->response->setOutput(json_encode($json));
			]]></add>
		</operation>
	</file>
</modification>

### DIFF
--- a/upload/admin/controller/extension/modification.php
+++ b/upload/admin/controller/extension/modification.php
@@ -292,7 +292,11 @@ class ControllerExtensionModification extends Controller {
 														$line_id += count($new_lines);
 														break;
 													case 'after':
-														array_splice($lines, ($line_id + 1) + $offset, 0, explode("\n", $add));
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
Well spotted! You are right, the $line_id needs indeed to be incremented:

$line_id += count($new_lines);

after an 'after' operation.

danielkerr:
yepo your right

PR #2553, opened 2015-01-11
Fixed the order of custom fields

mpestov:
The order of custom fields is working incorrect if exist hidden blocks. For example the block with customer group. I have not used  jQuery selector ":visible" because it does not return the objects


danielkerr:
if you had read the help on the sort order field you would see you can do -1

mpestov:
I have read the help. The sort order is different on the pages http://demo.opencart.com/index.php?route=checkout/checkout and http://demo.opencart.com/index.php?route=account/edit. I am sorry i have not described my issue fully.

PR #2558, opened 2015-01-13
Add the missing DIR_UPLOAD variable in cli_install.php

johnnyzheng99:
The DIR_UPLOAD variable is missing if it's installed via cli. When it's installed through the web wizard it's ok because it's defined in upload/install/controller/step_3.php. Just need to add the same line to the cli_install.php


jamesallsup:
Code is already in latest master.

PR #2566, opened 2015-01-16
display prices redundant condition

pine3ree:
if 
$this->customer->isLogged() === true 
then prices are automatically visible, 
no need to check $this->config->get('config_customer_price')


danielkerr:
i add this manually. thx!

PR #2583, opened 2015-01-21
Update summernote.js

saifwebberoo:
Padding issue fixed

### DIFF
--- a/upload/admin/view/javascript/summernote/summernote.js
+++ b/upload/admin/view/javascript/summernote/summernote.js
@@ -4004,7 +4004,7 @@
                    ) +
                    '<form class="note-modal-form">' +
                      '<div class="modal-body">' +
-                       '<div class="row-fluid">' + body + '</div>' +
+                       '<div class="row-fluid"><div class="col-xs-12">' + body + '</div></div>' +
                      '</div>' +
                      (footer ?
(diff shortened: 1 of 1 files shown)


gob33:
And also update from 0.5.8 to the latest (0.6.0 ?).
I need the new function "insertText (node, text)" to do at least like the good old CKEditor for moving extensions to OC2.x. This is a necessity.

gob33:
In fact, Summernote 0.5.8 exposes only 2 methods to outside: .code() and .destroy()
Summernote API is accessible inside plugins only, not from outside, it is in developpement, far from the point where it could be usable.
I suggest hardly dropping Summernote and revert back to CKEditor 4 + bootstrap style.
An OC extension already propose that for 2.x.

jamesallsup:
Summernote will probably be updated to the latest version shortly.

gob33:
Even in 0.6.2, Summernote is far away of CKEditor 4.
I will certainly drop certains functionalities (tags insertion in editor on double-click) in my extension(s) when passing from OC1.5.x to OC2.x because there is no API equivalent in Summernote.
This is a regression for merchants but I cant do it with your choice of Summernote.

danielkerr:
latest summernote has just been added.

PR #2594, opened 2015-01-22
Better file and folder names display in filemanager

yemaw:
If filename is more than 20 characters, only show first 20 characters, dot dot dot and extension.
Previously, it is adding a space every 14 characters in filenames for new line. It's a bit confusing for user as they wrongly put space in file and folder names.

### DIFF
--- a/upload/admin/controller/common/filemanager.php
+++ b/upload/admin/controller/common/filemanager.php
@@ -50,7 +50,9 @@ class ControllerCommonFileManager extends Controller {
 		$images = array_splice($images, ($page - 1) * 16, 16);
 
 		foreach ($images as $image) {
-			$name = str_split(basename($image), 14);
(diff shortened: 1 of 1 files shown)


eddieco:
I am setting up Opencart 2.0.1.1 and I must admit that with the space every 14characters the result is very strange as it can occur in the middle of the extension, e.g. "productimage.j pg" ! I have tried this fix and it seems good - thanks! 

Looking at the code of the controller filemanager.php I can't understand why the filename basename($image) is split into an array of elements length 14 characters with:


`53      $name = str_split(basename($image), 14);`

which is then rebuilt into the string with spaces at every 14 chars with:

`68      'name'  => implode(' ', $name),`
`83      'name'  => implode(' ', $name),`

Was it instead meant to be `'name'  => implode('<br\>', $name),` or similar? If so the space would at least become a line break but I think yemaw's solution is more elegant.

PR #2606, opened 2015-01-24
Disable maintenance check for API requests

JAY6390:
This should be part of the core as the payment one is. There are times when administrators need to update orders while in maintenance mode, and can't if the API is blocked by maintenance mode


dtjones:
Kl. I was just about to create a new issue. merge please

jamesallsup:
Definitely, cheers Jay.

PR #2612, opened 2015-01-27
why array_merge?

pine3ree:
...with a one-key/value array? why not setting the value for the key directly?

### DIFF
--- a/upload/admin/model/catalog/product.php
+++ b/upload/admin/model/catalog/product.php
@@ -292,20 +292,20 @@ class ModelCatalogProduct extends Model {
 			$data['keyword'] = '';
 			$data['status'] = '0';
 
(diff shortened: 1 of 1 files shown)


pine3ree:
sorry again, why using array_merge when setting one key at a time?

```php
$arr = array(
    'a' => 1,
    'b' => 2,
);

$arr1 = $arr;
$arr2 = $arr;

// 3 operations: create a new 1-key array, merge it with old one and assign back to old variable
$arr1 = array_merge($arr1, array('c' => 3)); 

// 1 opearation
$arr2['c'] = 3;

var_dump($arr1 === $arr2); // true
```

array_merge makes sense if you build a new array with all the remaing keys and then merge to $data:

```php
$data = array_merge($data, array(
    'product_attribute' => $this->getProductAttributes($product_id),
    'product_description' => $this->getProductDescriptions($product_id),
    // etc
));

PR #2613, opened 2015-01-27
Update cart.php

skybluesofa:
Add minimum quantity to the 'add to cart' functionality.


danielkerr:
it already does this

skybluesofa:
The minimum works on the product page. However, from the category page, it wasn't. In my development, only one item was being placed in the cart from the category page, even though the minimum was set for 40. I added this code and was able to successfully add the correct (40) number of items to the cart. 

Can you review the process of adding an item with a minimum from the category page?

PR #2617, opened 2015-01-28
Added a testing mail checkbox for configuration testing

danelsan:
This update is usefull because an administrator can be configure an email
and verify if email setting is good
This send an email for testing.


jamesallsup:
It's a good idea but the code needs to be tidied up a bit inline with the standards guide. Due to recent changes it is also not possible to merge this request automatically - please submit a new PR forked from the latest master.

https://github.com/opencart/opencart/wiki/Coding-standards

Thanks.

PR #2640, opened 2015-02-02
unused keys and joins

pine3ree:
weight_class and length_class don't seem to be ever used in /catalog, so maybe it's worth saving a couple of nested queries per product query.


danielkerr:
it is used

pine3ree:
where? 
i searched for 'weight_class', "weight_class"  in the whole application related to product result row(s).

Even so, a join with 2 more tables slow down queries and you can use weight / length service classes to give you the value just using weight_class_id and length_class_id in the result row and add the corresponding keys.

PR #2641, opened 2015-02-02
fix wrong orders count column in admin:report/customer_order

rb2:
Admin - Reports - Orders.

Column with orders count calculated wrong numbers (equal to products count).


danielkerr:
not sure if this is resolved or not but the sql has changed a lot since reported

PR #2650, opened 2015-02-06
OC 2.0.1.1 - Added a second placeholder for the store name instead of 'us'

gortsilas:
"This commit has broken then the account registration link.
In account.php sprintf expects 2 placeholders where there is only one in the new string"

It is commit https://github.com/gortsilas/opencart/commit/0073f03e6a8882973df776163e2afc11a8486844

It actually moved the error from affiliates to account since only one of the strings had 2 placeholders but their respective sprintf functions are identical.

$data['text_message'] = sprintf($this->language->get('text_message'), $this->config->get('config_name'), $this->url->link('information/contact'));

The second placeholder must fix the issue.

### DIFF
--- a/upload/catalog/language/english/account/success.php
+++ b/upload/catalog/language/english/account/success.php
@@ -3,7 +3,7 @@
 $_['heading_title'] = 'Your Account Has Been Created!';
 
 // Text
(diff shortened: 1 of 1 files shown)


jamesallsup:
Please submit complete pull requests were all code has been changed so it works. We test by pulling your PR into our build system automatically and this failed the check due to the missing params in the controller it requires. If you can resubmit complete we will test again.

Thanks.

PR #2664, opened 2015-02-11
Missing php closing tag in catalog/language/english/default.php

mathsolutions:
language file is missing php closing tag


osworx:
Why should it have one??

mathsolutions:
ok, it's optional

PR #2676, opened 2015-02-13
redundant product_attribute queries

pine3ree:
- when adding product there is no need to remove any attribute belonging to it (as there are none yet)
- when updating the product the attributes related to the product are already cleared with
```php
$this->db->query("DELETE FROM " . DB_PREFIX . "product_attribute WHERE product_id = '" . (int)$product_id . "'");
```
no need to issue more specific deletion queries inside the loop
```php
$this->db->query("DELETE FROM " . DB_PREFIX . "product_attribute WHERE product_id = '" . (int)$product_id . "' AND attribute_id = '" . (int)$product_attribute['attribute_id'] . "'");
``


jamesallsup:
Thanks.

PR #2677, opened 2015-02-14
config values are unserialized already

pine3ree:
btw, i am not sure that config key ("config_token_ignore") is even used anywhere.


jamesallsup:
That section of code has just been removed anyway as it isn't used now.

Thanks.

PR #2682, opened 2015-02-16
Add HTTPS support during installation

redochka:
When installing opencart in secure mode, page layout was broken because of browser security regarding non-secure resources (scripts, links, fonts...).

Thanks


jamesallsup:
Well spotted, I've never installed using SSL but since its "advised" to use only SSL on stores now I can see that being required much more.

Thanks.

PR #2691, opened 2015-02-20
2 modules - Most Reviewed, Top Rated

OC2PS-2:
This patch adds 2 modules to OpenCart. The Top Rated module lists the
products that have received the best ratings. The Most Reviewed module
lists the products that have received the highest number of ratings.


OC2PS-2:
The two modules are generally quite useful, and for 1.5.6.4 I distributed them in the OC Extension Directory, but feel that they are interesting to enough people that they deserve inclusion in core.
http://www.opencart.com/index.php?route=extension/extension/info&extension_id=8368
http://www.opencart.com/index.php?route=extension/extension/info&extension_id=8369

PR #2697, opened 2015-02-21
You save

OC2PS-2:
Savings are important for customers. Showing only the old price and new is not enough as it requires customers to make mental calculations as to how much they are saving.
Amazon understands this - that's why it shows the savings separately http://www.amazon.com/dp/B00CX5P8FC/ref=br_imp_ara-1/179-1035819-2800748?pf_rd_m=ATVPDKIKX0DER&pf_rd_s=desktop-hero-1&pf_rd_r=0YFM6R2R090AJAR40480&pf_rd_t=36701&pf_rd_p=2039920302&pf_rd_i=desktop

This patch adds savings ($ and %) information below the price on the product page if a valid special price exists for the product.

This is useful for many OpenCart installs as evidenced from my popular extension
http://www.opencart.com/index.php?route=extension/extension/info&extension_id=8319

### DIFF
--- a/upload/admin/controller/module/mostreviewed.php
+++ b/upload/admin/controller/module/mostreviewed.php
@@ -1,0 +1,167 @@
+<?php
+class ControllerModuleMostReviewed extends Controller {
+	private $error = array();
+
+	public function index() {
+		$this->load->language('module/mostreviewed');
+
+		$this->document->setTitle($this->language->get('heading_title'));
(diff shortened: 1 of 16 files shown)


OC2PS-2:
Aw, crap! I sent this pull request from a separate branch, and yet it includes both the you save commit and the 2 modules commit.

PR #2701, opened 2015-02-22
Fix default value of error_address

CZechBoY:
It's array, not string.


jamesallsup:
Thanks, accepted this as it is good for consistency. I think there is other areas needed to be checked for similar issues. It shouldn't cause any issues though.

PR #2702, opened 2015-02-22
Admin Tweak - Reverse Default Order of Reviews

OC2PS-2:
OpenCart lists reviews ordered by date added in ascending order (oldest first) by default. This patch reverses the default order (newest first) so that the reviews awaiting moderation are shown first.
This tweak in the Admin Panel can save us a click or two, and in day-to-day operations, these clicks add up.
This is one of the functions performed by one of the two functions performed by my extension http://www.opencart.com/index.php?route=extension/extension/info&extension_id=7841&filter_username=OC2PS&page=2

### DIFF
--- a/upload/admin/controller/catalog/review.php
+++ b/upload/admin/controller/catalog/review.php
@@ -182,19 +182,20 @@ class ControllerCatalogReview extends Controller {
 		} else {
 			$filter_date_added = null;
 		}
-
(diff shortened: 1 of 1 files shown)


jamesallsup:
I think its a matter of preference on the default sort order but I couldn't agree more from experience that this way is [probably] the more common and popular option.

Thanks.

PR #2705, opened 2015-02-22
Add product photos and links in order mail

OC2PS-2:
Our customers like it if the order confirmation email includes product photos - this way, they can quickly scan the email and know the products are correct, without having to read the mail.
Product photos in order emails are good for us as well: Due to local regulations we have a high turnover of inventory and packaging staff and giving printouts of order emails containing images helps get them off the ground quicker with less training required.
By clicking the link, they can easily check product description later.
This patch adds product photo and link to the order confirmation email.
Essentially it does the job of my popular extension (9000 downloads) http://www.opencart.com/index.php?route=extension/extension/info&extension_id=7102

### DIFF
--- a/upload/catalog/model/checkout/order.php
+++ b/upload/catalog/model/checkout/order.php
@@ -413,6 +413,8 @@ class ModelCheckoutOrder extends Model {
 				$subject = sprintf($language->get('text_new_subject'), $order_info['store_name'], $order_id);
 
 				// HTML Mail
+				$this->load->model('tool/image');
+				
 				$data = array();
 
 				$data['title'] = sprintf($language->get('text_new_subject'), html_entity_decode($order_info['store_name'], ENT_QUOTES, 'UTF-8'), $order_id);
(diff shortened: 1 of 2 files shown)


EveryAuction:
I removed the 
(OC2PS) Product pictures in order email VqMod.
I replaced the v.1.5.6.4 default order.php with your order.php

In Checkout Step 6:
http://www.bigmax.ch/shop/index.php?route=checkout/checkout
I have the following error:
---
Fatal error: Call to a member function trigger() on a non-object in /home/jacob/www/bigmax/shop/catalog/model/checkout/order.php on line 4
---
>> Line 4: $this->event->trigger('pre.order.add', $data); <<
---
did I miss something?
best regards,
Ernie (OC-Handle: IP-CAM)

OC2PS-2:
@EveryAuction This patch is for OpenCart 2.0.1.1 and will not work with older versions. 
For OpenCart 1.5.6.4, please continue using my extension.

EveryAuction:
Thank you, this Version Info is something I miss here very much, one never know's...
Ernie

OC2PS-2:
@danielkerr @jamesallsup I understand you have closed this merge request without comment.

Would greatly appreciate your feedback. Was the code buggy, inefficient or inelegant? Do you feel most OpenCart users won't benefit from this patch? etc. Thanks again.

garudacrafts:
I very much support the addition of product images and links to the order emails. I do this for all my projects. I also add them to the bottom of the checkout page ("confirm.tpl").

One thing this pull request is missing is a check on the existence of the product image:
```php
if ($prodquery['image'] && file_exists(DIR_IMAGE . $prodquery['image'])) {
	$thumb = $this->model_tool_image->resize($prodquery['image'], 60, 60);
} else {
	$thumb = $this->model_tool_image->resize('no_image.jpg', 60, 60);
}
```

and then simply `'image'    => $thumb,` instead of `'thumb'    => $this->model_tool_image->resize($image, 60, 60),`.

Personally, I'd go even further and recommend adding a column to the order_product table for image, because it's possible that a product will be deleted in the future. In addition, the additional SQL hit to the product table isn't required; image will be included in $order_product_query and so substitute `$prodquery['image']` with `$product['image']` above.

OC2PS-2:
@garudacrafts Good point Derek. I wanted to start small and not make database changes while managing backwards compatibility. But your solution is better. I am happy to submit another patch if @danielkerr @jamesallsup agree.

PR #2706, opened 2015-02-22
Fixes line endings in several files (CRLF are now LF)

schrodervictor:
Even having `.gitattributes` to normalize the line endings, it doesn't affect the old files that are still with CRLF instead of LF.

That becomes really annoying when the developer has autocrlf set to true or input, because `git status` will always say that those files have changes (of course, because git itself changed them...).

And, after all, it's good for the project to have this issue finally solved.

After this commit, grep'ing for ^M shows no files other them binaries. So I believe all of them has been fixed.

The command (Bash 4+) that generated the output was:

```bash
$ for f in $(git status --short | cut -b4-); do sed -i 's/^M$//' $f; done
```
Where ^M is not literal, but the actual byte (Crtl+V, Ctrl+M).

No content was changed, only the line endings.


jamesallsup:
Amazing, thanks!

schrodervictor:
My pleasure! If I can help with something else, just let me know.

PR #2723, opened 2015-02-28
Category menus with class

OC2PS-2:
OpenCart has already started started doing part of my extension
http://www.opencart.com/index.php?route=extension/extension/info&extension_id=6500
assigning classes to left category menu.

This patch offers store owners more granular control by adding category ID to the category module.

Further, this patch goes beyond the aforementioned extension by adding classes to the top/dropdown menu.


danielkerr:
not a chance. you dont need iDs just css selector.

PR #2726, opened 2015-03-01
Enable Facebook Social Plugins

OC2PS-2:
Daniel,

I know that your initial instinct will be to reject this patch immediately. However, I request you to please read on, let it simmer for a bit and then decide.

There is a wide range of popular modules that insert Facebook plugins into OpenCart shops.

Trouble is, all of them need to include the SDK in the template, especially if they go down the HTML5 or FBML route.

One issue is that just to add a module, there needs to be a modification.

Second issue is that if multiple such extensions are used, each adding the Facebook Javascript, it can lead to conflicts. Serious conflicts.

This patch adds the Facebook SDK Javascript into the Default theme supplied with core. Thus, the extensions now only need to worry about the module, and dont need to do any modifications or worry about whether FB JS has already been loaded.

Would be useful for a large proportion of OpenCart shops.

### DIFF
--- a/upload/catalog/view/theme/default/template/common/header.tpl
+++ b/upload/catalog/view/theme/default/template/common/header.tpl
@@ -39,6 +39,14 @@
 <?php echo $google_analytics; ?>
 </head>
 <body class="<?php echo $class; ?>">
+<div id="fb-root"></div>
+<script>(function(d, s, id) {
+  var js, fjs = d.getElementsByTagName(s)[0];
+  if (d.getElementById(id)) return;
+  js = d.createElement(s); js.id = id;
+  js.src = "//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.0";
+  fjs.parentNode.insertBefore(js, fjs);
+}(document, 'script', 'facebook-jssdk'));</script>
 <nav id="top">
   <div class="container">
     <?php echo $currency; ?>


nuke123:
Not every store wants facebook. Why can't you use ```ignoreif```?

OC2PS-2:
@nuke123 While not every store wants Facebook, most do. 

Adding the Facebook JS should not have any noticeable negative impact on the stores that don't use Facebook since any performance hit comes from the social plugin itself (and will only be felt if you add a Facebook module) not really from this JS.

Using this eliminates the need for modules to have to include modifications.

To be fair, OpenCart 1.5.x included AddThis, which is much heavier and was definitely not valued by all store owners.

What I could do, if needed, is add a config option to include this JS if needed.

manakill:
I'm with OC2PS, it's time OC included this.

Besides, If a store now days DID NOT have social buttons linked to a legit account, it would be a huge red flag that something fishy is going on. Why would a store not use social share buttons? If it was a scam site worried about word getting out..

osworx:
Social media is not all, or how much do you become back from them??

manakill:
It is a key indicator if a company is thriving and serving customers good.
Not having share buttons means it's not getting shared. Period. Happy customers naturally share items and help to promote to friends and family. Facts are facts - word of mouth is the most powerful form of marketing.
Our products have been shared millions of times and generate a nice traffic stream. If you're not using social media then you're just not trying to succeed at all.

sorcix:
Please do not merge this. It adds an external dependency to every opencart shop using the default theme even when not using facebook plugins at all. Adding the SDK should be the responsibility of those adding the facebook plugin, not the other way around.

schrodervictor:
I fully, fully agree with @sorcix. It would be a huge mistake to have this external dependency as a default setting.

manakill:
@sorcix @schrodervictor a simple Enable/Disable option fixes everything. I wasn't implying an option that couldn't be toggled.

manakill:
Upon further inspection this is a ridiculous pull.
For this to be considered it would have to include a toggle to Disable and Enable it in Admin

OC2PS-2:
As mentioned [above https://github.com/opencart/opencart/pull/2726#issuecomment-76658696], if @danielkerr agrees in principle, then I am happy to initiate a separate pull request, including a Settings toggle so that everyone gets what they want. It's easy enough to do. But I do want a green light before i waste any time on it.

sorcix:
> For this to be considered it would have to include a toggle to Disable and Enable it in Admin

Even that would be ridiculous. Remember that if you do this for facebook, you'll have to do it for any popular social network, including the ones that may be popular in the future. The only valid place for things like this is inside the actual plugin. You're suggesting a **very bad architecture** that does not scale at all.

What about other plugins that share something? Is your suggestion to create additional bloat in core for all of them?

OC2PS-2:
> if you do this for facebook, you'll have to do it for others...
No, we don't. That's a terrible argument. A very developer-centric argument that completely ignores the business reality on the ground. In business, we tend to use the Pareto principle. We aren't obliged to and should not treat all social networks the same, because they are not. We do know that Facebook has 1.4 billion users http://files.shareholder.com/downloads/AMDA-NJ5DZ/3907746207x0x805520/2D74EDCA-E02A-420B-A262-BC096264BB93/FB_Q414EarningsSlides20150128.pdf - that is 20% of the world's population (including men, women, children, newborns, old, inform, remote tribes, poor and people without internet access). It has a set of tools that are useful for businesses. When another social network crosses the 1B threshold, we can think about including that too.
> You're suggesting a very bad architecture that does not scale at all.
As discussed above, scaling to 1000 social networks is not the idea. We are not trying to build a perfect social machine here. Just trying to be practical about real business needs of real businesses.
> The only valid place for things like this is inside the actual plugin. 
> What about other plugins that share something? Is your suggestion to create additional bloat in core for all of them?
I don't think you understand what this patch does. This does NOT add any Facebook functionality to OpenCart at all. What it does do is add the code to include Facebook SDK.
Opposing that is a bit like saying we shouldn't include jQuery because 1.5% of people have JavaScript disabled.
Especially after adding the toggle, the job of the patch would be to simply add the Facebook SDK which can then be utilized by any modules that want to use it. Solves 2 problems:
1. Extension developers don't need to create modifications just to add FB modules.
2. If a store owner adds 2 Facebook modules (or more) he doesn't get into a holy mess because their modifications conflict or try to load the JS multiple times etc

sorcix:
> I don't think you understand what this patch does. This does NOT add any Facebook functionality to OpenCart at all. What it does do is add the code to include Facebook SDK.

And that is exactly what shouldn't happen. Loading the SDK takes bandwidth and processing power, thus time. Even if you don't use the actual SDK your browser has to load and interpret everything. Especially on mobile this kills performance. Why would you want to burden *every* user with something that may not be used at all?

Maybe you should ask the facebook module developers to fix their modules? Maybe get them to work together to solve this. If using two modules creates a mess, then it's up to the module developers to fix it, as it is clearly broken.

jamesallsup:
Sorry everyone but this is getting closed off. If you want to vote/discuss adding features please use:

https://opencart.uservoice.com/forums/52387-general 

That is what it is there for. Or feel free to create a topic on the forums for it too, if the majority of people get behind an idea then we will certainly look into it.

Github is for PR's on bug fixes or project changes that have been pre agreed with Daniel or myself.

Thanks.

OC2PS-2:
@jamesallsup Thanks for the guidance. How/where do I discuss ideas with you? I tried PMing you in OC forums earlier, but not sure if you regularly check those.

PR #2731, opened 2015-03-03
add slashes to escape javascript

veksen:
JS broke when language file had quotes. Fixed what I've found, potentially more across the system.


veksen:
closing to add more cases, and branching it

PR #2732, opened 2015-03-03
Fix event bug in manufacturer.php

haydenw:
Including $manufacturer_id as argument for event trigger for post.admin.manufacturer.edit as per the documentation.


jamesallsup:
Well spotted, thanks!

PR #2746, opened 2015-03-09
Update securetrading_pp.php

osworx:
Following vars are not needed, because already in main file:
$_['text_enabled'] = 'Enabled';
$_['text_disabled'] = 'Disabled';
$_['text_yes'] = 'Yes';
$_['text_no'] = 'No';

Only additional work for translators.


jamesallsup:
Good find, thanks

PR #2771, opened 2015-03-16
fix trailing whitespace

veksen:
Removing more trailing whitespace.


veksen:
As offered by @jamesallsup access to Jenkins build reviews would be very useful for future cooperation.

veksen:
Are we following PSR-1/PSR-2? You'll see that I'm removing spaces after the last php closing tag, but currently not adding a new blank line. However, PSR-2 states "All PHP files MUST end with a single blank line.".

veksen:
closing in favor of #2837

PR #2772, opened 2015-03-16
fix trailing whitespace

veksen:
Removing trailing whitespace from admin templates


veksen:
closing in favor of #2772

PR #2784, opened 2015-03-18
refactored address_data building

pine3ree:
maybe also worth returning an empty array instead of false in getAddress()


pine3ree:
on second thought i think it will be better to refactor using one query with left joins to the country and the zone tables and having the builder just takig care of casting the result values.

PR #2799, opened 2015-03-19
Replace MyIsam database with InnoDB

tmccaff:
Last one was wrong. This is the new fix. InnoDB is better for
transactions. Therefore, makes sense to change to InnoDB

### DIFF
--- a/upload/install/opencart.sql
+++ b/upload/install/opencart.sql
@@ -26,7 +26,7 @@ CREATE TABLE IF NOT EXISTS `oc_address` (
   `custom_field` text NOT NULL,
   PRIMARY KEY (`address_id`),
   KEY `customer_id` (`customer_id`)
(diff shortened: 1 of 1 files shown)


veksen:
Won't InnoDB screw up with up with full text search (searching in product description)?

Martynas-P:
Where does OpenCart use Full Text search? As far as I am aware, OpenCart uses LIKE %...% to search in descriptions and titles.

tmccaff:
Yes need to check for this. I have to look more deeply into this. However, this needs to be done asap. It's absolutely critical to move from MyIsam

mariosangiorgio:
I see this patch has not been merged into master and I totally agree with @tmccaff when he says that it's critical to move from MyIsam.

I am currently choosing an what e-commerce software I should use and I would really like to have InnoDB support. What is the current state of InnoDB support in OpenCart?

PR #2801, opened 2015-03-20
integer casting for id fields

pine3ree:


jamesallsup:
You don't need to do this. Thanks anyway.

pine3ree:
...but in my opinion is more consistent with the type of data the id fields represent.
the function name too: isLogged() suggests a boolean result, while it returns a string (empty or not) which eventually is evaluated truthy or falsy in a if condition.

Issue #2811, opened 2015-03-24
Html.tpl - Html Module. 


aWeirdo:
Hi,
First off, I'm sorry, 
but i have no clue how all that pull request and stuff works.. 
Minor bug, but annoying to look at.

When adding custom HTML code to your webstore via the HTML module, 
if you don't need a heading title for your code, the module still adds a empty ```<h2></h2> ```, which leaves very a ugly space.

upload/catalog/view/theme/default/template/module/html.tpl

Current code:
```
<div>
<h2><?php echo $heading_title; ?></h2>
<?php echo $html; ?></div>
```

Fix:
```
<div>
<?php if(!$heading_title){} else { ?>
<h2><?php echo $heading_title; ?></h2>
<?php } echo $html; ?></div>
```

123 and that ugly space is gone.

Keep up the good work!, just started using opencart, loving it  :) :+1:

beifloyd:
I don't think they accept code from here.
pull request created for you here https://github.com/opencart/opencart/pull/2813

PR #2813, opened 2015-03-24
html.tpl

beifloyd:
When adding custom HTML code to your webstore via the HTML module,
if you don't need a heading title for your code, the module still adds a empty <h2></h2>, which leaves a very ugly space.

credit  to  aWeirdo  issue #2811


veksen:
Why a negative and not the following?

    <?php if($heading_title) { ?>
    <h2><?php echo $heading_title; ?></h2>
    <?php } ?>

aWeirdo:
Thank you, @beifloyd.

@veksen,
It was just something fast, so I didn't give it much thought, both way works, 
but of course your's should be a tiny bit faster :-)

PR #2822, opened 2015-03-27
utility methods

pine3ree:
I've been using these for a very long time, always forgot to PR.
Caching results to static vars to avoid array access and already made comparisons

### DIFF
--- a/upload/system/library/request.php
+++ b/upload/system/library/request.php
@@ -5,6 +5,11 @@ class Request {
 	public $cookie = array();
 	public $files = array();
 	public $server = array();
+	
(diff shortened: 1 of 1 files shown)


VeeeneX:
Why static?

pine3ree:
@VeeeneX 
static vars are used as local cache. Once they are set their cached values are used, since i check against null. request and response objects in OC are shared services, so static function vars can be used in that way, otherwise i could have used private members for the same purpose, but i did not want to pollute the class.

danielkerr:
not a chance.

PR #2831, opened 2015-03-30
Fixed FontAwesome issue in Firefox

chipironcin:
Some firefox configurations blocking cross-domain requests for FontAwsome CDN.
This is a fix to allow those cross-domain requests ONLY for fonts files.
(First seen in Mozilla Firefox for Ubuntu 36.0 in Ubuntu Trusty 14.04)
More info:
http://wpvkp.com/font-awesome-doesnt-display-in-firefox-maxcdn/
http://stackoverflow.com/questions/10636611/how-does-access-control-allow-origin-header-work

### DIFF
--- a/upload/.htaccess.txt
+++ b/upload/.htaccess.txt
@@ -11,8 +11,8 @@ Options -Indexes
 
 # Prevent Direct Access to files
 <FilesMatch "(?i)((\.tpl|\.ini|\.log|(?<!robots)\.txt))">
- Order deny,allow
- Deny from all
+  Order deny,allow
+  Deny from all
 </FilesMatch>
 
 # SEO URL Settings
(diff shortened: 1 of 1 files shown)


jamesallsup:
This is a topic that keeps appearing about cross origin.

Using www and non-www is wrong. Pick a domain and use it, adding in hacks to allow things to work is not how things should be done, but if it is what the user wants then they can add in their own solution.

This is down to host to set up the server properly, to only response to a domain it is configured for. Sadly many solutions like WHM/cPanel allow this by default (responding to a request with or without the www and serving the same doc root). This leads people to believe it is acceptable behaviour, it is not!!

The solution is to pick your preference (www.example.com or example.com) and use htaccess to 301 redirect any traffic landing on the wrong url.

PR #2856, opened 2015-04-03
remove unneeded array_merge calls

pine3ree:
all 'product_*' keys are non existent in $data since it's a product table only query result row. I removed those calls more then 2 years ago in my code. array_merge is used for merging values for existing keys.

No point in initializing $data to an array and then overwriting it in the next line. btw, $query->row is always an array (at least an empty array)

(Techically, there is no need here to left join to product_description in the query either other then to ensure the product has a valid description)


pine3ree:
@mrg123 
ooopppsss...right, i think i committed to remote without saving the latest file changes.

pine3ree:
@mrg123 
Yes, I think it happened just that as my local repo is ok. I'll fix online.

pine3ree:
@mrg123 
....mmmm...weird...I believe it wasn't like this when it was merged....anyone would have spot those before merging...anyway I noticed it's been already fixed....

PR #2862, opened 2015-04-04
remove unneeded OR and ORDER

pine3ree:
ugly query:
```sql
SELECT * FROM setting WHERE store_id = '0' OR store_id = '0' ORDER BY store_id ASC
```
these look better:
```sql
SELECT * FROM setting WHERE store_id = '0'
SELECT * FROM setting WHERE store_id = '0' OR store_id = '1' ORDER BY store_id ASC
```

### DIFF
--- a/upload/index.php
+++ b/upload/index.php
@@ -45,7 +45,14 @@ if ($store_query->num_rows) {
 }
 
 // Settings
-$query = $db->query("SELECT * FROM `" . DB_PREFIX . "setting` WHERE store_id = '0' OR store_id = '" . (int)$config->get('config_store_id') . "' ORDER BY store_id ASC");
+$settings_sql = "SELECT * FROM `" . DB_PREFIX . "setting` WHERE store_id = '0'";
+
+$config_store_id = (int)$config->get('config_store_id');
+if ($config_store_id > 0) {
+    $settings_sql .= " OR store_id = '{$config_store_id}' ORDER BY store_id ASC";
+}
+
+$query = $db->query($settings_sql);
 
 foreach ($query->rows as $result) {
 	if (!$result['serialized']) {


jamesallsup:
Please use the http://opencart.uservoice.com site for suggestions with code. Github is for bugs only.

osworx:
@jamesallsup Do not undersatnd why this should be a recommandation?
The query is stupid as it calls `WHERE store_id = '0' OR store_id = '0' ORDER BY store_id`
That is all what he said ..

danielkerr:
there is no improvement here.

pine3ree:
@osworx 
:-)
I think it was about 1 year ago i had to debug queries with a few server administrators and i was feeling ashamed of the unused-unneeded query clauses appearing on top of every request. Seemed kind of weird so i removed at once. Last month I was browsing the oc repo and noticed it was still there in the master-branch, so i opened a quick pr.

Not considering the "visual" sql improvement and considering that 'how it is' is obviously not a real waiste of resources, still mysql performs an unnecessary order-by + unnecessary condition.

@jamesallsup 
i believe that http://opencart.uservoice.com is more the place for "bigger" design suggestions/requests.
In every other open source repo I contribute to PRs are used for much more than simply bug-fixes: coding style improvements, code readability, optimization, refactoring....so, sorry, i wasn't aware that prs in opencart are only meant to be bug-fixes.

Any way, no damage done here: PRs are mere suggestions from other ppl's point of view, repo owners are free to include/disregard/modify...and this was just a little candy pr, i didn't even remember about....

cheers to every1

osworx:
Well not only db queries are 'wasting' and 'unneeded'.
Same goes for the language strings and a few other (also functions) - always repeating the same - code over the whole cart.

pine3ree:
@osworx 
You could add a t() _t() __() method in the view taking care of calling the language translation service. You would get rid of a lot of code in your controllers. It was one of the 1st thing i did. But i agree with you, i had to refactor pretty much everything to remove code duplication and optimize data retrieving function calls with cached values..

osworx:
@pine3ree either that, but if the view should be the same code like before, 1 line in the controllers will do the same job.

pine3ree:
@osworx 
At 1st i did it like that. a translate method in the controller to set many language keys at once (i think i have some leftover code with this in the admin area). But then I dediced that it would be cleaner to have a plugin-like behaviour for that (something like ZF2) without using static methods. I simply inject the language service (using the registry locator in the constructor) into the view and short-named method acts as a proxy to Language::get(). If you prefer setter injection, you'll need to dapat the proxy method to check for the language service presence 1st.

cheers

osworx:
@pine3ree would you mind to contact me direct?
Would like to discuss that a bit ..
thx!

pine3ree:
Hi @osworx ,
since it's not a discussion about a security issue I think it's better to keep the discussion public so that other people can read and share their ideas.
:-) cheers!!

osworx:
no problem, but not here.

opencarthelp:
@pine3ree Strongly coupling the view to the framework/application has a lot of downsides.  It's much easier to break views and a lot harder to control the data going to them.  It's much better to move this sort of thing to a model or service the controller calls.

pine3ree:
@opencarthelp 
I completely agree about having de-coupled components. But opencart is not a decoupled framework. Everything is glued by the registry+loader that basically act together as a service locator. The dependencies are most of the times retrieved by the locator, not injected.

(Also if you really want a decoupled framework (but i doubt this is the goal of this project) you should write code to generally recognized interfaces and handle setter injection of dependencies/ initializers etc etc. I am not 100% sure, but i don't think there is 1 interface in the project and I also believe (not sure though) that opencart project leader wants to keep the project simple. I agree about keeping it simple. Nevertheless I think (well...more like "I know", as I already did a lot of refactoring) that there are many parts of the project that could be refactored and simplified. The translation texts being one of them as pointed out by @osworx) 

Anyway, opencart doesn't have a real separate View service. The loader locate a template, extract the data and capture the output buffer. In my implementation I have a (very basic php-based, no rendering strategies) View engine component, which can be enriched with plugins. Most View implementations do that (see Twig and i18n for example, or zf2-view + zf2-i18n + relative view plugin, even smarty handles translations if i remember it well). In my case plugins are called by a handful of proxy methods. 

I don't think it's a good idea to have to assign translatable texts in every controller action (and also have tell to load the translation file too). 

MVC controller-actions (Controller class public methods in opencart) should be as coincise as possible and manage input data, pass data to proper business logic object/services for data manipulation, and build/return responses.

Also translations messages are not real data, but can be confused with data in the template code.  For example
```php
<?php echo $column_left; ?>
<?php echo $column_price; ?>
```

it's not clear that $column_left is a view section and $column_price is a i18n text.

```php
<?php echo $column_left; ?>
<?php echo $this->_t('column_price', 'Price'); ?>
```
The second case, which is similar to what i implemented for the view component, is clearer to me....and in the relative controller method i didn't have to load anything or to write:

```php
$this->language->load(....);
//...
$data['column_price'] = $this->language->get('column_price');
$data['column_image'] = $this->language->get('column_image');
$data['text_this'] = $this->language->get('text_this');
//...
```
those parts distract me from what is the purpose of the action.

It is funny that a simple pr about a sql query brought us to talk about something totally different. Means the project is alive.

cheers guys.
I am finishing my psr-7 http-message implementation so i'm not sure i will be able to reply quickly to any new comment.

opencarthelp:
There's so much wrong there I don't even know where to begin.  You don't even seem to understand how OpenCart works internally.

pine3ree:
@opencarthelp 
Your reply is not helpful, and kind of arrogant and presumptuous. It's not the first time I have encountered this kind of attitude in this project. 

This kind of attitude won't help the project improve.

You are basically saying: "You don't understand anything" and "I will not bother to tell you why".

Let's assume it is true. In that case a constructive comment would be answering point by point with an explaination of what you think is wrong and why. I am willing and hoping to learn everyday from everyone else. Are You? Or do You believe you already know too much?

Let me use your same attitude towards you:
You said "I don't even know where to begin". About that, I guess you are probably right?
:-) this wasn't an helpful comment either, right?

You said:
 "It's much better to move this sort of thing to a model or service the controller calls."

Opencart already does that: the controller uses the language service (it's a shared service) that does provide translated text (after taking care of loading the right file if exists) and then assign translated text to data later extracted in template files. It does that message by message. 

This is a constructive reply. I didn't just say your comment didn't make any sense, I also explained why. :-) (attitude, attitude)

cheers, mr Guru!!!

opencarthelp:
Sorry if that came off as a personal attack.  I'll just skip over all the little stuff and just hit the two biggest issues:

1. Whether a framework is highly abstracted or tightly coupled has nothing to do with MVC as a design pattern.
2. Templates have never used the registry.  It was incidentally accessible in 1.x but never used and is not available at all in 2.x.

pine3ree:
@opencarthelp 
It's ok. :-) I'm glad you understand my ironic answer (It was a test)

Let me comment on your points:
1. I didn't say that. You said that the view shouldn't be strongly-coupled with the framework. But in fact in opencart (2+) it is, or better there isn't e real separate View service (with adapters/renderers as the db class for instance). All controllers (weather they feed the response body or returns strings as in module-controllers or page-section-controller or whatever you want to call them) load views.

they do that like this:

```php
$string = $this->load->view('path/identifier', $data);
// which basically is a shortcut for
$loader = $this->registry->get('load'); // fetch the loader service
$string = $loader->view('path/identifier', $data);
```

it's the loader responsible of locating the template file, start the ob, extract the data and finally capturing the buffer.
So if you consider that the view in oc is played by the loader, it's more than strongly-coupled to the framework.
There is no possibility of switching to another template engine. To achieve that we would need more abstraction and renderer-components that interface as proxies to external engines.

2. Yes, i know that. I think i didn't explain myself clearly or maybe I forgot to add some context. 
If i well remember, not 100% sure right now, in oc 1.5 the context in template files was the calling controller (i.e. $this resolved to the calling controller). So you could call services in there.

About the translation service in the view i was referring about how i did it, or could be done.

To do that i believe it's better to have a separate view component, into wich inject the needed services.
I created long time ago a simple View component, very very simple....it does what the loader does plus a few more vew related functions like partial rendering and so on. Plus it has access to the registry (so i can use the same shared services or plugins any other registry_aware class has). I could have simply skipped the registry and use setter injection for adding plugin-like features. But since i consider opencart a full-stack framework i didn't bother and i intentionally kept it simple as i also don't really need to be able to use any external component for translation. Otherwise i would need to write a more abstract translator service in wich the current Language service would be an adapter/driver. But again, i wanted it simple and just working. 

I've worked and still work with more than a few framework and view renderers. In each of them translations are not handled in controller. All use short and semantic proxy methods that call a translation service. Wether you fetch this service using a locator or use an internal reference, you still have have a dependency on the service itself and that need to be injected in some way. My way was, even if not the best option, to simply inject the service locator on instantiation, like it's done in models and the controllers. But in any framework i worked with, the translated text are not feeded into the template by the controller and i like this. Some frameworks use static class methods or functions that call the singleton application which in turn retrieves the translation service. At the end it's all the same: the view accesses the translation service. (Only i haven't liked statics in a long time)
  
I don't see any other solution: either you use the translator in the controller e put data ito the view or you use it in the view and get translated data in there. I prefer having shorter controller actions code and having something that hint me about translated text in templates.

If you have other ideas please share. As i said, i have much fun to learn from other people (just probably being bored by myself :-))

cheers

PR #2867, opened 2015-04-05
performance improvement lazy loading rates

pine3ree:
Not every route needs tax calculations (ajax generated zone/country listing, product review creation, login, logout)
By moving tax rates loading out of the constructors we can avoid (3) unused database query for those routes.

also, refactored setters....they are basically one method with a parameter ($based)

### DIFF
--- a/upload/system/library/tax.php
+++ b/upload/system/library/tax.php
@@ -2,6 +2,8 @@
 final class Tax {
 	private $tax_rates = array();
 
+	private $address = array();
+
 	private $initialized = false;
 
 	/**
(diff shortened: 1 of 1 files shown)


pine3ree:
also solving this issue for digital product tax:
[http://oc.pine3ree.com/oc/202/index.php?route=test/tax](http://oc.pine3ree.com/oc/202/index.php?route=test/tax)
calling $this->tax->set(Payment|Shipping|Store)Address() should always give consistent results if called on the same request, i.e it should force tax recalculation, not just adding new rates to the library/tax class "tax_rate" internal cache

danielkerr:
sorry but your tax changes keep causing errors appearing on the tax class also i don't think you realize all the possibilities for taxes based on different counties. what i have done is correct. it does not need updating.

pine3ree:
They work on different request (which are all of the cases in opencart code so far);

In the same request tax rates just continue to add up ....if set(Payment|Shipping)Address are called more than once (by more than once i mean for every different method)

This example calls setPaymentAddress a few times on the same request to explain that, 100% oc original code no changes:

http://oc.pine3ree.com/oc/202/index.php?route=test/tax

If set(Payment|Shipping)Address are called singularly only once per request, there is not going to be any issue ever.

I solved that in my code (I run a largely modified version of opencart) using a state variable in the tax class. Tax rates are retrieved from the database only at the first tax calculation request (no need to fetch them when generating a captcha, but that's another story....i use a simpler captcha front controller, instead of index.php and standard routes). In my case any double calls to any of those methods with a different country_id-zone_id combination force a reload of the tax rates.

I found that the same thing was happening to oc-2.x as well so i just wanted to report it.

That's the only thing i wanted to say. :-)

(about the error in the tax code, i explained what happened...i believed it happened 3 times to me, one of that the resulting code was really freaky... :-) )

cheers

PR #2917, opened 2015-04-15
Update skrill.php

stockmaster:
Caps in the 'text_skrill' was preventing the link & image to show. I do suggest as well to change the moneybrokers.png image to skrill.png


stockmaster:
skrill

That image to be used

PR #2924, opened 2015-04-16
Feature no zones

JTS-FIN:
This feature hides "Zone" -selection input from the forms in the following places:
   - My Account -> Address book -> Add or edit address
   - Checkout -> Register OR Guest checkout (for both payment and delivery address)
   - Register

When either one of these are met:
   - When in Admin area System-> Localisation -> Countries -> insert_country_name -> "Show/use zones" is set to false (new feature as well)
   - When country has no zones

Adds a new column to country -table, opencart.sql changed accordingly.

I created this feature is since at least in Finland we don't really use the Zones (lääni) for anything (same goes to at least Sweden that I know of), the basic address is composed of Post code + City (or postal area) + street address, so asking for Zone is kind of stupid here, and gives a little unprofessional image of the seller here in Finland. 

More about the subject I found from the forums:
http://forum.opencart.com/viewtopic.php?f=20&t=25808

This is my first pull request ever, so please let me know if there is something more I need to do, or some commits weren't on point. I did my best testing the features, but I didn't ran the automated tests that were provided.

ps. By the way I was rather pleasantly surprised about the high quality of the code/structure, keep up the good work :)

### DIFF
--- a/upload/admin/controller/localisation/country.php
+++ b/upload/admin/controller/localisation/country.php
@@ -276,6 +276,7 @@ class ControllerLocalisationCountry extends Controller {
 		$data['entry_iso_code_3'] = $this->language->get('entry_iso_code_3');
 		$data['entry_address_format'] = $this->language->get('entry_address_format');
 		$data['entry_postcode_required'] = $this->language->get('entry_postcode_required');
+		$data['entry_use_zones'] = $this->language->get('entry_use_zones');
 		$data['entry_status'] = $this->language->get('entry_status');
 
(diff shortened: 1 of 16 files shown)


danielkerr:
this will cause so many issues!

there is also asystemin place called all zones set for geo zones.

JTS-FIN:
I see that modifying database is an issue. However, if I make another pull request where Zones-selection is hidden with JS when country has no zones defined, would that be ok? 

Currently if country has no zones, "None" with zoneid 0  is already selected for the country, so just hiding the selection would clean the UI, and would not cause issues (as far as I see).

PR #2940, opened 2015-04-23
Fixed Early Reset Button Glitch

CoverUp:
On checkout process, when you go from step 2 to step 4 (or step 3 to step 4) and you get the shipping module information, the main ajax complete() function execute before the success() function in the child ajax call, thus, when you press the "next" button on checkout page, you will see a "Loading" message as expected, but the button will go to the reset state ("Next" text) before all the shipping method data appear. Most of the time this is not a problem if you are dealing with fixed data for shipping (instant data), but if your shipping module call a website in order to get real onlines prices, this may take almost 5 seconds, so the user will experiment a 5 seconds delay between the moment in which the button goes to reset state ("next" text) and the update of the shipping data. (Most of the time, the user may think that the "Next" button doesn't work and hence  press it again)
The proposed change adds an extra wait, that rely on the child ajax call in order to update the button to the reset state, so the user now will see the "Loading" state of the button all the time until shipping data is displayed. If any error occur, the system will work as usual.
This is a small fix, but it is big from the user experience point of view.

### DIFF
--- a/upload/catalog/view/theme/default/template/checkout/checkout.tpl
+++ b/upload/catalog/view/theme/default/template/checkout/checkout.tpl
@@ -428,6 +428,7 @@ $(document).delegate('#button-payment-address', 'click', function() {
 
 // Shipping Address
 $(document).delegate('#button-shipping-address', 'click', function() {
+	var wait_more = 0;
     $.ajax({
         url: 'index.php?route=checkout/shipping_address/save',
         type: 'post',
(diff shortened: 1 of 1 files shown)


danielkerr:
if there is an issue here or not the fix you have presented is badly coded.

CoverUp:
Actually it fix the problem, what kind of solution do you propose instead?

PR #2944, opened 2015-04-23
feature for show image in option for radio buttons and checkboxes

bezdna:


bezdna:
opencart option
if the image selected, it shows, if not, it displays only text

CoverUp:
any chance to disable this without deleting the image?
maybe add a new option kind is better, so instead select checkbox option you select a image checkbox option (or what ever name of the new option kind), oh but wait a minute, we already got the $option['type'] == 'image' kind of option.



But we can't  add this kind of option to a product ¿? in fact there are many option that can't be included on product as option ¿?



I think we got to start fixing this.. allow each option to be included on the product, otherwise what are the option for

bezdna:
I believe that if it is possible to specify an image for the option, then it should be displayed not only in the admin area, otherwise I do not see any reason the presence of this option in the admin panel

arturbonnett:
>I believe that if it is possible to specify an image for the option, then it should be displayed not only in the admin area, otherwise I do not see any reason the presence of this option in the admin panel
Totally agree

arturbonnett:
>I believe that if it is possible to specify an image for the option, then it should be displayed not only in the admin area, otherwise I do not see any reason the presence of this option in the admin panel

Totally agree

bezdna:
Another reason, to look at the number of downloads )
http://www.opencart.com/index.php?route=extension/extension/info&extension_id=20970&filter_search=image%20option%20change&filter_download_id=40&page=2

danielkerr:
added this manually. just added image to display under checkbox or radio

PR #2958, opened 2015-04-27
added fraudlabs pro module

chrislim2888:
Added a new fraudlabs pro anti-fraud module


CoverUp:
why this must be in the core and not be an extension?

stockmaster:
@CoverUp $

PR #2981, opened 2015-05-05
Add missing class to product wrapper 

marcos-paiva:
This commit inserts the class 'product-layout' for the following files:

* bestseller.tpl
* latest.tpl
* special.tpl

This prevents such problems when these modules are placed in the left and right columns.

### DIFF
--- a/upload/catalog/view/javascript/common.js
+++ b/upload/catalog/view/javascript/common.js
@@ -27,8 +27,10 @@ $(document).ready(function() {
 	cols1 = $('#column-right, #column-left').length;
 	
 	if (cols1 == 2) {
+		$('#content .product-layout').removeClass('col-lg-3 col-md-3');
 		$('#content .product-layout:nth-child(2n+2)').after('<div class="clearfix visible-md visible-sm"></div>');
 	} else if (cols1 == 1) {
+		$('#content .product-layout').addClass('col-sm-4').removeClass('col-sm-6 col-md-3 col-lg-3');
 		$('#content .product-layout:nth-child(3n+3)').after('<div class="clearfix visible-lg"></div>');
 	} else {
 		$('#content .product-layout:nth-child(4n+4)').after('<div class="clearfix"></div>');
--- a/upload/catalog/view/theme/default/template/module/bestseller.tpl
+++ b/upload/catalog/view/theme/default/template/module/bestseller.tpl
@@ -1,7 +1,7 @@
 <h3><?php echo $heading_title; ?></h3>
 <div class="row">
   <?php foreach ($products as $product) { ?>
-  <div class="col-lg-3 col-md-3 col-sm-6 col-xs-12">
+  <div class="product-layout col-lg-3 col-md-3 col-sm-6 col-xs-12">
     <div class="product-thumb transition">
       <div class="image"><a href="<?php echo $product['href']; ?>"><img src="<?php echo $product['thumb']; ?>" alt="<?php echo $product['name']; ?>" title="<?php echo $product['name']; ?>" class="img-responsive" /></a></div>
       <div class="caption">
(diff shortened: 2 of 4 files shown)


marcos-paiva:
When the left and right columns are enabled in layout, product list is squeezed, always showing 4 columns.

This commit solves this problem as follows:

* When there are two columns entitled, products items are arranged in two columns from the screens "small".
* When there is an enabled column will show three columns of product items.
* Only when there is no enabled column, we will be showing the default setting in this .tpl files

_The default setting item count continues in this way, even if the product items have different heights the layout appears not broken._

Thus, the default theme becomes more modular to different types of use that can have.

QualityWorks:
is there also an easy fix to show only 1 with when these modules are set in the left or right column? it now will also show 4 squeezed columns instead of only 1 (or perhaps your fix fixes that too?)

marcos-paiva:
Hey @QualityWorks.

The first commit that proposed fixes this problem. Always Products blocks are arranged one by one, _UNsqueezed_

With both the standard template works very versatile way.

veksen:
This shouldn't be dealt with with JS.

marcos-paiva:
Hey @veksen. Yes, you are correct: It's not the best way to treat this problem.

However, it is the best way now: :neckbeard:

In a perfect world, this information could be handled in a controller, but that's what we have today: People using the default theme will **have trouble customize their modules in columns**.

We think we can work together in a more efficient solution to this problem?

veksen:
I wholeheartly disagree. If a layout issue can be resolved with CSS alone, you fix it with CSS, not with JS.

Furthermore, this whole piece of JS should be taken out from the current code:

	// Adding the clear Fix
	cols1 = $('#column-right, #column-left').length;
	
	if (cols1 == 2) {
		$('#content .product-layout').removeClass('col-lg-3 col-md-3');
		$('#content .product-layout:nth-child(2n+2)').after('<div class="clearfix visible-md visible-sm"></div>');
	} else if (cols1 == 1) {
		$('#content .product-layout').addClass('col-sm-4').removeClass('col-sm-6 col-md-3 col-lg-3');
		$('#content .product-layout:nth-child(3n+3)').after('<div class="clearfix visible-lg"></div>');
	} else {
		$('#content .product-layout:nth-child(4n+4)').after('<div class="clearfix"></div>');
	}

Then, the columns should be handled on the controller (bestseller.php), somewhat like this:

    $route = isset($_GET['route']) ? $_GET['route'] : 'common/home';
    $route = $this->model_design_layout->getLayout($route);

    $data['column_left_active'] = count($this->model_design_layout->getLayoutModules($route, 'column_left'));
    $data['column_right_active'] = count($this->model_design_layout->getLayoutModules($route, 'column_right'));

bestseller.tpl:

    <?php if ((isset($column_left_active) && $column_left_active) && (isset($column_right_active) && $column_right_active)) { ?>
    <div class="col-lg-6 col-md-6 col-sm-12 col-xs-12">
    <?php } elseif ((isset($column_left_active) && $column_left_active) || (isset($column_right_active) && $column_right_active)) { ?>
    <div class="col-lg-4 col-md-4 col-sm-6 col-xs-12">
    <?php } else { ?>
    <div class="col-lg-3 col-md-4 col-sm-6 col-xs-12">
    <?php } ?>

default\stylesheet\stylesheet.css:

    /* fixed product layouts used in left and right columns */
    #column-left > .row > [class^='col-'],
    #column-left > .row > [class*=' col-'],
    #column-right > .row > [class^='col-'],
    #column-right > .row > [class*=' col-'] {
        width: 100%;
    }

This is how we fixed it for our clients (via VQMOD), I'm sure a more flexible approach could be used (the $column_left_active/$column_right_active should be accessible to any view, imo).

danielkerr:
this was added a while ago. if you guys still think there can be improvements using css then please post another commit.

PR #2990, opened 2015-05-06
Extension of config, request and session

rkgrep:
1. Implement a Bag class which provides useful methods for internal data handling.
2. Extend Config from Bag to avoid duplicate code
3. Use Bag as storage in Request so data can be pulled with default values to avoid lots of conditions in code (example of use in `admin/controller/catalog/attribute.php`)
4. Extend Session from Bag (with backward compatibility)
5. Add `pull` method to session to get data and unset it in one call


jamesallsup:
Please use the http://opencart.uservoice.com site for suggestions. Github is for bugs only.

Thanks.

PR #2995, opened 2015-05-08
Fixed security issue on log filename

mrnfrancesco:
I found a security issue on the setting controller that make you able to read and delete every file.
Admin panel access is a requirement so it is not so dangerous, but for just 2 line of code I think it's better to fix it.

### DIFF
--- a/upload/admin/controller/setting/setting.php
+++ b/upload/admin/controller/setting/setting.php
@@ -1404,9 +1404,11 @@ class ControllerSettingSetting extends Controller {
 		}
 
 		if (!$this->request->post['config_error_filename']) {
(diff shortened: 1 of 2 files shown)


opencarthelp:
Good catch.  Maybe an error would be more appropriate in this case than a silent conversion?  That way a simple typo like ``error..log`` wouldn't have unintended effects.

It would also be nice to have a standardized tool for directory traversal protection in OpenCart so we can eliminate all the duplicate code.  That's probably an issue for a different day though.

mrnfrancesco:
Yeah, a standardized tool would be great!
Actually I'm working on other similar security issues in OpenCart and many of them are related with the same missing input sanitization.

By the way, an error is probably the best choice. I'll update my pull request according to your suggestion.

PR #2998, opened 2015-05-09
Full path disclosure due to Undefined index

mrnfrancesco:
In admin panel, if you are not logged in, and generate a `404 Not Found` error with something like `<youdomain>/admin/index.php?route=common/login/XXX` an `Undefined index` warning will appears leading to `Full Path Disclosure` issue.

### DIFF
--- a/upload/admin/controller/common/login.php
+++ b/upload/admin/controller/common/login.php
@@ -101,19 +101,7 @@ class ControllerCommonLogin extends Controller {
 	}
 
 	public function check() {
-		$route = '';
-
-		if (isset($this->request->get['route'])) {
-			$part = explode('/', $this->request->get['route']);
-
-			if (isset($part[0])) {
-				$route .= $part[0];
-			}
-
-			if (isset($part[1])) {
-				$route .= '/' . $part[1];
-			}
-		}
+		$route = isset($this->request->get['route']) ? $this->request->get['route'] : '';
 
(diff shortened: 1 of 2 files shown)


Milow:
I don't see how this can happen, and I can't reproduce.

Line 158 of admin/index.php:
`$controller->addPreAction(new Action('common/login/check'));`

Which is going run the check function on line 103 of admin/controller/common/login.php. It's going to fail the check at line 124 because you are not logged-in and the route is not ignored, so now we are sent to the login page. There is no way to hit the error/not_found.php file if you are not logged in.

Please advise me if I am wrong, but like I said, I cannot reproduce this issue.

mrnfrancesco:
2015-05-09-193820_1365x435_scrot

As you can see in the image above I see the warning coming from `error/not_found.php`.

I debugged opencart and I think the problem is I'm asking for a page that does not require login (`route=common/login`), so OC start searching for a method named `XXX` and hit the `error/not_found.php`.

Does it make sense?
*(It is possible that you cannot reproduce this issue if you have a well configured server that does not print warnings/errors)*

mrnfrancesco:
You're right, it is a better solution. I'll update my pull request in minutes.

Milow:
Didn't think about the simple solution - nice.

jamesallsup:
Perfect, thanks.

PR #3017, opened 2015-05-14
Bug fix #3015

### LINKED ISSUE #3015: Bug in admin/model/setting/setting.php
This is for OC 2.0.2.0 and the current master branch:

In file admin/model/setting/setting.php, in method editSettingValue, the following line 
```
$this->db->query("UPDATE " . DB_PREFIX . "setting SET `value` = '" . $this->db->escape($value) . "' WHERE `code` = '" . $this->db->escape($code) . "' AND `key` = '" . $this->db->escape($key) . "' AND store_id = '" . (int)$store_id . "'");
```
should be replaced with this:
```
$this->db->query("UPDATE " . DB_PREFIX . "setting SET `value` = '" . $this->db->escape($value) . "', serialized = '0'  WHERE `code` = '" . $this->db->escape($code) . "' AND `key` = '" . $this->db->escape($key) . "' AND store_id = '" . (int)$store_id . "'");
```
Otherwise, if the previous setting value was a serialised value, and is about about to be updated with a non-serialised string value, it keeps the wrong 'serialised=1'.

Newman101:
Fixed #3015.

### DIFF
--- a/upload/admin/model/setting/setting.php
+++ b/upload/admin/model/setting/setting.php
@@ -36,7 +36,7 @@ class ModelSettingSetting extends Model {
 
 	public function editSettingValue($code = '', $key = '', $value = '', $store_id = 0) {
 		if (!is_array($value)) {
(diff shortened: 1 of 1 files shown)


jamesallsup:
Thanks for the PR.

Newman101:
You are welcome, happy to help. :)

PR #3023, opened 2015-05-17
Bug fix: if $mail->reply_to is set, use that (instead of $mail->from) in the email headers 

steefdw:
If someone would ever use
```php
    $mail = new Mail();
    $mail->setReplyTo($reply_to);
    ...
    $mail->send();
```
Make sure the reply-to email address that is being set in ```$mail->setReplyTo($reply_to);``` is being used, instead of the ```$mail->from```, when creating the email headers here:
```php
$header .= 'Reply-To: =?UTF-8?B?' . base64_encode($this->reply_to) . '?=' . 
           ' <' . $this->reply_to . '>' . $this->newline;
```
Bug was introduced here https://github.com/opencart/opencart/commit/7851910af42eaaee84492ca62d2bbce34434418e

Note: although setReplyTo() is not used anywhere yet, it's very useful when sending an email on somebody else's behalf. So if one would every use it, this bug is fixed here.

### DIFF
--- a/upload/system/library/mail.php
+++ b/upload/system/library/mail.php
@@ -103,7 +103,7 @@ class Mail {
 		if (!$this->reply_to) {
 			$header .= 'Reply-To: =?UTF-8?B?' . base64_encode($this->sender) . '?=' . ' <' . $this->from . '>' . $this->newline;
 		} else {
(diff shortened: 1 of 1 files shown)


danielkerr:
fine

avcadvaanz:
Dear All,
How to configure the mail or smtp settings in open cart 2.1.0.1
please let me know anything of thil mail configuration.awaiting for your reply.

PR #3024, opened 2015-05-17
DRY: less isset(this-request->get['foo'])

steefdw:
Because this is used VERY often in the codebase:
```php
if (isset(this-request->get['foo'])) {
	$bar = $this-request->get['foo'];
} else {
	$bar = 'baz';
}	
```
...why not use:
```php
$bar = $this->request->get('foo', 'baz');
```

It does exactly the same, but with these extra [3 lines of code](https://github.com/steefdw/opencart/commit/bb985e6e4c88ea770a7f8ad24f5d97e6fbe5b13b), we could have more [readable code](https://github.com/steefdw/opencart/commit/cd95152334ee4a9c89bd1ca49ffcef90d0d65478) AND remove +1000 lines of code.

This proposal does not replace functionality, but adds to it:
* ```isset($this->request->get['foo']``` can still be used for stuff like:
```php
	if (isset($this->request->get['foo'])) {
		$url .= '&foo=' . $this->request->get['foo'];
	}
```	
* but ```$this->request->get('foo', 'baz')``` is more readable (and preferred IMHO) than:
```php
	if (isset(this-request->get['foo'])) {
		$bar = $this-request->get['foo'];
	} else {
		$bar = 'baz';
	}	
```

### DIFF
--- a/upload/catalog/controller/information/information.php
+++ b/upload/catalog/controller/information/information.php
@@ -12,12 +12,7 @@ class ControllerInformationInformation extends Controller {
 			'href' => $this->url->link('common/home')
 		);
 
-		if (isset($this->request->get['information_id'])) {
-			$information_id = (int)$this->request->get['information_id'];
-		} else {
-			$information_id = 0;
-		}
-
+		$information_id = (int)$this->request->get('information_id', 0);		
 		$information_info = $this->model_catalog_information->getInformation($information_id);
 
 		if ($information_info) {
--- a/upload/catalog/controller/product/category.php
+++ b/upload/catalog/controller/product/category.php
@@ -9,35 +9,11 @@ class ControllerProductCategory extends Controller {
 
 		$this->load->model('tool/image');
 
-		if (isset($this->request->get['filter'])) {
-			$filter = $this->request->get['filter'];
-		} else {
-			$filter = '';
-		}
-
-		if (isset($this->request->get['sort'])) {
-			$sort = $this->request->get['sort'];
-		} else {
-			$sort = 'p.sort_order';
-		}
-
-		if (isset($this->request->get['order'])) {
-			$order = $this->request->get['order'];
-		} else {
-			$order = 'ASC';
-		}
-
-		if (isset($this->request->get['page'])) {
-			$page = $this->request->get['page'];
-		} else {
-			$page = 1;
-		}
-
-		if (isset($this->request->get['limit'])) {
-			$limit = $this->request->get['limit'];
-		} else {
-			$limit = $this->config->get('config_product_limit');
-		}
+		$filter = $this->request->get('filter', '');
+		$sort = $this->request->get('sort', 'p.sort_order');
+		$order = $this->request->get('order', 'ASC');
+		$page = $this->request->get('page', 1);
+		$limit = $this->request->get('limit', $this->config->get('config_product_limit'));
 
 		$data['breadcrumbs'] = array();
 
--- a/upload/system/library/request.php
+++ b/upload/system/library/request.php
@@ -28,4 +28,8 @@ class Request {
 
 		return $data;
 	}
+        
+	public function get($key, $default = null) {
+		return (isset($this->get[$key]) ? $this->get[$key] : $default);
+	}        
 }


danielkerr:
problem is sometimes you require direct access to the post vars.

you might need $this->request->post['foo'][$key]

this can not be done with $this->request->post('foo')[$key]

steefdw:
Hi Daniel,

You are totally right that ```$this->request->post('foo')[$key]``` (looks and) feels REALLY dirty. Like I said, this proposal does not *replace* functionality, but *adds* to it.

$this->request->**get**($key, $default) could be used instead of code blocks where we do:
```php
if(isset($request[$key])) $foo = $key; else $foo = $default;
```
but in other cases, we can still use the current way of doing it: 
```php
$this->model_catalog_product->addProduct($this->request->post);
$this->request->get['foo'][$key];
```
So $this->request->**get**($key, $default) ---or $this->request->**post**($key, $default) --- is a method for retrieving a specific key from the cleaned $_GET / $_POST that can default to anything you like. Without the need for ```isset()``` or ```if/else``` stuff in the controllers.

But we don't have to replace all the instances that we use
```php
if(isset($request[$key])) $foo = $key; else $foo = $default;
```
because this extra "helper" doesn't break anything.

jamesallsup:
OK, I was initially a big fan of this - even confirmed with Daniel i'd like to see it to potentially save some lines of code from the 5151 occurrences of $this->request->get. But looking more into it it would look ugly.

YES, technically you could have a method of "get()" and a property of "get" so you could be backwards compatible:

$this->request->get('var', 'default');

$this->request->get['var'];

But that is borderline, the worst standards ever (I think my IDE thought I was joking when i tried it). I think it requires more thought to work (or a lot of changes) but it is definitely along the right lines - looking at other frameworks most do this I can see.

@steefdw I'll take a look more into this though :)

steefdw:
Hi James,

I guess you're right. My proposed syntax is a bit confusing, and confusing stuff should be avoided at all times. This...
```php
$this->request->get('var', 'default');
$this->request->get['var'];
``` 
is confusing for newcomers, and for your IDE apparently ;)

But we still have all those:
```php
$this->config->get('foo');
$this->request->get('foo');
$this->request->post('foo');
et cetera
```

Maybe I should rewrite this proposal to something like:
```php
$this->config_get($key, $default)
$this->request_get($key, $default);
$this->request_post($key, $default);
```
The $key should be "dot notation", so The issue Daniel mentioned is solved:
```$this->request->post('foo.bar')``` to get the value of ```$this->request->post['foo'][$key]```

Those 3 (or more) functions could call a "global" function to get stuff from the registry. Let's say Registry::get('config', $key, $default), so we don't have to repeat ourselves.

In code (from the top of my head), ```$this->registry_get('post', 'settings.categories.product_limit')``` could be something like: 
```php
function registry_get($registry_key, $key, $default = null)
{
	$array = $this->$registry_key;

	if (is_null($key))       return $registry_key;
	if (isset($array[$key])) return $registry_key[$key];
	
	foreach (explode('.', $key) as $segment) {
		if ( ! is_array($registry_key) || ! array_key_exists($segment, $registry_key)) {
			return value($default);
		}

		$array = $registry_key[$segment];
	}
	
	return $array;
}
```
So for a post like this:
```php
    $_POST = array(
        'settings' => array(
            'categories' => array(
                'product_limit' => '15',
                'product_description_length' => '100',
            ),
            'foo' => array(
                'bar' => 'beer',
                'baz' => 'coffee',
            )
        ),
    );
```
we can do this ```$this->registry_get('post', 'settings.categories.product_limit')``` to get the result of "15". That's a lot easier (and easier to read!) than a lot of issets and if/else.

So to conclude: is that more readable / cleaner than my proposal of this pull request? I like easy syntax to get thinks done, so I don't mind to rewrite this concept with other naming.

danielkerr:
actually normally code igniter uses

to get

$this->request->get($key);

to set

$this->request->get($key, $value);

you could always use a magic method to return null if a key does not exist.

opencarthelp:
I'd recommend checking out https://github.com/opencart/opencart/pull/2990 from @rkgrep.  It's similar to Symfony's HTTPFoundation library.  It would be easy to add a toArray() method to the Bag class or even directly inject the class into the model.

steefdw:
Okay, this pull request is becoming a discussion/issue. 
So let's close it.

Like James said, we're doing stuff very verbose 5151 times for $this->request->get().
And if we look at all those $this->request_post() and $this->config->get() issets, that's +20k LOC that could be thrown away AND make things more readable.

I will close, and rewrite this PR with the input from you guys in this PR.

@opencarthelp: this wheel was already invented by some frameworks. 
* Arr:get($array, $key, $default) from Kohana
* array_get($array, $key, $default) from Laravel (http://laravel.com/docs/4.2/helpers#arrays)

PR #3029, opened 2015-05-20
 DRY: don't do file_exists() for every template file

steefdw:
In every controller, this is used to render the page:
```php        
if (file_exists(DIR_TEMPLATE . $this->config->get('config_template') . '/template/foo/bar.tpl')) {
    $this->response->setOutput($this->load->view($this->config->get('config_template') . '/template/foo/bar.tpl', $data));
} else {
    $this->response->setOutput($this->load->view('default/template/foo/bar.tpl', $data));
}   
```
This can also be done with:
```php
$this->response->render('foo/bar', $data);
```

And for partials, this is used every time:
```php
if (file_exists(DIR_TEMPLATE . $this->config->get('config_template') . '/template/foo/bar.tpl')) {
    return $this->load->view($this->config->get('config_template') . '/template/foo/bar.tpl', $data);
} else {
    return $this->load->view('default/template/foo/bar.tpl', $data);
}                     
```
It's easier to read/write with this syntax:
```php
return $this->response->compile('foo/bar', $data);
```
It does exactly the same, but with [a few extra lines of code](https://github.com/steefdw/opencart/commit/717188838a953b0d4202517afe3a92e490b2be45), we could have more readable code AND remove (169x4=) 676 lines of code. It's used in 132 files (169 occurrences).

In this PR, I cleaned 39 files, so 93 files (122 occurrences) to go. After this improvement is implemented, I will clean those as well.

This proposal does not replace functionality, but adds to it. So you could still use the current way of setting the template file.

### DIFF
--- a/upload/catalog/controller/checkout/cart.php
+++ b/upload/catalog/controller/checkout/cart.php
@@ -245,11 +245,7 @@ class ControllerCheckoutCart extends Controller {
 			$data['footer'] = $this->load->controller('common/footer');
 			$data['header'] = $this->load->controller('common/header');
 
-			if (file_exists(DIR_TEMPLATE . $this->config->get('config_template') . '/template/checkout/cart.tpl')) {
-				$this->response->setOutput($this->load->view($this->config->get('config_template') . '/template/checkout/cart.tpl', $data));
-			} else {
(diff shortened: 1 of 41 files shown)


danielkerr:
response has nothing to do with the template so it should not be able to call it.

PR #3058, opened 2015-05-26
Moved Validate cart has products and has stock on library/cart.

eka7a:


upload/system/library/cart.php
```diff
@@ -443,4 +443,16 @@ public function hasDownload() {
 
 		return $download;
 	}
+
+	public function hasCart() {
```
skyosev: *isEmpty()* would be more logical name, comparing $this->cart->hasCart() and $this->cart->isEmpty()

PR #3066, opened 2015-05-30
Add function to develop extensions and OCMod

valdeir2000:
### Add function to develop extensions and OCMod #3066
The when function enabled deletes the cache files and rebuild all modifications
installed, it makes developing simpler extensions.

The proposed modification is to automate some functions.

Clear the logs (System, OCMod and Openbay)
Refresh modifications
Removes old cache files

**How to enable the function**
Administration Panel > System> Configuration > Tab Server

#### Map correction of Brazil (jqvmap).
The previous file was incomplete.


danielkerr:
i think this can be done in a better way. why not just change the current core files to refesh the mod system on upload

valdeir2000:
Yes of course! We can always improve. The initial idea is that updating the sistem file modification, it would be better to address possible conflicts between modules.

Issue #3069, opened 2015-06-01
Maxmind Error !


haxcop:
HI, i just create an order with paypal express and with the maxmind service active on oc2.0.3.1 and i just got this error message after the paypal process , but  when i click back it shows me the successful page of paypal. any ideas? 
![6ee1a90d40cf66db1a6ad5993e86a836 copy](https://cloud.githubusercontent.com/assets/7826906/7904590/188a5646-07fa-11e5-832c-acc91f486e41.png)

pekka2:
Is this old code? You try line 6 in file catalog/model/fraud/maxmind.php:
`$fraud_info = $this->getFraud($data['order_id']);`
change to:
`//$fraud_info = $this->getFraud($data['order_id']);`

haxcop:
Sorry for the delay @pekka2  what do you mean "is this old code?". and also that error that i get i think will be for everyone here using oc.2.0.3.1 , have you make the pull request yet?

pekka2:
The variable "$fraud_info" does not refer anywhere else in the whole file. It gives the impression that this line 6 is unnecessary.

haxcop:
Hi, thanks for your comments. @danielkerr  Hi, can you please check this error and let us know if you are agree with or there's other Reason>? Thanks @pekka2

mrnfrancesco:
The `getFraud` method was deleted in commit https://github.com/opencart/opencart/commit/a0f3a115fce8dd9b55e121af2e4a464316b540a2

I'll make a pull request fixing it in minutes.

haxcop:
Hi @mrnfrancesco & @pekka2 Thanks for your Support! Now let's see if Daniel think is right or not.
Best Regards

jamesallsup:
Merged.

44kaine:
Did we get some kind of solution for this issue, encountering exactly the same using OC 2.0.3.1 just recently downloaded and installed. Weird that a default module doesnt work

PR #3076, opened 2015-06-03
Fix issue #3069

mrnfrancesco:
`getFraud` method was still invoked, but the method does not exists anymore by commit https://github.com/opencart/opencart/commit/a0f3a115fce8dd9b55e121af2e4a464316b540a2 .

`query` SQL is generated using `order_id` variable, but it does not exists (it was the `getFraud` parameter). Replaced with the correct one.


jamesallsup:
Thanks.

PR #3077, opened 2015-06-03
Fix maintenance mode always true

ManoelVidal:
Fix maintenance mode always true in refresh the cache of modification.


xguntis:
Is this fix implemented in master branch? I look at the current code of master branch for this file `opencart/upload/admin/controller/extension/modification.php` and the 2 lines after line 67, which are added in this commit, are not at master branch. Also, when downloading the latest code, **the problem of *maintenance mode always true* still exists**.

This line is missing in the latest code:

```
$this->config->set('config_maintenance', true);
```

bijak8:
A simple fix:-
http://www.opencart.com/index.php?route=extension/extension/info&extension_id=22694&filter_license=0&filter_download_id=41&page=8

Issue #2722, opened 2015-02-27
Wrong order status if clause


GomDani:
In  opencart/upload/admin/model/sale/order.php

public function getTotalOrders($data = array()) {}

is

`if (!empty($data['filter_order_status'])) {`

must be

`if (isset($data['filter_order_status'])) {`

Now displays all orders, if filtered by order_status_id = 0.

munirasim:
isset() is implicitly checked by empty() so I don't see any improvement there

GomDani:
The improvement is, that with isset() total_orders display right values, with if !empty() missing orders with 0 value for order_status_id is not displayed. Go to demo page and in order list filter by missing orders and you see that filtered total orders is the same that not filtered.

munirasim:
No such issue in Opencart 2.0.1.1 as you claim. Attached is the proof
![opencart](https://cloud.githubusercontent.com/assets/11026716/6424762/c085a0c4-bf0e-11e4-9f4b-a4376d04aae3.png)

GomDani:
Not total of individual order, but the number of orders in the list is wrong. Look down.

munirasim:
'Missing Order' filter is for those orders which customers put in cart but did not 'confirm'. Down in the list I have such orders. Would you please establish your point more clearly and with an image if possible?

GomDani:
Look, go to demo page in opencart.com and filter order list by order_status_id=0. Go down to pagination and look how many orders is in the list. Go to page 15 and you see.

Showing 281 to 300 of 965 (49 Pages)

but the list is empty...

munirasim:
It is installation specific issue not Opencart issue in general

danielkerr:
the demo is the first 2.0 version released.

GomDani:
Yes is the first, but the function structure is the same. Download 2.0.0.0 and 2.01.1 and no difference.

mrnfrancesco:
@GomDani is right:

if `$data['filter_status_id']` is equal to `'0'` *(Missing orders)*, `!empty($data['filter_order_status'])` returns `false` skipping the `if` statement and going down into the `else` which do:

```
$sql .= " WHERE order_status_id > '0'";
```

You can reproduce the same issue if you filter by order ID: if you type, for example, 1669, you will have the 1669 order as unique result. If you type 0 you will have all the orders as results!

PR #3079, opened 2015-06-04
Fixed issue #2722

mrnfrancesco:
When filtering orders, if argument is '0', the `empty` method returns `true` and the query returns wrong results. I changed it with the `isset` method which returns the right boolean.

You can easily reproduce the issue filtering by *order ID*: If you type 0 you will have all the orders as results.


danielkerr:
yes you want an argument of 0

PR #3095, opened 2015-06-09
Gift Voucher Email Fix

OpenCartAddons:
Fix for gift voucher confirmation email.  Wrong order_status_id variable was being used.


OpenCartAddons:
This appears to be corrected in the latest version.

PR #3096, opened 2015-06-10
HTML5 Input Types

OpenCartAddons:
Updated some of the front end input types to be HTML 5 compliant.


danielkerr:
not all html 5 types are supported by firefox.

nuke123:
You should be at least using the basic ones: email/tel/number. They are supoprted in all browsers and the advantages they give on mobile changing the keyboard layout. 

Any browser which doesn't support it will revert back to text anyway.

OpenCartAddons:
@danielkerr The pull request was geared more towards mobile keyboards as @nuke123 said.

These input types are already in some areas of the default template, I just carried this standard through the rest of the of it.

Joel.

Issue #3116, opened 2015-06-19
Fatal error: Call to undefined method ModelAffiliateAffiliate::deleteLoginAttempts()


Uniarts:
When a customer wants to register to affiliate program there is:
Fatal error: Call to undefined method ModelAffiliateAffiliate::deleteLoginAttempts() in /home/.../public_html/vqmod/vqcache/vq2-catalog_controller_affiliate_register.php on line 21

and after when someone affiliate login: 
Fatal error: Call to undefined method ModelAffiliateAffiliate::getLoginAttempts() in /home/.../public_html/vqmod/vqcache/vq2-catalog_controller_affiliate_login.php on line 124

and there is no possibility to login

-------------------EDIT ------------------------------------
opencart ~~2.0.3.1~~  not affected with this bug
**Latest OpenCart 2.0.3.2_rc** until commit 4,474 https://github.com/opencart/opencart/commit/feffa3a2d82e6176b554e3717df8917a1aa30dae  has this error >>> but **@mrnfrancesco** applied this fix https://github.com/opencart/opencart/pull/3117 [Fixed call to undefined methods and calling method on null object] (https://github.com/mrnfrancesco/opencart/commit/f9105bf21b9b3d77cfde69578a46f0e39e6bfdfc). I tested on a live site and the fix worked.

kanenas:
@Uniarts this is a bug tracker. Please ask for help here http://forum.opencart.com/ or you could just delete the **vQmod** that generates the fatal error.

mrnfrancesco:
@Uniarts is right. I do not have vQmod installed and can reproduce the error.

Uniarts:
@kanenas Sorry for my ignorance but are you sure that this is not a bug?

mrnfrancesco:
@Uniarts yep, it is a bug. It depends on these lines (as you suggested):

catalog/controller/affiliate/register.php
```
// Clear any previous login attempts in not registered.
$this->model_affiliate_affiliate->deleteLoginAttempts($this->request->post['email']);
```
and
catalog/controller/affiliate/login.php
```
// Check how many login attempts have been made.
$login_info = $this->model_affiliate_affiliate->getLoginAttempts($this->request->post['email']);
```

The problem is that in affiliate model (catalog/model/affiliate/affiliate.php) there are no methods like `deleteLoginAttempts` and `getLoginAttempts`.

milimar:
[...]

PR #3117, opened 2015-06-19
Fixed issue #3116

mrnfrancesco:
In the following controllers are used two method not belonging to the used model.
In particular:

in `catalog/controller/affiliate/register.php`:
```
// Clear any previous login attempts in not registered.
$this->model_affiliate_affiliate->deleteLoginAttempts($this->request->post['email']);
```
and
in `catalog/controller/affiliate/login.php`
```
// Check how many login attempts have been made.
$login_info = $this->model_affiliate_affiliate->getLoginAttempts($this->request->post['email']);
```

The problem is that in affiliate model (`catalog/model/affiliate/affiliate.php`) there are no methods like `deleteLoginAttempts` and `getLoginAttempts`, because they belongs to customer model (`catalog/model/account/customer.php`).

### DIFF
--- a/upload/catalog/controller/affiliate/login.php
+++ b/upload/catalog/controller/affiliate/login.php
@@ -120,7 +120,7 @@ class ControllerAffiliateLogin extends Controller {
 
 	protected function validate() {
 		// Check how many login attempts have been made.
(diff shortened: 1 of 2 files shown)


mrnfrancesco:
Bug introduced with commit 32cb600a2621c00a300bbac1cf17721d9a5cda59

PR #3120, opened 2015-06-21
fix for shipping to countries with no zones

Dania02525:
Fixed problem with zone selection being required for countries without zones (Eg., Singapore).  If no zones found for country, create 'dummy' zone called No zone.

### DIFF
--- a/upload/catalog/model/localisation/zone.php
+++ b/upload/catalog/model/localisation/zone.php
@@ -14,6 +14,10 @@ class ModelLocalisationZone extends Model {
 
 			$zone_data = $query->rows;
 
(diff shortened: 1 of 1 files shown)


danielkerr:
system already does this!

Dania02525:
I ran into this while using v2.0.3.1 and the kingstore light free theme. I suspect that it might be common for themes to be missing the logic for giving the zero/nil option in the templates that you have for example on lines 322 to 337 on \view\theme\default\template\checkout\payment_address.tpl.  Although not as elegant, it does eliminate the need for this logic in themes :-).  Would it be worthwhile to move this logic to the core and out of the theme?

PR #3128, opened 2015-06-23
adds db_port to database connection

raventid:
If you have to start new connection with different port (not 3306), index.php does not create connection and tells you that database user password is not correct, so it is a bit difficult to figure out what is going on, if you do not know exactly where to find the problem.

### DIFF
--- a/upload/index.php
+++ b/upload/index.php
@@ -28,7 +28,7 @@ $config = new Config();
 $registry->set('config', $config);
 
 // Database
-$db = new DB(DB_DRIVER, DB_HOSTNAME, DB_USERNAME, DB_PASSWORD, DB_DATABASE);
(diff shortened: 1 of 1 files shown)


danielkerr:
so why are you requesting ports be removed altogether?

what about remote DB's?

raventid:
Perhaps I experience some kind of gap in my understanding of how this line of code works, but I am pretty sure it should be DB_PORT here.

 Last day we (2 devs) had to build a small web-site and we took opencart, I've connected to remote mysql on default port 3306 and my co-worker on port 3333 via SSH. He got db_user_password is wrong, but we could avoid this by sending DB_PORT, while creating new DB(). That is what I am trying to do ^_^

After some research I discovered that my pull request is not complete, should add DB_PORT in admin/index.php too.

raventid:
Thank you, jamesallsup! You've dramatically improved my commit!

webxpro:
when you want to create database and if you not create oc_  then also got error , we can not use any other than oc_

PR #3149, opened 2015-06-29
Fix bad link to contact us on order successful page

richard-kong:
When an order is completed successfully as a logged in user. The "contact us" link goes to the download page in the account area. This is due to the incorrect variable being used for the contact address.


jamesallsup:
https://github.com/opencart/opencart/blob/master/upload/catalog/language/english/checkout/success.php#L9

3rd reference is the download link, 4th is the contact link. If you have a custom language / theme installed check that it is correctly.

PR #3156, opened 2015-07-02
Remove PayPal Pro "Please Wait..." message after completion

jaywilliams:
Added the missing "attention" class to the alert div.


danielkerr:
it should be  $('.alert').remove();

PR #3165, opened 2015-07-05
Fix reply_to in header

detergen:


danielkerr:
except i removed reply to as it would get flagged as spam

detergen:
So opencart admin can't find email to reply to real sender f.e. from "Contact Form" ...  Ok, as is.

danielkerr:
fine

PR #3181, opened 2015-07-14
New error message when uploading extension via Extension Installer

chipironcin:
Add more information about the error when uploading an extension via Extension Installer.


danielkerr:
invalid file type is enough. people will start renaming normal extension they download and try to upload.

chipironcin:
I had a hard time before I realize some extensions require to end like that (ocmod.zip or ocmod.xml). Some of them do not specify that in the description and the OC will keep saying "invalid file type" when in fact its just naming.

PR #3188, opened 2015-07-20
Bunch of old bugs that never got fixed

dhaupin:
- Removed redundant/duplicate product url generation (no need for cats and mfg products since there are already canonicals). **"Only include canonical URLs. A common mistake is to include URLs of duplicate pages. This increases the load on your server without improving indexing."** http://googlewebmastercentral.blogspot.com/2014/10/best-practices-for-xml-sitemaps-rssatom.html

 - Added ampersand escape on all urls: **"All data values in a Sitemap must be entity-escaped"** **"In addition, all URLs (including the URL of your Sitemap) must be URL-escaped and encoded for readability by the web server on which they are located"** http://www.sitemaps.org/protocol.html#escaping

 - Added replace spaces with `%20` on all output since MANY people use spaces in their image folder names and "dumbcrawlers" will hang on parse or complain on SEO scans. This is not technically required by the spec but experience and 404 logs show tons of incomplete uri requests if no `%20` exists in paths.

 - Added `<lastmod>` for categories and products in W3C datetime format. MFG and informations are lacking a modified column so didn't add those. (They should have it too for many semantic/push/ping/sanity reasons, but that's another commit...).

 - Removed image resizing. This causes insane load on servers when a large sitemap is generated. Side effects: script timeouts beyond what default php.ini allows, connection timeouts or multi-fetch retry A.D.D. after ~60-90 seconds (can cause EP caps or super easy DoS on CL enabled servers/VPS), and possible hosting penalties for CPU abuse on shared servers with popular site(s) being hit by Search bots, scrapers, scanners, etc. Google prefers full size images anyways and has its own caching and limiting mechanisms that decide what to do per the device requesting via search. We live in a world with unlimited hosting bandwidth as the norm....its just logical in general to provide high resolution and allow endpoints to scale down.

 - Added full URL for images instead of just the URi. Again, for dumbcrawlers and stuff that may not be aware of the TLD from just the map.

 - Also did some admin product image changes to move to correct tab.

 - Added the category canonical and breadcrumb chain fixes too.


smilesrg:
+1

CoverUp:
hope this dont go to the bunch of old bugs fixes that never got merged

garudacrafts:
A lot of the 400+ highlighted line changes under "Files Changed" appear to be triggered by whitespace differences. Would it be possible for you to separate these from the legitimate modifications by modifying your pull request?

dhaupin:
@garudacrafts Sure i can separate them when i get a minute. Curious though -- should i be fixing wrong whitespace at all? It annoys the heck out of me (the wrong/extra whitespace all over the codebase)

garudacrafts:
@dhaupin I'm not sure which is correct, but I'd recommend separating it out whitespace/indention corrections from the substantive code changes, for readability. The wiki has a section on coding standards, https://github.com/opencart/opencart/wiki/Coding-standards, in case you'd like to check.

dhaupin:
@garudacrafts Alright, good to go on the revert. Extra/inconsistent whitespace is another day

@allFolks, just noticed its reporting the pull merged into my repo via github ui. I can revert that too if its BS and trying to push to ocmaster, did it via phone to sync dhmaster (via wtfui ... was the right head...), my bad.

garudacrafts:
Much easier to read. All the suggested changes to google_sitemap.php sound good, except "Added ampersand escape on all urls" (e.g. "str_replace('&', '&amp;'...") is redundant as the link() method in /system/library/url.php already performs this operation.

garudacrafts:
I think you'd have a better chance at getting this commit merged if you split it into 3, because they are all independent and deal with different issues

danielkerr:
ok lets take a look at this.

danielkerr:
the tab stuff is not required as the system changes the spaces to tabs and also removes white spaces.

danielkerr:
what advantage is this?

str_replace(array('&', ' '), array('&amp;', '%20'), $this->config->get('config_url') . 'image/' . $product['image'])

so having a image with & in the name will just cause the image not to show.

better to rename the image.

danielkerr:
again not required

str_replace('&', '&amp;', $this->url->link('product/category', 'path=' . $new_path))

the url class already does this.

danielkerr:
Category canonical cateogry id is already taken care of in another issue

danielkerr:
I'm going toclsoe this pull request. there are to many problems this will cause if it was merged. I really dont trust your coding.

if you want to resubmit please make sure you have set everythign out clearly.

PR #3190, opened 2015-07-21
Fix invalid reference to $option['option_value']

richard-kong:
When checking out via paypal express with a product in the cart that contains an option (and is not a file). There is an undefined indexed error for $option['option_value'].


danielkerr:
i think already added

PR #3194, opened 2015-07-23
Fixed 'option' variable for OCv2.0.x

hydrowire:
The array variable for option values in OpenCart v2.0.x should be 'value', not 'option_value'.


richard-kong:
I did the same fix :)

https://github.com/opencart/opencart/pull/3190

danielkerr:
already fixed

PR #3215, opened 2015-08-05
Contact POST url with regards to SSL

adrianrosian:
Fixing contact action for SSL

### DIFF
--- a/upload/catalog/controller/information/contact.php
+++ b/upload/catalog/controller/information/contact.php
@@ -80,7 +80,7 @@ class ControllerInformationContact extends Controller {
 
 		$data['button_submit'] = $this->language->get('button_submit');
 
-		$data['action'] = $this->url->link('information/contact');
+		$data['action'] = $this->url->link('information/contact', '', $this->config->get('config_ssl'));
 
 		$this->load->model('tool/image');
 


ADDCreative:
This is wrong. 'config_ssl' contains a URL not a flag to say if SSL is enabled or not. If you want to make the change, shouldn't it be.
```$data['action'] = $this->url->link('information/contact', '', 'SSL');```

adrianrosian:
@ADDCreative , I already tested this on an actual site. AFAI remember, I also took this model from other parts of teh code.

ADDCreative:
It will only of worked as ```https://somedomain.com``` equates to not false or 0. Please look at all other instances of $this->url->link in the code, you will see they all us 'SSL' as the 3rd argument (execpt for a few that use true).

adrianrosian:
@ADDCreative , you are mostly (but partially) right. I used the incorrect setting because I inspected the form (which doesn't contain a field with the same name in the database). The setting I used indeed contains the SSL url, not a boolean. I corrected this in PR #3261 , thank you very much. OTOH, the use of `SSL` as the third aregument is the same, it is not a boolean. I now use `config_secure` setting, which is what you want to use.

ADDCreative:
Sorry @adrianrosian, why should this link use ```$this->config->get('config_secure')``` when all the other 2000+ occurrences in the code use 'SSL'. Some examples of the 2000+:

https://github.com/opencart/opencart/blob/master/upload/catalog/controller/account/login.php#L150-L152

https://github.com/opencart/opencart/blob/master/upload/catalog/controller/checkout/cart.php#L231

It's not about it being functional but about being consistent with the rest of the code.

PR #3218, opened 2015-08-06
replace old jquery delegate() calls with on()

pine3ree:


danielkerr:
there is nothing on the jquery site to indecate delegate is not going to be used anymore.

why would you use .on() when it wont do work on dynamically updated html.

http://api.jquery.com/on/

The .on() method attaches event handlers to the currently selected set of elements in the jQuery object. As of jQuery 1.7, the .on() method provides all functionality required for attaching event handlers. For help in converting from older jQuery event methods, see .bind(),  .delegate(), and .live(). To remove events bound with .on(), see .off(). To attach an event that runs only once and then removes itself, see .one()

pine3ree:
Hello Daniel,

"on" works on dynamically updated elements if you use it like this

$('containserSelector').on('event', 'selector', handler). where containserSelector is a wrapper always present in the html, it could also be the document...but it's preferable to have more specifc selector.

i stopped using delegate in opencart themes 3 years ago.

http://api.jquery.com/delegate/
As of jQuery 1.7, .delegate() has been superseded by the .on() method. For earlier versions, however, it remains the most effective means to use event delegation. More information on event binding and delegation is in the .on() method. In general, these are the equivalent templates for the two methods:
```
// jQuery 1.4.3+
$( elements ).delegate( selector, events, data, handler );
// jQuery 1.7+
$( elements ).on( events, selector, data, handler );
```

regards

PR #3229, opened 2015-08-08
Issue with Tax calculation in system/library/tax

Dreamvention:
Hi

I have noticed that the system/library/tax.php when it comes to setting the tax_rates only uses the system config customer group id.

yet we have two more options that we want to use - that is 
calculate the tax rate based on the customer group id of a logged_in customer
or if the customer select a group id when registering. 

because of this issue, when a customer is guest - he will not see an update of the tax rate because they rely only on the config option.

thank you

### DIFF
--- a/upload/system/library/tax.php
+++ b/upload/system/library/tax.php
@@ -23,7 +23,15 @@ final class Tax {
 	}
 
 	public function setShippingAddress($country_id, $zone_id) {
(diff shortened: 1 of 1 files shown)


Dreamvention:
as I see it was moved to the index.php

yet because we are setting the session value after the we have loaded the application, the customer_group_id is not re

PR #3240, opened 2015-08-12
renamed single voucher item to $voucher

pine3ree:


danielkerr:
this would have caused issues with the voucher. i have renamed the totals to modules when loading discount totals should give the same effect.

PR #3242, opened 2015-08-13
Enhanced OC Admin User login with PCI compliance.

mdutra555:
I found out the hard way that our production OC Admin security was compromised by hackers some how got access to admin username/password. I had to act quickly to implement PCI compliance security measures. I also added a 2nd tier security by adding a security question/answer if the login attempt comes from an IP address outside the corporate office IP. Suffice to say after implementing this security enhancement in production the hackers have been deterred. I hope this can be a valuable enhancement to OpenCart – I am a huge fan!


danielkerr:
this is not pci compliance!

danielkerr:
only the audit trail you have got right for pci compliance. the other stuff is not required.

PR #3261, opened 2015-08-21
Corrected url generation on contact

adrianrosian:
The correct setting is `config_secure`, not `config_ssl`

### DIFF
--- a/upload/catalog/controller/information/contact.php
+++ b/upload/catalog/controller/information/contact.php
@@ -80,7 +80,7 @@ class ControllerInformationContact extends Controller {
 
 		$data['button_submit'] = $this->language->get('button_submit');
 
-		$data['action'] = $this->url->link('information/contact', '', $this->config->get('config_ssl'));
+		$data['action'] = $this->url->link('information/contact', '', $this->config->get('config_secure'));
 
 		$this->load->model('tool/image');
 
--- a/upload/catalog/controller/information/contact.php
+++ b/upload/catalog/controller/information/contact.php
@@ -80,7 +80,7 @@ class ControllerInformationContact extends Controller {
 
 		$data['button_submit'] = $this->language->get('button_submit');
 
-		$data['action'] = $this->url->link('information/contact', '', $this->config->get('config_secure'));
+		$data['action'] = $this->url->link('information/contact', '', 'SSL');
 
 		$this->load->model('tool/image');
 


ADDCreative:
The change should be ```$data['action'] = $this->url->link('information/contact', '', 'SSL');``` to be consistent with the rest of the 2000+ uses of secure links in the code.

adrianrosian:
@ADDCreative , that might be so, but that also means that the link is always SSL, irrespective of whether the user activated the setting or not from the admin panel. The only thing keeping the url from not being an actual SSL url would be the $ssl param passed to the `Url` class constructor, which I assume comes from the config (and which you don't normally modify unless you use SSL), so to me that is incorrect.

Am I wrong?

ADDCreative:
@adrianrosian I'll try to explain how the secure links work OpenCart. The 3rd argument of the Url link method means that the link needs to be secure if ```config_secure``` is set.

Whether to use HTTPS is decided by  ```config_secure```. As you can see in index.php line 64 https://github.com/opencart/opencart/blob/master/upload/index.php#L64. If ```config_secure``` is set  ```config_ssl``` is passed as $ssl to the Url class constructor. If ```config_secure`` is not set ```config_url``` is passed as $ssl instead (effectly making $ssl and $domain of the Url class the same).

adrianrosian:
@ADDCreative , thank you for the explanation, but I figured this as well. What I am saying is the following scenario:

1. user goes on and enables ssl in the opencart admin
2. user goes on and modifes the SSL url in the opencart config file to the `https://..` variant
3. later on he decides he doesn't need SSL, so he disables SSL from the opencart admin

Expected behavior: SSL is turned off, all links are now non-https. Real behavior:

1. with actions like `$data['action'] = $this->url->link('information/contact', '', 'SSL');` form is still sent to the https variant, which is no longer valid
2. with actions like `$data['action'] = $this->url->link('information/contact', '', $this->config->get('config_secure'));` form is sent to the correct url

I understand what you are saying, the first option is how secure links are generated all over. What I am saying is that it might not be the best option.

ADDCreative:
@adrianrosian You have not understood my explanation. Your real behaviour example 1 is incorrect.
```$data['action'] = $this->url->link('information/contact', '', 'SSL');``` returns the correct URL due to how the Url class is constructed. See line 64 of index.php. https://github.com/opencart/opencart/blob/master/upload/index.php#L64

This line unsures that if SSL is disabled in the admin the link method of the Url class will never return the https variant.

adrianrosian:
@ADDCreative , you are correct. Still, using a string where boolean is expected feels wrong. I will modify this for consistency, with strong objections.

jamesallsup:
https://github.com/opencart/opencart/commit/f64120d484cb34effcaf3b73f1a55f951b729f8a

PR #3265, opened 2015-08-22
Added global variables to templates

smilesrg:
* Language global variable (```$lng```). There is no more need to set ```$data``` with language strings: language string can be retrieved in the template by calling ```$lng->get('language_string')```. So all code pieces like ```$data['entry_email'] = $this->language->get('entry_email');``` can be removed from controllers.
* Request service global variable(```$request```). POST and GET variables can be retrieved directly in the template by calling ```$request->post('variable_name', 'default_value');``` or just ```$request->post('variable_name')```. All code pieces like below can be removed from controllers:
```php
if (isset($this->request->post['email'])) {
    $data['email'] = $this->request->post['email'];
} else {
    $data['email'] = '';
}
```

And in the templates it could be used like:
```php
<input type="text" name="email" value="<?= $request->post('email') ?>" />
```

So construction like ```<?= $request->post('email') ?>``` or ```<?= $request->get('email') ?>``` can be used.

* Url service global variable (```$router```). I remaned it from ```$url``` global variable to keep paypal modules work well. Now urls can be generated righ in the templates:
```php
<?= $router->link('account/forgotten', '', 'SSL') ?>
```

* Loader service global variable (```$loader```). Now columns, header, footer and even any controller can be loaded right into template. Example: ```$load->controller('common/content_bottom') ```

### DIFF
--- a/upload/system/engine/loader.php
+++ b/upload/system/engine/loader.php
@@ -77,12 +77,12 @@ final class Loader {
 			exit();
 		}
 
-		if (isset($data['url']) || isset($data['lng']) || isset($data['load']) || isset($data['request'])) {
-			trigger_error('Error: you must not override url, lng, load and request parameters in the $data array!');
+		if (isset($data['router']) || isset($data['lng']) || isset($data['load']) || isset($data['request'])) {
+			trigger_error('Error: you must not override router, lng, load and request parameters in the $data array!');
 			exit();
 		}
 
-		$data['url'] = $this->registry->get('url');
+		$data['router'] = $this->registry->get('url');
 		$data['lng'] = $this->registry->get('language');
 		$data['load'] = $this->registry->get('load');
 		$data['request'] = $this->registry->get('request');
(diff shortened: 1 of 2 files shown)


danielkerr:
sorry but not a good idea.

smilesrg:
@danielkerr what exactly didn't you like? The idea of global variables at all? Or some global variables like "request" and "loader" ?

Why don't you want to use something to get language strings directly in the templates, like it's done in Wordpress and some other CMS?

4pxs:
It will not work in template, where loaded other module. It will always be overridden by the variable. For example 'heading_title' in module and at page. Variable will be taken from the last loaded language file.  For example, instead of "Heading Checkout" will appear "Heading module Also Purchased"

arnisjuraga:
You can modify your Template loader, to access `$this->registry`  methods in template directly. 
I has been available in older OC, but have been removed starting some 2.x version. 
But if you need, you can do it. 
https://github.com/copona/copona/commit/d661cbf387e0d80ee093f0639cfa17e99721a4ab?w=1
https://github.com/copona/copona/commit/5274dd5c7cd2b26bbefe26c8025a283dcdd200da?w=1

PR #3282, opened 2015-08-29
Selects the lowest price in the event of overlapping special and disc…

chrisrollins65:
…ount prices. This confuses a lot of users (including me), both customer and seller because they expect the lowest price to be used, and currently, that's not always the case.


danielkerr:
i cant accept this. first this should be done at the sql level because otherwise it will not be able to sort by price.

also it should be down to the store owner to set the prices correctly. its not normal for the special price to be higher then the normal price.

PR #3310, opened 2015-09-09
Update order.php

OpenCartAddons:
Moved tool/upload model load before product data call.  Fixes error when products have upload file.


jamesallsup:
Great, thanks!

PR #3312, opened 2015-09-09
Added Guest Checkout as Captcha Option

OpenCartAddons:
Added Guest Checkout to the captcha list and modified checkout/guest to look for guest_checkout in array instead of register.

Also modified text_register to capitalize the "R".  Now says Register instead of register in the captcha list.


danielkerr:
might be better to use the key guest instead of guest checkout

danielkerr:
can you resolve the conflicts?

OpenCartAddons:
@danielkerr #3356

PR #3315, opened 2015-09-09
Change SEO Keyword Text to SEO URL

OpenCartAddons:
The term "SEO Keyword" has been causing a lot of confusion for OpenCart shop owners.  I've lost count the number of times I've worked on a customer's site only to find their SEO Keyword field stuffed with Google keywords.

It's pretty obvious why they believe that the SEO Keyword field is to input Google keywords. If you Google the term "SEO Keyword" the results are all related to Google keywords, and nothing to do with SEO friendly URLs.

I am proposing that we finally eliminate this issue by changing the field text to read "SEO URL".

This will lessen the issues of shop owners using the field for Google keywords.  It also now makes the field text correspond with the system setting "Use SEO URLs", so it's easier to identify that the two fields work together.

Joel.


GizMecano:
That may seem a little more explicit indeed.

PR #3356, opened 2015-09-18
Added Guest Checkout as Captcha Option

OpenCartAddons:
@danielkerr Revised to change the key to guest instead of guest_checkout.


danielkerr:
had to add manually.

PR #3358, opened 2015-09-18
Added template service

smilesrg:
Right now this piece of code:

```php
		if (file_exists(DIR_TEMPLATE . $this->config->get('config_template') . '/template/common/success.tpl')) {
			$this->response->setOutput($this->load->view($this->config->get('config_template') . '/template/common/success.tpl', $data));
		} else {
			$this->response->setOutput($this->load->view('default/template/common/success.tpl', $data));
		}
```

can be replaced with:

```php
$this->template->render('common/success.tpl', $data);
```

in every frontend controller.


smilesrg:
@danielkerr can I ask you why don't you like this change?

PR #3361, opened 2015-09-19
readd missed commit

pine3ree:


danielkerr:
did you not read that your last tax commit caused a lot of errors to appear?

thgis commit offers no improvement.

PR #3384, opened 2015-09-23
Payments

GomDani:


milimar:
I am not a native English speaker, but the sentence "If your hosting on godaddy you might need to rename the php.ini to user.ini " cannot be correct. 

I suggest "If you're hosting.....' or "If you are hosting......"

PR #3416, opened 2015-10-03
Voucher -> Gift Certificate

OpenCartAddons:
I've gone through and updated the language files in the front end to consistently use the word Gift Certificate instead of using a mixture of Gift Voucher and Gift Certificate.


danielkerr:
i prefer the word voucher but nevermind.

PR #3419, opened 2015-10-05
support for multiple upload

valdeir2000:


danielkerr:
i can not merge this as it has changes to files that are not related to support for multiple upload.

also it looks like an unfinished feature.

PR #3422, opened 2015-10-06
checkout.tpl enable payment method button on error

mgraph:
if you forgot to check "I have read and agree to the Terms & Conditions" the button get disabled and can't submit the form even after ticking the checkbox


danielkerr:
works fine for me.

PR #3433, opened 2015-10-07
Update backup.tpl

OpenCartAddons:
Fixed missing variable entry_export.  The wrong variable was being used in the template file, should have been entry_backup according to the language and controller files.


danielkerr:
i changed it to export import and not backup and restore

Issue #3380, opened 2015-09-22
Extention PayPal PayFlow template $text_loading Error


Krishan19:
Extention PayPal PayFlow template ( payment/pp_payflow.tpl ) file line no 76, data not recived from ajax.

PR #3450, opened 2015-10-08
Update pp_payflow.tpl

OpenCartAddons:
Fix for issue https://github.com/opencart/opencart/issues/3380


danielkerr:
alreayd fixed. also it should have been loading not waiting

Issue #3447, opened 2015-10-08
2.1.01 admin/controller/design/layout.php


chulcha:
$this->load->language('design/layout'); (ver 2.0)
=>
$this->language->load('design/layout'); (ver 2.1)

PR #3460, opened 2015-10-10
Solved issue #3447

Newman101:
Issue number #3447 solved. 

Changed: $this->load->language('design/layout');
To: $this->language->load('design/layout');

Across whole file upload/admin/controller/design/layout.php.

Thank you.


danielkerr:
i think we already went through all of this.

PR #3480, opened 2015-10-13
Fixed multiple variable overwrites inside a few foreach loops

matrunchyk:
Hi guys,

I just noticed that there is a bug in modification system - some array/string variables are being rewritten/redeclared many times within a loop. 

Could you please take a look on it and merge these changes?
Thank you very much!


danielkerr:
thats not how php works. nothing is being overwriten.

danielkerr:
as you can see both the default xml and the loaded xml files still work fine.

PR #3493, opened 2015-10-14
Add `use Cart\User` namespace/class

matrunchyk:
Hi guys,

Please merge this fix.
It fixes `PHP Fatal error:  Class 'User' not found in .../catalog/controller/common/maintenance.php on line 16`.

[Here](https://github.com/opencart/opencart/blob/master/upload/catalog/controller/common/maintenance.php#L16) is user a `User` class which has not been imported.

Thanks.

### DIFF
--- a/upload/catalog/controller/common/maintenance.php
+++ b/upload/catalog/controller/common/maintenance.php
@@ -1,4 +1,6 @@
 <?php
+use Cart\User;
+
 class ControllerCommonMaintenance extends Controller {
 	public function index() {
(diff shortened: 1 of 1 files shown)


danielkerr:
@matrunchyk are u a seller?

matrunchyk:
@danielkerr no, I does not sell extensions, I just work with my client who starts to use OpenCart instead of Magento. I like OpenCart and I'd be happy to contribute my part to it :)

PR #3501, opened 2015-10-15
product report total

pine3ree:
Daniel, please verify:

According to Cart/Cart  and ControllerCheckoutConfirm  the order_product table rows are populated like this:
total = cart product total = unit price * quantity
tax = unit tax amount ( see ControllerCheckoutConfirm populating$order_data['products'] ~ line 224) 

so we need to sum op.total + op.tax*op.quantity to get the order product total tax included instead of 1st adding up total + single tax

i believe (not 100% sure) that some time ago (maybe version 1.5.1~2 the tax column in order_product held the total product tax)

regards

### DIFF
--- a/upload/admin/model/report/product.php
+++ b/upload/admin/model/report/product.php
@@ -37,7 +37,7 @@ class ModelReportProduct extends Model {
 	}
 
 	public function getPurchased($data = array()) {
-		$sql = "SELECT op.name, op.model, SUM(op.quantity) AS quantity, SUM((op.total + op.tax) * op.quantity) AS total FROM " . DB_PREFIX . "order_product op LEFT JOIN `" . DB_PREFIX . "order` o ON (op.order_id = o.order_id)";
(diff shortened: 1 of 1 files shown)


Badgesec:
Hello
I too haveing trouble to get total product correct 
in report.   I have copied the file and it made no diffent to the total. 
Can any have a look at the file below and advise me if i done the change incorrect
[product.txt](https://github.com/opencart/opencart/files/38012/product.txt)
the file above is the same as product.php
regards

pine3ree:
@Badgesec 

Hello,

the only change is in the sql string:

from
```
SUM((op.total + op.tax) * op.quantity)
```

to

```
SUM(op.total + (op.tax * op.quantity))
```

In recent versions of opencart op.tax represents the unit tax so we need to multiply by the order_product quantity to have the total product tax.
Another correct sql string would be
```
SUM((op.price + op.tax) * op.quantity)
```
The only case in which the result are the same is when op.quantity is 1. Is this your case, maybe you are testing with simple orders?

kind regards

Badgesec:
Thanks for your feed back, Have checked sql string and is as per your line and sad to say it didn't make any diffents.
Only when qty is 1 it's correct.  Sample say i have a qty of 41 I get £631.33 but should be £34.85.
I do not know where to go from here.
Kind regards

PR #3503, opened 2015-10-16
Correct wrong variable index

revoke:


danielkerr:
ok added all this manually.

PR #3505, opened 2015-10-16
Branding and version changes.

maddes:


maddes:
This is a really dumb test, just ignore.

PR #3513, opened 2015-10-17
The Composer Autoload and Namespaces Support

prhost:
# The Composer Autoload and Namespaces Support #

* To get started, download the [Composer](https://getcomposer.org/) and run on your terminal `composer update`

#### Namespaces PSR-4 predefined: ####

* `Opencart\Admin`
* `Opencart\Catalog`
* `Opencart\System`

Ps.: You can add more namespaces in `upload/composer.json` and then run `composer dump-autoload` to generate your new namespaces.

#### Example of use in classes that you create in OpenCart: ####

* Creating a class with namespace, sample path `catalog/controller/example/myclass.php`:

```php
<?php

namespace Opencart\Catalog\controller\example;

class Myclass
{
        public static function foo() {
                 return 'foo';
        }

        public function bar() {
                 return 'bar';
        }
}
```

* Using your new class elsewhere, path `catalog/controller/example/other_class.php`:

```php
<?php

use Opencart\Catalog\controller\example\Myclass;

class ControllerExampleOtherClass extends Controller
{
        public function index() {

                 echo Myclass::foo();

                 $my_class = new Myclass();
                 echo $my_class->bar();

        }
}
```

* Now only access route `upload/index.php?route=example/other_class` and good work (:

### DIFF
--- a/upload/composer.json
+++ b/upload/composer.json
@@ -1,0 +1,17 @@
+{
+    "name": "opencart/opencart",
+    "authors": [
+        {
+            "name": "Opencart",
+            "email": "opencart@opencart.com"
+        }
+    ],
+    "require": {},
+    "autoload": {
+        "psr-4": {
+            "Opencart\\Admin\\": "admin/",
+            "Opencart\\Catalog\\": "catalog/",
+            "Opencart\\System\\": "system/"
+        }
+    }
+}
--- a/upload/index.php
+++ b/upload/index.php
@@ -2,6 +2,9 @@
 // Version
 define('VERSION', '2.1.0.2_rc');
 
+// Composer Autoload
+require_once('vendor/autoload.php');
+
 // Configuration
 if (is_file('config.php')) {
 	require_once('config.php');


danielkerr:
never planning to autoload models or controllers

prhost:
That's why Opencart is dying because they seek or do not accept new technologies do not want evolution, that not Codeigniter. The Composer is not only for use as autoload, but allow the community to manage your packages more easily than unzip a zip in the library folder.

chienhoang:
Composer is tended using for lib dependencies and OpenCart is not a library at all then what do you think @prhost? https://getcomposer.org/doc/00-intro.md#introduction
```
Composer is a tool for dependency management in PHP. It allows you to declare the libraries your project depends on and it will manage (install/update) them for you.
```
my 2 cents.

prhost:
Dependencies composer means that you are dependent on that library or Quaker another package, Opencart is not 100% modular, but if you create a certain extent you can depend on something, such as a slug generator if the Opencart does not provide it in your core, you need to do in the hand or uncompress a library in system / library, for use in its length, if that library has a bug, you need to manually correct or uncompress a newer version, it is laborious, is massante understand that the composer will not replace the market of you, on the contrary, will be able to develop extensions enable higher quality.

hjfigueira:
@chienhoang  but the pourpose of this pull request is add composer compatibility for libraries that IT could depend on, not transform itself in a library.

PR #3517, opened 2015-10-18
Master

deepvision7:
Select All (Orders) Not Activating 'Print Invoice' and 'Print Shipping List'

http://forum.opencart.com/viewtopic.php?f=191&t=150448


danielkerr:
this has been fixed

PR #3523, opened 2015-10-19
Fix pagination

Spell6inder:
Remove `?page=1` from paginate urls


ADDCreative:
Doesn't work when SEO URLs are on as it doesn't replace ```?page=1``` only ```&page=1```. For example a URL or ```example.com/desktops?page={page}``` doesn't get changed.

Spell6inder:
#3617

PR #3551, opened 2015-10-25
Fix empty array error if no product selected

diyweb:
An error message is shown (Warning: Invalid argument supplied for foreach() in .../controller/module/featured.php on line 124) if there is no product selected.


OpenCartAddons:
You can find the fix here https://github.com/opencart/opencart/pull/3552

PR #3563, opened 2015-10-26
Fix order not disappearing from the list after order delete

matrunchyk:
Hi @danielkerr, this commit fixes a small issue — when a user deletes an order (which is done by AJAX) the order list is not to be refreshed (a correspondent table row is still remains on the table)


GomDani:
Why not just reload page. Orders are not deleted so frequently.

iofen:
Replace:
    location.reload();
with:
    $(node).closest('tr').addClass("danger");

Set order red color after order delete,don't need reload

danielkerr:
it already does rediect

PR #3564, opened 2015-10-27
Add MySql DbEngine Option on Install

andre-m-pereira:
On install add MySql db engine configuration. Now we can choose between InnoDB and MyISAM.
Already developed a jquery function that only shows this options when the db drivers are 'mysqli' or 'mpdo'.


danielkerr:
if you want innodb tables just alter the tables using phpmyadmin operation.

this is not required!

andre-m-pereira:
Ok, It was just to simplify life to the common user.

PR #3565, opened 2015-10-27
[geo_zone_form.tpl] Fix Select Value

mgraph:
Change select_box.attr("value") to select_box.val() because .attr("value") returns "Undefined"


danielkerr:
already fixed.

PR #3601, opened 2015-11-05
[order_shipping.tpl] remove unused Script & CSS

mgraph:
No need to load "JQuery" or "font-awesome.min.css"


danielkerr:
its there in case some one wants to use them with the shippign invoice

PR #3611, opened 2015-11-09
Button Reset

valdeir2000:
The problem occurs when the user does not mark the option to accept the terms.


chienhoang:
Is it related to this fix? https://github.com/opencart/opencart/commit/e895ed0742b543088e1f87ced335b7e690c67017

PR #3617, opened 2015-11-10
Udp pagination seo fix

Spell6inder:


ADDCreative:
This, of something similar, needs to be merged. It was a fix for a previous update to the pagination that added a bug https://github.com/opencart/opencart/pull/3523.

PR #3630, opened 2015-11-16
Rawurlencode for non-URL letters in image URLs

toporchillo:
Non-ASCII letters in URLs should be encoded through %-encoding (see RFC). Users with non-ASCII alphabet still use non-ASCII letters and whitespaces in images and image folders. Modern browsers is not take care of wrong image URLs, but image URLs into data feeds are still incorrect. Data feed readers raise image URL error.

### DIFF
--- a/upload/catalog/model/tool/image.php
+++ b/upload/catalog/model/tool/image.php
@@ -34,6 +34,10 @@ class ModelToolImage extends Model {
 			}
 		}
 
+ 
++		$imagepath_parts = explode('/', $new_image);
++		$new_image = implode('/', array_map('rawurlencode', $imagepath_parts));
+
 		if ($this->request->server['HTTPS']) {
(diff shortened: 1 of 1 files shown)


danielkerr:
ok so thats the URL what about the file system! shoudl have url encoded filenames or directories. thx for posting a waster of space pull request that would have caused issues

danielkerr:
"but image URLs into data feeds are still incorrect. "

so convert them at the datafeed level and not directly on the a server folders!

do not post here again unless you know what you are doing because you just mess everything up for every user except your self!

toporchillo:
Have a look again. My patch does not convert image file names. It fixes wrong image URLs.
http://site.ru/image/data/картинка.jpg - is incorrect URL
http://site.ru/image/data/%D0%BA%D0%B0%D1%80%D1%82%D0%B8%D0%BD%D0%BA%D0%B0.jpg - this is correct.

PR #3652, opened 2015-11-21
no need for substr()

pine3ree:
just use token(...)

### DIFF
--- a/upload/catalog/controller/api/order.php
+++ b/upload/catalog/controller/api/order.php
@@ -226,7 +226,7 @@ class ControllerApiOrder extends Controller {
 					foreach ($this->session->data['vouchers'] as $voucher) {
 						$order_data['vouchers'][] = array(
 							'description'      => $voucher['description'],
-							'code'             => substr(token(32), 0, 10),
+							'code'             => token(10),
 							'to_name'          => $voucher['to_name'],
 							'to_email'         => $voucher['to_email'],
 							'from_name'        => $voucher['from_name'],


alancheongkl:
Hi there,
I've error for Fatal error: Call to undefined function token() in /admin/model/user/user.php on line 12.

similar function error for token() at other page, example, register customer.
Do you have any idea which file I miss out ?

pine3ree:
@alancheongkl 
the token() helper method is located in `/system/helper/general.php` (i believed it was added in v 2.1.x).

alancheongkl:
yes, it was added, but still getting the error. possible to advise ?

the coding for the general.php as following 
<?php
function token($length = 32) {
	// Create token to login with
	$string = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
	
	$token = '';
	
	for ($i = 0; $i < $length; $i++) {
		$token .= $string[mt_rand(0, strlen($string) - 1)];
	}	
	
	return $token;
}

the example of the script with error is 

public function editPassword($email, $password) {
		$this->event->trigger('pre.customer.edit.password');

		$this->db->query("UPDATE " . DB_PREFIX . "customer SET salt = '" . $this->db->escape($salt = token(9)) . "', password = '" . $this->db->escape(sha1($salt . sha1($salt . sha1($password)))) . "' WHERE LOWER(email) = '" . $this->db->escape(utf8_strtolower($email)) . "'");

		$this->event->trigger('post.customer.edit.password');
	}

pine3ree:
@alancheongkl 
it's very strange, the file is included in system/startup both for catalog and admin applications.
Just to be sure, temporarily add:
```php
require_once(DIR_SYSTEM . 'helper/general.php');
```
before the class definition in admin/model/user/user.php and see what happens.

alancheongkl:
it temporary solved. any further advise ?

pine3ree:
you should check that system/startup.php is actually loading the file (at the end of startup.php)
```
// Helper
require_once(DIR_SYSTEM . 'helper/general.php'); // check for this
require_once(DIR_SYSTEM . 'helper/utf8.php');
require_once(DIR_SYSTEM . 'helper/json.php');
```
if the line is not present, add it.

If you are using extensions which adds modifications (vqmod style) check if one of them is overriding the startup.php file and excluding the needed line.

The line i suggested to add shouldn't be needed there at all as all the generally required files are loaded during startup.

kind regards

alancheongkl:
Hi there, it solved perfectly. I miss out the line at startup.php.

BTW, do you have any idea for following 

Notice: unserialize(): Error at offset 0 of 40 bytes in /admin/controller/dashboard/activity.php on line 19Warning: vsprintf(): Too few arguments in /admin/controller/dashboard/activity.php on line 19
Notice: unserialize(): Error at offset 0 of 54 bytes in /admin/controller/dashboard/activity.php on line 19
Warning: vsprintf(): Too few arguments in /admin/controller/dashboard/activity.php on line 19

pine3ree:
@alancheongkl 

`vsprintf(string $format , array $args)`
the error is due to the fact that for some reason the number of placeholders in $format is higher than the replacements privided by $args, so you should check why $args is not filled correctly.

If you believe it's an issue with the code open a new issue, but be sure to run tests against the master branch without any external mods.

kind regards

PR #3660, opened 2015-11-24
simplified format method

pine3ree:
Hello Daniel/James,
please verify carefully my changes and thoughts.

1. No need to set $value = $value

2. Using $value both as the conversion parameter name and the soon-to-be formatted value name can be a little confusing. I changed it to $amount (I could have re-used $number as well)

3. If we are not returning a formatted string we could simply return the rounded value without calling number_format.

regards


danielkerr:
ok added

PR #3662, opened 2015-11-24
Made height parameter optional in 'tool/image/resize' method

Linguisto:
Now, if we call "resize" method of "ModelToolImage" class with two parameters (filename and width), the third (height) will be calculated automatically according to the original image ratio.


danielkerr:
thx but not required.

PR #3696, opened 2015-12-03
refactor common code into private method

pine3ree:


danielkerr:
astop messing wityh the tax class\! it works fine! the last time you did something it caused errors!

pine3ree:
Not exactly true,

Last time i just corrected a class variable name and the PR was merged. :-)
About a prevoius PR i explained what happened.  My last commit for some reason didn't reach github, so a variable was missing. That caused the error.

This time i refactored 3 almost identical methods into one private method to be called by the original ones. The 3 original public methods are basically the same method, with only the `based` column value different in the sql string.

Everything can be made better, always. I don't believe to be enough that it just works.....If you don't trust my changes, you could do the refactoring yourself, now that I explained the reason for doing it and if of course you also believe it's worth doing it.

respectfully

PR #3710, opened 2015-12-05
add manufacturer field to product filter in admin panel

savage4pro:
Alt text


savage4pro:
overwritten by pull request #3711

PR #3711, opened 2015-12-05
add category and manufacturer fields to product filter in admin panel

savage4pro:
Alt text

### DIFF
--- a/upload/admin/controller/catalog/product.php
+++ b/upload/admin/controller/catalog/product.php
@@ -46,6 +46,14 @@ class ControllerCatalogProduct extends Controller {
 				$url .= '&filter_status=' . $this->request->get['filter_status'];
 			}
 
+			if (isset($this->request->get['filter_category'])) {
+				$url .= '&filter_category=' . $this->request->get['filter_category'];
+			}
+
+			if (isset($this->request->get['filter_manufacturer'])) {
+				$url .= '&filter_manufacturer=' . $this->request->get['filter_manufacturer'];
+			}
+
 			if (isset($this->request->get['sort'])) {
 				$url .= '&sort=' . $this->request->get['sort'];
 			}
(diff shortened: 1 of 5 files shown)


drkskwlkr:
Hi, I see tremendous value in the feature you have implemented in this branch, and I tried to install it by hand over an OC 2.1.0.1 demo to test it. However I can't get it to work.

I am not very good at this and I apologize in advance for any silly questions.

Am I wrong to assume that OC 2.1.01 can safely get patches from the OpenCart master branch?

Here is what/how I did:
~ I created local copies of all five original files that have to be patched according to the 'Files changed' page
~ I used wget to download each of them to my server using the URLs to the 'Raw' views
~ I then checked by hand whether the downloaded tpl/php files were readable and contained the changes

However I get an error message when I open admin > products:
```
Notice: Error: Could not load template /home/opencart/public_html/admin/view/template/catalog/product_list! in /home/opencart/public_html/system/engine/loader.php on line 86
```

This error can be fixed by reverting to the unpatched version of system/modification/admin/controller/catalog/product.php
but then of course the change would not work.

Please do not take this as a support request – it is almost 100% certain I am doing something wrong. However I will appreciate you letting me know if patching the files the way I did is the right thing to do and my result is something unexpected that needs to be troubleshot.

Thank you!
Ivan

savage4pro:
hi, Ivan!

OpenCart master branch at this moment containts 2.1.0.2_rc version
it has critical differences compairing to 2.1.0.1 in some files
so you shouldn't just overwrite your files or apply all changes

by the way the changes I've done to files in #3711 you can apply to 2.1.0.1 and additional functionality would work well in this version
you can do it by hand savage4pro#1 or get branch https://github.com/savage4pro/opencart/tree/2.1.0.1-patch-1

best regards
Alexander

drkskwlkr:
Thank you for the explanation, Alexander - and doubly so for back-porting to 2.1.0.1.

I did as advised but unfortunately the product.php controller does not seem to pick up the new language strings although they are in the language file, I checked. See:
filters-2101


I also checked to make sure an extension is not patching admin/language/english/product/product.php and caching the result but this is not the case.

I will be scratching my head about this some more, I am in the process of filling a catalog with products and your code will save me lots of time, so if I get it running it will get well tested for sure.

Thank you and all the best
Ivan

savage4pro:
if you use any ocmods try to clear modifications cache

drkskwlkr:
I did do a Refresh (blue button) before I wrote, and it did nothing.
But now I also I did a Clear (orange button) and that did it.
Life is good.

Thank you so much!

savage4pro:
ur welcome

RamyaKrishna1:
Hai sir i'm ramya my problem is when user is clicking any category for ex: rice, if his requirement is for 5kg rice bag it is showing same 1 kg photo for both 1kg & 5kg what i have to do now please leave a comment solve my problem .
[problem (1).PDF]


Thank you,
Ramya krishna.

savage4pro:
hi Ramya!

Please contact the "options" module author to solve your problem.

best regards
Alexander

PR #3713, opened 2015-12-06
there is no sense to use InnoDB

savage4pro:


renothing:
I don't think there is any reason that need keep myisam engine. 
from mysql5.5 the default engine changed innodb, because the its data security much better than myisam.

because it's huge disaster for myisam data recovery.

danielkerr:
both myisam and innodb have different advantages and disadavatanges!

i dont need to change to innodb! the fact that it takes less then a second to change in phpmyadmin means wasting my time!

PR #3730, opened 2015-12-10
No more isset on request

prhost:
Helper to help pick up the values without the need to check the position in the array exists or not, and still you can have a value if there is no default.

Example:

```php
$telephone = $this->request->post('telephone', false);
```

### DIFF
--- a/upload/system/library/request.php
+++ b/upload/system/library/request.php
@@ -28,4 +28,28 @@ class Request {
 
 		return $data;
 	}
+
+	public function post($key, $default = null) {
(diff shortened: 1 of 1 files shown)


OpenCartAddons:
+1 Good Idea

danielkerr:
not a chance. this causes as many problems as issues fixed.

I have explained this now about 3 times.

$this->request->post('telephone', false);

can not do 

$this->request->post['address'][0];

prhost:
but can do so:
`$this->request->post('address')[0]`

this will only help but be checking all the time if the position exists, would only help.

PR #3735, opened 2015-12-11
Change quantity to input type="number"

DaveMcW:


danielkerr:
can not as firefox does not support a lot of the new html 5 form attributes

nuke123:
http://caniuse.com/input-number 
Yes but number is well supported and will revert back to text anyway. It gives a massive UI improvement on mobiles so why not?

DaveMcW:
It works very well with Firefox, even on PC you get a useful up/down arrow.  And of course a much nicer keypad on mobile.

PR #3742, opened 2015-12-16
Added generator meta tag

artem328:


nuke123:
Nobody really needs to know this information, well apart from making it easier for hackers

artem328:
@nuke123, may be you're right. But I think it looks better, more official.

nuke123:
Add it to a humans.txt file under Software

jamesallsup:
This is not a bug, please use the suggestion site.

PR #3744, opened 2015-12-17
Dreamvention

mif32:
We  need this function for our Ajax Quick Checkout.
Please add this to new version.


danielkerr:
there will be away of doing what you need to do without this feature.

PR #3747, opened 2015-12-18
Update header.tpl

newerton:
Searching script only for the header


GomDani:
Don't need, if you look the function

```
public function getScripts($postion = 'header') 
```

header is predefined.

newerton:
But the functions return this:
```php
return $this->scripts[$postion];
```
In catalog/controller/common/header.php(line 34), return;
```php
$data['scripts'] = $this->document->getScripts();
```

Result in the layout is:
```php
array(1) {
  ["header"]=>
  array(2) {
    ["catalog/view/javascript/bootstrap/dist/js/bootstrap.min.js"]=>
    string(58) "catalog/view/javascript/bootstrap/dist/js/bootstrap.min.js"
  }
}
```

newerton:
Sorry!

The vqmod of Nitro Pack, alter the function.

PR #3760, opened 2015-12-28
Update Old MD5 password

amdev:
Update Old MD5 password


danielkerr:
no because there will be people who have upgraded ages ago

PR #3762, opened 2015-12-28
Fix wrong modified controller/model/view being loaded

Bremaweb:
In the modification function !defined('DIR_CATALOG') can cause an admin controller/model/view to be loaded instead of the catalog controller/model/view it should be if (defined('DIR_CATALOG'))

### DIFF
--- a/upload/system/startup.php
+++ b/upload/system/startup.php
@@ -66,7 +66,7 @@ if (isset($_SERVER['HTTPS']) && (($_SERVER['HTTPS'] == 'on') || ($_SERVER['HTTPS
 
 // Modification Override
 function modification($filename) {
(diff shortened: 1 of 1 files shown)


Bremaweb:
But this only works if DIR_CATALOG is not defined in admin/config.php

Bremaweb:
Upon further testing this pull request fixed my front end problem but removing DIR_CATALOG from admin/config.php breaks things on the backend. It seems maybe "if ( DIR_CATALOG != DIR_APPLICATION )" may be a better solution

danielkerr:
DIR_CATALOG is only defined in the admin/config.php

wroks for every one else. check your config.php are setup correctly.

PR #3773, opened 2016-01-04
Custom Directory to File Cache System

eka7a:
Examples
$this->cache->get('key', 'custom_directory');
$this->cache->set('key', 'value', 'custom_directory');
$this->cache->delete('key', 'custom_directory');


danielkerr:
I think this can be done another way. all we want to do is stop people being able to traverse the directories outside of the cache directory.

although if this is the case what about memcache.

i think the key can be directory/file.

so for page cache maybe page/product

PR #3778, opened 2016-01-05
Update ups.php

billynoah:
Add error handling in the event UPS servers return invalid XML (which occasionally does happen).  If XML is invalid return boolean false.


danielkerr:
added your fix manually

billynoah:
Thanks - UPS had a major outage a couple days ago which caused many OpenCart stores to fail checkout since this extension had no way to deal with malformed XML.  You might also consider reducing the timeout to 15 or 20 seconds - 60 seconds seems like a long time to wait for a response but that's a bit more subjective.

PR #3784, opened 2016-01-06
Installer & SSL repairs

dhaupin:
The installer doesn't work right behind SSL, especially if is some kinda rev proxy. Also there is no IP rewrite support for IPs behind rev proxy. Here are a few tweaks and a rework of the switches to have broader compatibility with more headers. What do you think?

### DIFF
--- a/upload/install/controller/install/step_3.php
+++ b/upload/install/controller/install/step_3.php
@@ -15,7 +15,7 @@ class ControllerInstallStep3 extends Controller {
 			$output .= 'define(\'HTTP_SERVER\', \'' . HTTP_OPENCART . '\');' . "\n\n";
 
 			$output .= '// HTTPS' . "\n";
-			$output .= 'define(\'HTTPS_SERVER\', \'' . HTTP_OPENCART . '\');' . "\n\n";
+			$output .= 'define(\'HTTPS_SERVER\', \'' . HTTPS_OPENCART . '\');' . "\n\n";
 
 			$output .= '// DIR' . "\n";
 			$output .= 'define(\'DIR_APPLICATION\', \'' . DIR_OPENCART . 'catalog/\');' . "\n";
--- a/upload/install/controller/install/step_4.php
+++ b/upload/install/controller/install/step_4.php
@@ -58,7 +58,7 @@ class ControllerInstallStep4 extends Controller {
 		
 		curl_setopt($curl, CURLOPT_POST, true);
 		curl_setopt($curl, CURLOPT_HEADER, false);
-		curl_setopt($curl, CURLOPT_URL, 'http://www.opencart.com/index.php?route=extension/json/extensions');
+		curl_setopt($curl, CURLOPT_URL, 'https://www.opencart.com/index.php?route=extension/json/extensions');
 		curl_setopt($curl, CURLOPT_USERAGENT, 'Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.8.1.1) Gecko/20061204 Firefox/2.0.0.1');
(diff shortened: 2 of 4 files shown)


danielkerr:
some pretty bad coding here

danielkerr:
also i think you copied it from here without proper testing

https://3v4l.org/9Ai34

dhaupin:
Read all of this twice before you freak out and start typing before thinking:

I agree about the seemingly unneeded function. I went back and forth with that (function or just loop). Honestly i chose function so other mods can use it before request bootstraps it, or in other parts of the core/index(es), or anything else that only needs to call startup. A couple examples of this: a rework on robots.txt to MIME-mask from php, it needs a protocol switch. Random cron sync mod (vtiger) rolls its own index, again calling startup and registry without loading out everything. There are other uses too such as just checking proto from a less standard port header (ie in a tweaked app server, docker, who knows), but again im down either way, its not a large deal to just loop it....ya know? So as far as "bad coding" goes, besides taking out function call, what would you change?

Now lets move on. Perhaps you didnt take a look at the original? Have you ever tried to install OC under SSL, especially something like Cloudflare with SSL? Go ahead, ill wait. Did you get it? Notice how your config constants are wonked (dual https or dual http?). Now try to run OC SSL mode full. Having troubles? That's expected in an UNTESTED circumstance. Consider this your first test of the SSL system.

I (and many others) feel that its ironic that you would run full SSL (cloudflare) everywhere on OC basecamp yet not even build in basic support in the project? Seriously, a 1 liner looking for HTTPS only to trigger an SSL fwrite that is supposed to dynamically write a protocol? Is that a joke line or are you for real in love with that? You and I know you wont get an accurate `$protocol` from that. Further on, did you notice that there isnt even a way to dyn fwrite in that https://? That sir, in todays day and age, seems totally untested and much like bad coding when compared to the stuff i am showing you that **actually works**.

But thats just installer index writing into a config, not like thats super important or anything right. Its only allowing you to install the cart and setting critical protocol based constants that drive untold amounts of the cart later, no biggie right? So lets talk about this in startup:

	// Check if SSL
	if ((isset($_SERVER['HTTPS']) && (($_SERVER['HTTPS'] == 'on') || ($_SERVER['HTTPS'] == '1'))) || $_SERVER['SERVER_PORT'] == 443) {
		$_SERVER['HTTPS'] = true;
	} elseif (!empty($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' || !empty($_SERVER['HTTP_X_FORWARDED_SSL']) && $_SERVER['HTTP_X_FORWARDED_SSL'] == 'on') {
		$_SERVER['HTTPS'] = true;
	} else {
		$_SERVER['HTTPS'] = false;
	}

 - What happens when HTTP_X_FORWARDED_SSL is set to 'https'?
 - What happens when `HTTP_X_FORWARDED_PROTO` is set to 'on'?
 - What happens if the revprox uses  'HTTP_X_FORWARDED_PROTOCOL'?
 - What happens if your server runs no cert on origin and just a cert on front?
 - What happens in a variety of less common setups that you obviously have never encountered?

But wait, we are totally forgetting about IP support, which as you know (since you RUN CLOUDFLARE ON BASECAMP) is a super easy fix to rewrite. I dont think we need to talk about that, its completely obvious. Unless of course you have some grand idea about how we dont need that either.

https://3v4l.org/9Ai34 <--- PS: OMFG dude you are soooooo pro. Call me out for copying? **THAT IS MY SNIPPET** that i started one eve messing around, its not done, not tested, and not the same as what i provided you. Yes indeed it doesnt pass tests because.......its not completed there lol. Im not trying to mess with you or your cart, im an ally, calm your ass down.

PR #3788, opened 2016-01-07
Allows menu to show just options where user has access

mahomedalid:
This way we give to the user a cleaner interface of he has access (read or modify). Notice I didn't change all links is just a proof of concept for master due the menu changed several items in branches 2.1.0.x


danielkerr:
i will added something like this myself later.

mahomedalid:
Don't you wanna help?

PR #3802, opened 2016-01-11
allow passing extra $data via render method

pine3ree:
Hello Daniel, James. This is a very common pattern for template renderers. Think it could be useful for opencart?

what about renaming adaptor to renderer in this case?
and maybe Basic to PhpRenderer?

in the code i do not modify (merge with extra $data) the instance->data, i think this should be kept as set via the set() method and only a merged array should be sent to the extract method.


danielkerr:
not required

PR #3817, opened 2016-01-15
Fix for db wrapper not passing the $params argument to the mpdo driver…

hisorange:
The PDO driver expects the $params argument but since the wrapper only passing the $sql argument makes it generaly useless, this PR does not affect any other driver's behavior.


danielkerr:
cheers

PR #3820, opened 2016-01-17
Fix possible issue in ebay_product.php

OldCoder:
There are three occurrences of the string 'now()' in the file that should probably read now() instead as 'now()' is a literal string and not the function call that the original author intended.


jamesallsup:
Thanks.

OldCoder:
You're welcome. I'm doing some minor work related to OpenCart and will 
Pull-Request again if I notice other issues.

PR #3839, opened 2016-01-20
Dev

shaunfreeman:
Using Sha1 for passwords are now considered very insecure and we should not be using them at all.
I have updated the password system to allow for the new PHP password functions which use the best algorithms for passwords. 
I have also tried to keep it backwards compatible as possible.


shaunfreeman:
Is there a reason you closed this without commenting.

jamesallsup:
Read the readme, under "Making a suggestion"

DanielRuf:
Uservoice is not for suggesting code changes.

GitHub issues are not just meant for bugs. Especially PRs.

You clearly missed the point when it comes to GitHub issues and PRs. How would you propose code changes with a diff on Uservoice?

jamesallsup:
@DanielRuf Considering I wrote the readme for OpenCart and set-up the suggestions site, I think I know exactly where WE want suggestions and bugs reported. If you took the time to actually read the readme you will realise that - but as with most people who just jump straight in and create issues/bugs and random pull requests we just waste further time having to explain this over and over; hence, close the issue. Just because you might do things one way, doesn't mean we do - at least we spent the time explaining how and where we would like to organise things. If people choose not to read that then that isn't really our fault now is it? 

If you find a bug, great! Report it. If you find a bug and choose to do a PR for it, even better! But that doesn't mean we're going to start accepting pull requests from random people who want to change things without it being discussed by the OpenCart team.

The current system that this pull request was made for already works, if it needs improvement then that is a suggestion that needs to be put forward - it is not a bug. We then discuss it internally and add to a release.

PR #3840, opened 2016-01-20
Propper order history ordering

toporchillo:
When you add several order history (f.a. via api) it could be with a same date_added value. And OpenCart display such order history entries in a wrong order.


danielkerr:
shoudl be date desc

in fact this whole commit is a waste

PR #3841, opened 2016-01-20
fix credit_status to be reward_status

qphoria:
incorrectly checks "credit_status" instead of "reward_status" in the reward total


jamesallsup:
Thanks @qphoria but dev branch isn't in use yet, looks like this has already been updated in the master by dan.

I'll take a look over any pr's that have been done into dev and merged and put in master correctly. After 2.2 there will be dev and dev feature branches in use.

PR #3859, opened 2016-01-25
Update the default template image to opencart version 2

ocextension:


ocextension:
please ignore the search.tpl. it was from my patch-1

PR #3862, opened 2016-01-25
Update several icon for better view for dashboard area in backend

ocextension:
World Map -using fa-globe instead of fa-eye
People Online - using fa-users instead of fa-eye


OpenCartAddons:
+1

PR #3863, opened 2016-01-25
Calculation error

mikehins:
``admin/index.php?route=report/product_purchased``

Right now in the model ``admin/report/product.php``, the ``getPurchased()`` method calculate the total by multiplying the

``(total + tax) * the quantity``

it should be 

``(price of the product + tax) * the quantity``


mikehins:
`SUM((op.price + op.tax) * op.quantity)`
not
`SUM(op.price + (op.tax * op.quantity))`

PR #3856, opened 2016-01-24
Remove Alert

valdeir2000:

PR #3867, opened 2016-01-25
remove alert inside summernote

qahar:
Fix js alert wherever summernote used


ocextension:
https://github.com/opencart/opencart/pull/3856 <- yeah, the guy just post about this. but as danielkerr said, it may be not a full fix

qahar:
ok

PR #3888, opened 2016-01-28
Added Warning: You added before the language!

eka7a:


upload/admin/language/en-gb/localisation/language.php
```diff
@@ -25,6 +25,7 @@
 
 // Error
 $_['error_permission']  = 'Warning: You do not have permission to modify languages!';
+$_['error_exists']      = 'Warning: You added before the language!';
```
greencopper: I am testing OpenCart for the first time, trying to add the Danish language pack, and I got this error "You added before the language", please consider changing that to something helpful and meaningful. What does it mean? Added before the language, added what, where?

Kind regards

danielkerr:
U can only use the code once. If u already set the code folder in another
language entry then just use that one.

upload/admin/language/en-gb/localisation/language.php
filipmilasinovic: It should say: "You are trying to add language before installing language extension."

PR #3904, opened 2016-02-02
Skip module if not return

danijelGombac:
If module not return, empty column left or right is visible.


qahar:
Another reason for this issue is also because sidebar file have empty new line, which is return string " " (single space)

PR #3911, opened 2016-02-03
Dynamicly add admin main menu through Event

qahar:
Using Events with current admin common/menu code require to outputing custom template in order adding menu. This can raise issue when using multiple Event to add menu; also possibly issue with extension use OcMod or vQmod.

In order to add admin main menu dynamicly through Event with minimum conflict issue, common/menu need to be refactored.

Example how we can use Events to add new menu

    $data['navigations']['catalog']['child'][] = array(
        'title'     => 'Hello Extension',
        'url'       => $this->url->link('extension/helloworld', 'token=' . $this->session->data['token'], true),
    );

Also possible to control menu visibility

    if ($this->user->hasPermission('access', 'extension/helloworld')) {
        $data['navigations']['catalog']['child'][] = array(
            'title'     => 'Hello Extension',
            'url'       => $this->url->link('extension/helloworld', 'token=' . $this->session->data['token'], true),
        );
    }

I think I'm going to check hasPermission to all main menu item once this proposal is accepted

### DIFF
--- a/upload/admin/controller/common/menu.php
+++ b/upload/admin/controller/common/menu.php
@@ -3,211 +3,510 @@ class ControllerCommonMenu extends Controller {
 	public function index() {
 		$this->load->language('common/menu');
 
-		$data['text_analytics'] = $this->language->get('text_analytics');
-		$data['text_affiliate'] = $this->language->get('text_affiliate');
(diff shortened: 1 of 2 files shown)


valdeir2000:
Good

danielkerr:
ok just added this using a different way

PR #3917, opened 2016-02-06
Fix startup/event pre_action order

qahar:
Event on front site is not trigered because it's registered after the _startup/seo_url_.
Both _startup/seo_url_ and _maintenance_ have side effect by returning new action.

After change the sequence, Events is working as expected.

Note: the reason _action_event_ in config/catalog.php working because it's registered before the _pre_action_ on framework.php


victorwbuzz:
With that fix, friendly urls not triggering events anymore.

PR #3922, opened 2016-02-08
Added customer language.

eka7a:
Customer login on store to change language automatically.


eka7a:
@nuke123 not required for checkout/register or account/register
When customer register, it will be registered with current language.

PR #3928, opened 2016-02-10
Update cart products with SEO url for checkout/cart

dennisbuijsgmailcom:
When using SEO urls for checkout/cart there is a problem with refreshing the cart page.

When you add a product to the cart and want to remove this product, the cart page must be refreshed. When using SEO url this is not hapening, because it checks the URL with -> getURLVar('route') == 'checkout/cart' 

This change must be in combination with a change in common.js in https://github.com/opencart/opencart/pull/3927 .

### DIFF
--- a/upload/catalog/controller/checkout/cart.php
+++ b/upload/catalog/controller/checkout/cart.php
@@ -421,6 +421,8 @@ class ControllerCheckoutCart extends Controller {
 			unset($this->session->data['payment_method']);
 			unset($this->session->data['payment_methods']);
 			unset($this->session->data['reward']);
+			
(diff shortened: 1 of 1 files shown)


danielkerr:
seo urls do not work on the cart page. the system was designed this way.

dennisbuijsgmailcom:
I don't understand why, a lot of people want seo urls and not a few urls but all urls. Now some urls are unchangable because this issue, and this fix is not harm normal functionality, maybe check the redirect url between the if statement is even better. 

With vqmod i can change a lot but i can't change javascript files.

Please reconsider!

Issue #3793, opened 2016-01-08
[2.1.0.1] Ocmod bug for error="skip" attribute


gob33:
Suppose i have an ocmod with 5 operations on file like that:
```
<operation>...1-blablabla...</operation>
<operation>...2-blablabla...</operation>
<operation error="skip">...3-blablabla...</operation>
<operation>...4-blablabla...</operation>
<operation>...5-blablabla...</operation>
```
then when refreshing in administration, operations 3,4 and 5 are not done. All after "skip" is ignored.
Because in OPENCART-2.1.0.1\upload\admin\controller\extension\modification.php, we have at line 664
```
	// Skip current operation
	if ($error == 'skip') {
		break;
	}
```
The break ends the loop.
Think it should be at least (moved after "abort' case):
```
	if (!$status) {
		// Log
		$log[] = 'NOT FOUND!';

		// Abort applying this modification completely.
		if ($error == 'abort') {
				$modification = $recovery;
				// Log
				$log[] = 'ABORTING!';
				break 5;
		}

		// Skip current operation
		if ($error == 'skip') continue;
		else break;
	}
```
for the loop on <operation> to be executed well.

gob33:
Allready signaled in #2972.

deepvision7:
Yes it is very frustrating bug making it unable to use skip at all.

nuke123:
Yes it is defintely similar bug I opened https://github.com/opencart/opencart/issues/3532

PR #3950, opened 2016-02-17
Correction for error="skip" attribute in ocmods

gob33:
Correction for #3793. <operation error="skip"> then will behave like in vqmod and will skip if explicitely written.


danielkerr:
if this break loads of stuff its on your head!

PR #3962, opened 2016-02-19
Admin templates backward compability

### LINKED ISSUE #3951: OC2.2 $this->load->view(<TEMPLATE>) without .tpl extension
Hi

I noticed that from OC2.2 it is required to remove the extension .tpl from `$this->load->view()` this will force developers to update "all" modules that call a template.

Also this change is not backwards compatible with OC2.1 so either two separate module versions or a conditional variable with the ".tpl" extension will be needed from OC2.2

Is there a reason for this change?

Many thanks!

qahar:
Fix #3951


danielkerr:
thx for this one!

PR #3981, opened 2016-02-22
Update worldpay.php - fix total sent to payment processor

chrisdempsey:
If a user changes the store currency display, adds items to cart, checks out and pays with Worldpay Online the store sends the incorrect order total to the payment processor.

Actual data sent:
- order amount - eg. 10 GBP (incorrect)
- currency - user selected eg. EUR (correct)

What it should be:
- order amount - cart default currency converted to user selected currency eg. 10 GBP becomes 11.29 EUR

This results in the user's card being charged 10 EUR for what should be a 11.29 EUR sale.


danielkerr:
thx

PR #3997, opened 2016-02-24
Update modification

ocextension:
update wrong sort by version

### DIFF
--- a/upload/admin/controller/extension/modification.php
+++ b/upload/admin/controller/extension/modification.php
@@ -736,7 +736,7 @@ class ControllerExtensionModification extends Controller {
 
 		$data['sort_name'] = $this->url->link('extension/modification', 'token=' . $this->session->data['token'] . '&sort=name' . $url, true);
 		$data['sort_author'] = $this->url->link('extension/modification', 'token=' . $this->session->data['token'] . '&sort=author' . $url, true);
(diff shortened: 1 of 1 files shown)


danielkerr:
this peice of code is crap! and removes the functionality of switching routes!!!!!!!

`-	protected function callback($registry, $route) {
 -		return function($args) use($registry, &$route) {			
 +	protected function callback($registry, $route, $class_instance) {
 +		return function($args) use($registry, &$route, $class_instance) {			
  			// Trigger the pre events
  			$result = $registry->get('event')->trigger('model/' . $route . '/before', array_merge(array(&$route), $args));
  			
  			if ($result) {
  				return $result;
  			}
  			
 -			$file = DIR_APPLICATION . 'model/' .  substr($route, 0, strrpos($route, '/')) . '.php';
 -			$class = 'Model' . preg_replace('/[^a-zA-Z0-9]/', '', substr($route, 0, strrpos($route, '/')));
  			$method = substr($route, strrpos($route, '/') + 1);
 -	
 -			if (is_file($file)) {
 -				include_once($file);
 -			
 -				$model = new $class($registry);
 -			} else {
 -				throw new \Exception('Error: Could not load model ' . substr($route, 0, strrpos($route, '/')) . '!');
 -			}`




next time think before you so something like this!

PR #4000, opened 2016-02-24
[2.2.0.0.b1] locale is missing in cache

gob33:
If 'directory' has been replaced by 'code', we always need 'locale' which is not in the cache.

### DIFF
--- a/upload/admin/model/localisation/language.php
+++ b/upload/admin/model/localisation/language.php
@@ -325,6 +325,8 @@ class ModelLocalisationLanguage extends Model {
 						'name'        => $result['name'],
 						'code'        => $result['code'],
 						'locale'      => $result['locale'],
+						'image'       => $result['image'],
+						'directory'   => $result['directory'],
 						'sort_order'  => $result['sort_order'],
 						'status'      => $result['status']
 					);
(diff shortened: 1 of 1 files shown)


eka7a:
@gob33 removed local column in language table on opencart v2.2.x

gob33:
I dont understand.
At install i read
```
DROP TABLE IF EXISTS `oc_language`;
CREATE TABLE `oc_language` (
  `language_id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(32) NOT NULL,
  `code` varchar(5) NOT NULL,
  `locale` varchar(255) NOT NULL,
  `image` varchar(64) NOT NULL,
  `directory` varchar(32) NOT NULL,
  `sort_order` int(3) NOT NULL DEFAULT '0',
  `status` tinyint(1) NOT NULL,
  PRIMARY KEY (`language_id`),
  KEY `name` (`name`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;

--
-- Dumping data for table `oc_language`
--
INSERT INTO `oc_language` (`language_id`, `name`, `code`, `locale`, `image`, `directory`, `sort_order`, `status`) VALUES
(1, 'English', 'en-gb', 'en_US.UTF-8,en_US,en-gb,english', 'gb.png', 'english', 1, 1);
```
So after reflexion,  the cache should have the same fields as we were reading in the DB, i.e. ALL the fields. You cant have a missing difference if datas come from the cache versus the DB.
You must revert getLanguages() as it was in OC2.1.0.2

PR #4004, opened 2016-02-25
.htaccess.txt: Comment out RewriteBase line.

BenUniqcode:
This line is not required for most installations, and its presence breaks SEO URLs when OpenCart is installed in a subdir of the document root, which is a far more common occurrence than the rare situations in which it's actually necessary to set RewriteBase. I therefore propose it should be commented out by default.

### DIFF
--- a/upload/.htaccess.txt
+++ b/upload/.htaccess.txt
@@ -17,9 +17,11 @@ Options -Indexes
 
 # SEO URL Settings
 RewriteEngine On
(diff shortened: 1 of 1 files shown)


BenUniqcode:
Hi Daniel
I'm new here so apologies if I've done something wrong - can you explain why you've closed this or if I'm going about contributing the wrong way then point me in the right direction please? Thanks.

PR #4005, opened 2016-02-25
[2.2.0.0.b1] History lines ordered ascending

gob33:
Suggest reverting lines order to ASC (same as OC1.5)
This is a more natural way to read the lines from top (old ones) to bottom (last ones).
And when adding new history lines, prevent srolling the page to see what just has been added.
Note that returns were always ASC.


danielkerr:
ill ad your changes after the next release

PR #4012, opened 2016-02-26
Fixed session assignment

sajalsoni:
Tried to do fix so that subsequent API calls have access to proper session token. Without this fix, it always gives "permission denied" error even though login token was successfully created.


sajalsoni:
Hi Danielkerr,

I would appreciate your comment on this one. Is there any issue with the API really Or there's a mistake on my end.

PR #4058, opened 2016-03-05
Update header.php

PhamToan:
Fixed error Access-Control-Allow-Origin with ajax action.

How to the error caused.

In config.php file we set: 
define('HTTP_SERVER', 'http://opencarthost.com/');

But we acces to http://www.opencarthost.com/

it will show error like below (we can't run ajax)
XMLHttpRequest cannot load http://opencarthost.com/index.php?route=checkout/cart/add. Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://www.opencarthost.com' is therefore not allowed access.

### DIFF
--- a/upload/catalog/controller/common/header.php
+++ b/upload/catalog/controller/common/header.php
@@ -19,6 +19,12 @@ class ControllerCommonHeader extends Controller {
 		} else {
 			$server = $this->config->get('config_url');
 		}
+		//fixed error add to cart acction(and all ajax Action) between www and non www 
+		$server_url_info = parse_url($server);
+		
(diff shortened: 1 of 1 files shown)


danielkerr:
se the correct HTTP_SERVER then!!!

PhamToan:
It showed error @danielkerr . 

You can check at http://www.opencarthost.com/ (can't add cart) and http://opencarthost.com/ (can add cart)

danielkerr:
u need to put a redirect in your htacess to redirect from non wwww to www! this is web programming 101!

PR #4100, opened 2016-03-11
Add a UNIQUE Key for prodict_id and image

Rpent001:
Adding this UNIQUE key prevents duplicates from being added to a product's additional images list.

ALTER TABLE `oc_product_image` ADD UNIQUE `product_id_image` (`product_id`, `image`);


Rpent001:
Are duplicate images in the same product being handled in another way?

PR #4101, opened 2016-03-11
Remove Formatting from Descriptions When Creating Google Base Feed

Rpent001:
Google limits the descriptions field to 5,000 characters and, last time I wrote a product export, they went out of their way to say HTML formatting was NOT allowed.

Personally, I like to jazz up my descriptions with bold lettering, images and other formatting.  strip_tags() removes all of that, plus any other tags left by the WYSIWYG editor, during the creation of the feed to keep it compliant.


danielkerr:
it should have been

$output .= '<description><![CDATA[' . strip_tags(html_entity_decode($product['description'], ENT_QUOTES, 'UTF-8')) . ']]></description>';

next time dont do a pull request unless you have tested the code!!!!!!!!!!

Rpent001:
The data I was using to test didn't contain any encoding like &gt; because I merged from another system that didn't use the same WYSIWYG editor so I missed that.

The way you suggested it be written converted &nbsp; into Å's for some reason though.

PR #4105, opened 2016-03-12
Use Lower Price of Specials and Discounts

Rpent001:
If you have a Special Price for a product it completely ignores the Discount Price

Example:
Base Price is $10/ea
Special Price is $8/ea
Buy 4 or More is $6/ea

The special price of $8 overrides the $6 lower quantity discount because of the order of execution.  This addition of code will choose the lower price of the two and give that price.

### DIFF
--- a/upload/system/library/cart/cart.php
+++ b/upload/system/library/cart/cart.php
@@ -187,6 +187,15 @@ class Cart {
 					$price = $product_special_query->row['price'];
 				}
 
+				// Use Lower Price of Specials and Discounts
+				if ($product_special_query->num_rows && $product_discount_query->num_rows) {
+				    if ($product_special_query->row['price'] < $product_discount_query->row['price']) {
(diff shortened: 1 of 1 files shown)


Rpent001:
@danielkerr This is the pull I was referring to that is an actual problem with the functionality of OpenCart.  Both methods of pricing should be able to coexist.  The store owner shouldn't have to pick one over another when they might want to have a special on a product with tiered pricing.  The lower price of the two should prevail after the values are determined.

Maybe add a setting called "Use Lower Price of Specials and Discounts" where the store owner can decide how to handle it and wrap the code i proposed in another if statement, but the way the code is currently written is incomplete.  You are overwriting the variable because it executes in order and obviously you did not test the system with both pricing methods in place, so it seemed to be working just fine when you gave it to go ahead.

Take a look.  You'll see what I mean.

PR #4108, opened 2016-03-13
Search Anywhere In Name or Model

Rpent001:
This makes searching within a product name or model easier and doesn't require you to add on the % wildcard to the beginning of your search term when you don't know the exact way a product is named.  When you have 50,000+ products in a catalog you tend to only remember parts of their names.

It still works as it did if you begin typing the full product name.


Rpent001:
Is there any specific reason as to why you only use the % at the end of a string and not encapsulate a search term in them?  It seems kind of obvious to do the latter so users can get optimal results.

PR #4109, opened 2016-03-13
Site Search by Partial SKU, Model, UPC, etc.

Rpent001:
Allows for customers searching on the site to only know partial SKU's, Models, UPC's, etc.

This is useful when you may have your Model set to something similar to the company SKU, but not exactly.  It also doesn't assume that a customer knows exactly what they are looking for all of the time.

### DIFF
--- a/upload/catalog/model/catalog/product.php
+++ b/upload/catalog/model/catalog/product.php
@@ -127,13 +127,13 @@ class ModelCatalogProduct extends Model {
 			}
 
 			if (!empty($data['filter_name'])) {
-				$sql .= " OR LCASE(p.model) = '" . $this->db->escape(utf8_strtolower($data['filter_name'])) . "'";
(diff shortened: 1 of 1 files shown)


danielkerr:
maybe if the wild card was at the end.

Rpent001:
Assuming that every customer knows the beginning of everything, that might actually work 100% of the time.  In a real world situation, the wildcards should be surrounding the search terms so that the maximum number of possible results will show up.  Not zero results like most of the cases I tested.

Rpent001:
Here's a real world example for you:

Let's assume a t-shirt shop is running on OpenCart.  They have multiple logo designs on the same 979 style shirts.  The skus all contain "-979-" to indicate the shirt style on the back end.  DESIGN1-979-L, DESIGN2-979-XL.  I am a customer and realize this and want to search for other designs only in the 979 style, so I do a search for "979".  The way it is right now that will show zero results.  With my modification it will show all of the skus containing "979".   They way you say it should be done would also show zero results because the sku doesn't begin with "979".  

Follow what I'm saying here?  Be the customer, not the developer.

Issue #4074, opened 2016-03-07
OC 2.2.0.0 Wrong event unregister


mhcwebdesign:
This code in system/engine/event.php looks strange:

```
	public function unregister($trigger, $action) {
		if (isset($this->data[$trigger])) {
			unset($this->data[$trigger]);
		}
	}
```

This would remove all actions for a given trigger. I think it should only remove it for a certain action.

danielkerr:
already fixed not sure why the issue is still open

PR #4139, opened 2016-03-16
Allow unregister of single event actions

skyosev:
Currently the event implementation removes all of the actions for a given event trigger. Related to #4074

This pull request makes possible to unregister a certain action only. It also adds typehint for the action argument to allow registering of Action class instances only. The current possibility to unregister the entire trigger has been kept as the action argument is optional.

The $data class variable has been made protected in order to prevent unneeded modifications through direct access (they should happen via register/unregister methods only). A getData() method has been added if someone wants to inspect the events though.


danielkerr:
was gong to add somethign like this myself

PR #4144, opened 2016-03-17
Redundant variable removal

skyosev:
`&$data` is not needed unless I'm missing something.


danielkerr:
yes you missed something!

PR #4148, opened 2016-03-19
Update product.tpl

newerton:


danielkerr:
advantage?

newerton:
It is no advantage, is correction!

See that starts with '$i = 0;' has a 'for()' within the 'foreach()', in the end if will never follow the sequence of '$i = 0'.

Sorry my English!

PR #4169, opened 2016-03-28
Change config variables to fix full url paths

shoeyn:
Change unused $_['site_base'] and $_['site_ssl'] to $_['config_url'] and
$_['config_ssl'] which is used throughout the project. Fixes full url
path not being called in on SSL, which consequently fixes issue with
always-on SSL with SEO_links not working.

### DIFF
--- a/upload/.htaccess.txt
+++ b/upload/.htaccess.txt
@@ -19,6 +19,10 @@ Options -Indexes
 RewriteEngine On
 # If your opencart installation does not run on the main web folder make sure you folder it does run in ie. / becomes /shop/
 
+# Enable always-on SSL
+# RewriteCond %{SERVER_PORT} 80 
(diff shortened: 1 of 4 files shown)


danielkerr:
dont posty here again!!

you have no right to put 

+# Enable always-on SSL
 +# RewriteCond %{SERVER_PORT} 80 
 +# RewriteRule ^(.*)$ https://www.yourdomain.com/$1 [R,L]
 +

because all that will happen is I will get a load of complaints from people who dont have SSL!!~!~!!

shoeyn:
That wasn't the point of the fix, that was just left in because it was something I used, and I still just commented it out with an explanation of what it is. The variables in the config file is what is important because they are actually redundant and the variables needed aren't there.

shoeyn:
I can remove the htaccess change and just leave the fix in. No need to get aggressive.

PR #4218, opened 2016-04-08
fix typo, remove redundant stat call and more...

pine3ree:
- typo in getRoute()
- file existence is already checked in the constructor, if file exists the $route property is set, so we just need to check for $this->route in execute()

this also pointed out a contradiction in the code (please Daniel verify)

1. If file exists and is file => $route is set
2. In the execute() method if file does not exists  $this->route (=== null) is used in the exception message: i added a validation test at the beginning of execute for $this->route, using the input route (stored as the action $id property) in the error message

kind regards

### DIFF
--- a/upload/system/engine/action.php
+++ b/upload/system/engine/action.php
@@ -1,9 +1,12 @@
 <?php
 class Action {
+	private $id;
 	private $route;
 	private $method = 'index';
 
 	public function __construct($route) {
+		$this->id = $route;
+		
 		$parts = explode('/', preg_replace('/[^a-zA-Z0-9_\/]/', '', (string)$route));
 
 		// Break apart the route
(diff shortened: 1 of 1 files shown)


danielkerr:
why use 

null === $this->route

when

!$this->route

does the same thing!

pine3ree:
@danielkerr 
Hello Daniel,
it's an habit of mine. I always prefer to use strick type comparison (and it's also a bit faster). In this case I also use `null` value to highlight the fact that the property has not been assigne a value. 
You can change it if you wish.

btw,
i still some prefer some traits of the older (v 1.5) Action class architecture, where the file (path) was cached in a property and the route (full route) was used as the identifier for the action.

kind regards
maks

Issue #4246, opened 2016-04-14
Opencart Order edit from admin shows product in cart in front end.


sureshmg23:
What if the admin just wants to edit the order status from pending to "Completed" ? In this situation also the product shown in customer cart? how to stop this when editing an order from admin? what is the use of this functionality, when the customer has already made the purchase? 
Customers are calling me, and asking me to stop this, what should i do to stop this activity to the customer cart?

hutbert:
We are seeing the same issue. We have a temporary fix by commenting out line 19 in system/library/cart.php. When an admin edits an order - additional rows are added to the cart table for that customer. When the customer returns to the site - the items in the temporary cart from the admin edit get pulled into the current customer session and end up in their cart.  Removing line 19 will cause items in the customers cart to drop off if the customer returns in a future session but we've decided that is probably better than having unwanted items show up in their cart.

OpenCartAddons:
@sureshmg23 This won't resolve your issue regarding editing the customer's phone number, but when updating the order status you should always VIEW the order and add a new order history.  This allows you to keep a history of the order statuses and you can also choose to notify the customer of the order status update.

Cheers,
Joel.

sureshmg23:
@OpenCartAddons  Thanks for the quick reply. Then how can i edit the phone number alone for that order without adding the product to the cart?

OpenCartAddons:
@sureshmg23 I haven't looked into the code to see exactly what it's doing, but you could try @hutbert solution.

sureshmg23:
@hutbert  Thank you. Removing line 19 from system/library/cart.php shows "Undefined variable: cart_query..". Am i commenting the correct line?

sureshmg23:
Below code comes between line 16 to 29:
if ($this->customer->getId()) {
			// We want to change the session ID on all the old items in the customers cart
			$this->db->query("UPDATE " . DB_PREFIX . "cart SET session_id = '" . $this->db->escape($this->session->getId()) . "' WHERE customer_id = '" . (int)$this->customer->getId() . "'");

			// Once the customer is logged in we want to update the customer ID on all items he has
			$cart_query = $this->db->query("SELECT * FROM " . DB_PREFIX . "cart WHERE customer_id = '0' AND session_id = '" . $this->db->escape($this->session->getId()) . "'");

			foreach ($cart_query->rows as $cart) {
[...]

PR #4247, opened 2016-04-14
adjust to add session source

hutbert:
This request addresses #4246. 

I believe the issue here is that when an admin user edits or creates and order for a customer the products in that order get added to the cart table. When the customer logs in and their cart object is initialized it captures all of the items in the cart table for that customer session. All items in the cart the admin is editing will appear in the customers session unexpectedly. 

To reproduce: 

1 - add or edit an order for a registered customer in the admin dashboard
2 - in a new browser session log into the catalog view as that customer 
3 - all of the items from the order edited by the admin will show up in the customer cart without the customer having added them. 

The pull request adds a field to the cart table for a session source which will be either "api" or "catalog". Cart items generated via an API call with an api_id in the session will be flagged as api generated cart items and will not be collected into the customer cart in a regular catalog session.

### DIFF
--- a/upload/install/opencart.sql
+++ b/upload/install/opencart.sql
@@ -356,6 +356,7 @@ CREATE TABLE `oc_cart` (
   `cart_id` int(11) UNSIGNED NOT NULL AUTO_INCREMENT,
   `customer_id` int(11) NOT NULL,
   `session_id` varchar(32) NOT NULL,
+  `session_source` varchar(32) NOT NULL,
   `product_id` int(11) NOT NULL,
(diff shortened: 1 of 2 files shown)


hutbert:
Looks like I may have the database adjustment in the wrong place - please let me know if it would be helpful to remove my change from opencart.sql and add  install/model/upgrade/1008.php instead

danielkerr:
not the correct way of doing things!

hutbert:
Any pointers on where to find the "correct way" would be appreciated. 
Thanks

PR #4272, opened 2016-04-20
fix `typo`

pine3ree:


danijelGombac:
There is also need to move 

$method = substr($route, strrpos($route, '/') + 1);

before

if (!isset($model[$route])) {}.

pine3ree:
@danijelGombac 
actually $method is not used inside that if block, it should be moved before
```php
if (method_exists($model[$route], $method)) {
```

danijelGombac:
Yes, you are right.

PR #4277, opened 2016-04-21
Fix Undefined variable: method

danijelGombac:


elcopdoo:
I can confirm this fix...

PR #4279, opened 2016-04-21
check model method visibility

pine3ree:
using is_callable as method_exists also includes private methods


danielkerr:
good point

PR #4287, opened 2016-04-24
Add a really simple redis adapter with the predis library.

mattythebatty:
Also add a suggestion for predis/predis to composer.json


danielkerr:
not required. also i dont like using 3rd party tools unless i have to.

PR #4291, opened 2016-04-25
explicitly hoist scope var declarations

pine3ree:
- JS will hoist var declaration so it's better to put at the beginning of scope. It's also better to define local scope variables (ex the html string)
- also declare loop iteration  variables (when all used browsers will be compatibile with emca 6 we will be able to use `let` for the loop block scope)

Please Daniel verify:
- category has string keys so it's better to define it as a simpler standard object (we are using for/in not using Array.length() in a std for loop)
- we are using category names as keys so we can simplify the category object: each key will represent an array of items related to the category represented by the key itself.

```js
// category will be something like 
{
    'Category 1': [
        {
            label: 'item 1-1',
            value: 11,
        },
        {
            label: 'item 1-2',
            value: 12,
        },
        //...
    ],
    'Category 2': [...] // category 2 items
}

### DIFF
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -289,7 +289,7 @@ $(document).ready(function() {
 			this.click = function(event) {
 				event.preventDefault();
 
-				value = $(event.target).parent().attr('data-value');
+				var value = $(event.target).parent().attr('data-value');
 
 				if (value && this.items[value]) {
 					this.select(this.items[value]);
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -331,17 +331,14 @@ $(document).ready(function() {
 
 				if (json.length) {
 					for (i = 0; i < json.length; i++) {
+						// update element items
 						this.items[json[i]['value']] = json[i];
-					}
 
-					for (i = 0; i < json.length; i++) {
 						if (!json[i]['category']) {
+							// ungrouped items
 							html += '<li data-value="' + json[i]['value'] + '"><a href="#">' + json[i]['label'] + '</a></li>';
-						}
-					}
-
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by veksen** [comment]
jsperf is down at the moment, but keep in mind that a regular `for` loop is over 10 times faster than a `for..in` loop

**Comment by pine3ree** [comment]
@veksen 
Yes, but i didn't change the type of loop. I just merged 3 loops into 1 because they are the same loop.
The same about the category loop. I didn't change the type.

**Comment by pine3ree** [comment]
@veksen 
In the original code:
```js
category[json[i]['category']] = {
    name: json[i]['category'],
    item: []
};
```

the inner object is redundant, 

json[i]['category'] represents the category name, when you loop
```
for (i in category) {
```

`i` is a property that represents  `json[i]['category']` the `i` is confusing because it has different meaning

kimhoo:
@pine3ree Boyscout rule: always leave the camp site clean. In other words: if you saw it, fix (or improve) it.

pine3ree:
@kimhoo Yes, basically for the category object

```js
var category = [];

// original
for (i = 0; i < json.length; i++) {
	if (json[i]['category']) {
		if (!category[json[i]['category']]) {
			category[json[i]['category']] = {
				name: json[i]['category'],
				item: []
			};
    	}

		category[json[i]['category']]['item'].push(json[i]);
	}
}

// simplified
for (i = 0; i < json.length; i++) {
	if (json[i]['category']) {
		name = json[i]['category'];
		if (!category[name]) {
			category[name] = [];
		}

		category[name].push(json[i]);
	}
}
```

no need for inner objects, and clearer code using name, but of course Daniel will decide at the end.

pine3ree:
..and of course 

```js
var category = {}
```
in the original code we are creating `category` as an array, but then we use it via properties (category names) as an object

PR #4310, opened 2016-05-02
Simplify template code

komirad:
Reduce redundant code


danielkerr:
this is how you think something should and just wasting my time as a non fix

Issue #4178, opened 2016-03-31
Currency format arguments to voucher.php


pekka2:
File catalog/controller/account/voucher.php line 16;

`'description'      => sprintf($this->language->get('text_for'), $this->currency->format($this->currency->convert($this->request->post['amount'], $this->session->data['currency'], $this->config->get('config_currency'))), $this->request->post['to_name']),`
change to:
`'description'      => sprintf($this->language->get('text_for'), $this->currency->format($this->currency->convert($this->request->post['amount'], $this->session->data['currency'], $this->config->get('config_currency')), $this->session->data['currency']), $this->request->post['to_name']),`

pekka2:
Finally repaired:
https://github.com/opencart/opencart/pull/4312

PR #4312, opened 2016-05-02
Fix error on voucher purchase

komirad:
Fix missing argument 2 on currency->convert when purchasing voucher.


pekka2:
Thanks.
https://github.com/opencart/opencart/issues/4178

PR #4329, opened 2016-05-05
Don't download currency update if nothing is going to be updated

CZechBoY:


danielkerr:
wastigne of time! no bug, no errors, no point

CZechBoY:
What is "wastigne"?
This PR solves ultra slow dashboard loading.

kuritkaj:
Yes, it is really a waste of time, when somebody wants to improve the performance.

CoverUp:
Actually would be better don't load the model and don't run the refresh in first place, if is not needed
(implement a flag for once a day)

(dashboard.php)
// Run currency update
		if (($this->config->get('config_currency_auto')) AND (RUN_ONCE)) {
			$this->load->model('localisation/currency');
			$this->model_localisation_currency->refresh();
		}

PR #4334, opened 2016-05-06
replace jquery deprecated delegate() with on()

pine3ree:
`$(outerContainerSelector).on(events, innerPresentAndFutureElementsSelector, handler);`

out of curiosity: why escaping single quotes ` 'a[data-toggle=\'image\']'` ? Why not just  `a[data-toggle="image"]`?


danielkerr:
just leave the \'image\' alone. i not sure changing is the best way.

PR #4370, opened 2016-05-14
Robots.txt and favicon

LaconicTranslator:
One can add (and edit) these files by himself, but it is easy to forget about them in a hurry. I suggest to include them by default.


ADDCreative:
Using robots.txt in this way can lead to pages still being indexed by Google. Robots.txt does not stop pages being index, it only stops the being crawled.

See https://support.google.com/webmasters/answer/6062608 and https://www.beussery.com/blog/index.php/2014/06/robots-txt-disallow-20/.

LaconicTranslator:
The main idea was to allow indexing for pages we want to be indexed. Google is not the only search engine ("User-agent") in the world. And some humble search engines do not index website at all if not evidently allowed to through "Allow" directive/
The idea was not to prevent indexing, also it helps a little, but to permit indexing to whose search engine that would not ever index the site without direct permission.
Also a sitemap directive in robots.txt is quite a common practice.

Of course a website will survive without favicon.

PR #4408, opened 2016-05-25
Reduced menu

danijelGombac:


danielkerr:
sorry i think the current way i did it is more straight foward

danijelGombac:
No problem. I do this way in my admin, because for me is easier to add links and is all more clear, because the code is shorter.

PR #4409, opened 2016-05-25
avoid adding duplicate geo zones

Jermim:


pine3ree:
I believe that the same `country_id|zone_id` combination can belong to different geo_zones for different purposes. For example i can define a shipping geo_zone for Italy containing italy `country_id|0=all-zones` combination and e Europe EU VAT zone containing italy `country_i|0=all-zones`. I will use the first geo-zone for shipping methods and the second for tax-rates.

kind regards

Jermim:
@pine3ree unique check is applied per geo_zone.

pine3ree:
@Jermim 
oh yes, right, we are managing a single geo_zone...

danielkerr:
not the best way of doing thngs

danielkerr:
in future try using the same coding style.

Jermim:
@danielkerr what coding style?

PR #4437, opened 2016-06-04
Wrong tax calculations

eka7a:
If apply a coupon, reward or credit, wrong tax calculations

**Wrong Tax**
wrong

**True Tax**
true


danielkerr:
discounts need to be done before tax is calculated. otherwise its illgeal. your discoutning tax! soem coupons are fixed amount.

PR #4442, opened 2016-06-05
Update return_reason.php

eka7a:


danielkerr:
wow didnt know i left this code in there

eka7a:
@danielkerr wow I gave the same reaction when I first saw :)

PR #4474, opened 2016-06-11
Update module.tpl

p0v1n0m:
text_confirm already is missing


opencarteli:
My suggestion is keep the confirm msg before del. So it should add "text_confirm" on controller instead of remove the confirm function
<a href="<?php echo $module['delete']; ?>" onclick="confirm('<?php echo $text_confirm; ?>');" data-toggle="tooltip" title="<?php echo $button_delete; ?>" class="btn btn-danger"><i class="fa fa-trash-o"></i></a>

PR #4486, opened 2016-06-11
Fix sending voucher if her does not belong to an order

p0v1n0m:


opencarteli:
`$data['text_greeting'] = sprintf($this->language->get('text_greeting'), $this->currency->format($voucher_info['amount'], (!empty($order_info['currency_code'])?$order_info['currency_code']:$this->config->get('config_currency')), (!empty($order_info['currency_value'])?$order_info['currency_value']:$this->currency->getValue($this->config->get('config_currency')))));`

p0v1n0m:
Above there is a condition:
`if ($order_info) {`
...
`}  else {`
> here fix

there may not be $order_info.

PR #4495, opened 2016-06-14
Fix sending voucher if her does not belong to an order #4486

opencarteli:


ADDCreative:
This looks wrong. The oringnal problem was with this line https://github.com/opencarteli/opencart/blob/master/upload/admin/model/sale/voucher.php#L143 not the line modified in the pull request. The original pull request https://github.com/opencart/opencart/pull/4486 looked correct to me.

PR #4503, opened 2016-06-14
Fix 'Error: Could not load model catalog/model/total/{total}.php'

shoputils:
Fix 'Error Could not load model catalog/model/total/{total}.php'.
It appears when the file does not exist.


opencarteli:
Please check your version, total file was put in extension folder
`if (is_file(DIR_APPLICATION . 'model/total/' . $order_total['code'] . '.php')) {`

PR #4538, opened 2016-06-21
Repeated inclusion of datetimepicker files

p0v1n0m:
Necessary scripts and styles included in the header.


danielkerr:
I need to do some testign before i merge this. I think there maybe a better way.

PR #4565, opened 2016-06-24
Update search.php

danijelGombac:


p0v1n0m:
For statistics need a user language, but not default language.

danijelGombac:
$this->session->data['language'] is language code, not id. User language_id is $this->config->get('config_language_id') which is set in startup.

`$this->session->data['language'] = $code;`

PR #4580, opened 2016-06-28
smtp password protection

eka7a:
Mail password are recorded as plain text. if provided access to the database, it is a security risk.


danielkerr:
i dont think this is required. the store owner should portion off the settings page.

PR #4587, opened 2016-06-29
Fix extension error/permission action.

danijelGombac:
Prevent adding part 2 to route when install or uninstall extension, but dashboard errors remain.


danielkerr:
i think ur right

PR #4614, opened 2016-07-05
add the "Categories" search function

bau720123:
before
before

after
after

when the "Categories" have a lot of data
it will cause users that difficult to find the one of them，so I made this
use the keyword to find them
no mater which the "Categories" level is


bau720123:
hi @danielkerr 
if there is anything trouble，please let me know
tks

PR #4640, opened 2016-07-10
why update all token in getCustomerByToken function?

majeed21:
why update all token in getCustomerByToken function?
what does it mean?


danielkerr:
its a one off token that is created and deleted and will be used by the admin to login into a custoemrs account. 

DO NOT SUBMIT A COMMIT A FIX FOR CODE YOU DO NOT UNDERSTAND!!!!

PR #4648, opened 2016-07-12
2 small "bug" fixes

raxbg:
These are 2 small changes:

1. Fixes an issue where OCMOD does not work properly for people that have the DIR_CATALOG constant defined in their main config.php file. I know this constant is not meant to be there, but some users have it and they get bad experience with installing modules that need OCMOD for the catalog dir.

2. The OCMOD log gets quite large if not cleared manually and it really slows down the Modifications page. Since it does not really make sense to have a log for any refresh event other than the last one, I suggest we clear the log on each refresh. The clear log button may even be removed.


danielkerr:
there is no DIR_CATALOG in the main file! it shoud not hbe there! and you should not be posting fixes for issues people cause themselfs!

florinsith:
Its those automated cpanel installers, they mess up the configs real bad.

danielkerr:
ok

danielkerr:
i already added a fix for this

PR #4651, opened 2016-07-12
Add ean and upc code for availability in view

gnovaro:
Make product barcodes availability in view


CoverUp:
and we need this because...
I mean, if we gonna add this, why not to add all the available fields.

danielkerr:
im going to set the system so people can set which of the UID fileds can be shown or not.

gnovaro:
I have a lot of ideas to improve the code i think use a many old practices, if better you pass the entire product object to the view and you can reduce a lot of line of code in the controller, and you make wharever you want with this information in the view, maybe can add some helper classes to transform some values.

CoverUp:
True, but we need to consider that we don't want to move much code from controller to the view neither, after all that's why controller exists in MVC scheme. We need to provide ready-to-use data as possible.

danielkerr:
not required

Issue #4652, opened 2016-07-12
Question - Twig design considerations


sergei-maertens:
Pardon me if this is not the correct place for this question, and I also want to mention beforehand that I did not check the current Twig implementation in depth, however I have a few questions:

1. Will it be possible to modify the Twig environment that's used? For example, I want to be able to register custom extensions to the Twig environment that's used. I feel this should be possible and supported in core without any OCMode/VQMod shenanigans.

2. The templates I saw that are being redone in Twig don't seem to use the inheritance system (`{% extends 'foo.html' %}`), but other blobs of HTML rendered via different templates. The `{% block <blockname> %}` system is one of the most powerfull features in decent template engines, will this be supported? And did you consider replacing the `{{ header }}` stuff with includes, e.g. `{% include 'header.html' %}`?

danielkerr:
its not finished yet. we can have an array of config settings maybe fed through thr constructor. which features do you want to use?

maybe not though because i dont want any executable code in the templates.

sergei-maertens:
I do realize this is in development, which is why I am asking if it will be
supported or not.

The features I would need are the ability to register my custom Twig
extension and/or add TokenParsers:

    $twig = new Twig_Environment($loader);
    $twig->addExtension(new Project_Twig_Extension());

    $twig = new Twig_Environment($loader);
    $twig->addTokenParser(new Project_Set_TokenParser());

Other features would be to use different loaders, but that's probably less
important.

I don't see how this has anything to do with executable code in the
template, care to elaborate?

Also, could you comment on point 2. with regards to the inheritance system?

danielkerr:
because if you look at the dev branch i have built a template editor. i dont want people being able to save php code and it being executable because its insecure which would not be good for opencart cloud.

danielkerr:
i can maybe add this.

sergei-maertens:
It's clear that being able to execute arbitrary PHP code is a bad thing and
should be avoided, but I fail to see how this relates to the ability to
register custom extensions. Mind explaining that link?

PR #4657, opened 2016-07-14
Many improvements in the template engine and twig

prhost:
Hello,

I made enough improvements and modifications to the engine and twig template, I believe that was pretty cool. When I saw that you put twig I was very happy. I made for the safety of Opencart cloud and extensive maximum possible.

The right would make an interface and implement the engines, to force a default, since both are used in the same way, not because I think it should be in a folder in so `system/instancies` and not within the`library`, and autoload only load the `system/library` folder.

I hope they accept, or at least discuss the pull request.

also solves the problem of issue: https://github.com/opencart/opencart/issues/4652

I'm doing a twig extension especially for Opencart, it will help a lot in the templates.


danielkerr:
im not going to accept this because coding like this.

public function render($template, $registry) {

danielkerr:
i will finish the twig integration today then we can come up with some solutions to access twigs extension system.

prhost:
Need the registry in the engine to gain access to opencart if you need, in this case I am using the event to modify the twig if necessary, but can be other future needs such as access to settings for example.

danielkerr:
What about adding a DI container?

prhost:
You want to use a design patterns?
https://github.com/domnikl/DesignPatternsPHP/tree/master/Structural/DependencyInjection

danielkerr:
maybe not specifically this implementation.

prhost:
I'll think of something better, needed to change the autoload to support all system folder and not just the library.

prhost:
I changed the issue of registry and created an interface to standardize engines. From a look.

A doubt you intend to launch this modification of the twig and complete convert themes to version 2.3?

danielkerr:
its not worht it

prhost:
because? should not be using events for anything, will eventually become another problem. have to provide events for developers only, and do not use, you have the possibility of leaving the core. as a developer will modify something template engine without modifying the theme event? show me what's wrong in my pull request so that I can improve

PR #4658, opened 2016-07-14
Divido admin twig

davidteal:


prhost:
It was cool, to use the ternary checked than repeat the code. Example:
`{{ plan.id in divido_plans_selected ? 'selected="selected"' : '' }}`

PR #4660, opened 2016-07-14
Porting admin payment extensions to twig

davidteal:


danielkerr:
your still not sticking to the standards!

and this code will probably not event work!

 {{ globalpay_account.visa.default is defined and globalpay_account.visa.default == 1 ? 'checked="checked" ' : '' }} />

look at my other files!

PR #4712, opened 2016-07-31
Backwards Compatibility Fix

OpenCartAddons:
- Added language directory fallback to compatibility event
- Added controllers to action_event in catalog config
- Added fallback to older extension directory for cart page module loading
- Fixed install/uninstall calls for older extensions (https://github.com/opencart/opencart/pull/4715)
- Fixed permissions for older extensions
- Changed default language directory object to public
- Removed unnecessary compatibility code in language library

PR #4715, opened 2016-07-31
install and uninstall functions not supported in compatibility

Dreamvention:
Hi

I noticed that functions install and uninstall were not supported for backwards compatibility. 

the issue lies in this line at /admin/controller/event/compatibility.php

line 10 - route = $part[1] . '/' . $part[2];

it simply drops the method after the route. 

so i have decided to add this fix - hope it gets added to the release since all my modules use theres nice methods.

thanks

### DIFF
--- a/upload/admin/controller/event/compatibility.php
+++ b/upload/admin/controller/event/compatibility.php
@@ -7,7 +7,8 @@ class ControllerEventCompatibility extends Controller {
 		$part = explode('/', $route);
 				
 		if (!is_file(DIR_APPLICATION . 'controller/' . $route . '.php') && is_file(DIR_APPLICATION . 'controller/' . $part[1] . '/' . $part[2] . '.php')) {
-			$route = $part[1] . '/' . $part[2];
+			unset($part[0]);
+			$route = implode('/', $part);
 		}
 	}
 	


OpenCartAddons:
Good catch

OpenCartAddons:
Your idea of unset and implode might actually work better than the current system.  The way the backwards compatibility it currently set assumes that the route consists of a maximum of 2 parts to the route which isn't always true.

I'll add this to my current PR https://github.com/opencart/opencart/pull/4712

florinsith:
In case an old file exists, examaple module/some_module.php and there is also an extension/module/some_module.php. Then you cant access any public method from the new version of the module because compatibility check will be valid since extension/module/some_module/some_method.php doesnt exist.
I think that since this compatibility is limited to extension, that should be the first check on the part[0] then see if the part count is atleast 3, then check if the first 3 part file exists then the  2 part file. Or something else?

Dreamvention:
Hi everyone
made a vqmod file for the compatibilty issue based on OpenCartAddons and my pull requests 

https://gist.github.com/Dreamvention/7794f7708956a6574dcd356457c8f189

feel free to use it untill opencart adds it to core

OpenCartAddons:
@danielkerr @jamesallsup Not sure why the compatibility changes haven't been implemented into the core yet.

I hope they are implemented before 2.3.0.3 is released.

straightlight:
This file no longer exists in the master branch and does not exist in the pre-release version. Therefore, this commit will now be closed.

PR #4722, opened 2016-08-02
- banner_image_description reference in language model removed

qphoria:
as that table was removed.


fido-x:
Since the "banner_image_description" table has been removed, the language entry for the banner is now in the "banner_image" table.

The query at line 222 in admin/model/localisation/language.php should be changed to:
`$this->db->query("DELETE FROM " . DB_PREFIX . "banner_image WHERE language_id = '" . (int)$language_id . "'");`
Not removed.

qphoria:
Good catch !

PR #4746, opened 2016-08-10
Fix for Secure Trading

ry-welford:
Fix for problem appearing with value being passed to the format method in Currency class and the last zero being dropped when amount was passed into round function.
- Added new format method into the Secure Trading model.
- Replaced round functionality with number_format and result conforms with XML spec Secure Trading require


danielkerr:
ilready been added

PR #4748, opened 2016-08-10
Master

ecsiron:
Filters products for customers and administrator


ecsiron:
Filters

danielkerr:
not using this crap! irt would also crash a server setting the limit to 999

ecsiron:
50?

Issue #4764, opened 2016-08-15
Double code


kagan94:
Hi, guys.

You have two the same lines in the styles of default template.
"catalog/view/theme/default/stylesheet/stylesheet.css"
lines:
#content .col-lg-2:nth-child(6n+1),
#content .col-lg-2:nth-child(6n+1),

I think one line is enough.

PR #4766, opened 2016-08-16
Fixed Issue #4764

andrei-bogdan:
Fixed issue by removing same line of CSS code existing twice in the file


evgenekorolev:
on line 698 (#content .col-md-2:nth-child(6n+1)) and 707 (#content .col-sm-2:nth-child(6n+1)) please delete doubles too.

PR #4662, opened 2016-07-14
Update pagination.php

eka7a:
Fixed ?page={page}
https://github.com/opencart/opencart/issues/4637

PR #4800, opened 2016-08-24
Adding the case where page is the only one query parameter

nchankov:
This will give possibility to have urls like:
http://server.com/news/archive?page=1
At the moment if there is a case like this, the first page is rendered as 
http://server.com/news/archive?page={page}

### DIFF
--- a/upload/system/library/pagination.php
+++ b/upload/system/library/pagination.php
@@ -33,10 +33,10 @@ class Pagination {
 		$output = '<ul class="pagination">';
 
 		if ($page > 1) {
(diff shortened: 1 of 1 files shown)


danielkerr:
was fixed before here:

https://github.com/opencart/opencart/pull/4662

I have no idea why you need to repost!

nchankov:
Simple, haven't seen it in the main branch and added it. Good that some one else fixed it.

danielkerr:
i will accept the pull request when i have time to go over it

PR #4805, opened 2016-08-25
Update google_base.php

billynoah:
Add feed products to `$product_data` array to prevent duplicates.


billynoah:
Why no merge Daniel?  This was clearly the original intention of `$product_data` and if not what's the purpose of that array?  It gets initialized and compared against but never populated...

PR #4822, opened 2016-08-28
Update voucher.php

osworx:
Sending a voucher per Email leads to an error like: `PHP Notice:  Undefined index:  in {server}\system\library\cart\currency.php on line 25` and lines 26,27 and 30.
This because the voucher is not created from an order.
Now it takes the currency value from config.


side-by-side:
@danielkerr, please merge this!

danielkerr:
i have merged it but i bet it will cause some issues.

PR #4826, opened 2016-08-28
Problem with VAT calculation right after customer registration.

jeffz2012:
Wrong place, I moved it to issues.
Sorry


danielkerr:
i have no idea what the changes are here

PR #4834, opened 2016-08-30
font-awesome replaced with material-design-icons

3zzy:
- Replaced font-awesome with material-icons EVERYWHERE. 
- Classes not removed since some are being used as selectors to do stuff with jQuery
- font-family changed to a better system default


pekka2:
Font image is not display in admin menu

danielkerr:
you need to ask before doing something like this.

3zzy:
I did it for myself but thought I'd also send a pull request, it almost works but needs a bit more work so I moved on for now.

PR #4869, opened 2016-09-09
fix upgrade error

brunoxu:
avoid unnecessary drop primary key operations

### DIFF
--- a/upload/install/model/upgrade/1000.php
+++ b/upload/install/model/upgrade/1000.php
@@ -269,9 +269,11 @@ class ModelUpgrade1000 extends Model {
 				$last_key_name = '';
 				if ($query->num_rows) {
 					foreach ($query->rows as $result) {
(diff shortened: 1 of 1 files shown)


danielkerr:
it might be necessary if a tables changes its primary key

brunoxu:
If there's a table has no primary keys, but some indexes, upgrade program will abort with error: "Can't DROP 'PRIMARY'; check that column/key exists".

table ddl:
```
CREATE TABLE `oc_blog_product` (
  `blog_id` int(11) NOT NULL,
  `related_id` int(11) NOT NULL,
  UNIQUE KEY `blog_id` (`blog_id`,`related_id`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```

brunoxu:
@danielkerr  Please check this commit again.

PR #4838, opened 2016-08-31
Paypal standard from issue #4608

pekka2:
https://github.com/opencart/opencart/commit/5ba0b5c3d8bde659bdb76c1f1458e92bfebe9dd2 https://github.com/opencart/opencart/commit/f1d41e76c4b7379537c4b845c25a38f6a8689ca https://github.com/opencart/opencart/commit/eb5f88dc16a677cc5dde1a9bea9bae721539ce7 https://github.com/opencart/opencart/commit/eb5f88dc16a677cc5dde1a9bea9bae721539ce7 https://github.com/opencart/opencart/commit/9f16bdb78b77e697bb2c6b36ce930dfb2d10545 https://github.com/opencart/opencart/commit/6da1eb7f22cae978cba8771c51c8f97a343397f https://github.com/opencart/opencart/commit/e960711a7bb0b794286190a9d075c472ca9d80c https://github.com/opencart/opencart/commit/e517dfb7b5c529a74392753b16635893d3be0ce4 https://github.com/opencart/opencart/commit/a32d0ca202e92d675b900bba588406c71283d6a https://github.com/opencart/opencart/pull/4838/commits/998ce010ef0a1b3ac8f7217260f0920665bfdd18

PR #4877, opened 2016-09-11
PP_Standard #4680 #4838

pekka2:


straightlight:
These files are no longer in the master branch nor in the pre-release version. Therefore, this commit will now be closed.

PR #4888, opened 2016-09-16
Price mulptiplying options

zvookiejoo:
I've added some code for price modifying options, so now 'select' option can multiply product price by some value. This should be helpful for shops, selling goods that not counting by whole pieces.


straightlight:
Should be suggested as an extension, not for the core Therefore, this commit will now be closed.

PR #4899, opened 2016-09-21
fix forbidden to load image and font file

smartdatasoft:
If any module put there font and css inside config folder then it restricted to load.


smartdatasoft:
hello,  danielkerr  when this will marge in master. so that people can got the file from master.

PR #4901, opened 2016-09-22
Request::clean() early return for array types

pine3ree:
...and remove unnecessary assignment for scalar types.

### DIFF
--- a/upload/system/library/request.php
+++ b/upload/system/library/request.php
@@ -19,13 +19,12 @@ class Request {
 		if (is_array($data)) {
 			foreach ($data as $key => $value) {
 				unset($data[$key]);
-
 				$data[$this->clean($key)] = $this->clean($value);
 			}
-		} else {
-			$data = htmlspecialchars($data, ENT_COMPAT, 'UTF-8');
+			
+			return $data;
 		}
 
-		return $data;
+		return htmlspecialchars($data, ENT_COMPAT, 'UTF-8');
 	}
 }


danielkerr:
so you want to put unsanitised request data on users web sites?

pine3ree:
@danielkerr 

NOPE...please re-read the code..... :-)

array data are still sanitized as well as scalar data....the only difference is avoiding an extra assignment if the data is a scalar....early returns are a common practice nowadays...and

pine3ree:
check it out: http://www.pine3ree.com/test/oc-clean.php

```php
<?php

function oc_clean($data) {
    if (is_array($data)) {
        foreach ($data as $key => $value) {
            unset($data[$key]);
            $data[oc_clean($key)] = oc_clean($value);
        }
    } else {
        $data = htmlspecialchars($data, ENT_COMPAT, 'UTF-8');
    }

    return $data;
}

function p3_clean($data) {
    if (is_array($data)) {
        foreach ($data as $key => $value) {
            unset($data[$key]);
            $data[p3_clean($key)] = p3_clean($value);
        }

        return $data;
    }

    return htmlspecialchars($data, ENT_COMPAT, 'UTF-8');
}

$data = array(
    '&' => '&',
    '<' => '>',
    '<' => '<',
    '"' => '"',
    'nested' => array(
        '&' => '&',
        '<' => '>',
        '<' => '<',
        '"' => '"',
    ),
);

$oc_data = oc_clean($data);
$p3_data = p3_clean($data);

header('Content-Type:text/plain; charset=utf-8');
echo 'RESULTS ARE ' . ($oc_data === $p3_data ? '' :  'NOT ') . 'IDENTICAL' . PHP_EOL . PHP_EOL;

echo 'oc_clean($data) => ' . print_r($oc_data, true) . PHP_EOL;
echo 'p3_clean($data) => ' . print_r($p3_data, true) . PHP_EOL;
```

PR #4916, opened 2016-09-28
skip loading common controller views with empty data

pine3ree:
in case there is no module data to be rendered we can just skip loading the view since the result will be an empty string anyway.
This saves a few included files for templates without modules or with modules in only a few positions (which represents most of the cases)

### DIFF
--- a/upload/catalog/controller/common/column_left.php
+++ b/upload/catalog/controller/common/column_left.php
@@ -69,6 +69,10 @@ class ControllerCommonColumnLeft extends Controller {
 			}
 		}
 
+		if (empty($data['modules'])) {
+			return '';
+		}
+
 		return $this->load->view('common/column_left', $data);
 	}
 }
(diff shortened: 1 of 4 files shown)


pine3ree:
Hello Daniel,

I have applied a few more improvements in my implementations:

column_(left|right) and content_(top|bottom) are basically the same controller with different parameters anbout which layout modules to load and which template to render.

in my implementation i defined a base abstract controller ControllerAbstractModuleContainer so that controllers can be this simple:

```php
class ControllerCommonColumnLeft extends ControllerAbstractModuleContainer
{
    protected $position = 'column_left';
    protected $template = 'common/column_left';
}

```

The determination of the current $layout_id should also be done once (when rendering modules we are still on the same route). This also can be easily achieved by using a common base controller and a static $layout_id class var.

kind regards,

maks

PR #4919, opened 2016-09-29
rename user_permission controller

pine3ree:
to `user_group` for "module" naming consistency.

maybe we should also add an upgrade utility to rename u'ser/user_permission' in the u`ser_group `db table , `permission` json_encoded column. 

Hello @danielkerr ,
does thes changes make sense to you?


danielkerr:
can not use user group as some web hosts block this word in the url

pine3ree:
@danielkerr ,

oh, ok, sorry. I just thought it was an old partial renaming leftover....
kind regards

pine3ree:
@danielkerr 
..or we could rename it to `user_role`...

danielkerr:
its fine!

PR #4927, opened 2016-10-03
Removed redundant delete call for product_attributes

qphoria:
Removed redundant delete call for product_attributes


straightlight:
Added in the pre-release version. Thanks for posting this.

straightlight:
Didn't noticed this was reverted. Will revert it back. Better to close this commit then.

PR #4939, opened 2016-10-11
Capture the folder name if the user rename

valdeir2000:
https://forum.opencartbrasil.com.br/viewtopic.php?p=70464#p70464

### DIFF
--- a/upload/admin/model/tool/image.php
+++ b/upload/admin/model/tool/image.php
@@ -39,9 +39,9 @@ class ModelToolImage extends Model {
 		}
 
 		if ($this->request->server['HTTPS']) {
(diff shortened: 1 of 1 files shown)


jamesallsup:
duplicate.

valdeir2000:
It is not duplicated. I do not know if you noticed that there are two different files.

One is admin and is another catalog.

I made by the browser and can only send a commit by file.

jamesallsup:
ok thanks, didnt see that it was different locations

straightlight:
This will cause increased error messages showing in the logs while the basename is correctly set with the DIR_IMAGE. Therefore, this commit will now be closed.

PR #4969, opened 2016-10-23
Twig Extension with Debug Module 

rodrigomanara:
Just create a template handler for twig,  set the environment also the debug on the twig file. 
Setup twig handler is very easy, append "TWIG" on the view "$this->load->view('path\view\template', $data , "TWIG")" this will call twig extension and render the content in twig {{ content }} will be handled , also I able the debug, so developers can debug "var" on the page.


hauso:
not sure if I got a goal, but for debuging twig is enought to add following 2 lines into twig.php (\system\library\template)
`		$this->twig->addExtension(new \Twig_Extension_Debug());
		$this->twig->enableDebug();`

so constructor wil looks like:
`	public function __construct() {
		// include and register Twig auto-loader
		include_once DIR_SYSTEM . 'library/template/Twig/Autoloader.php';
		
		\Twig_Autoloader::register();	
		
		// specify where to look for templates
		$loader = new \Twig_Loader_Filesystem(DIR_TEMPLATE);	
		
		// initialize Twig environment
		$this->twig = new \Twig_Environment($loader, array('autoescape' => false));
		$this->twig->addExtension(new \Twig_Extension_Debug());
		$this->twig->enableDebug();
    }`

and in template use 
{{ dump() }} - this show all variables (same as var_dump) or
{{ dump(varName) }}

danielkerr:
thx but i have other solution

rodrigomanara:
no worries.. thanks for looking

PR #4979, opened 2016-10-27
change extension type filter method

brunoxu:
change from select to btn-group

### DIFF
--- a/upload/admin/view/template/extension/extension.tpl
+++ b/upload/admin/view/template/extension/extension.tpl
@@ -19,17 +19,11 @@
         <fieldset>
           <legend><?php echo $text_type; ?></legend>
           <div class="well">
-            <div class="input-group">
-              <select name="type" class="form-control">
-                <?php foreach ($categories as $category) { ?>
-                <?php if ($type == $category['code']) { ?>
-                <option value="<?php echo $category['href']; ?>" selected="selected"><?php echo $category['text']; ?></option>
-                <?php } else { ?>
-                <option value="<?php echo $category['href']; ?>"><?php echo $category['text']; ?></option>
-                <?php } ?>
-                <?php } ?>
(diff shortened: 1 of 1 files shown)


OpenCartAddons:
Looks good.  I would also argue that it's more user friendly as I've had a few people get confused and not realize that the drop down menu was for filtering the type of extensions that are visible in the list.

[This is how the modified filter looks](http://imgur.com/a/18dCv)

One change I would recommend is removing the filter icon from each button.

Cheers,
Joel.

upload/admin/view/template/extension/extension.tpl
```diff
@@ -19,38 +19,36 @@
...
-                <?php } else { ?>
-                <option value="<?php echo $category['href']; ?>"><?php echo $category['text']; ?></option>
-                <?php } ?>
-                <?php } ?>
-              </select>
-              <span class="input-group-addon"><i class="fa fa-filter"></i> <?php echo $text_filter; ?></span>
+            <div class="btn-group ext-types">
+              <?php foreach ($categories as $category) { ?>
+              <?php $active = ($type == $category['code'])?' active':''; ?>
+              <button type="button" class="ext-type btn btn-default<?php echo $active; ?>" data-value="<?php echo $category['href']; ?>"><i class="fa fa-filter"></i> <?php echo $category['text']; ?></button>
```
OpenCartAddons: I would recommend removing the filter icon from each button:
`<i class="fa fa-filter"></i> `

brunoxu: The filter icon is used for showing loading progress. Unless there's a better way.

straightlight:
The extension in the admin has been modified on the master branch recently as it is still on-going process. Therefore, this commit will now be closed since there are too many variances involved.

PR #4984, opened 2016-10-31
admin section. fix tooltips

savage4pro:


savage4pro:
just closed?
btw current solution fixes this bugs:

before uninstall module
after uninstall
add new row with not working tooltip

PR #4988, opened 2016-11-01
fixed error days when date_modified is 0000-00-00 00:00:00

PhamToan:
When the product has no modified the date will be 0000-00-00 00:00:00. it cause error in google site map. We should change to date_added
$date_modified = $product['date_modified'] != '0000-00-00 00:00:00' ? $product['date_modified'] : $product['date_added'];
$output .= '<lastmod>' . date('Y-m-d\TH:i:sP', strtotime($date_modified)) . '</lastmod>';


melvin16:
Hello PhamToan,
In which file should be change to date_added
Best regards,
melvin16

straightlight:
No longer useful. Both startup files now uses the date('P') function in order to prevent this sort of insertion on the database to occur nowadays, either from the Master Branch or from the pre-release version. Therefore, this commit will now be closed.

Issue #4762, opened 2016-08-15
Some Payment Methods not working


saulfautley:
I just installed a fresh copy of the latest OpenCart 2.3.0.2, and for some reason I can't get some native payments methods working, such as Free Checkout.

You can see it is enabled here:
![2016-08-15_133841](https://cloud.githubusercontent.com/assets/2036475/17663580/d3d3a4e4-62ed-11e6-84c0-453c1a3a6690.png)

Yet it doesn't show up at checkout:
![2016-08-15_133936](https://cloud.githubusercontent.com/assets/2036475/17663585/d752d590-62ed-11e6-8520-9bb90f52be6d.png)

Has anyone else experienced this?

trueliarx:
Verify if it's enabled and usable for your configuration. Is it a fresh install?

saulfautley:
@truliarx As you can see in my screenshot it is enabled. I've also stated that it is a fresh install. Is there something I'm missing?

tylerbenzing:
Have you made sure that the product you are trying to checkout with has the "Requires Shipping" option checked in the admin panel? I had this same issue and that turned out to be the problem. Easy to miss but makes sense, you can't get Cash On Delivery if your product isn't deliverable

saulfautley:
I tested with Free Checkout, not Cash On Delivery, so not sure why it would need Required Shipping enabled. Will test it out though.

uchk:
Voguepay payment method not working....no order confirmation button
![ww](https://cloud.githubusercontent.com/assets/23273773/20028157/3e3f4d20-a32a-11e6-9c13-4d466fcd2fb7.PNG)
![capturedd](https://cloud.githubusercontent.com/assets/23273773/20028158/43ccfed6-a32a-11e6-9756-9230ee4478ca.PNG)

danielkerr:
free checkout works if the price is 0.00

also cash on delivery can only be used for products that require shipping

PR #4989, opened 2016-11-01
Update cod.php

p0v1n0m:
I think it is excess condition, which results the error #4762


tylerbenzing:
I don't think this check should be removed but rather a warning be added to checkout that informs the user that COD is disabled because the items in the cart are non deliverable

PR #5000, opened 2016-11-05
Disable order_id input text on return_form.tpl

labrg:
Leaving it enabled makes it possible for users/customers to change the order id when submitting return for an order, which has a fixed order id.


labrg:
Resubmitting with more changes!

PR #5001, opened 2016-11-05
Disable various input text on return_form.tpl

labrg:
Input text like first name, last name, email, order id, order date, product name, product code, quantity is filled with specifics loaded according to order/customer. These fields shouldn't be changeable by the customer when submitting returns. 

Enabled fields are telephone (user might want a different phone number to be contacted on for the return), reason for return, product is opened, faulty or other details.

### DIFF
--- a/upload/catalog/view/theme/default/template/account/return_form.tpl
+++ b/upload/catalog/view/theme/default/template/account/return_form.tpl
@@ -25,7 +25,7 @@
           <div class="form-group required">
             <label class="col-sm-2 control-label" for="input-firstname"><?php echo $entry_firstname; ?></label>
             <div class="col-sm-10">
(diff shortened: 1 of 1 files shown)


ADDCreative:
There is a returns link in the footer. If a customer clicks on this link they need to supply all the information as it's not linked to any order.

labrg:
In my opinion, that encourages spam submits; could be avoidable work for new businesses who are trying to compete with other giants when starting out. It would probably be better to have users submit returns once they've logged in.

danielkerr:
what about guest checkouts?

if there is an issue with spam submits it should be handled another way using tokens in a hidden form field or url.

can some one build this into the form?

open an issue if you are getting a spam submit issue.

PR #5010, opened 2016-11-08
Added Timezone

eka7a:
http://stackoverflow.com/a/17355238


danielkerr:
can you match the code up to opencart style?

add the time zones procedural in the sittings page since this is only going to ever be used in one place. we dont need to add a whole function in the settings model just for time. then use the startup folder to set this using the setting config data.

PR #5016, opened 2016-11-11
access permission fix for old extensions

givanz:
When saving group permission old modules permissions are saved with "module/module_name" path and permission check fails when the edit link is generated with "extensions/module/module_name" I changed the code to generate the edit link with the right permission path.


danielkerr:
remove the compatibility seems to cause more issues.

PR #5033, opened 2016-11-17
Directories

tblanchard:
I have had a lot of problems deploying opencart to cloud services out of git because it fails to create certain directories before trying to write files to them.  These changes fix that problem.

### DIFF
--- a/upload/admin/controller/extension/modification.php
+++ b/upload/admin/controller/extension/modification.php
@@ -399,16 +399,11 @@ class ControllerExtensionModification extends Controller {
 			foreach ($modification as $key => $value) {
 				// Only create a file if there are changes
 				if ($original[$key] != $value) {
-					$path = '';
 
-					$directories = explode('/', dirname($key));
+					$directory = DIR_MODIFICATION . dirname($key);
 
-					foreach ($directories as $directory) {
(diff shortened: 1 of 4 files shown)


danielkerr:
im not going to go into the long list of reason why you should not have the system creating directories on the fly if they dont exist.

tblanchard:
It already does it - cache directories etc....

And then fails when it actually matters.

I suppose you are not interested in opencart working on cloud deployments.

danielkerr:
already have it working on a cloud deployment.

the directories are already there so no need to create them.

sees like you have done this pull request to for your own projects not any one elses. also if u built a cloud system that require these directories t be built on the fly then you have not set it up properly!

stop wasting m time! this is an issue only for what ever you have built and not opencart! pull this crap again u will be blocked from posting issues.

tblanchard:
Wow. You're a prick. Go suck an egg you miserable cunt. At no time did I do anything to deserve that level of hostility now go fuck yourself

danielkerr:
your a prick for doing a pull request for somthing you want added to the project that nobody wants and would cause issues if config directories got messed up!

use you brain! i use my time to fix issues and add features. it should not be wasted!

PR #5035, opened 2016-11-17
Admin Product List : add Category Filter

shawnnoh:
always hard to find products. So I added category filter


straightlight:
No longer required. Recent versions of OC already captures the category path from the URL. Therefore, this commit will now be closed.

PR #5036, opened 2016-11-20
Report Customer_order results not right

liquidacid:
Total came from the wrong table resulting in very high amounts in the reports

### DIFF
--- a/upload/admin/model/report/customer.php
+++ b/upload/admin/model/report/customer.php
@@ -163,7 +163,7 @@ class ModelReportCustomer extends Model {
 	}
 
 	public function getRewardPoints($data = array()) {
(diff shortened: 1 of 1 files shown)


liquidacid:
Now i review it, it would be even better to do:

$sql = "SELECT cr.customer_id, CONCAT(c.firstname, ' ', c.lastname) AS customer, c.email, cgd.name AS customer_group, c.status, SUM(cr.points) AS points, COUNT(o.order_id) AS orders, o.total AS total FROM " . DB_PREFIX . "customer_reward cr LEFT JOIN `" . DB_PREFIX . "customer` c ON (cr.customer_id = c.customer_id) LEFT JOIN " . DB_PREFIX . "customer_group_description cgd ON (c.customer_group_id = cgd.customer_group_id) LEFT JOIN `" . DB_PREFIX . "order` o ON (cr.order_id = o.order_id) WHERE cgd.language_id = '" . (int)$this->config->get('config_language_id') . "'";

This will not give rounding issues.

straightlight:
This model file does not exist in the master branch. Therefore, this commit will now be closed.

PR #5040, opened 2016-11-23
Admin panel Product list making usefull update

enesfnd:
-Automatic filter when selected auto complete fields (name, model) selected in product list
-Automatic filter when enter key pressed in product list


enesfnd:
Hello sir, I'm currious about making opencart better.

And also I couldn't unsderstand why did you rejected my pul request, what is wrong? if you tell me; I can fix my errors.

Thanks

PR #5056, opened 2016-11-30
Change extension/openbay/fba charset & collation to utf8_unicode_ci

dvlprs:
- confirmed that utf8 is widely used elsewhere in Opencart
   (checked with grep)

 - utf8_unicode_ci because http://stackoverflow.com/q/766809


jamesallsup:
@dvlprs OpenBay Pro files are maintained in a separate fork so have updated there, there was also quite a few instances of it. It will be merged before a new release into OC.

Thanks for the info.

PR #5062, opened 2016-12-03
Comparing with wrong total

samirkumardas:
Since OC 2.2.x, $total is an Array so $discount_total should be compared with $total['total'] instead of $total which is an Array.


danielkerr:
cheers

samirkumardas:
Thank you Daniel. Good day!

PR #5076, opened 2016-12-09
FIX analytics.tpl

cbekir:
Only one extension can be installed.
Fixed endforeach.


revington:
Is it possible to create a 2.3.0.3 release that includes this pach?

PR #5085, opened 2016-12-16
Add catalog/model/catalog/option.php

mehov:
The file I'm submitting is an edited down copy of [`/admin/model/catalog/option.php`](https://github.com/opencart/opencart/blob/master/upload/admin/model/catalog/option.php), so it's 100% your code.

There are scenarios where people will need access to options while in catalog, like building a product filter module. 

For example, we needed a search by colour feature, and to make it work we had to add this file via vQmod. I believe it's much better and more reliable to have it there by default.

Please review this pull request.

### DIFF
--- a/upload/catalog/model/catalog/option.php
+++ b/upload/catalog/model/catalog/option.php
@@ -1,0 +1,116 @@
+<?php
+class ModelCatalogOption extends Model {
+	public function getOption($option_id) {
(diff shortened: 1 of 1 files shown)


straightlight:
Recently, the conventional method names have changed. While the idea might still be a great one for others to use as an extension, please ensure the codes are up-to-date in the PR. As an alternative, you could always publish this type of idea as an extension.

mehov:
Thank you for your reply.

> please ensure the codes are up-to-date in the PR

I made this pull request 3.5 years ago, in December 2016. The code must have been up-to-date back then.

PR #5096, opened 2016-12-23
Update forgotten.php

DMConklin:
``&amp; ``appearing in reset URL


arnisjuraga:
@DMConklin Did you fixed it differently?

DMConklin:
No, I goofed, html_entity_decode was already added in the mail->setText, I had added mandrill API instead of mail() to an install of opencart and took this down once I noticed it was already added.  Would be nice for the string to have been preped earlier like where I had it rather than when added to mail that way I wouldn't have had the an issue, or maybe if I just payed more attention lol.

PR #5109, opened 2016-12-29
Update edit.php

sergiosinicin:
Check key before set value


danielkerr:
no reason for this change

PR #5115, opened 2017-01-03
Fix: division by zero, if store has only Missing orders

arnisjuraga:


straightlight:
Issue already fixed in the master branch and the pre-release version. Therefore, this commit will now be closed.

PR #5124, opened 2017-01-05
Invalid data return from the function

ajaysharavat:
It should return total count of emails. But it returns single email.
Due to this Admin->Sales->Marketing->Email for products not working.


DMConklin:
While I understand that the name of the function has "total" in it I don't think the purpose was to return a number.  Since it's in marketing I would believe a list of distinct emails that purchased that product would be of more use than a number.

straightlight:
Already fixed in the master branch and the pre-release version. Therefore, this commit will now be closed.

PR #5129, opened 2017-01-06
fix #5128

### LINKED ISSUE #5128: Error when installing via CLI: Undefined index: SERVER_PORT
PHP Notice:  Undefined index: SERVER_PORT in /var/www/opencart.dev/public/upload/system/startup.php on line 40

sofyansitorus:


arnisjuraga:
`cli_install.php` is already trying to _validate_ slash at the end of `http_host` variable.

Unfortunately - it has no effect. Look here: https://github.com/opencart/opencart/blob/master/upload/install/cli_install.php#L122 

So, IMHO, better fix is just changing `valid()`  function definition to 

~~~php
function valid(&$options) { 
...
}
~~~

Then, it will work both and correctly add `/` if missing.

PR #5136, opened 2017-01-10
Remove PostgreSQL option when install opencart.

yishiashia:
Remove the PostgreSQL option when installation.


lucasjkr:
If you're going to remove postgresql option, also remove it from languages, controllers, etc.

Also consider removing ordinary 'mysql' option, as it is deprecated in later versions of PHP5 and gone in PHP7

danielkerr:
why?

PR #5144, opened 2017-01-17
Update product.php

billynoah:
Rolling back to previous method (which was fine) as the recent change is wrong.  See comments:
https://github.com/opencart/opencart/commit/528625adb2581d2fcaf2d943b29a5b5c78b4a0a3


arnisjuraga:
+1 for this solution, as it changes less code.
Tested and seems to be correct.

danielkerr:
@arnisjuraga this is why u have been banned because u dont know what u are talking about!

Issue #4954, opened 2016-10-18
[AUTHORIZE.NET SIM] transaction key error


sajicodea:
Authorize.net SIM Payment is not working. Showing "Error 99: This transaction cannot be accepted."

Need to update code : \opencart\upload\catalog\controller\extension\payment\authorizenet_sim.php

Replace :
        $data['x_fp_hash'] = hash_hmac('md5', $data['x_login'] . '^' . $data['x_fp_sequence'] . '^' . $data['x_fp_timestamp'] . '^' . $data['x_amount'] . '^' . $data['x_currency_code'], $this->config->get('authorizenet_sim_transaction_key'));

with : 

```
    $data['x_fp_hash'] = hash_hmac('md5', $data['x_login'] . '^' . $data['x_fp_sequence'] . '^' . $data['x_fp_timestamp'] . '^' . $data['x_amount'] . '^' . $data['x_currency_code'], $this->config->get('authorizenet_sim_key'));
```

JesseOC:
Checked and fixed already.
This issue will be closed.

PR #5148, opened 2017-01-18
Authorize.net SIM Fix

robinflyhigh:
SIM transaction key error #4954
Error 13: The merchant login ID or password is invalid or the account is inactive.

Have tested this with a live account.


danielkerr:
thxc added manually

PR #5154, opened 2017-01-22
Adds download link near upload button for custom fields

cartbinder:
For custom field with type file.
No download links are available in customer form.

Its adds download button for account and address both.


straightlight:
This has been suggested recently on the forum. I might add this feature in the features-pre in the future.

danielkerr:
manually added

PR #5162, opened 2017-01-24
Add persistent option to config

svirusxxx:
We don't want use a persistent connections on our servers.
So I wrote a little changes to opencarta, and now it can be changed.


danielkerr:
not a chance

arnisjuraga:
@svirusxxx is there any use for that?

danielkerr:
set the persistence via php.ini

PR #5165, opened 2017-01-27
Display language name in language selector during install.

olegsv:
Relies on intl extension.


arnisjuraga:
I couldn't test it, because I don't have `intl` loaded on my dev server, but seems legit. :)
Could be too fancy for Opencart.

olegsv:
@arnisjuraga this will work without the extension as well. More reliable way will be to add a constant with language name to all language files, like in osCommerce.

Issue #5151, opened 2017-01-19
OpenCart 2.3.0.2 CSRF vulnerability - User account takeover


open-security:
Hello,
There is a security vulnerability in OpenCart 2.3.0.2 which allows a
hacker to break into a customer account.
The bug exists in "My Account Information" page. The form is not protected
with a token id, so a hacker can change user's information silently.
Here is a demonstrative video to explain this security vulnerability:
http://opensecurity.ca/media/opencart-csrf.mp4

With regards,
Open Security

uksitebuilder:
How ridicuous - the hacker would need to be on the same machine and the user being hacked would need to have left browser window open and not have logged out of OpenCart

Much simpler for the hacker to use the browser history to go back to the PpenCart My Account page - Or press CMD / (Ctrl) + T until the tab re-opens

arnisjuraga:
From the video, the hacker `doesn't need to be` on the same machine!

**This is VERY CRITICAL security issue, correct me if I'm wrong:**

1) User have to be logged into Opencart
2) If user now opens hacker link (sent in email, hidden in any web page) from the same browser in ANY tab, then hacker silently changes the user settings for the user.

open-security:
Yes, the hacker doesn't need to be on the same machine.
CSRF is a dangerous security vulnerability and is one of OWASP's top 10 attacks:
https://www.owasp.org/index.php/Top10
https://www.owasp.org/index.php/CSRF

arnisjuraga:
This has been discovered already 2 years ago. 
So, @danielkerr - this IS an issue:

https://www.curesec.com/blog/article/blog/OpenCart-2031-CSRF-66.html
http://seclists.org/fulldisclosure/2015/Nov/21

danielkerr:
forget it! this guy has only registered an account to promote his own crappy services! 

Ive know about this for years but its always been at the bottom my priority list as its not worth fixing. I have explained this many times! 

but any reason i give this guy will argue with me rather than holding his hand up and admitting it, its always the "what if scenario!".

so lets do the maths

317,000 live web  sites using opencart.

1st how are you going to find which users are members of which store since no personal information about its users are listed on any opencart site!!! 

2nd even if you found a user the user would have to be logged into a specific store you are targeting while visiting a compromised page on another site. I dont know about you but i must spend a few minutes a year logged into sites i buy off. Im not logged in 24 hours a day while im, reading emails! 

[...]

PR #5172, opened 2017-02-03
Implement CSRF protection in user setting forms

shinkbr:
Hi,

This is a fix for a CSRF vulnerability discussed in #5151.
(I'm aware that this is a closed issue but I believe that this needs to be fixed and therefore I wrote a patch.)

There are still more endpoints that should be protected from CSRF (such as the checkout form), but I will keep this PR to the most important two and cover the rest later if this seems OK.

Please correct me if there is a problem with the way I implemented this.

Thanks,

### DIFF
--- a/upload/catalog/controller/account/edit.php
+++ b/upload/catalog/controller/account/edit.php
@@ -73,6 +73,12 @@ class ControllerAccountEdit extends Controller {
 		$data['button_back'] = $this->language->get('button_back');
 		$data['button_upload'] = $this->language->get('button_upload');
 
+		if (isset($this->error['csrf_token'])) {
+			$data['error_csrf_token'] = $this->error['csrf_token'];
+		} else {
+			$data['error_csrf_token'] = '';
+		}
+
 		if (isset($this->error['warning'])) {
 			$data['error_warning'] = $this->error['warning'];
 		} else {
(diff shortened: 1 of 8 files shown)


arnisjuraga:
BUG: csrf token is not generated uppon account/register!

shinkbr:
@arnisjuraga Thanks for pointing out! I will fix it, as we were going to use this for our service regardless.

arnisjuraga:
Please, give a note here too, if someone else is willing to use iet.

akzogibbs:
@shinkbr @arnisjuraga @danielkerr  - Does anyone know if this has been addressed in the most recent V2 build 2.0.3.0? I've been having massive issues with malware injections to point where i'm about to drop Opencart and move to another solution. When i've run scans on the site i'm being told the vulnerabilities are in various input forms?

RaidoDev:
Hi @akzogibbs ! 

Available vulnerabilities do not guarantee, that your site has been infected through them. There can be multiple reasons, why your site has the problem.

You can contact me and I can have a look, what exactly has been broken.

akzogibbs:
@RaidoDev - Yes please. I'll send you an email

shinkbr:
@akzogibbs CSRF is not the kind of vulnerability that can directly result in a malware injection. Perhaps there is another issue in the system.

danielkerr:
This does not sound like an OC issue. If files are being added u can check the logs where there IP is uploading the files

akzogibbs:
thanks for the feedback @danielkerr , @shinkbr ...much appriciated

danielkerr:
also yopur issue is nothing to do with CSRF .

schiggi:
Is this patch available as an vqmod or ocmod file?

PR #5173, opened 2017-02-03
Undefined index: custom_field in catalog/controller/account/edit.php on line 166

arnisjuraga:
If validation error occures (e.g. - no email address), then you will see notice error, if no custom fields ar used.


danielkerr:
already fixed

PR #5190, opened 2017-02-09
Options. Sorting

savage4pro:
#### option list
option list

#### option values
option values

#### product options values
product options values

#### product options values after save
product options values after save

#### product options list
product options list

#### product page
product page

#### product page after save option list
product page after save option list

> As we can see since the _sort_order_ values is not given (and even if it does somewere) there is many **unpredictable** results.

#### option list
option list
order by **sort_order & name**

#### option values
option values
order by **sort_order & option_value_id**

#### product options values
product options values
order by **sort_order & product_option_value_id** (in select-box - same as we have on the Option Values page)

#### product options list
product options list
options ordered by **sort_order & product_option_id** (in ajax-return-box now more results, ordered by **"type" & sort_order & name**)

#### product page
product options list
options and values sorted the **same way as in Admin section**

### DIFF
--- a/upload/admin/controller/catalog/option.php
+++ b/upload/admin/controller/catalog/option.php
@@ -470,7 +470,8 @@ class ControllerCatalogOption extends Controller {
 			$filter_data = array(
 				'filter_name' => $this->request->get['filter_name'],
 				'start'       => 0,
-				'limit'       => 5
+				'sort'        => 'o.sort_order',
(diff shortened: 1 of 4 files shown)


straightlight:
Way too complexed. Better to use this suggestion as an extension from the Opencart Marketplace, not to fit in the master branch in this case. Therefore, this commit will now be closed.

Issue #5208, opened 2017-02-17
Custom customer address fields


fagner-alves:
Implementation of custom fields from the customer address on the invoice and the shipping note.

PR #5209, opened 2017-02-17
Update order.php ( Implementing custom fields )

fagner-alves:
Implementation of custom fields from the customer address on the invoice and the shipping note. #5208


straightlight:
See this solution instead: https://forum.opencart.com/viewtopic.php?f=201&t=209266&start=60#p782020 . Therefore, this commit will now be closed.

PR #5210, opened 2017-02-18
Update store.php

osworx:
If store is running in SSL-mode (https) the display of the url may confuse because it is always http:// only!


straightlight:
No longer the case since the .htaccess file, once enabled, should use the HTTPS protocol by default nowadays which will also reflect on the HTTP_CATALOG constant from both config.php files. Therefore, this commit will now be closed.

PR #5221, opened 2017-02-24
Fixed issue with gift voucher for failed orders

cartbinder:
If gift voucher was added to order. And suppose order was cancelled at payment page or it failed.
So the gift voucher should be allowed to use again since order is not made.

But the negative if statement is doing the reverse. So it should be made positive.

If there is some other logic to this. Then i am sorry for this commit.


ADDCreative:
Your logic is wrong. The query is checking the order for the voucher, not the order the voucher is to be use to pay for.

With this change vouchers that were not paid for can be used and a paid for voucher will be rejected.

cartbinder:
Yes you are correct. I was thinking in wrong way so ended up here.
Thanks !!

PR #5222, opened 2017-02-24
Don't show tax if it was not applied

ink-ru:
Don't show tax if it was not applied


danielkerr:
not requiired the are mutiple settings top show not show tax!

PR #5223, opened 2017-02-24
Don't show tax if it not was aplicated

ink-ru:
Don't show tax if it not was aplicated


jamesallsup:
This is down to preference, not a bug.

Also, please read the standards document before creating PR as the if statement code is never used like that.

PR #5227, opened 2017-02-26
No base href bug fix

ink-ru:
No base href bug fix


danielkerr:
It's not an issue unless u messed something up

PR #5230, opened 2017-02-27
We have to apply OCMOD ordered by it's date.

toporchillo:
When we cleaning OCMOD cache, OpenCart applies all OCMOD-modifications. We have to keep OCMOD apply order or we can get unexpected OCMOD compatibility errors.


howdu:
Sometimes you need a modification to be called last if it's modify other modifications. At the moment I do this by prefix name with 'Z' sorting by date loosing this.

danielkerr:
sort order was added in the last release

PR #5232, opened 2017-02-28
Update pagination.php

billynoah:
lines 36, 39 & 69: Fix replacements in pagination url for page 1 when SEO Urls are enabled and current page is higher than 1
lines 38 & 68: use loose type comparison == instead of === since $i becomes type `double` and not an integer
fix indentation on line 69


straightlight:
No longer relevant. system/library/pagination.php file has been removed from the master branch and now being handled from common/pagination loading the common/pagination TWIG file. Therefore, this commit will now be closed.

PR #5247, opened 2017-03-07
Marketplace Update

OpenCartAddons:
Added support For `/image/extension/`, `/javascript/extension/` & `/stylesheet/extension/` directories for extension download packages.

The `admin/view/template/extension/` directory will get cluttered with images, .js and .css files.  This change will allow developers to keep extension files organized in their corresponding directories.

Front end already supports `/image/extension/` and `/stylesheet/extension/` as these are located within the theme directory.

### DIFF
--- a/upload/admin/controller/marketplace/marketplace.php
+++ b/upload/admin/controller/marketplace/marketplace.php
@@ -930,9 +930,13 @@ class ControllerMarketplaceMarketplace extends Controller {
 				$allowed = array(
 					'admin/controller/extension/',
 					'admin/model/extension/',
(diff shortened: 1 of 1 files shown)


dmitriyzhuk:
@OpenCartAddons  i think you are absolutely right, but i went a bit further in this pull request. https://github.com/opencart/opencart/pull/5281

because we also miss out on autoload if we don't have access to system/library 
config 
shared files between admin and catalog
shared images ...

OpenCartAddons:
https://github.com/opencart/opencart/pull/5281

Issue #5248, opened 2017-03-07
Invoice Prefix still using year 2013 for new opencart install


trinkaljuneja:
While installing new opencart , Invoice Prefix still start for year 2013

**config_invoice_prefix** is variable 

File 
[opencart/upload/install/opencart.sql](https://github.com/opencart/opencart/blob/master/upload/install/opencart.sql)

line number (3189) to edit 

(0, 'config', 'config_invoice_prefix', 'INV-2013-00', 0),

pelagise:
I looked at the code and see the same, it looks like the code needs to be updated to the current year of install.

arnisjuraga:
Wouldn't it be better to use current year always?

~~~sql
(0, 'config', 'config_invoice_prefix', CONCAT('INV-', YEAR(CURDATE()), '-00'), 0),
~~~

https://github.com/opencart/opencart/pull/5250/commits/44ea59cf71019af9f212b84fff4014b5a8ba3bc9#diff-e1c5ff998f578ec755fca08d0c454667R3189

pnair39:
If this is a live store with invoice prefix set and they do an upgrade will this script overwrite the current settings in settings>options?

arnisjuraga:
No, it should not, because this script is executed ONLY on a new install with a fresh database. 
Upgrades are taken only from here: https://github.com/opencart/opencart/tree/master/upload/install/model/upgrade

PR #5250, opened 2017-03-07
Set Invoice Prefix during install to current Year

arnisjuraga:


arnisjuraga:
#5248

straightlight:
During a fresh installation, that might be ideal. Added in the features-pre release. Thanks for posting this.

PR #5281, opened 2017-03-15
Update Marketplace  - folders that allow all extension to work

dmitriyzhuk:
For most of the modules the list of allowed folders is not enough. it will create duplicate code which is not good practice. 

to avoid this we would suggest adding the follwoing folders 

$allowed = array(
    'admin/controller/extension/',
    'admin/model/extension/',
    'admin/view/image/extension/',
    'admin/view/javascript/extension/',
    'admin/view/stylesheet/extension/',
    'admin/view/template/extension/',
    'catalog/controller/extension/',
    'catalog/model/extension/',
    'catalog/view/javascript/extension/',
    'catalog/view/theme/',
    'system/config/extension/',
    'system/library/extension/',
    'system/storage/logs/extension/',
    'system/extension/',
    'image/catalog/extension/',
);

from the code I assume language is not an issue - it has its own logic.

### DIFF
--- a/upload/admin/controller/marketplace/marketplace.php
+++ b/upload/admin/controller/marketplace/marketplace.php
@@ -930,10 +930,19 @@ class ControllerMarketplaceMarketplace extends Controller {
 				$allowed = array(
 					'admin/controller/extension/',
 					'admin/model/extension/',
+					'admin/view/image/extension/',
+					'admin/view/javascript/extension/',
+					'admin/view/stylesheet/extension/',
 					'admin/view/template/extension/',
 					'catalog/controller/extension/',
 					'catalog/model/extension/',
-					'catalog/view/theme/'
+					'catalog/view/javascript/extension/',
+					'catalog/view/theme/',
+					'system/config/extension/',
+					'system/library/extension/',
+					'system/extension/',
+					'system/storage/logs/extension/',
+					'image/catalog/extension/'
 				);
 
 				// Language Admin


dmitriyzhuk:
Many of our extensions are quite complex and rely on thrid-party libraries that are used both in admin and category. 

placing them in each of those folders simply duplicates the code and makes no sense.

Also we use PSR  Autoload and losing this option is really a bad way to develop code.

we often use config files to create a basic setup and we have files that are shared between the admin and catalog. 

so with this in mind we would be happy if we could keep all these features in Opencart

OpenCartAddons:
I've submitted a PR (https://github.com/opencart/opencart/pull/5247) for this already, but I like your idea of including a system/library directory.

danielkerr:
im going to have an override folder called vendor for the 3rd party libraires. it will still work like load libraries.

dmitriyzhuk:
Super! 
what about **system/config** - we use it for almost every product. and we load the configs via the standard load method which is very convenient. 

would also love to see this option available

danielkerr:
I'll add some of the folders

danielkerr:
Also once this Dev branch is ready we are going to give Devs a cloud store each for testing there mods.

danielkerr:
ok added this manually

danielkerr:
i dont think its require to haver a seperate directory for extension logs. its just working space needed for logs

PR #5293, opened 2017-03-20
Fix 'Undefined index: password' in customer model

auralon:


ADDCreative:
This is wrong. If the password is blank. Which is very likely if just updating name, email, etc. It will mess up the customer's password.

If you are getting 'Undefined index: password' it's probable a messed up admin theme or due to a modification.

auralon:
@ADDCreative in which case, shouldn't it be...

    if (isset($data['password']) && $data['password']) {

This would get rid of the undefined index error and ensure that blank passwords are ignored.

samchen945:
Why the tpl has no <input name="password"> ?

auralon:
@samchen945 is it not this? https://github.com/opencart/opencart/blob/master/upload/admin/view/template/customer/customer_form.tpl#L266

danielkerr:
@auralon dont post fixes again you dont know what you are doing!

auralon:
I'm working on an inherited site that uses the `editCustomer` method in an extension module, where at times the password key/value pair isn't in the `$data` array passed through to the model.  I traced the undefined index error to this line and spotted what I believed to be a genuine bug.  The amendment to that initial patch would work perfectly well, however, even if it would be surplus to requirement in a default opencart installation.  I'll be sure to not post any future fixes though, thanks.

PR #5298, opened 2017-03-22
Voucher description correction

ADDCreative:
Voucher value in description is wrong if not default currency.

### DIFF
--- a/upload/catalog/controller/account/voucher.php
+++ b/upload/catalog/controller/account/voucher.php
@@ -13,7 +13,7 @@ class ControllerAccountVoucher extends Controller {
 
 		if (($this->request->server['REQUEST_METHOD'] == 'POST') && $this->validate()) {
 			$this->session->data['vouchers'][mt_rand()] = array(
-				'description'      => sprintf($this->language->get('text_for'), $this->currency->format($this->request->post['amount'], $this->session->data['currency']), $this->request->post['to_name']),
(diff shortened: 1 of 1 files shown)


straightlight:
Not sure why the 1.0 would always need to be used ...
With the integration of the currency fixer, the automation for the currency can be handled with the $this->currency->format and the $this->currency->convert ...

```$this->config->get('config_currency')``` could then be used instead of 1.0 ... any reason why the default value of 1.0 needs to be hardcoded while, nowadays, the pre-release version now instructs the user to put 1 in their admin currency form as their default value?

ADDCreative:
The value of the default currency will be 1, but other currencies will different values. See the screenshot https://github.com/opencart/opencart/issues/6319#issuecomment-358535564 where a second currency EUR have been added and the customer has selected this. So default will be USD = 1 and second EUR = 0.85. It wrongly has the description "€12.75 Gift Certificate" where it should be €15 to match the actual cost. This is because the $this->request->post['amount'] is already in the customer's currency. As no conversion is needed, only formatting we need to set the conversion value to 1.0.

Looks @Felix-Ho has already committed my fix to the master branch. https://github.com/opencart/opencart/commit/9bda8ae81f182dcfb27c60f9258f877e2d9f70b8

Just need to add to the 3.0.3.4 Pre-release.

straightlight:
Ok, found it in the master branch: https://github.com/opencart/opencart/blob/master/upload/catalog/controller/account/voucher.php . Will add it to the pre-release.

straightlight:
Ok, issue has been resolved. Code is already in master branch and now added to pre-release. This commit will now be closed.

PR #5303, opened 2017-03-24
Update store.php

DMConklin:
I know it's purely aesthetic but hey why not?


ADDCreative:
$this->config->get('config_secure') is actually redundant. It's not actually used anywhere any more (apart from maybe klarna_checkout.php). See https://github.com/opencart/opencart/search?utf8=%E2%9C%93&q=config_secure&type=

It should either be removed from the settings or actually used somewhere.

Maybe this line https://github.com/opencart/opencart/blob/master/upload/catalog/controller/startup/startup.php#L36 should be.
```php
$this->registry->set('url', new Url($this->config->get('config_url'), $this->config->get('config_secure') ? $this->config->get('config_ssl') : ''));
```

In which case the home page will not be HTTPS anyway.

PR #5313, opened 2017-03-30
Update account.php

kooy:
in reg page,  /index.php?route=account/register

404 Error when select country

json query:  /index.php?route=account/account/country&country_id=


ADDCreative:
I think it's been moved to here https://github.com/opencart/opencart/blob/master/upload/catalog/controller/localisation/country.php.

See https://github.com/opencart/opencart/commit/8647e6c097afba292b1c20a5205b5e536712a76c.

PR #5330, opened 2017-04-04
Update session.php

danijelGombac:
I think we also need to check if isset api_token.


danielkerr:
i think already fixed with my last commit

danijelGombac:
No, this part is mising.

danielkerr:
i it required?

PR #5332, opened 2017-04-05
Add support for vendor libraries in system/vendor/

raxbg:
In this implementation the **vendor/** directory is being checked before the **library/** one, to allow replacing native libraries with 3rd party ones in an "elegant" way. Not sure if this is okay though. The order might need to be reversed if such an override is not going to be allowed.


danielkerr:
im going to allow extensions to use the library folder i think

PR #5336, opened 2017-04-06
Discounts/Offers Module

patricmutwiri:
A module for displaying products with discounts.. 
image


danielkerr:
thx for the code but i can not accept this. in future please ask before coming up with these types of features. maybe its better to post in the marketplace.

PR #5344, opened 2017-04-07
Update api.php

### LINKED ISSUE #5341: Api bug
If not added ip to api_ip table the error appear:

> PHP Notice:  Indirect modification of overloaded property ControllerStartupStartup::$session has no effect

In startup session, if api_query return nothing, the session missing. There should be an error for ip or remove api_ip table and look only in api_session table for ip.

danijelGombac:
Add ip when session added. Fix #5341


danielkerr:
added already

PR #5349, opened 2017-04-08
Fix send mail

Gendrop:
Message denied for spoofing attempt via SMTP Auth (From address: John Doe <john.doe@gmail.com> IP: xxx.xxx.xxx.xxx AuthenticatedID: store@mail.com Account owner: username)

The SMTP it’s blocking all emails because the FROM email, it’s different from the Authenticated one (<john.doe@gmail.com> != <store@mail.com> )

Reply-To header remains with the email address of the sender, so you can reply directly.

### DIFF
--- a/upload/system/library/mail.php
+++ b/upload/system/library/mail.php
@@ -92,7 +92,7 @@ class Mail {
 		}
 
 		$header .= 'Date: ' . date('D, d M Y H:i:s O') . PHP_EOL;
-		$header .= 'From: =?UTF-8?B?' . base64_encode($this->sender) . '?= <' . $this->from . '>' . PHP_EOL;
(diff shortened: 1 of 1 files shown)


danielkerr:
this might work only for you. it works fine for me on gmail

Gendrop:
It's not a fix only for me, it's a fix for all the new SMTP with spoofing validations.
The sender email MUST be the same authenticated email.

ADDCreative:
@Gendrop The smtp_username is not always a valid email address. So your suggested change could cause problems for some set ups.

Gendrop:
@ADDCreative, that make sense. Let me figure a fix without messing with other configurations

PR #5356, opened 2017-04-11
route=common/header is blank fixed

aarontong00:
when I visit url+?route=common/header is blank ,If you think it's ok , can you add it


danielkerr:
That's fine header is a module not set to output. Stop wasting my time

aarontong00:
I'm just a little confused, because 1.5.6 will enter the not found page

PR #5365, opened 2017-04-13
Added docker-compose w/ nginx and php7-fpm

muhammaddadu:


prhost:
@arnisjuraga please, copy to copona, nginx support is important to copona.

PR #5368, opened 2017-04-14
Fixed typo in addModification method

hauso:
`code ` contain space in field name

### DIFF
--- a/upload/admin/model/setting/modification.php
+++ b/upload/admin/model/setting/modification.php
@@ -1,7 +1,7 @@
 <?php
 class ModelSettingModification extends Model {
 	public function addModification($data) {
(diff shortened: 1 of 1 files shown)


uksitebuilder:
Please see #5370 duplicated, but committing this will invalidate an additional fix in #5370

hauso:
there is more problems with installer, as well missing 'status' parameter in	$modification_data array
what cause PHP notice, but I'm not sure what "kenius" plan to do, so I didn't created stupid fix...

uksitebuilder:
Yes, I forgot about that status error, I think I must have hard-coded the value as '1' in the SQL statement - Been trying for ages to test/install an updated extension.

Didn't think the installer would be so broken after devs were asked to test/check their extensions for OC 3k

Once our commits and the status error is sorted, I believe the installer will work though.

PR #5370, opened 2017-04-14
Fix Extension Installer OC3.0.0.0

uksitebuilder:
Moved setting_extension load model to correct an error

Fixed table column 'code' in setting_modification Insert SQL statement


hauso:
there is more problems with installer, as well missing 'status' parameter in	$modification_data array
what cause PHP notice, but I'm not sure what "kenius" plan to do, so I didn't create stupid fix...

danielkerr:
already done thx

PR #5429, opened 2017-05-16
Extra API features

RichardMarchant:
Adds the following useful features to the API

- Get product information.
- List orders (can be filtered in the same way they can in admin).
- Get list of order products for an order.

Tested and works correctly with 2.3.0.2
Should be non-breaking as it only adds new features, nothing else changes.
A use case is being able to retrieve the information needed to forward onto a warehouse to deal with the fulfilment.


willfaulds:
This looks good Richard, are you using this in a live project? I will test over the next month and suggest inclusion.

straightlight:
Denied. Against GDPR since you're suggesting to use automated scripts to capture specific information from the users without their consent. Therefore, this commit will now be closed.

mehov:
> Against GDPR since you're suggesting to use automated scripts to capture specific information from the users without their consent. 

I have looked through the code several times and I can't see any *automated scripts* or anything suspicious. All the code does is it introduces endpoints letting authenticated users list the store orders, instead of getting those one by one using IDs. It does not *capture* any more user information than had already been collected elsewhere before.

Getting a list of orders from the webshop API is a much needed feature that would allow us to do richer integrations with e.g. external warehouse software.

Seeing that neither @straightlight nor @willfaulds who replied here are active, maybe @danielkerr can step in and comment?

ultrono:
I'm 100% confused as to where the automated code is in the PR. 

The PR appears to add some basic API endpoints to retrieve order information (very common requirement for online stores). I don't see any code other than standard PHP, following the Opencart format, added here.

PR #5444, opened 2017-05-21
Language class -> get() functions was removed from controllers [small part]

hkaraoglu:
get() function callings was removed  from Controller classes which uses get function in the Language class .Instead of this, data array variable in Language class, was encapsulated to get it from another class. And this was used to merge with controller data. After merging, no need to call get function of Language class. Notice: This commit just contains small part of refactoring of controllers which use get function of Language class. If this commit is accepted, we will continue to refactor controllers.

### DIFF
--- a/upload/catalog/controller/account/account.php
+++ b/upload/catalog/controller/account/account.php
@@ -6,9 +6,7 @@ class ControllerAccountAccount extends Controller {
 
 			$this->response->redirect($this->url->link('account/login', '', true));
 		}
-
-		$this->load->language('account/account');
-		$this->data = array_merge($this->data, $this->language->getData());
+		$this->language->load('account/account', $data);
 
 		$this->document->setTitle($this->language->get('heading_title'));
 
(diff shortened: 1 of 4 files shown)


upload/catalog/controller/account/account.php
```diff
@@ -8,6 +8,7 @@ public function index() {
 		}
 
 		$this->load->language('account/account');
+		$this->data = array_merge($this->data, $this->language->getData());
```
osworx: How many times do you want to merge?
Easier and simplier:

> $this->data = $this->load->language('account/account');
if first,
> $this->data += $this->load->language('account/account');
if data already has assigned vars.

Revenge94: Good idea. I didn't realize that language method returns data.

Revenge94: I tried your advice. But it didn't work.

Revenge94:
I updated my code by your advices and load function.

PR #5449, opened 2017-05-24
Check prior to adding to user_group table to avoid duplicate entries

scottcwilson:
I was looking at this table and noticed a lot of duplicate entries for software I was working on.  The reason is that the controllers in `admin/controller/extension/extension` do a call to `addPermission()` in their `install()` function, without a corresponding `deletePermission()` in the `uninstall()` function.


danielkerr:
Doesernt matter. It's stored as a array

scottcwilson:
I should have provided a picture - here's what happens when you install and remove the coupon order total module half a dozen times. You can see there are a bunch of entries in the access array for the Administrator record for the user_group table.  Duplicate keys just keep getting added.

screen shot 2017-05-27 at 6 18 12 am

PR #5455, opened 2017-05-27
Added addStyleVersioned and addScriptVersioned methods to document.php

krylov123:
Added methods to automatically versioning css and js files.

For example, convert

- catalog/view/javascript/common.js

string to

- catalog/view/javascript/common.js?v=18469e0552

based on a hash of the file.


danielkerr:
this does not really fit into opencart cart as an end product.

PR #5463, opened 2017-05-30
Admin - Validate the file max size on Filemanager

arturbonnett:
It is in the settings but is not implemented in practice


danielkerr:
i think max size really needs to be done with js first

arturbonnett:
I totaly agree it would be better to have js first as well! It's a good thing to implement.
But this simple validation in php is still necessary

danielkerr:
No it's not because like most PHP systems it will throw an error if u
uploaded a file large than the max upload file size PHP allows is used.

arturbonnett:
The point is not to let anyone send files in the maximum size that the server's PHP allows, but the maximum size that administrator limit on OpenCart settings.
This field even already exists in settings but it's there only to deceive the administrators, since it is not being applied!
Allow very large file upload, can consequently cause 'ModelToolImage::resize' to result in 500 error.
Making it possible to only fix via FTP. Unnecessary bug...

danielkerr:
Yes and to get the size of the upload it needs to be uploaded first. There
by throwing errors.

danielkerr:
It needs to be done with js. Doing with PHP does no make a difference.

Issue #5461, opened 2017-05-29
Opencart - Customer orders report is working improperly.


prakannino:
Dear Opencart Team, 

At the moment, I'm using Opencart VERSION : 2.0.1.1 and I am experiencing a problem with **Customer orders report.** 

I've tried to fix this issue but seems it still has a bug that I've found was when I reached 20 orders, the customer orders report doesn't work anymore. 

Could you please take a look on to this matter, please?  

Refer to the link below : 
https://github.com/opencart/opencart/pull/3105
https://github.com/opencart/opencart/commit/e30de743e09537ffdb03455f8a0768dc6342d89f

_Noted: I've tried to fix the file at path : upload/admin/model/report/customer.php by these solution above. But it doesn't work._ 

![customer orders report 1](https://cloud.githubusercontent.com/assets/29037308/26555865/5a0476b4-44c1-11e7-95b8-6c672f60e5fc.png)

![customer orders report 2](https://cloud.githubusercontent.com/assets/29037308/26555868/5fd51eb8-44c1-11e7-9c1c-da2c4af46a95.png)

Thank you for your help in advance.

mrnfrancesco:
The two values showed in your images come from different model and different queries.

The **TOTAL ORDERS** value in the dashboard comes from the `getTotalOrders` function in `upload/admin/model/sale/order.php`, which do the following query:

`SELECT COUNT(*) AS total FROM order`

The **No. Orders** value comes from `getOrders` function in `upload/admin/model/report/customer.php` with the following query which is a bit more complex:

```
SELECT
    t.customer_id,
    t.customer,
    t.email,
    t.customer_group,
    t.status,
    COUNT(DISTINCT t.order_id) AS orders,    <== THIS IS THE PRINTED VALUE
    SUM(t.products) AS products,
    SUM(t.total) AS total
FROM (

    SELECT
        c.customer_id,
        CONCAT(c.firstname, ' ', c.lastname) AS customer,
        c.email,
        cgd.name AS customer_group,
        c.status,
        o.order_id,
        SUM(op.quantity) as products,
        SUM(o.total) AS total
    FROM order o
    LEFT JOIN order_product op ON (o.order_id = op.order_id)
    LEFT JOIN customer c ON (o.customer_id = c.customer_id)
    LEFT JOIN customer_group_description cgd ON (c.customer_group_id = cgd.customer_group_id)
    WHERE o.customer_id > 0 AND cgd.language_id = '<your language id>'

) AS t
[...]

Issue #2874, opened 2015-04-06
Wrong query in admin/model/report/product


znvPredatoR:
In function getPurchased total sum is calculated as (**total** + tax) * quantity, when it should really be (**price** + tax) * quantity. 

```
public function getPurchased($data = array()) {
		$sql = "SELECT op.name, op.model, SUM(op.quantity) AS quantity, SUM((op.total+ op.tax) * op.quantity) AS total FROM " . DB_PREFIX . "order_product op LEFT JOIN `" . DB_PREFIX . "order` o ON (op.order_id = o.order_id)";
```

danielkerr:
which version??

the latest version does not have this code! also i suggest you look for a fix on the current branch

znvPredatoR:
Got it, will look next time.

iLevye:
v2.0.1.0

Issue #2588, opened 2015-01-21
OC 2.0.1.1 Reports>Products>Purchased - Inflated amounts in Totals column


shalom-cohen:
When running reports to view Purchased Products, the Totals column duplicates the amounts charged.

The file responsible for this calculation is located in:
webroot / admin / model / report / product.php

The original line of code (line 34), sums up the model count, but then also multiplies the total + tax  and multiplies it with the quantity once again, inflating the results as shown below:

$sql = "SELECT op.name, op.model, SUM(op.quantity) AS quantity, SUM((op.total + op.tax) * op.quantity) AS total FROM " . DB_PREFIX . "order_product op LEFT JOIN `" . DB_PREFIX . "order` o ON (op.order_id = o.order_id)";

To correct this issue, you need to remove the second instance of multiplying by quantity as follows:

$sql = "SELECT op.name, op.model, SUM(op.quantity) AS quantity, SUM(op.total + op.tax) AS total FROM " . DB_PREFIX . "order_product op LEFT JOIN `" . DB_PREFIX . "order` o ON (op.order_id = o.order_id)";

Tea56:
I am seeing the total inflation when editing same order. (quantity is going up and sometimes a second instance of the product gets added.) 
As i'm new to opencart and have some modifications, i'm not really sure if it's a bug, but it sounds exactly like what you're describing.

Any help would be appreciated.

shalom-cohen:
The problem you are describing seems to be with another module of the OC system, you may want to create a completely new ticket for this one.

Tea56:
Yeah, i already did :) Thank you!

danielkerr:
@shalom-cohen this is not the case. I think you have your prices and tax setup wrong maybe.

shalom-cohen:
I've reviewed the results of my findings before posting them online. I stand behind my original post. If you feel I am incorrect then I recommend you substantiate your claims.

milimar:
I tried to check this issue myself, so I found out that shalom-cohen has right, the totals in the total column are calculated incorrectly. The issue appears only if more then one product is bought at a time. I used opencart 2.0.1.2_rc from about one week ago. I hope this helps.

ADDCreative:
This is an issue. It was cause by https://github.com/opencart/opencart/commit/b22942a149fba290e1379b8ce2e653717320beb1, which was tying to fix another issue.

The calculation should be using op.price not op.total, assuming op.tax is the same as in 1.5.x.
```php
[...]

Issue #1014, opened 2013-11-21
Totals incorrect in Reports>Products>Purchased


stokeyblokey:
The sum is incorrectly working out product prices as shown below:

Example Product A is £15.99 including 20% VAT - net price is £13.3250 and VAT is £2.665

The sum in the report is doing this:

SUM(op.total + op.total \* op.tax / 100) AS total where op.total is op.price \* op.quantity

This is calling the op.tax field value and it should not be! This field is a value, not a tax rate, so it varies for every item!

So for the above example it calculates thus:

SUM(13.3250 + ((13.3250*2.665)/100)) = £13.68

It SHOULD be calculating like this:

SUM(13.3250 + ((13.3250*20)/100)) = £15.99

Potential fix:

Open admin\model\report\product.php

Find:

public function getPurchased($data = array()) {
$sql = "SELECT op.name, op.model, SUM(op.quantity) AS quantity, SUM(op.total + op.total \* op.tax / 100) AS total FROM " . DB_PREFIX . "order_product op LEFT JOIN `" . DB_PREFIX . "order` o ON (op.order_id = o.order_id)";

Change to:

public function getPurchased($data = array()) {
$sql = "SELECT op.name, op.model, SUM(op.quantity) AS quantity, SUM(op.total + op.tax) AS total FROM " . DB_PREFIX . "order_product op LEFT JOIN `" . DB_PREFIX . "order` o ON (op.order_id = o.order_id)";

LennyLip:
+1 opencart 2.0.3 Totals incorrect

arnisjuraga:
Huge!

PR #5467, opened 2017-06-01
Fixed a calculation error on the product purchased report page

Skeletbossen:
The calculation on the page was wrong, but would only appear wrong if
the quantity of the product was more than one per order.

### DIFF
--- a/upload/admin/model/report/product.php
+++ b/upload/admin/model/report/product.php
@@ -37,7 +37,7 @@ class ModelReportProduct extends Model {
 	}
 
 	public function getPurchased($data = array()) {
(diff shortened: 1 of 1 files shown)


RaidoDev:
This has been a problem for a long time now, and tried to change, discussed multiple times:

#5144
#2874  
https://github.com/opencart/opencart/issues/2588#issuecomment-73496710
#1014
#909



with different solutions

528625adb2581d2fcaf2d943b29a5b5c78b4a0a3
663d2319e978d6d1303ec75a09147f0ed1bb561f
b22942a149fba290e1379b8ce2e653717320beb1
4505399df38227af3207c8d79dfee4c6b6890c8b



but still having problems...
Is there too many modules out there, which depends on these line of codes or what?

prakannino:
Hi @RaidoDev, 

I think it not the same problem with #5461

Let's take a look onto the link as below. I think you could find the solution out there! 

https://forum.opencart.com/viewtopic.php?t=114760

Good luck.

RaidoDev:
This one is still not fixed, I suppose.

PR #5481, opened 2017-06-06
SagePay Server - Prevent Order ID mismatch

jtaylor69:
When a customer proceeds to SagePay and then hits the back button in their browser, the same order ID is kept in session.
The current addOrder function allows another database entry with the same order ID, but a different SecurityKey and other details are generated at SagePay. Future calls within the callback function tries to retrieve by Order ID, which will cause an issue as multiple Order IDs are present. This can cause an issue where the signatures do not match, due to a certain detail like the SecurityKey being different.
By removing old orders using the additional database query within this PR, this ensures that the order is current and allows the transaction to complete without error (providing the customer's card details are correct!)

### DIFF
--- a/upload/catalog/model/extension/payment/sagepay_server.php
+++ b/upload/catalog/model/extension/payment/sagepay_server.php
@@ -70,6 +70,7 @@ class ModelExtensionPaymentSagePayServer extends Model {
 	}
 
 	public function addOrder($order_info) {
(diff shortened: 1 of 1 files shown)


danielkerr:
i think u have messed this fix up. if some one clicks the back button a new order needs to be created not overridden!

PR #5484, opened 2017-06-06
Regex add ( After OR Before ) - without ( Replace )

RegEx20170606:
it's helping developer to write less lines & code , Regex after or before that mean searching easier ! less search token "search tag" with same "add tag" codes .
e.g :
The following regex it could replace 8 different lines :
``~$this->load->language('(catalog|soconfig)/(mvd_|vdi_|m|)product~``
but if you don't want to replace , want to add before or after , this update for this purpose !
You can adapt many themes with your module in OCMOD !

Please @danielkerr Confirm it.
Say goodbye to VQMOD

### DIFF
--- a/upload/admin/controller/extension/modification.php
+++ b/upload/admin/controller/extension/modification.php
@@ -225,8 +225,7 @@ class ControllerExtensionModification extends Controller {
 
 										$status = false;
 
-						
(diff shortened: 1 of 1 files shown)


straightlight:
Cannot find this line entry to look for from the master branch. Therefore, this commit will now be closed.

PR #5502, opened 2017-06-10
Update header.twig

OpenCartAddons:
Added back separate store dropdown list.  Profile is under it's own dropdown.


danielkerr:
my graphic design has been re workign the admin. hopefully have a new one soon.

PR #5537, opened 2017-06-19
hidden opencart font-awesome icon on mobile view

JesseOC:
hidden opencart font-awesome icon on mobile view


RaidoDev:
Why not `display: none;`

PR #5569, opened 2017-06-21
twig library move to composer

chenly:


danielkerr:
maybe at a later stage

PR #5597, opened 2017-06-23
Dashboard map update

JackWillDavis:
**Issue**
The map on the dashboard is slightly misleading as it includes country summaries across all orders including failed/declined etc. This pull request adds an order_status_id filter to the getTotalOrdersByCountry model function and passes the config_complete_status details to only pass back amounts from orders which have actually been completed/shipped.

Store owners can then see a country breakdown from successful orders and get a better insight into actual totals.

**Original Commit Details**
Updated the getTotalOrdersByCountry function to use an order_status_id filter.

Applied to dashboard map model call to ensure that only completed orders show up on the map dashboard extension.

### DIFF
--- a/upload/admin/controller/extension/dashboard/map.php
+++ b/upload/admin/controller/extension/dashboard/map.php
@@ -114,7 +114,7 @@ class ControllerExtensionDashboardMap extends Controller {
 
 		$this->load->model('report/sale');
 
-		$results = $this->model_report_sale->getTotalOrdersByCountry();
+		$results = $this->model_report_sale->getTotalOrdersByCountry($this->config->get('config_complete_status'));
 
 		foreach ($results as $result) {
 			$json[strtolower($result['iso_code_2'])] = array(
--- a/upload/admin/controller/extension/dashboard/map.php
+++ b/upload/admin/controller/extension/dashboard/map.php
@@ -114,7 +114,7 @@ class ControllerExtensionDashboardMap extends Controller {
 
 		$this->load->model('report/sale');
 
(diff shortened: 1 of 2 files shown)


CoverUp:
no drop-down menu to choose what to see?

JackWillDavis:
Personally, I'm not sure what the value would be to allow store owners to change it to different order types. They could see chargebacks by country I supposed but for the average user, the dashboard offers a snapshot view of orders.

I could see that having a quick view of where 'cash-in-the-bank' orders have come from as being far more useful than anything else. Certainly having all orders lumped together, even if they were cancelled/refunded/failed etc.

CoverUp:
Including the failed/declined order also provides insights about customers location. For some store owners this provide feedback about what payment methods to implement. For example, in one of our customer's stores, all completed orders are from the same country only using a local payment method (3rd party gateway not native to opencart), and most of the missing/declined order come from other countries, this lead to enable paypal for those countries as a sale desision. Just looking the dashboard, if we fix the dashboard to only completed orders we lose this information. Thus a drop-down menu or dash config for "completed only", "missing orders" or "all orders" make sense to us.

JackWillDavis:
Hmm OK, I can see where you're coming from. That would probably be something I would personally drill down into using the reports section.

If there is support for this from other people, I'd be happy to amend this pull request.

danielkerr:
it needs to be done from the new stats table as stores with large number of orders will be slow when visiting the dashboard page

danielkerr:
some bad hackable code there!

JackWillDavis:
OK. Changed to a boolean and ensures only integers are used for checking against order status ID in the model.

In terms of the stats table, I don't believe that is being used in the default dashboard map model call?

PR #5609, opened 2017-06-25
Return error if not isset g-recaptcha-response

danijelGombac:
Return error if not isset g-recaptcha-response, because maybe captcha was removed from the form.


ADDCreative:
That would not return an error. It would mean captcha would pass validation if isset g-recaptcha-response was no set.

Even without this change there looks to be an issue that the captcha only has to be passed once per session.

PR #5640, opened 2017-06-30
Fix backup.php for Windows and Linux

ManoelVidal:
Replace:
ini_get('upload_tmp_dir')

For:
DIR_UPLOAD

Same as installer.php


danielkerr:
why is this not working on your server?

ManoelVidal:
In Windows Server does not
With the fix... OK 👍

PR #5656, opened 2017-07-03
Fix installer

ManoelVidal:
Fix installer for not install zip filename already installed

If zip filename already exists in extension_install table, extension not install


ManoelVidal:
Not approved by beta testers

PR #5668, opened 2017-07-04
Fix install step_3.php in write config.php

ManoelVidal:
After install...

Parse error: syntax error, unexpected 'language' (T_STRING), expecting ',' or ')' in

Error:
config.php
admin/config.php


danielkerr:
fixing now. and rereleasing

PR #5675, opened 2017-07-05
Fix table responsive in order_list.twig

ManoelVidal:
Add in table:
`<div class="table-responsive"></div>`


danielkerr:
yo! if u commit something make sure u dont remove all the js tabs instead of spaces! wasting my time fixing it.

PR #5693, opened 2017-07-07
Consistent delete cache, event methods and code clean up 

howdu:


howdu:
foo2 should not be deleted. Testing caching between file, memcache php5.6 and memcached php7.0. 

```php
$this->cache->set('foo', 'bar');
$this->cache->set('foo2', 'bar2');

$this->cache->delete('foo');

$foo = $this->cache->get('foo');
var_dump($foo);

$foo2 = $this->cache->get('foo2');
var_dump($foo2);
```

PR #5701, opened 2017-07-08
Admin panel design improvements

evpadallas:
Added button grouping.
Added Row link from http://www.jasny.net/bootstrap/javascript/#rowlink.
Added Input mask to ISBN field from http://www.jasny.net/bootstrap/javascript/#inputmask


danielkerr:
im not keen on the changes but thx.

PR #5706, opened 2017-07-09
Fix strange codes in pp_braintree.php

ManoelVidal:


danielkerr:
maybe i slept on the keyboard.

ManoelVidal:
Already happened to me :)

PR #5747, opened 2017-07-15
Set date_modified adding new product

howdu:


danielkerr:
added manually because u included all the past merged fixes.

PR #5774, opened 2017-07-20
Simplified code in coupon and reward totals

gunjanpatel:
There is no functionality bug here just improved the code to make it simple to understand.


danielkerr:
its already simple enough!

PR #5776, opened 2017-07-21
Update image.php

thesamarkand:
These changes will make images in OpenCart 100% optimized for Google Page Speed Insights from the box.
https://developers.google.com/speed/pagespeed/insights/
imageinterlace($this->image, true);  // reduces image size up to 10%
$img->stripImage(); // removes GD banner

### DIFF
--- a/upload/system/library/image.php
+++ b/upload/system/library/image.php
@@ -118,14 +118,23 @@ class Image {
 
 		if (is_resource($this->image)) {
 			if ($extension == 'jpeg' || $extension == 'jpg') {
-				imagejpeg($this->image, $file, $quality);
+				imageinterlace($this->image, true);
+				$result = imagejpeg($this->image, $file, $quality);
 			} elseif ($extension == 'png') {
-				imagepng($this->image, $file);
+				$result = imagepng($this->image, $file);
 			} elseif ($extension == 'gif') {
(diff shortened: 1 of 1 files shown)


danielkerr:
"These changes will make images in OpenCart 100% optimized for Google Page Speed Insights"

i diagree. it looks like u have added compatibility issue for servers requiring them to have image magic.

zemguru:
Actually, the idea of interlacing is quite good. And that does not require `Imagick`
https://github.com/opencart/opencart/pull/5776/files?diff=split#diff-485cbf7119ac6afdc150f24673daef30R121

`Imagick` is used additionally for stripping metadata from images. And is done correctly by checking class availability - no problems with servers without `Imagick`.
https://github.com/opencart/opencart/pull/5776/files#diff-485cbf7119ac6afdc150f24673daef30R133

zemguru:
My bad - just checking of `Imagick` class availability is not enough. It will still throw exception, if server has no support of image encodings for Imagick.

thesamarkand:
> My bad - just checking of Imagick class availability is not enough
I don't have hosting without Image Magick, so I can't test this, but testing extension_loaded('imagick') may help.
https://stackoverflow.com/questions/1475830/why-would-a-loaded-php-extension-fail-on-class-instantiation

PR #5787, opened 2017-07-24
Free checkout status missing payment prefix

howdu:


danielkerr:
added manually as i can not see from your commit the full changes

vaughnchill:
Also was facing the same problem, and this added these changes which fixed the issue.  Thank you.

ProfessorLinebeck:
Hm seems not to work for me :/ I have made all the changes and then updated via the install page. It also said update 1010 was succesful. Do i maybe need to do something with the sql file?

PR #5797, opened 2017-07-26
dashboard.php > check for install dir ..

osworx:
.. does not work, because DIR_APPLICATION contain admin and the install dir is usually above.
Recommend to use DIR_SYSTEM (and replace system/), because admin could be changed while system not.


danielkerr:
catalog would be better

PR #5798, opened 2017-07-26
Update password hashing to use PHP's password_hash() function

goofygoblin:
PHP added a password_hash() function in version 5.5 which over time will automatically update the hashing method used as more secure options become available. This modification will allow backward comparability with old OpenCart hashes, which will be automatically updated when a customer or admin user logs in. This modification does require the password field in the customers and users database tables be modified to 255 in length for current compatibility and future proofing.


danielkerr:
good one!

danielkerr:
i will test before release, but please help and just double check everything.

goofygoblin:
Thanks.

I ran through the following tests and everything seems to work, but no harm in someone else double checking;

- [ ] Register a new customer
- [ ] Login as customer
- [ ] Login as customer with password set as SHA1() type
- [ ] Login as customer with password set as MD5() type
- [ ] Change customer password from account settings
- [ ] Login to Admin panel
- [ ] Login to Admin panel with password set as SHA1() type
- [ ] Login to Admin panel with password set as MD5() type
- [ ] Change Admin user password
- [ ] Create Admin user
- [ ] Login as customer from Admin panel

I think that covers everything.

If anyone is doing testing here is the SQL function for the SHA1 passwords so you can manually set them. Just remember to set the salt field as well. MD5() passwords were are salted.
SHA1(CONCAT("salt", SHA1(CONCAT("salt", SHA1("password")))))

goofygoblin:
Found an error in /catalog/model/account/customer.php. In the function editPassword() I've used $data['password'] instead of $password in the query.

I've fixed it in my fork, should I put in another pull request or raise a new issue? (yeah, I'm new to github)

JJKalnins:
Just a quick question. Isn't this the same thing? Because this seems identical to #5234

danijelGombac:
$password_hashed in

```
if (password_needs_rehash($password_hashed, PASSWORD_DEFAULT)) {
	$new_password_hashed = password_hash($password, PASSWORD_DEFAULT);
}
```

is Undefined variable.

goofygoblin:
See pull request #5844 for fixes to the undefined variables mentioned above.

danielkerr:
u forgot the installer. im adding it now

photographer-mememe:
Would it be better to use html_entity_decode both at the password check and database entry?

Let's say the password limit at registration is 40 characters, you enter 40 characters, and a special character pushes it over 40 and triggers an error at validate. html_entity_decode would bring it back down to 40.

Special characters without html_entity_decode would also artificially increase short passwords over the minimum character limit.

But existing passwords with special characters already in the database that weren't hashed with html_entity_decode would break and need to be reset.

PR #5812, opened 2017-07-28
Fix possible SQL injection vulnerability

PetrHeinz:
should fix vulnerability to SQL injection via multi-byte attack when improper character set is set at the server level,
see http://php.net/manual/en/mysqlinfo.concepts.charset.php for details

I have spotted this issue while answering a question on Stack Overflow https://stackoverflow.com/questions/45378258/php-sql-injection-possibility/45379001.

I hope I am not mistaken as it is some time I've worked with `mysql_*` functions.

### DIFF
--- a/upload/system/library/db/mysql.php
+++ b/upload/system/library/db/mysql.php
@@ -14,7 +14,7 @@ final class MySQL {
 		}
 
 		mysql_query("SET NAMES 'utf8'", $this->connection);
-		mysql_query("SET CHARACTER SET utf8", $this->connection);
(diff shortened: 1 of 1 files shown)


PetrHeinz:
If my suspicion is correct, it's been there since the initial commit: https://github.com/opencart/opencart/blob/b9542883d77a6188462814cbdf6ba0b250b40d1b/upload/system/database/mysql.php#L17

danielkerr:
ill merge your pull request but like i tell many people test the code before you start posting SQL injection vulnerability. as all ur doing is guessing that opencart is vulnerable!

PR #5818, opened 2017-07-30
Fixed #5817 - Correct install config.php permissions an existence check

### LINKED ISSUE #5817: Incorrect error variable and permissions check order on install step_2
When you run install on 3.0.2.1.b, without created `config.php` files, you receive an incorrect message:

image

Even though - both files `config.php` and `admin/config.php` does not exists!

I will create PR for this, please, review.

zemguru:


danielkerr:
already added in last commit

Issue #4295, opened 2016-04-26
HTML Content Module - Code View Changes Are Not Saved


codesticle:
When editing an HTML Content module, if changes are made in the code view and then the module is saved while still in code view, the changes are not saved. 

In order to make the changes stay, the changes need to either be made in the regular view, or made in the code view, and then the code view needs to be toggled off before saving.

kstampo:
Great!!! I lost so much time, wondering where is my code.... If i only knew that prior to save I had to toggle off code view....

danielkerr:
you need to complain to summer not because if i implemented a fix it would not be a very good one.

danielkerr:
or find me another solution out there that does not require me to build my own.

a-salihov:
Is this issue fixed? There isn't patch code in OC 3.0.3.2

ADDCreative:
Summernote was dropped and replaced with CKEditor in the 3.1 alpha and the patch never made it into 3.0.3.2. You could try this workaround. https://forum.opencart.com/viewtopic.php?f=201&t=213287&hilit=summernote#p763240

PR #5851, opened 2017-08-04
Update summernote.js

billynoah:
Addresses https://github.com/opencart/opencart/issues/4295.  Patch to update contents from code view without needing to switch back to wysiwyg mode.


danielkerr:
Big thx

PR #5859, opened 2017-08-07
Added missing 'required' to custom fields

howdu:
Also remove duplicate time and use twig elseif syntax


danielkerr:
all custom fields are required for these pages becasue the system hides or displays the fields depending on which customer group is selected

howdu:
The changes are for when you have a default customer group selection and no selection. On first load there's no customer group id passed and missed required data from custom_field_customer_group

```php
$custom_fields = $this->model_account_custom_field->getCustomFields();```

danielkerr:
It forces default any way.

PR #5864, opened 2017-08-08
Missing language value

evpadallas:


ADDCreative:
I believe this is wrong. It will show the message to the admin in the customer's language not the admin's.

This is because mail/voucher is loaded based on the order's language.
https://github.com/opencart/opencart/blob/master/upload/admin/controller/sale/voucher.php#L588-L590

The text is already in https://github.com/opencart/opencart/blob/master/upload/admin/language/en-gb/sale/voucher.php#L10.

Just need to change https://github.com/opencart/opencart/blob/master/upload/admin/controller/sale/voucher.php#L550 to ```$this->load->language('sale/voucher');```.

This will also fix the missing error_permission caulse by https://github.com/opencart/opencart/blob/master/upload/admin/controller/sale/voucher.php#L555, as error_permission in not in mail/voucher either.

PR #5867, opened 2017-08-09
mysqli error message check for connect_errno

howdu:
Required with mysql5.7


howdu:
https://github.com/opencart/opencart/pull/5867

PR #5880, opened 2017-08-17
Add validations and friendly message in view for the featured module.

calevano:
This change fixes the featured module to show right elements on the home page.

cc @danielkerr


danielkerr:
a completly useless fix for a non issue!

PR #5904, opened 2017-08-24
Enabling code completion for common magic properties

filiplikavcan:
This PR enables code completion for common magic properties by adding the @property in the DocBlock of the Controller class.

https://stackoverflow.com/questions/3814733/code-completion-for-private-protected-member-variables-when-using-magic-get/3815198#3815198

If you agree I will add @property annotations to other classes.

### DIFF
--- a/upload/system/engine/controller.php
+++ b/upload/system/engine/controller.php
@@ -9,6 +9,15 @@
 
 /**
 * Controller class
+*
+* @property Document document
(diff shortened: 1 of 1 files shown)


filiplikavcan:
screen shot 2017-08-24 at 13 04 30

It would allow us to use code inspection and auto complete which every decent IDE provides.

PR #5918, opened 2017-08-29
fix display rating

AndreyZaets:
duplicated from product/product.twig


danielkerr:
issue? trhe issue is the size? works fine for me and how it should!

PR #5956, opened 2017-09-08
bootstrap toggles2 - use opacity in btn-group only for btn-success & btn-danger

stalker780:
use opacity in btn-group only for btn-success & btn-danger
in this case on|off are contrast and marketplace group switch doesn't change


danielkerr:
bootstrap was updated a long time ago. I'm pretty sure this is fixed.

PR #5958, opened 2017-09-09
Update contact.php

DMConklin:
$text_message not set


danielkerr:
there is no text_success in the language file

PR #5960, opened 2017-09-09
Update order.php

DMConklin:
An artifact was showing up in account > order history > order information right before the order id.

order-artifact

Now, I'm unfamiliar with twig and it's quirks but to me, it seemed as if maybe it was attempting to pull in a missing variable so I looked in the controller and noticed text_order_id wasn't defined and after defining it the artifact disappeared.

Also, I noticed that
```php
 $data['text_invoice_no'] = $this->language->get('text_invoice_no');
```
has been removed from github.
Why?  Am I missing something?


danielkerr:
can replicate. m,aybe your using an old language pack

PR #5976, opened 2017-09-18
Update smtp.php

Shinrai:
Fixes issues with SMTP Servers running policy servers which make the SMTP servers hang for a few milliseconds. In my specific case the server was hanging for 50 milliseconds to 100 milliseconds before a reply was sent. After this change the emails were sent without an issue.


CoverUp:
seems good

PR #5981, opened 2017-09-20
support billing address and ship-to name validation

chrislim2888:
Updated the fraudlabspro extension to support validation for
* Billing street address
* Ship-to first name and last name


chrislim2888:
a few more update needed

PR #5992, opened 2017-09-23
Added Edit button to Menu extensions

zijege:
menu_edit


danielkerr:
if u are modifying the menu try using the events to override the headers menu var that goes in the template.

PR #6044, opened 2017-10-03
Update backup.php

danijelGombac:


straightlight:
The 'ser_group' should be replaced with: 'user_group' from the replacement code.

PR #6046, opened 2017-10-03
text_message missing

cyberfountain:
Missing data for the text_message variable after form submission


danielkerr:
there is already a fix in the master

cyberfountain:
Cool, thanks Daniel!

Issue #6045, opened 2017-10-03
Modification can not be freshed.


vingoyuan:
i am a beginner for OCMOD.
first time i upload a  install.xml, just want to change some code in catalog/view/theme/default/template/account/order_list.twig.

at first i want to change status field  to print "abc", it succeed,  but after some  operations, like update and delete in Extension.   now i want to see " 1234" in status field.  but it still shows "abc", when i enable a  new Extension( with  1234.).   when i disenable new Extension, it comes back to original. 

[t9.ocmod.zip](https://github.com/opencart/opencart/files/1352434/t9.ocmod.zip)

vingoyuan:
i mean,  right now, whatever i want to show "123" or something else, it still shows "abc", it seems like , some cache still there.  even after i checked database , table "modification", there is no record about "abc"

vingoyuan:
now i checked, in database  , table modification,    field xml has been changed to 1234, but order history of front side  still shows  abc.     should i   refresh any cache?

straightlight:
I did reported odd activities last week about the auto-refresh feature from the admin extensions page, as I was able to reproduce the issue on my end. The cache may be one of the source of problems where the content does not change in realtime.

vingoyuan:
if the auto-refresh feature can not works in realtime,   where should i test my xml file during development phase?   is there any other ways, that i can test my files without multiple  uploading in the admin extensions page?  because this is only way i know  right now.

stalker780:
@vingoyuan
If you are a developer you should have your own instruments for developing/testing XMLs.
Installer/Modifications was build basically for customers to let them install extension and forget.

Try to search at Marketplace https://www.opencart.com/index.php?route=marketplace/extension&filter_download_id=56&filter_search=ocmod%20editor

There are several OCMOD editors, even free ones. Some of them allow save+refresh+clear twig&sass cache.

vingoyuan:
@stalker780 ， thank you for advices.    

i just reused vqmod, and it works for opencart 3.x.    during developing/testing phase, should i use vqmod ?    

but  i still don't understand, why there is  the problem i described. 

thanks a lot

stalker780:
@vingoyuan
VQMod is a different thing with a different syntax. You told here about OCMod...

Try clearing or disabling twig & sass cache during development. Dashboard -> gear icon on right top.

vingoyuan:
@stalker780 , thank you very much, it is working.  i just cleared sass  cache, and Extension works in realtime。 


[...]

PR #6049, opened 2017-10-03
Clear twig & sass cache on modifications refresh

stalker780:
Clear twig & sass cache on modifications refresh. 
A lot of users and developers still don't know how to clear template cache in OC3.
This will solve most issues on installing/disabling modifications.

#6045


straightlight:
This controller file is no longer in-use on the master branch. Therefore, this commit will now be closed.

PR #6051, opened 2017-10-04
Adds a root folder constant in config.php

NikoChiu:
Adds a root folder constant in the config files to make migration easier.


danielkerr:
not required.

PR #6085, opened 2017-10-13
Update the marketing report page to support the new marketing report …

AlexChanHK:
…table for the click count and sales of that marketing object.

Also, adding new filter such as country and group by function.
However, I did not group the result of click and sales of the maketing record into one table since there are some techial issue when I try to join order and marketing report together.


straightlight:
If only a partial solution is provided due to technical issues, you either need to report them or provide a complete solution. Therefore, this commit will now be closed.

PR #6087, opened 2017-10-13
Fix a simple typos in Admin / Model / Catalog / Recurring.php

lucasjkr:


danielkerr:
i fixed how it should be coded.

PR #6103, opened 2017-10-15
Deleted two unnecessary queries

lucasjkr:
The delete query in addGeoZone is unnecessary because the related record won't exist - the zone_id it's was just generated in the preceding query.

In editGeoZone  - there is already a DELETE query preceding the foreach loop which would have deleted anything that the more specific DELETE in the loop would have found.

Both unnecessary.


danielkerr:
no this is so the zones with in the geo zones are not duplicated

PR #6124, opened 2017-10-20
Update modification.php

stalker780:
Logs are written not correctly

If file was already loaded by one of modifications, another modification changing this file will never write it to log and we get strange log like this:

```
MOD: My mod for OC3
CODE: $data['add'] = $this->url->link(
LINE: 326
```

But it should be:

```
MOD: My mod for OC3

FILE: admin/controller/catalog/product.php
CODE: $data['add'] = $this->url->link(
LINE: 326
```

It's hard to debug if you have many modifications. Also I'm working on ocmod.log parser extension and this problem doesn't allow me to parse what file was changed by what modification.


straightlight:
This file is no longer listed in the master branch. Therefore, this commit will now be closed.

PR #6128, opened 2017-10-22
removed .thumbnails and .image-additional duplicate selectors

enescala:


danielkerr:
i think was remove a while ago

PR #6129, opened 2017-10-22
fix repeated query criteria

lucasjkr:
Not an error, it worked before, but no reason to repeat the select criteria

(look at WHERE on line 13 - cf.status = '1' appears twice)


straightlight:
Not happening. This commit will now be closed.

PR #6131, opened 2017-10-23
Hide the discounts if the user is not logged in and config_customer_p…

tom-it:
I noticed discounted prices were being displayed when the option show prices when logged in is on, this small change hides the discounts if the user is not logged in and config_customer_price = 1


straightlight:
Might be useful for the features-pre in the future.

PR #6162, opened 2017-11-02
Add filter by category to products page in admin

pcmanik:
Filtering by category is really useful and it should be packed in default opencart installation.


pcmanik:
Hi could these be merged already? It's more than one year...

PR #6182, opened 2017-11-09
 Changes to be committed:

ichenyonghua:
fixed install errors.
	modified:   upload/install/controller/install/step_3.php
	modified:   upload/install/model/install/install.php


danielkerr:
i added it manually.

PR #6194, opened 2017-11-13
Remove 'height: ...' to allow editor to expand as needed

Mortuus-Medicus:
in 
admin/view/javascript/codemirror/lib/codemirror.css

in the class CodeMirror, please remove height property, so that when you code in the admin panel, the size of the editor is adjusted to the size of the code, instead of having only about 16 lines visible.


danielkerr:
u need to clean up the commit

danielkerr:
also it does expand by height

Mortuus-Medicus:
Sorry, first time doing this, by 'clean up the commit' you mean, so that all these other commits aren't visible?

If it does expand, than that's grand, Mine wasn't, mine had a fixed height. But if it does, that that's awesome ^^

PR #6196, opened 2017-11-14
Cleanup

AJenbo:
Hi, heres some quick cleanups I did to the url class. Depending on how you feel about them I might do some of the changes on a wider scale.

P.s. it would be nice if /tests/phpcs/OpenCart/ruleset.xml was moved to /phpcs.xml so that it was auto detected by PHPCS. Also a .editorconfig would be nice and is simple to make: http://editorconfig.org/.

### DIFF
--- a/upload/system/library/url.php
+++ b/upload/system/library/url.php
@@ -13,7 +13,7 @@
 class Url {
 	private $url;
 	private $ssl;
-	private $rewrite = array();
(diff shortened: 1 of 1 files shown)


danielkerr:
we dont use short hand arrays.

danielkerr:
also needs some documentation writing not lines removed

PR #6205, opened 2017-11-16
Remove link SSL state from controllers

ChrisLancs:
This pull request affects most of the files in the project where the SSL state is being set on a per link basis, meaning all links should now be SSL if SSL is enabled in the admin area.

All changes in controllers on lines with **$this->url->link**

-   Removed parameter 3 on all links
-   Removed parameter two where it was only used to allow parameter 3.

Also changed

store_root/system/library/url.php

- Removed the third $secure parameter from the link method constructor
- Removed the $secure check from the if block in the link method

In the final commit I fixed what i missed in the startup and framework files and so added a config_secure check to both startup and framework. that i assumed incorrectly was already happening.

Because of the checks in startup and framework only one of the two urls is passed is passed into the url class at any one time. This resulted in being able to remove the ssl object entirely including the previously edited if block from the link method.

I just noticed also that I have not modified the doc comments within the Url class, due to this being already more commits than i would have liked i will leave them until this request has been handled..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,5 +60,5 @@
 
 # IDE Project files
 /.idea
-/nbproject/
+/nbproject/*
 /nbproject/private/
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@
 
 # IDE Project files
 /.idea
+/nbproject/private/
--- a/.gitignore
+++ b/.gitignore
@@ -60,4 +60,5 @@
 
 # IDE Project files
 /.idea
+/nbproject/
 /nbproject/private/
--- a/nbproject/project.properties
+++ b/nbproject/project.properties
@@ -1,0 +1,7 @@
+include.path=${php.global.include.path}
+php.version=PHP_70
+source.encoding=UTF-8
+src.dir=.
+tags.asp=false
+tags.short=false
+web.root=.
--- a/nbproject/project.xml
+++ b/nbproject/project.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://www.netbeans.org/ns/project/1">
+    <type>org.netbeans.modules.php.project</type>
+    <configuration>
+        <data xmlns="http://www.netbeans.org/ns/php-project/1">
+            <name>opencart</name>
+        </data>
+    </configuration>
+</project>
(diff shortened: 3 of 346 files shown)


danielkerr:
if u want SSL through out the site u set it in config.php!

ChrisLancs:
Hi daniel thanks for your response. If thats the case, whats the point in having the setting in the admin area?

danielkerr:
i cant remember the reason but there is one.  something to do with remote ajax requests or image or html from 3rd party sites that are not ssl. either way there is a fix for people who want full ssl through out the site so no need to to post this as an issue!

danielkerr:
actually i will remove it. i dont know maybe the issue was shared ssl. or ssl on sub domains for multi stores.

ChrisLancs:
In 2.1 (my current live version) i had some SSL issues that I fixed by checking against that setting to set the urls accordingly. It was indeed insecure content reports. If those are now fixed then there is no need for a check against the setting and it sounds like a good idea to remove it.

I previously posted it as an issue and closed it when i submitted this pull. I will try to find other areas to be helpful in the future.

ADDCreative:
The setting in the admin area (config_secure) used to work, but the functionality got lost somewhere in the changes to version 2. What looks to be missing is the test of config_secure when creating the Url object.

The line https://github.com/opencart/opencart/blob/3.0.2.0/upload/catalog/controller/startup/startup.php#L39 would need changing to somthing like bleow and a simlar line adding to the admin startup.php.
```php
$this->registry->set('url', new Url($this->config->get('config_url'), $config->get('config_secure') ? $this->config->get('config_ssl') : ''));
```

ChrisLancs:
As daniel has said he is going to remove it then I would hold off on that change, the pull was to to remove setting of SSL on the links which i still believe should be looked at, but as the preference is now to set full SSL from the config files (I suspect by setting both urls to HTTPS, the pull request is I submitted is no longer required.

ravilrrr:
What to do then if I need ssl only for account and cart?

ChrisLancs:
I am not sure what reason you could have for wanting SSL in only one cart, but my PR would have allowed for it.

ravilrrr:
I agree. This is not so important and you can remove the ssl state.

danielkerr:
just remembered one of the reasons.  ssl remote content and emails.

ChrisLancs:
Im sure there were and still are perfectly valid reasons, I'm just still trying to weigh the benefit of doing it from the config files when multi stores don't have their own file but can support ssl.

Actually come to think of it, it did puzzle me a little why the default store is hard coded the way it is and not added the same way its added for multi store, is that a left over from an earlier single store version?

danielkerr:
its being removed.

PR #6217, opened 2017-11-23
Automatically determine TPL or TWIG

stalker780:
Now 'template_engine' setting have no sense. Nobody will ever set it to 'template' and edit all default templates to revert them from twig.

This small fix allows to use both twig and tpl files at the same time.
If .tpl file is found it will be used, otherwise loader will try to load .twig template.
This will help developers and users move from OC 2.x tpl to OC 3.0 twig more easily. Most modules and templates will work on 3.0 with minor changes.

### DIFF
--- a/upload/system/engine/loader.php
+++ b/upload/system/engine/loader.php
@@ -120,7 +120,11 @@ final class Loader {
 		if ($result && !$result instanceof Exception) {
 			$output = $result;
 		} else {
-			$template = new Template($this->registry->get('config')->get('template_engine'));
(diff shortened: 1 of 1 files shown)


stalker780:
It seems that oc3 will be in beta for a long time. 
Therefore, a transition period is needed. 
Now there is no sense in writing and supporting extensions both for OC2 and OC3. 
Why not make it easier for everyone and support both template engines at the same time?

danielkerr:
this feature was removed because its not required in 3.0!!!! just use twig

PR #6222, opened 2017-11-25
Update setting.twig

faca5:
Fixed sending to additional emails.


danielkerr:
the key config_alert_email is not used any where! so that would have caused the system to stop working.

PR #6231, opened 2017-11-27
Remove deprecated Twig code

lucasjkr:
There is an old Twig installation in:

upload/system/librart/template/

This is unused, Twig is now loaded from the vendor directory.


danielkerr:
done manually

Issue #6118, opened 2017-10-19
[v3.0.2.0 - Fix] - admin/controller/user/api.php


straightlight:
In the validateForm() method, the user/user is being validated with the modify method. Although, the user/api should be validated.

AlexChanHK:
Commit a fixed version with also fixing another bug already. 
Maybe you can try to fix and commit it next time.
This issue can be closed.

Issue #5876, opened 2017-08-15
Rating stars didn't display on Featured products 3.0.2.1b


Manitoou:
catalog/view/theme/default/template/extension/module/featured.twig
Change this line:  {% for i in 5 %} 
to:  {% for i in 1..5 %}

AlexChanHK:
Just commit a edition for fixing this mistake.
And **this issue can be closed.**

Gotolr:
Problem is still there on all default product modules (latest, featured, bestsellers..) ocStore 3.0.2.0b1 and default template.

PR #6233, opened 2017-11-28
Edition for issue #6118 & #5876

AlexChanHK:
Also adjust for the twig prevent the ip list disappear
when the user submit the form without modify permission.


danielkerr:
@AlexChanHK 

u made a mistake. check my commit

AlexChanHK:
Thanks, I will refer and be careful next time.
since I am not sure different the ip_row and using []'s and seem working after these edition.
Mostly this is for the issue ip list disappear issue, seen you had fixed in another way.
(although not very important)

danielkerr:
Testing would be better

PR #6239, opened 2017-11-29
Fix issue with random string breaking the request data array

JonnyCarter:
Removed the string mexico44 from the request data array. It was causing a PHP Parse error


cjgk:
ping @danielkerr

PR #6251, opened 2017-12-04
Update install.php

eka7a:
We are not allowed to add a new product page (e.g. bestseller page). We can upload in catalog/controller/extension/ but this address is not a very good link.

Should be added as an allowed directory on installer.


danielkerr:
use the event system to over ride the call requests! only folders that are allowed to be written to are extension folders!

PR #6253, opened 2017-12-07
Dont show disabled custom fields on order form page in admin: fix

cartbinder:
If custom field is disabled it should not be shown on order form page. Though no error is generated as on front end. But it is extra field added on order form page on admin side which is not needed.


cartbinder:
Some thing wrong. I just want to merge just one commit. It has send 6 commit.

PR #6258, opened 2017-12-08
Facebook Share Meta tag add

JesseOC:
1. Add the meta tag on header
2. Setup set / get function in library/document
3. Product, information add the function to get the shared image, page title, URL, description for Facebook share, tested and works normally


danijelGombac:
There is an error in 

`return $this->$url;`

should be 

`return $this->url;`

JesseOC:
yes, forgot to fixed on submitted file, changed, thanks!

pcmanik:
og:type for product should be product and not article.

agatha65:
https://www.opencart.com/index.php?route=marketplace/extension/info&extension_id=22963

JesseOC:
yes, it has already been updated, thanks for the comment and feedback

PR #6263, opened 2017-12-13
Add seo url autogeneration

pcmanik:
Automatic generation of seo url for product, category and information page.
Supports multiple languages and multiple stores.

### DIFF
--- a/upload/admin/controller/catalog/category.php
+++ b/upload/admin/controller/catalog/category.php
@@ -499,6 +499,8 @@ class ControllerCatalogCategory extends Controller {
 		} else {
 			$data['category_seo_url'] = array();
 		}
+
+		$data['category_seo_autocomplete'] = count($data['category_seo_url']) == 0;
 				
 		if (isset($this->request->post['category_layout'])) {
 			$data['category_layout'] = $this->request->post['category_layout'];
--- a/upload/admin/controller/catalog/information.php
+++ b/upload/admin/controller/catalog/information.php
@@ -398,6 +398,8 @@ class ControllerCatalogInformation extends Controller {
 		} else {
 			$data['information_seo_url'] = array();
 		}
+
+		$data['information_seo_autocomplete'] = count($data['information_seo_url']) == 0;
 		
 		if (isset($this->request->post['information_layout'])) {
 			$data['information_layout'] = $this->request->post['information_layout'];
--- a/upload/admin/controller/catalog/product.php
+++ b/upload/admin/controller/catalog/product.php
@@ -1119,6 +1119,8 @@ class ControllerCatalogProduct extends Controller {
 			$data['product_seo_url'] = array();
 		}
 
+		$data['product_seo_autocomplete'] = count($data['product_seo_url']) == 0;
+
 		if (isset($this->request->post['product_layout'])) {
 			$data['product_layout'] = $this->request->post['product_layout'];
 		} elseif (isset($this->request->get['product_id'])) {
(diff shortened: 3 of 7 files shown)


zemguru:
Neat! Really nice, but there are some comments:
- does not support Cyrillic
- what if You want to change product name a bit, but don't want  to lose previously indexed seo urls? I would vote to have SEO url generated only on `product add` or if keyword is empty.
But if it's already added, then keyup would not change it.

pcmanik:
@zemguru I have added support for Cyrillic characters!

Your point on loosing previously indexed seo url is also valid. I have implemented check to generate it only on new items or on items that don't already have seo url  :)

pcmanik:
@danielkerr There is one more edge case I have noticed. If you have 2 languages with the same product name the resulting seo urls will also be the same.
So should I just append lang code to the seo url like product-name-en or will you update the opencart to allow the same seo url for multiple language mutations?
There is one problem with that approach. It will block search engines from crawling other language mutations.

I would like to hear more opinions on this from you and the rest of devs here.

danielkerr:
yes fixing it.

v5manoel:
keyup does not work on Chrome browser for Android.

Alternatively you can use blur

pcmanik:
Thank you @v5manoel it's fixed.

Shinrai:
This is great besides the fact it only works when you blur the inputs. This will work when you've changed the name. However it will not work on existing items. Or items which were inserted previous to adding a language to the system. Unless the user goes to each name and then exits it. In most cases this will do. But in a general sense this still is flawed.

Additionally appending the language to the name is viable in Opencart's current handling of language codes. But opencart should be prepending the language code to the URLs for better SEO URL traction.

IE /en-gb/product-name/

Additionally the model or UPC (not both, only one) should be appended to the seo-url. IE: product-name-UPC or product-name-MODEL. Though this poses some issues in itself. While it's a best practice to including an identifier if the identifier changes you pose the same issue if the name changes.

pcmanik:
This pull request is aiming at new items, not the old ones. It's just filling the seo url for the user. On existing items it's desired to not change the url since we want to preserve the same url for the product even if its name slightly change. Generally you don't want to break old urls pointing to the product.

What's flawes with this pull in general sense? It does it job - auto create seo url for new products (and other pages) and for the old ones when they don't have filled it.

Also I don't think that UPC and MODEL should go to url. But that's up to Daniel.

And you can easily write your own script to automate fixing missing seo urls before this pull is merged.

Shinrai:
I already have my own script for Opencart 1.4, 1.5, 2.0, 2.3 and 3.0 which auto generates SEO URLs for items and languages and stores which do not have them upon saving a product, manufacturer, information, as well as a few more.

I was commenting here on to point out you are addressing a single use case and not the issue as a whole. The single use case you are addressing is when a new item is created. You are not addressing when a new language is added, even at a fundamental level of saving the item afterwards.

Basically you should be addressing backwards compatibility. When an item's save button is clicked you could address this code by simply bluring all seo url fields which are empty and focusing them then bluring them prior to the submission itself. Though this is not 100% cross browser as some browsers don't like to trigger focus and bluring of fields without user input. This has gotten better over the years but may still pose an issue in some use cases.

The better tactic would be to keep the javascript as so. This will give the admin the instant gratification of an auto-generated URL. But then add in actual model programming to account for empty fields and assign those in the model code itself. This way when it's submitted and the user has not changed the name the additional languages or even items which were present prior to this feature will be able to auto-generate urls as well.

As to the UPC/Model being in the URL. If I've bought a product and that product doesn't show what it is, which is common, but does list a model number or UPC. I can search for that UPC or model number. Which in turn having those values in the URL will produce a higher weight for search engines. As a UPC is a higher value of a unique identifier than a model number. As well as the fact Opencart by default does not show any other identifier than the model of a product. To me it makes perfect sense to place the UPC in the url or if a UPC isn't set the model number.

PR #6268, opened 2017-12-19
Adds filters in user group

niko-chiu:
Adds filters in user group to filter out the permission you need


straightlight:
Not very useful since if, and when, permissions may not match at specific times due to technical issues, users still won't be able to filter what they can't find. Therefore, this commit will now be closed.

PR #6299, opened 2017-12-31
Fixed list Time Zone

v5manoel:
The previous function that display the time zone difference in hours it is wrong.

### DIFF
--- a/upload/admin/controller/setting/setting.php
+++ b/upload/admin/controller/setting/setting.php
@@ -332,28 +332,22 @@ class ControllerSettingSetting extends Controller {
 		// Set Time Zone
 		$data['timezones'] = array();
 
+		$timestamp = time();
+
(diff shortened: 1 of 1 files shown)


danielkerr:
makes no sense.  does not change the value of the time stamp.

i think what i have commited is fine!

danijelGombac:
Once you set the time zone the timezone list options text change.

UTC selected ->Pacific/Wallis (+12:00)

Europe/Zurich selected ->Pacific/Wallis (+1:00)

v5manoel:
@danijelGombac Test my fix and make sure the issue is resolved.

v5manoel:
@danielkerr You may not accept my correction, but you will have to correct the time zone listing because it has the problem reported by @danijelGombac

danijelGombac:
@v5manoel 

With your fix the issue is resolved. Thanks.

PR #6301, opened 2018-01-02
Error from admin/controller/startup/startup.php

JesseOC:
Check config_timezone is exists or not
Fix the error Notice: date_default_timezone_set(): Timezone ID '' is invalid in C:\xampp\htdocs\OpenCart\master-test\admin\controller\startup\startup.php on line 17
And JSON error popup

### DIFF
--- a/upload/admin/controller/startup/startup.php
+++ b/upload/admin/controller/startup/startup.php
@@ -3,7 +3,7 @@ class ControllerStartupStartup extends Controller {
 	public function index() {
 		// Settings
 		$query = $this->db->query("SELECT * FROM " . DB_PREFIX . "setting WHERE store_id = '0'");
-		
+
 		foreach ($query->rows as $setting) {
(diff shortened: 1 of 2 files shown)


fabiodemelo:
When things get fixed here after a bug report does the files (the download) file get updated as well? I am wondering if a user downloads the 3.0.2.0 does he gets these fixed already included or he needs to come back here and download the fixes?  Thank you..

ADDCreative:
@fabiodemelo The downloads are not updated. You need to apply the updates yourself or wait for the next version release.

DariyGRAY:
You can download the master brunch, where all updates are applied

fabiodemelo:
Thank you.

sazpurua:
@DariyGRAY Are the master brunch files for v3.0.2.0? for all the files here I don't see anywhere an indicator of which version they belong.
Thanks

PR #6304, opened 2018-01-02
Config getter improvement

alexslipknot:
Allow to set default value when default setting does not exists

### DIFF
--- a/upload/system/library/config.php
+++ b/upload/system/library/config.php
@@ -17,11 +17,12 @@ class Config {
      * 
      *
      * @param	string	$key
-	 * 
+	 * @param mixed $default
+	 *
 	 * @return	mixed
      */
-	public function get($key) {
-		return (isset($this->data[$key]) ? $this->data[$key] : null);
+	public function get($key, $default = null) {
+		return (isset($this->data[$key]) ? $this->data[$key] : $default);
 	}
 	
     /**


zemguru:
+1

alexslipknot:
Hello, @danielkerr. 
Could you please say something? Maybe it's duplicate or you just has another opinion? 
Thank you anyway. You're doing great work.

danielkerr:
not required.

zemguru:
Not required by who? This is great, simple peace of code, which greatly reduces unneeded if/else checks on variable existence. And does not brake anything already available in Opencart. 
Good practice by Convention over configuration.

danielkerr:
im not wasting my time with every one whose issue i reject.

alexslipknot:
I'ts ok, I don't blame you. It's your project and it's not community version. 
But anyway it would be nice if you can have a staff for requests like that. Sometimes it may contents useful features.

danielkerr:
explain the difference!

if i added every bit of code that people push the system would be fucked.

I have more experience with running an open source project than you do!

alexslipknot:
Indeed. I agree with your opinion about security issues. That's why I suggested a staff for reviewing requests.
And one more thing. Don't judge people by their labels. I don't think that you know about my experience at all. Listen, sir, I don't want to offend you or your knowledge. Please take my apologizes if I hurt you somehow. I just wanted you to know about the idea. Thanks.

PR #6309, opened 2018-01-04
Update filemanager.php

DariyGRAY:


krokodylowy:
Hi.
Above patch has no sense.  Filemenager doesn't display all directories and images because sum is invalid.
Instead  $pagination->total = count(array_merge((array)$images, (array)$files));
it should be   $pagination->total = count(array_merge((array)$directories, (array)$files))+sizeof($data['images']);

danielkerr:
i just added another fix. yeah he should not hacve messed with the pagnation

PR #6315, opened 2018-01-06
Fixed error in SQL install

v5manoel:
File:
opencart.sql

Before:
INSERT INTO `oc_seo_url` (`seo_url_id`, `store_id`, `language_id`, `query`, `keyword`, `push`) VALUES

After:
INSERT INTO `oc_seo_url` (`store_id`, `language_id`, `query`, `keyword`, `push`) VALUES


danielkerr:
already added thanks

PR #6316, opened 2018-01-06
Added missing permissions in opencart.sql

v5mvidal:
Added:
Design->SEO Regex
Marketing->Affiliate

Tested!
Installation succeeded after change in opencart.sql file.


straightlight:
Not happening. There are too many variances between the releases. These permissions needs to be looked accordingly between each releases and not as a master branch solution. Therefore, this commit will now be closed.

PR #6320, opened 2018-01-07
Fix admin customer login by store

danijelGombac:


straightlight:
Already fixed in the pre-release. Besides, now uses a token to login as customer from the admin. Therefore, this commit will now be closed.

PR #6326, opened 2018-01-08
Update encryption.php

billynoah:
The current encryption class contains some legacy code and fails to implement an iv in such a way to satisfy the openssl functions.  Because of this, most modern versions of php will throw a warning which is visible to the admin user: "Warning: openssl_encrypt(): Using an empty Initialization Vector (iv) is potentially insecure and not recommended"

This is more or less a complete rewrite using all openssl library functions, sha256 digest and (currently) the strongest encryption cipher I know of 'aes-256-ctr'.  Both the cipher and digest are now class properties which should make this easy to update in the future should these become obsolete.

I've tested this on php 5.6, 7.0 and 7.1 with success.

### DIFF
--- a/upload/system/library/encryption.php
+++ b/upload/system/library/encryption.php
@@ -39,11 +39,15 @@ final class Encryption {
      * @return	string
      */
 	public function decrypt($key, $value) {
+		$result    = NULL;
 		$key       = openssl_digest($key, $this->digest, true);
 		$iv_length = openssl_cipher_iv_length($this->cipher);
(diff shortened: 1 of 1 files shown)


billynoah:
Update - tested on PHP 7.2 and no conflicts.

For the sake of anyone still on OC1.5, here's a drop in replacement of the version provided here that will work on older versions of Opencart: https://forum.opencart.com/viewtopic.php?f=181&t=199924&p=722688#p722648

danielkerr:
cheers.

PR #6332, opened 2018-01-09
Fix bug when attach image on email (gmail.com)

toporchillo:
Сorrect fix of email image attachment bug. Image filename may contains spaces AND nonASCII-symbols (UTF-8). This symbols we have to urlencode, not only spaces.

### DIFF
--- a/upload/catalog/model/tool/image.php
+++ b/upload/catalog/model/tool/image.php
@@ -38,7 +38,7 @@ class ModelToolImage extends Model {
 			}
 		}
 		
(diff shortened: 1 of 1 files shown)


toporchillo:
Сorrect fix of email image attachment bug. Image filename may contains spaces AND nonASCII-symbols (UTF-8). This symbols we have to urlencode, not only spaces.

straightlight:
array_map is not the solution to use with images. Therefore, this commit will now be closed.

PR #6335, opened 2018-01-10
Update admin/model/catalog/information.php

dariygray:
Add data for «push» column in `seo_url` table


straightlight:
The SEO has been re-created in the master branch since this commit was created. Therefore, this commit will now be closed.

PR #6336, opened 2018-01-10
Update category

dariygray:
• Replace «category_id» to «path» in seo_url validation
• Add data for «push» column in `seo_url` table
• Replace «category_id» to «path» for «query» column in `seo_url` table

### DIFF
--- a/upload/admin/controller/catalog/category.php
+++ b/upload/admin/controller/catalog/category.php
@@ -547,6 +547,9 @@ class ControllerCatalogCategory extends Controller {
 		}
 
 		if ($this->request->post['category_seo_url']) {
+			$this->load->model('catalog/category');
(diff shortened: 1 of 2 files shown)


zemguru:
Can You, please, explain the motivation behind this?
As far as I can see, the only benefit would be, to be able to select path by full category path?
but this means also, that you will not be returned seo url link, if you request it for specific category id?

There is little use of replacing "category_id" as key to "path", don't You think? 

I mean, what's the benefit from this change?

DariyGRAY:
I did not solve it) I just fixed the errors, because my seo_url worked incorrectly. The replacement of "category_id" with "path" occurred in the database. According to this, I made changes

image 1

Sorry for my bad english)

straightlight:
The SEO URLs and profile have been re-coded in the master branch since this commit was created. Therefore, this commit will now be closed.

PR #6340, opened 2018-01-12
callback for alipay, alipay-cross-border and wechat-pay

mengwb:
Change the callback method for alipay, alipay-cross-border and wechat-pay to allow to that apache rewrite not required.


danielkerr:
this was removed from the master. going forward there will be no new extensions 3rd party extensions in the core.  you need to post these fixes to the opencart.com marketplace.

PR #6342, opened 2018-01-13
Removed unnecessary paratmer while product option insert

cartbinder:
Removed **product_option_value_id** saving on edit insert SQL query for product option. It is an auto-increment field. So don't know why it is here. If it is for some other purpose I apologize for this commit.


straightlight:
This model file no longer exists in the master branch. Therefore, this commit will now be closed.

PR #6348, opened 2018-01-16
Update Account module

dariygray:


straightlight:
This one seem to be in-demand from the users. Considering this for the features-pre release but the account.twig file on the catalog-end side also needs to be dealt with in order to integrate such modifications.

danielkerr:
not sure why u want multiple account modules when u can just set them on the design layout page

PR #6349, opened 2018-01-16
Update Filter module

dariygray:


straightlight:
Too complexed to deal with sub-sets of filters. Besides, it could also impact too many extensions having to deal with filter groups as it is. One wouldn't need to imagine the impact for multiple rows of filter IDs that could already be regrouped with filter groups at this point.

PR #6350, opened 2018-01-16
Update Information module

dariygray:


straightlight:
Might be useful for the features-pre in the future in order to set this with each locations and the custom fields.

danielkerr:
theres no need for this.

PR #6351, opened 2018-01-16
Update Store module

dariygray:


straightlight:
Suggesting the use of new cart\user without a login or customer token cannot be taken into account. Therefore, this commit will now be closed.

PR #6376, opened 2018-01-29
Fix email delivery

Shinrai:
Creates a centralized email delivery function. This function does most of the leg work in sending the email through the mail library.

It will attempt to auto correct email headers for from, reply_to, cc, to.

Issue arises when sending SMTP email. Most providers will not allow the FROM header to be anything other than the logged in SMTP user. Thus this patch sets the FROM to be that of the SMTP username if it qualifies as a full email (as most services require). The reply_to header is automatically set to the original FROM header.

### DIFF
--- a/upload/admin/controller/mail/affiliate.php
+++ b/upload/admin/controller/mail/affiliate.php
@@ -31,28 +31,22 @@ class ControllerMailAffiliate extends Controller {
 			$language = new Language($language_code);
 			$language->load($language_code);
 			$language->load('mail/affiliate_approve');
-							
-			$subject = sprintf($language->get('text_subject'), $store_name);
 			
 			$data['text_welcome'] = sprintf($language->get('text_welcome'), $store_name);
-						
+
 			$data['login'] = $store_url . 'index.php?route=account/login';
 			$data['store'] = $store_name;
-	
-			$mail = new Mail($this->config->get('config_mail_engine'));
-			$mail->parameter = $this->config->get('config_mail_parameter');
-			$mail->smtp_hostname = $this->config->get('config_mail_smtp_hostname');
-			$mail->smtp_username = $this->config->get('config_mail_smtp_username');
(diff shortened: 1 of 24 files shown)


Jrijke:
I need this fix, when is it expected to be released?

Shinrai:
Actually had a test case for this fix the other day. Specifically when someone enters an AOL email address on the contact form. AOL's DMARC does not allow the email to be delivered. Which in turn leads to unhappy customers who can't get a hold of the site's admin.

Jrijke:
When is this fix expected to be released? I'm looking forward to it. At the moment it isnt possible to recieve notifications and e-mails form opencart v3.x. Also when a client registered a new account, no mails are send to the client or the admin. 

it's essential for a shop to send notifications to clients as a marketing strategie by e-mail or as a admin just simply being noticed when a new account is made. otherwise OC3 is useless.

is it possible to just copy some files to my store to get the mail issue fixed?

Thanks!

Shinrai:
You should verify your emails are working in the first place. This pull request fixes providers like aol who have a restrictive DMARC record.

To verify your emails work use the contact form as a guest and enter the email address of the smtp username you have setup in your admin settings. If the email goes through than your emails work. If not than this Pull request isn’t going to fix your issue.

To be clear this pull request changes the from header to be that of the smtp username if it’s a fully qualified email. Then sets the reply-to to the email that it was originally supposed to be from.

Shinrai:
An Example would be the following:

**Presuming the following settings:**
> config_mail_smtp_username: example@example.com
> config_email: sales@example.com
>
> user's email: example@aol.com

### Opencart as is (without this pull request):
> To: sales@example.com
> From: example@aol.com
> Reply-To: example@aol.com
The above will fail due to AOL's DMARC rules

### Opencart after this pull request:
> To: sales@example.com
> From: example@example.com
> Reply-To: example@aol.com
The above will be delivered and any emails replying to the delivered email will be sent to example@aol.com

PR #6379, opened 2018-02-01
Mysql error upgrading step 12

mikehins:
Fix error updating setting values with single quote


ADDCreative:
You need to use $db->escape() not addslashes(). The other values in that file could also do with escaping or casting.

ADDCreative:
You also need the $this. So ```$this->db->escape($result['value'])```.

PR #6399, opened 2018-02-09
2Checkout and language fixes

stalker780:
1. fixed SSL links in admin
2. fixed USD conversion. Cause 2Checkout works only with USD
3. fixed digital products checkout. They don't require shipping address now on 2Checkout gateway page
4. Fixed redirects
5. Updated form field names according to latest 2Checkout documentation https://www.2checkout.com/documentation/checkout/inline-checkout


danielkerr:
does not need the true arg!

its now ssl or not!

danielkerr:
please re push with the url stuff removed.

ADDCreative:
Still some url->link with the ture argument.

https://github.com/opencart/opencart/blob/master/upload/admin/controller/extension/payment/twocheckout.php#L42
https://github.com/opencart/opencart/blob/master/upload/catalog/controller/extension/payment/twocheckout.php#L81

PR #6408, opened 2018-02-12
Fix HTTPS issue admin order edit. Change API URL based on request admin URL

howdu:
Fixes error:
Mixed Content: The page at 'https://www.example.com/admin/index.php?route=sale/order/add&user_token=123' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://www.example.com/index.php?route=api/currency&api_token=123&store_id=0'. This request has been blocked; the content must be served over HTTPS

### DIFF
--- a/upload/admin/controller/sale/order.php
+++ b/upload/admin/controller/sale/order.php
@@ -371,7 +371,7 @@ class ControllerSaleOrder extends Controller {
 		$data['order_statuses'] = $this->model_localisation_order_status->getOrderStatuses();
 
 		// API login
-		$data['catalog'] = HTTP_CATALOG;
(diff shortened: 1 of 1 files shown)


danielkerr:
make sure the default config is for the front and the admin are the same!!

danielkerr:
it will still have issues because of cross domain

howdu:
The error is clear. The change is simple. If the admin is https then ajax request must be also.

https://developers.google.com/web/fundamentals/security/prevent-mixed-content/what-is-mixed-content

howdu:
Or could be fixed by removing the http so it starts '//www.example.com'.

```php
$data['catalog'] = preg_replace("(^https?://)", "", HTTP_CATALOG)';
```

PR #6425, opened 2018-02-26
[fraud_fraudlabspro] added IPv6 supports

fraudlabspro:
added IPv6 supports for FraudLabs Pro fraud extension


fraudlabspro:
Will create another pull request with another new update.

PR #6432, opened 2018-02-28
adding newsletter feature

Felix-Ho:
adding newsletter


straightlight:
Might be considered for the features-pre version.

PR #6439, opened 2018-03-02
Update manufacturer.php

BuslikDrev:
чтобы формировалась чпу ссылка
---------------------------------------
so that a ccu link is formed


straightlight:
Valid point. Without doing this, it will affect the SEOs. Fixed in the pre-release. Well done!

PR #6440, opened 2018-03-02
Update manufacturer_list.twig

BuslikDrev:
чтобы формировалась чпу ссылка
---------------------------------------
so that a ccu link is formed


straightlight:
The manufacturer_list.twig has now been fixed in the pre-release version. Well done!

danielkerr:
added manually

PR #6442, opened 2018-03-02
Clear button For Product filter form

rajeshsingh520:
Clear button For Product filter form


straightlight:
Might be useful for the features-pre version.

danielkerr:
it could be done a lot better than this.

PR #6461, opened 2018-03-14
Ocmod should allow more than one install.xml

xtensions:
Ocmod should allow more than one install.xml files, maybe getting away from naming convention install.xml and should support any name ending with .ocmod.xml

Reasoning: Most of the modules that we create need to give compatibility for other addons. Like a checkout module which modifies many files will have their own install.xml but they may want to add a compatibility for other developers module like abandonedcart, smsshare, postcode manager, gift wraps, shippingpros etc. We can not put these changes in main install.xml because not all customers will have these running on their store. So we need to put these in zip folder along with install.xml and name them compatibility_abaondonedcart.ocmod.xml, compatibility_xshippingpro.ocmod.xml extra. The extension installer can prompt user that which files they need to install along. 

In current system this can be achieved by adding a new zip file named compatibile_addons.zip which user can install manually, but looking at how marketplace installation works online in version above 3.x, it will require us to create as many zip files for each xml. And when you have just one file in zip. The opencart system does not approve the zip file because of its very small size(thats what I have seen when I uploaded one separate zip that contains changelog, installation and licensing file).

Original discussion: https://forum.opencart.com/viewtopic.php?f=110&t=200631

One ocmod file attached for testing. [testing.ocmod.zip](https://github.com/opencart/opencart/files/1810531/testing.ocmod.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@
 
 # IDE Project files
 /.idea
+/.settings
--- a/upload/admin/controller/marketplace/install.php
+++ b/upload/admin/controller/marketplace/install.php
@@ -255,95 +255,280 @@ class ControllerMarketplaceInstall extends Controller {
 		} elseif (!is_dir(DIR_STORAGE . 'marketplace/' . 'tmp-' . $this->session->data['install'] . '/')) {
 			$json['error'] = $this->language->get('error_directory');
 		}
-
+		$hasOptional =  false;
 		if (!$json) {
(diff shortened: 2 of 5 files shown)


trinkaljuneja:
Yes , this will surely help for cloud version of opencart.

we need to provide separate zip for other templates / extensions compatibility and this will lead to longer extension approval time

danielkerr:
sorry.

PR #6467, opened 2018-03-16
issue required class added to <fieldset>

mgraph:
.parent().parent() adds ".required" class to wrong container <fieldset id="address"> instead of <div class="form-group">


straightlight:
Considered for testing on the features-pre version. Will leave this commit opened for now.

danielkerr:
added manually.

PR #6478, opened 2018-03-21
Update modification.php

billynoah:
Enable trim attribute for regex replacement.  It makes sense to ignore this for the regex itself, but it's not a problem to leave replacement content untrimmed.


danielkerr:
modification system has been removed

PR #6483, opened 2018-03-23
Fix bug with Theme Editor part 1

AlexDWZ:
When using the theme editor, OCMOD cache files are not considered and not checked.
And if the file changes both in the Theme Editor and through modifiers, changes from OCMOD are ignored.

This patch fixes it (part 1)


pcmanik:
Please change comments to something more descriptive. Because current comments are useless and don't provide any clues about what's been fixed with your change.

straightlight:
Irrelevant. The template path is still used as user defined while the config.php and admin/config.php files uses the DIR_TEMPLATE dynamically. Therefore, this commit will now be closed.

PR #6484, opened 2018-03-23
Fix bug with Theme Editor part 2

AlexDWZ:
When using the theme editor, OCMOD cache files are not considered and not checked.
And if the file changes both in the Theme Editor and through modifiers, changes from OCMOD are ignored.

This patch fixes it (part 2)


straightlight:
No longer in the master branch. Therefore, this commit will now be closed.

PR #6486, opened 2018-03-25
Update product_option_form.twig

Gorobez:
When creating new options, the date and time is not shown.
Need input-group-addon
http://bootstrap-datepicker.readthedocs.io/en/latest/index.html#data-api


danielkerr:
input-group-addon is no longer used in bootstrap 4

maybe u should do some research before trying to sabotaging my time! the datepicker will be updated when i have finished updating all the html elements!

PR #6488, opened 2018-03-26
Use native recursive directory creation

Cosmologist:


danielkerr:
i dont think this works!

Cosmologist:
Why? :)

danielkerr:
doesn't work for all server configurations. read comments of mkdir

PR #6494, opened 2018-03-28
Update common.js

sincromaster:
updated modal title Agree to Terms


straightlight:
I don't see the difference in those replaced codes ... both lines looks the same.

sincromaster:
@straightlight,
 apparently makes no difference, but

Before replacing
before

After replacing

after

danielkerr:
already fixed a while ago. cheers.

PR #6531, opened 2018-04-02
Escape XML entities in Google Sitemaps

squarepegsys:
I have a lot of products with `&` in the name and was surprised to get an invalid Google Sitemap from the extension because it didn't escape it or many other characters. This is a quick and dirty fix to make it work.

### DIFF
--- a/upload/catalog/controller/extension/feed/google_sitemap.php
+++ b/upload/catalog/controller/extension/feed/google_sitemap.php
@@ -1,5 +1,19 @@
 <?php
 class ControllerExtensionFeedGoogleSitemap extends Controller {
+
+    private function xml_entities($string) {
+        return strtr(
+            $string, 
+            array(
+                "<" => "&lt;",
+                ">" => "&gt;",
+                '"' => "&quot;",
+                "'" => "&apos;",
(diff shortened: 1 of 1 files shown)


straightlight:
For which OC version are you posting these modifications?

squarepegsys:
Version 3.0.1.1. Do you have a branch for that?

straightlight:
There are no specific branch by versions but it is mainly focused on the latest releases or if the code to be replaced has never been replaced before and was meant to be changed to rectify a reported issue.

However, by looking at your codes, $name and $product['name'] in the XML schema, I do not see the difference between these two ... any purpose on creating a $name variable compared to the already existing $product['name']?

squarepegsys:
Dang it -- copy-paste error. Look at it again.

straightlight:
This line:

`private function xml_entities($string) {`

should be:

`protected function xml_entities($string) {`

however.

straightlight:
In addition, any string parsing functions should be declared as an helper file rather than a protected method (system/helper)

danielkerr:
not accepted since there is a built in php for this.

PR #6533, opened 2018-04-02
Various fixes and improvements to Amazon Login/Pay.

simo-vt:
- Improvement: The language 'American' is renamed to 'American English'.
- Improvement: Error reporting in Sandbox mode is now in the JavaScript console instead of HTML elements.
- Improvement: ModelExtensionPaymentAmazonLoginPay:validateResponse now skips debug logging if the request is for details validation.
- Improvement: Debug logger no longer throws an error in case the class index is not defined.
- Improvement: Set a maximum of 150 characters for the IPN URL.
- Improvement: Action buttons in Payment Address and Payment Method are now the same width as the main container.
- Improvement: IPN token and IPN URL inputs have been moved after the mandatory app credentials.
- Bug fix: OffAmazonPayments.Button now uses the Merchant ID instead ot the Client ID.
- Bug fix: The redirect after save is now working properly.
- Bug fix: Fixed broken customer/address creation calls.
- Bug fix: Fixed validator for empty IPN token.
- Bug fix: On install/uninstall, the event model is now loaded properly.
- Bug fix: Added missing $error parameters to front-end controller classes. This was causing issues on PHP 7.

### DIFF
--- a/upload/admin/controller/extension/module/amazon_login.php
+++ b/upload/admin/controller/extension/module/amazon_login.php
@@ -1,9 +1,10 @@
 <?php
 class ControllerExtensionModuleAmazonLogin extends Controller {
-
+	private $version = '3.0';
 	private $error = array();
 
 	public function index() {
+
 		$this->load->language('extension/module/amazon_login');
 
 		$this->load->model('setting/setting');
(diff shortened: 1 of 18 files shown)


straightlight:
When using any test mode, implicitly the Amazon Login Pay, the user_token should also be validated in the if statement since the debug mode is providing API responses that customers should not see. Granted, the maintenance mode can be activated to avoid that kind of issue. Although, it would even be better to add the user_token check in case a store admin would not set his store under maintenance when these tests occurs.

simo-vt:
@straightlight 

Thank you for reporting this. I have a few comments and a question regarding your remark.

The messages are outputted in the JavaScript console. They are not directly visible by the customers unless they open up their Debug Tools, which is rarely the case. In any case, anyone without technical knowledge would not be affected. Likewise, anyone with technical knowledge would be able to see the error messages with a JavaScript debugger, even if console.log is suppressed. Also, the extension would be in test mode only for initial testing and debugging.

Given the above conditions, I personally believe it to be an overkill to require an active admin panel session for debug messages. Note that I shall soon do a pull request with the change you suggested because it would not hurt to do it. What I need to know is if there is a security vulnerability I may be missing to address.

PR #6560, opened 2018-04-04
Update loader.php

MikhailShemelin:
to avoid creating multiple model class instances the static variable 'model' should be defined outsite of the anonymous function 
(otherwise every instance of the anonymous function have its own variable 'model' and finally every proxied model method is linked with its own model class instance)

### DIFF
--- a/upload/system/engine/loader.php
+++ b/upload/system/engine/loader.php
@@ -221,8 +221,9 @@ final class Loader {
 	}
 
 	protected function callback($registry, $route) {
+		static $model;
 		return function ($args) use ($registry, $route) {
-			static $model;
+			global $model;
 
 			$route = preg_replace('/[^a-zA-Z0-9_\/]/', '', (string)$route);
 


danielkerr:
your right. but i think it needs to be done in the registry maybe.

Tnkr:
we could make it working with registry (put the array of class instances to registry) using the methods
```
$registry->set('model', array());
$registry->get('model');
```

straightlight:
Should definitely not be used with global . Through the OC registry would be the right method.

MikhailShemelin:
> Should definitely not be used with global . Through the OC registry would be the right method.

At the same time it looks like registry actually contains only callable instances, and the collection of not proxied model instances (currently stored in "static $model;") could be unsuitable there. 
Maybe it would be better to use just a private property of the loader instance for this.

danielkerr:
already fixed. its stored in the registry.

PR #6570, opened 2018-04-09
escape XML in Google Sitemap output

### LINKED ISSUE #6532: Invalid XML in Google Sitemaps if `&` is in then product name
I have a lot of products with & in the name and was surprised to get an invalid Google Sitemap from the extension because it didn't escape it or many other characters. This is a quick and dirty fix to make it work.

see my pull request https://github.com/opencart/opencart/pull/6531

miklcct:
This fixes #6532


ADDCreative:
@alaplap As I commented here there are a few things wrong with the pull request. https://github.com/opencart/opencart/issues/6532#issuecomment-426675610

As well as the three issues I pointed out in the comment. I've just also noticed that htmlspecialchars does not heave the encoding set to UTF-8.

PR #6575, opened 2018-04-11
[v3.0.3.0b - Concept] - Custom Field Required not only on customer group

straightlight:
In system/helper/db_schema.php file, on the custom_field_value table, the:

`array(
				'name' => 'required',
				'type' => 'int(1)',
				'not_null' => true,
                                'default' => '0'
			),`

should be added for checkout customer information compliance.


jamesallsup:
Stale branch, create new PR if issue on master.

PR #6580, opened 2018-04-12
eway refund amount rounding issue

### LINKED ISSUE #6579: Incorrect refund amount on eWAY payment module
I'm using Opencart 2.3.0.2, and have found a problem with the native eWay payment module - when refunding orders, the system seems to be only refunding part of the entered amount. 

For example: I entered "6.95" into the refund amount field, and it only refunded $6.00. When I entered "2.60", it only refunded $2.00. In both cases, the "captured amount" was the same as what I entered. Even weirder, when I tried to refund the rest of the captured amount (95c and 60c, respectively), the system suddenly refunded the full amount again - meaning that all up I had refunded nearly double the order amount!

miklcct:
fix #6579

### DIFF
--- a/upload/admin/model/extension/payment/eway.php
+++ b/upload/admin/model/extension/payment/eway.php
@@ -142,7 +142,7 @@ class ModelExtensionPaymentEway extends Model {
 
 			$refund_data = new stdClass();
 			$refund_data->Refund = new stdClass();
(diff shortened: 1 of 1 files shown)


straightlight:
Deprecated from the master branch and the pre-release version. Therefore, this commit will now be closed.

PR #6583, opened 2018-04-12
Add button generate auto geocode in localisation/location

royera:


straightlight:
Outstanding. However:

`if (isset($this->request->get['address']) && !empty($this->request->get['address'])) {`

is no longer necessary to be used together. Simply using the !empty portion already implicit the isset function.

royera:
Ajust condition to if (!empty($this->request->get['address'])) {

straightlight:
The location and geo codes might be considered for the features-pre version in the future.

AJenbo:
@royera in case you are still relevant to you could you please resolve the conflicts with master?

mScott224 (approved):

PR #6585, opened 2018-04-13
Update filemanager.php

fallax010:
a space was added inside directory or file name every 14 characters if the name was at least 14 character long.


zemguru:
+1

PR #6590, opened 2018-04-17
handle order status in pp_express ipn

### LINKED ISSUE #6582: Paypal Express module - no status update after an eCheck clears
I'm using Opencart 2.3.0.2, and have found a problem with the Paypal Express Checkout module. I don't believe that IPNs after the first one (when an order is first made) are being received and acted on by Opencart. This seems to affect eChecks (which are not paid right away), as well as disputes by the customer. Both of these events could occur a number of days after an order has been originally made, and - I would think - update an order's status in Opencart.

For eChecks, I have checked the IPN log in Paypal, and it has shown that two IPNs are sent to "/index.php?route=extension/payment/pp_express/ipn" (one when the eCheck is registered, and one when it is cleared), and that they both get a "200" response. However, it doesn't seem to lead to any kind of update on the order itself (I would expect it to be set to "Complete", if we're dealing with a cleared eCheck).

miklcct:
fix #6582

### DIFF
--- a/upload/catalog/controller/extension/payment/pp_express.php
+++ b/upload/catalog/controller/extension/payment/pp_express.php
@@ -1088,9 +1088,9 @@ class ControllerExtensionPaymentPPExpress extends Controller {
 
 			if ($result['ACK'] == 'Success') {
 				//handle order status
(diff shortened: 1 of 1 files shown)


straightlight:
Deprecated from the master branch and from the pre-release. Therefore, this commit will now be closed.

PR #6594, opened 2018-04-18
Bug Fixed

JesseOC:


ravilrrr:
bad fix #742ff42 
It was all right.  When fighting spam, almost all hosts forbid sending an email from an arbitrary address.

ADDCreative:
742ff42 is likely to cause problems with emails being rejected as spam if Sender Policy Framework is used.

PR #6595, opened 2018-04-18
fix paypal environment variable name

miklcct:
the previous fix from howdu is not consistent with the variable name used in the controller, this change the fix to use the correct variable.


straightlight:
Deprecated from the master branch and from the pre-release. Therefore, this commit will now be closed.

PR #6597, opened 2018-04-18
do not hardcode decimal places in pp_express currency

miklcct:
The pull request eliminates cases when a currency uses more than 2 decimal places (such as XBT) which would make the calculation wrong.

### DIFF
--- a/upload/catalog/model/extension/payment/pp_express.php
+++ b/upload/catalog/model/extension/payment/pp_express.php
@@ -1,4 +1,12 @@
 <?php
+
+use Cart\Currency;
+
+/**
+ * Class ModelExtensionPaymentPPExpress
+ *
+ * @property Currency $currency
+ */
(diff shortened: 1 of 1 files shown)


eivydas:
1. You could just replace second number_format parameter with $decimals value
2. Line [95](https://github.com/opencart/opencart/commit/f2473cc04be7b6eab0b704019347c2b5dcd1c027#diff-713b1d266f6a1bd1bb9d54fe892c9de1R114) is incorrect

miklcct:
fixed incorrect line

It's not good to use number_format here because the intended purpose of number_format is to generate a printable string, but the use here is to create a number suitable for further calculation, therefore we should avoid string manipulation.

straightlight:
Deprecated from the master branch as well as from the pre-release version. Therefore, this commit will now be closed.

PR #6598, opened 2018-04-18
clarify help message for value settings

miklcct:
The value field settings (for length, weight and currency) is confusing and unintuitive. For example, if my store is in HKD as the main currency, when I create USD as an alternative currency, I am tempted to type in 7.8 as the "value" because the value of USD is 7.8 HKD. This is the exact opposite of how the system work.

This pull request clarify the help message in order to assist the user to type in the correct value.


straightlight:
Agreed. Fixed in the pre-release version. Well done!

PR #6604, opened 2018-04-24
code updated

JesseOC:
code updated


danijelGombac:
@JesseOC 

I tried the changes in catalog/controller/startup/seo_url.php and the seo urls stop working.

ADDCreative:
@JesseOC You can't just remove the die function. You need to prevent execution of the rest of the code. Maybe one option would be to replace with return.

PR #6607, opened 2018-04-25
removes the overal page zooming on mobile and tablet

ThaFridge:
Some minor changes to the admin and default theme to remove mobile and tablet page zooming when insert/input box is hit

### DIFF
--- a/upload/admin/view/template/common/header.twig
+++ b/upload/admin/view/template/common/header.twig
@@ -2,6 +2,7 @@
 <html dir="{{ direction }}" lang="{{ lang }}">
 <head>
   <meta charset="UTF-8"/>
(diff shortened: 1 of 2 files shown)


pcmanik:
This no longer works on iOS and btw blocking user from zooming is bad practice.

ThaFridge:
Your wrong sir, this still works on IOS and Android! Im still using it to make sure bootstrap use the right scalling on mobile devices since opencart doenst pratice it. And yes its even still working on my master branch fork

pcmanik:
@ThaFridge Since iOS 10 user scalable attribute is ignored in Safari on iOS.

ThaFridge:
@pcmanik it isnt, just test it your self on your IOS devices before we run into a endless discussion. Just remove both lines and just hit a input box. IOS will still do a page zoom around the input-box.

danielkerr:
im sticking wit the bootstrap one.

PR #6613, opened 2018-04-27
Do not read expired sessions

florinsith:


ADDCreative:
This one still may need looking at.
https://github.com/opencart/opencart/blob/master/upload/system/library/session/db.php#L45

Issue #6480, opened 2018-03-21
Cron jobs require admin login


howdu:
The cron job url displays the admin login - admin/index.php?route=common/cron
Is it missing from the first `$ignore` admin/controller/startup/login.php

Also it could do with some form of security to prevent the cron from being easily run by anybody. Maybe a token check you pass as a parameter.

AlexChanHK:
I also can't find a way to ignore the login or auto login with the cron job function.
The only solution I can think is placing it to somewhere that don't need to login and adding a secret key for that.
For example:
wget "http://localhost/opencartlatest/upload/index.php?route=common/cron&secret=???" --read-timeout=5400
Don't know much knowledge about server, if there is possible solution, please suggest.

danijelGombac:
@AlexChanHK 

I use cron jobs without problems.

I change $ignore array:

upload/admin/controller/startup/login.php
upload/admin/controller/startup/permission.php

and added cron_token in config settings. The only problem with cron jobs is date_modified.

miklcct:
@howdu `common/route` is present in `$ignore` in the master branch in both `login.php` and `permission.php`

AlexChanHK:
@danijelGombac
Thanks, as I did not notice that we can ignore the login and permission protect in this way.><
I checked the code in the related to the date_modified of corn job,
seem Daniel had added a protection for the all the corn job,
`if ($result['status'] && (strtotime('+1 ' . $result['cycle'], strtotime($result['date_modified'])) < time())) {`
therefore if last update of corn object is less than 1 cycle(default of currency corn is day), it will not run.
If I manually edit the date in database and run Corn URL, then the day will update.

Also, did you created a filed for user to create the cron_token, or you directly added in DB for testing?
___
On the other hand, refer the current  master version(29/3/2018)
The 'common/cron' only exist in the 2nd $ignore in ControllerStartupLogin,
Therefore it will still exist login page as the access was rejected in the
`if (!$this->user->isLogged() && !in_array($route, $ignore)) {`
Should we add 'common/cron' in the first $ignore also?

danijelGombac:
I use cron for daily backup. But with current code it work every two days.

In admin/controller/startup/login.php add 'common/cron' to the first $ignore array

and i change admin/controller/common/cron.php

```
<?php
class ControllerCommonCron extends Controller {
    public function index() {
        $timestamp = time();

[...]

PR #6620, opened 2018-05-01
 Fix missing event cart/edit/after. Fix admin cron exclude login and added token. Plus admin bootstrap 4 fixes

howdu:


howdu:
https://github.com/opencart/opencart/issues/6480

danielkerr:
does not need cron token. u can hit the cron url as many times as u want its not going to run anything thats not set to run until date modified has past its expired period..

PR #6625, opened 2018-05-04
Admin view/*/before event redefined

samchen945:
`view/*/before` is redefined, and `event/theme` won't be called.

### DIFF
--- a/upload/system/config/admin.php
+++ b/upload/system/config/admin.php
@@ -43,9 +43,9 @@ $_['action_event'] = array(
 		999  => 'event/language',
 		1000 => 'event/theme'
 	),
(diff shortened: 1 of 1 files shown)


samchen945:
Hi @danielkerr 

I think I didn't make it clear. In this file `system/config/admin.php`, `view/*/before` defined twice in the array (one line: 42 and line 46) as you can see below.

image

It turns out that the later one overwrites the first `view/*/before`, which makes the `1000 => 'event/theme'` never be called, see below.

image

danielkerr:
ahh ok

danielkerr:
it will be in the next commit.

PR #6627, opened 2018-05-05
Update product.php

fallax010:
maintaining the comment pattern


straightlight:
Agreed. Fixed in the pre-release version. However, there are also more comments that could be added on that controller file. So far, well done!

PR #6630, opened 2018-05-07
Update image.php

faca5:
A lot of users have problem with memory restriction.
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 19648 bytes) .

### DIFF
--- a/upload/system/library/image.php
+++ b/upload/system/library/image.php
@@ -10,6 +10,9 @@
 /**
 * Image class
 */
+
+ini_set('memory_limit','-1');
+
 class Image {
 	private $file;
 	private $image;


danielkerr:
it needs to be done on the js level.

faca5:
k

but this is php restrictions when someone upload big image exp. 2 MB and need more resources to resize.

osworx:
And are you asking why so many users have troubles while uploading?
Instead of a 'normal' size and bitrate (e.g. max. 1200 w) the upload images made with cams with 4K and higher!
That should be avoided, not setting memory limits > unlimited!

faca5:
I understand. 

End user don't have expert skills to use 3rd party program first to reduce image and then upload.

danielkerr:
@faca5 how is php going to be able to tell a file size before its uploaded! js level! if they remove the js they will still get a php error.

danielkerr:
please look for a js fix online and commit that,

osworx:
@faca5 But those 'end users' can operate a webshop ??
Just use your common sense.

Daniel: something like this: https://stackoverflow.com/a/47786555/3450837

danielkerr:
@faca5 and yet your giving advice to some one who does!

faca5:
OpenCart v3
Settings >System > Image > Max File Size

If you set 1 byte, you can still upload files larger then 1 byte exp. 1 MB.


You can upload 10 MB image without any problem and errors.

Request:
/admin/index.php?route=common/filemanager/upload&user_token=gWjZmTH3QfgiAvoPOBDNX9pddgzxpVPu&directory=

Response:
{"success":"Success: Your file has been uploaded!"}

Problem appear on re-size:
admin/index.php?route=common/filemanager&user_token=gWjZmTH3QfgiAvoPOBDNX9pddgzxpVPu&target=input-image5&thumb=thumb-image5
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 41260 bytes) in\system\library\image.php on line 47

If you have any larger file in folder with images you got this error if you not increase memory on PHP level. JS won't fix the problem. Example: User allow all images via FTP.


Default PHP.ini has set 64 MB for memory limit but many hosting doesn't allow custom PHP files in root folder. New modern was is to use USER.ini file.

memory_limit = 64M;

magic_quotes_gpc = Off;
register_globals = Off;
default_charset	= UTF-8;
max_execution_time = 36000;
upload_max_filesize = 999M;
safe_mode = Off;
mysql.connect_timeout = 20;
session.auto_start = Off;
session.use_only_cookies = On;
session.use_cookies = On;
session.use_trans_sid = Off;
session.cookie_httponly = On;
session.gc_maxlifetime = 3600;
allow_url_fopen = on;
;display_errors = 1;
;error_reporting = E_ALL;

Problem appear in v2 and v3.

PR #6636, opened 2018-05-13
repair for custom_field db table

klavatron:
Repair for "Custom field" page. 
Set DB column size to fit values (from 7 to 9 chars). Fix existing "cuted" values.


klavatron:
https://forum.opencart.com/viewtopic.php?f=201&t=204329

danielkerr:
the default is 10 and is upgraded under help db schema.


			array(
				'name' => 'location',
				'type' => 'varchar(10)',
				'not_null' => true
			),

PR #6649, opened 2018-05-24
SQL Optimisation for getProducts(...)

fblaser:
While optimising the database for a customer having a large OpenCart shop, I noticed that the query from getProducts() is sometimes running slowly (1.5 to 2 seconds query time).

This happens when the number of products selected by the WHERE clause is high (in this case, it was more than 10,000 rows). MySQL is not smart enough to take the LIMIT clause into account and run the three subqueries for the 10,000 rows instead of just the few rows needed.

I've modified the query by adding another subquery in the FROM clause that returns only the product ids we are interested in before running the other three dependant queries on those rows.

With the modification, the query time has dropped to 0.4 seconds on the shop of our customer.

### DIFF
--- a/upload/catalog/model/catalog/product.php
+++ b/upload/catalog/model/catalog/product.php
@@ -57,7 +57,7 @@ class ModelCatalogProduct extends Model {
 	}
 
 	public function getProducts($data = array()) {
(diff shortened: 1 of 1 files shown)


schiggi:
I can confirm the results! Great work, @fblase.

fblaser:
Thanks Schiggi for the test!

Do you know if there is anything I can do to increase the chances of this patch being included?

jacob-v-dam:
It works but not when you are sorting on a price. Then the query crashes and calls that `special` is missing.

PR #6653, opened 2018-05-25
duplicate query on admin/model/catalog/product.php  related

f4rr3ll1990:


danielkerr:
more untested code that will mess upo some ones site

PR #6679, opened 2018-06-17
Update footer.php

osworx:
Same line 2 times


straightlight:
Since duplication has been found, this commit will now be closed.

PR #6690, opened 2018-06-22
add * wildcard to api session

bakr-aj:
allow using of * wildcard to allow connection from any ip in order to extends opencart API.


straightlight:
Not sure if this is really useful in this case since there might also be other ways to bypass this. Therefore, this commit will now be closed.

PR #6693, opened 2018-06-25
Update order_form.twig

osworx:
Fix required.
Fields are always required, no matter how the setting is.


straightlight:
Agreed. Fixed in the pre-release version. Well done!

PR #6697, opened 2018-06-27
Update language.php

osworx:
Only 2 parameters allowed


danielkerr:
already added i think

PR #6698, opened 2018-06-27
Update forgotten.php

osworx:
Check for email is made, but if not exist, error is not set.


straightlight:
I am not sure I understand this change ... why is the 'else' statement without an IF statement from the above position during the change?

danielkerr:
already validated.

PR #6699, opened 2018-06-27
Update cod.php

osworx:
Replace `&amp; > &` otherwise the returned URL will be like this:
`route=checkout/success&amp;language=de-de`


straightlight:
Confirmed. This issue has now been fixed in the pre-release. Also discovered in the master branch. Well done!

PR #6703, opened 2018-06-29
Fixed error in Pagination in Filemanager

ilcomp:
Replace function array_splice to array_slice.


straightlight:
Not entirely sure this is useful. I will keep this commit opened and see if it's a relevant feature to add in the future.

PR #6708, opened 2018-07-03
One extra request

ArtemPitov:
If we use $this->request->get['store_id'], we send the wrong request


straightlight:
Similar changes have already been done on the master branch recently. Therefore, this commit will now be closed.

PR #6711, opened 2018-07-03
improving code when database has .

PhamToan:
when database = Opencart3.1beta
it shows error. that code fixs that bug


straightlight:
Might be considered for the features-pre version.

danielkerr:
i have updated the code.

PR #6724, opened 2018-07-09
gracefully hide the carousel if need

kemoycampbell:
add twig statement to hide the carousel if the banners passed from the controller is empty. This prevent a white carousel from appearing on pages where the carousel module are loaded and prevent unprofessional look of the site.
The change is as follow {% if banners is defined and banners is not empty %} at the begin of the line with {% endif %} at the end


straightlight:
Due to CSS changes with Bootstrap versions on the Master Branch, these changes are no longer relevant. Therefore, this commit will now be closed.

PR #6727, opened 2018-07-11
Update sale_order.php

osworx:
Status or Order Status .. this is the question ?


straightlight:
Issue fixed in the pre-release. Still on the master branch. Therefore, this commit will remain opened.

danielkerr:
already fixed in previous push

PR #6733, opened 2018-07-12
 Fixing API login notice when the username and key are missing

binarweb:
If you paste the login api link in the browser you get 2 errors:

<b>Notice</b>: Undefined index: username in <b>D:\xampp\htdocs\opencart\catalog\controller\api\login.php</b> on line <b>11</b>
and
<b>Notice</b>: Undefined index: key in <b>D:\xampp\htdocs\opencart\catalog\controller\api\login.php</b> on line <b>11</b>

that triggers 1 MySQL error:

<b>Fatal error</b>:  Uncaught Exception: Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '`key` = '' AND status = '1'' at line;<br />Error No: 1064<br />SELECT * FROM `oc_api` WHERE `username` = '' `key` = '' AND status = '1'

### DIFF
--- a/upload/catalog/controller/api/login.php
+++ b/upload/catalog/controller/api/login.php
@@ -7,37 +7,43 @@ class ControllerApiLogin extends Controller {
 
 		$this->load->model('account/api');
 
-		// Login with API Key
-		$api_info = $this->model_account_api->login($this->request->post['username'], $this->request->post['key']);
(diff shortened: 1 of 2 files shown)


straightlight:
Not suggested to show the IP address in JSON response as an error message being sent throughout the browser. Therefore, this commit will now be closed.

PR #6740, opened 2018-07-15
removed date_default_timezone_set in framework.php

v5mvidal:
It is no longer necessary in:
system/framework.php

Also cause conflict when:
admin/controller/startup/startup.php
date_default_timezone_set ($this->config->get('config_timezone'));
It is different from UTC.

Important: Tested at PHP 5.6 to 7.2


straightlight:
Does no longer matter since the date('P') for the database has been added in both startup files recently. Therefore, this commit will now be closed.

PR #6745, opened 2018-07-17
Removed redundant SQL call

MrPlow254:
order_status is already defined in:
model/sale/order.php
function getOrder
$order_query


straightlight:
The redundant SQL link does address a good point where the order_status_info['name'] should rather display the order_info['order_status'] instead. This may be considered for the core-pre or features-pre in the future. Therefore, I will leave this commit opened.

Issue #6752, opened 2018-07-23
Twig & modification is broken


osworx:
Line 21 of the modification.xml tries to change system/library/template/twig.php.

https://github.com/opencart/opencart/blob/3d25cf3fb796a81f282db25bce319e2f0449028c/upload/system/modification.xml#L21

But that line does not exist anymore, therefore any template modifications do not work anymore.
Could that be fixed as soon as possible, otherwise a serious esting is not really possible.

See also pull request: https://github.com/opencart/opencart/pull/6741

howdu:
Simplier to pass it as array.
https://twig.symfony.com/doc/2.x/api.html#twig-loader-filesystem

```xml
<file path="system/library/template/twig.php">
        <operation>
            <search>
                <![CDATA[$twig = new \Twig_Environment(new \Twig_Loader_Filesystem(DIR_TEMPLATE), $config);]]>
            </search>
            <add position="replace">
                <![CDATA[$path = array();

		if (defined('DIR_CATALOG') && is_dir(DIR_MODIFICATION . 'admin/view/template/')) {
			$path[] = DIR_MODIFICATION . 'admin/view/template/';
		} elseif (is_dir(DIR_MODIFICATION . 'catalog/view/theme/')) {
			$path[] = DIR_MODIFICATION . 'catalog/view/theme/';
		}

		$path[] = DIR_TEMPLATE;

		$twig = new \Twig_Environment(new \Twig_Loader_Filesystem($path), $config);]]>
            </add>
        </operation>
    </file>
```

osworx:
Finally Daniel has to decide, but then the correct code ($twig changed to $this->twig)

> <file path="system/library/template/twig.php">
        <operation>
            <search>
                <![CDATA[$this->twig = new \Twig_Environment(new \Twig_Loader_Filesystem(DIR_TEMPLATE), $config);]]>
            </search>
            <add position="replace">
                <![CDATA[
        $path = array();

        if (defined('DIR_CATALOG') && is_dir(DIR_MODIFICATION . 'admin/view/template/')) {
            $path[] = DIR_MODIFICATION . 'admin/view/template/';
        } elseif (is_dir(DIR_MODIFICATION . 'catalog/view/theme/')) {
            $path[] = DIR_MODIFICATION . 'catalog/view/theme/';
        }

        $path[] = DIR_TEMPLATE;

        $this->twig = new \Twig_Environment($path, $config);]]>
            </add>
        </operation>
    </file>

[...]

PR #6753, opened 2018-07-23
twig & modification fix

osworx:
Due latest changes, modification on twig will not work anymore.
See also: https://github.com/opencart/opencart/issues/6752


straightlight:
The master branch already contains the bug-fixes to load XML files. With the current development of the DIR_MARKETPLACE, this will also differ from these modifications. Therefore, this commit will now be closed.

PR #6764, opened 2018-07-25
Added svg support to file manager

Teoxoy:
I added svg support for the file manager.
You can now upload and use an svg logo for your store.


Teoxoy:
Looks like this is an unwanted feature. Will close this for now...

PR #6811, opened 2018-08-14
Update column_left.php

eka7a:


straightlight:
Don't see the goal of this replacement. Therefore, this commit will now be closed.

eka7a:
install

straightlight:
You're probably talking about the master branch version. In the pre-release version, not much related language keys that can be found. I will re-open the commit for further investigation.

eka7a:
Yes, master branch.

@danielkerr  Instead of adding code for each language, it would be more useful to check language directories and list languages automatically. In this way, it is necessary to add code in every new language.

Issue #6744, opened 2018-07-17
PHP Fatal error:  Cannot declare class Event


bmjoan:
Opencart's `class Event` defined in `system/engine/event.php` clashes with `event` PHP [extension](https://pecl.php.net/package/event).

Full error message:

> PHP Fatal error:  Cannot declare class Event, because the name is already in use in /var/www/htdocs/opencart/system/engine/event.php on line 17

`event` extension is enabled in many hostings incl. Heroku and its `class Event` is definitely of more general use then very specific one at `OpenCart`.

I'd suggest put it into its own namespace, like `\Engine\Event` or `\OpenCart\Event` or something else of the sort, to prevent the name collision.

The fix is fairly trivial:
 - declare namespace in `system/engine/event.php`
 - specify global ns for `\Action` type hint of `Event->register` function argument
 - use new ns for `Event` class in `system/framework.php`

I can provide my pull request for described fix, when needed.

bmjoan:
Also renaming `Event` [class](https://github.com/opencart/opencart/blob/4d0b84ee940537115c99ba9cbf3bd1e28ce8fcc2/upload/system/engine/event.php#L17) to `OCEvent` and update its name in the sole [instantiation](https://github.com/opencart/opencart/blob/master/upload/system/framework.php#L55)  is an another solution. Even more simple and straightforward.

danielkerr:
fixed since adding name spaces.

PR #6824, opened 2018-08-20
Fix of issue report #6744. Class `Event` name conflict.

bmjoan:
Fixes issue report #6744 - class name conflict with `event` PHP extension.

Solved by putting Opencart's class `Event` in its own namespace, in a similar way like other system classes.


straightlight:
Might be considered for the features-pre release.

danielkerr:
i think fixed

Issue #6830, opened 2018-08-22
Twice {% if custom_field.type == 'time' %} in register.twig


optimlab:
In [catalog/view/theme/default/template/account/register.twig](https://github.com/opencart/opencart/blob/master/upload/catalog/view/theme/default/template/account/register.twig) 
twice   `{% if custom_field.type == 'time' %}`.

Needs `{% if custom_field.type == 'datetime' %}`.

howdu:
Could do with using elseif too.

jamesallsup:
Resolved with #6835

PR #6835, opened 2018-08-28
updated a few bits and bolts

gerardadamczewski:


gerardadamczewski:
re: https://github.com/opencart/opencart/issues/6830

PR #6838, opened 2018-08-29
It is necessary to delete the log

ArtemPitov:
A large log causes a "memory limit" error, we do not need to store as much information


straightlight:
This file is no longer listed in the master branch. Therefore, this commit will now be closed.

PR #6841, opened 2018-08-30
3.0.2.1 rc

straightlight:
PHP 7.3 features that may be useful to know for the next upcoming OC stable release: https://ayesh.me/Upgrade-PHP-7.3


jamesallsup:
Why are you trying to merge an old release which is being patched into the latest master which is radically different?

straightlight:
I couldn't find the way to create a new pull request in this section so I just went to the most relative one. Besides, Opencart does operate with PHP as 7.3 has not been released as we speak yet.

Issue #6854, opened 2018-09-10
Manual Order Create Does Not Use Cust GRP Product Discount


Dabnis:
The client has a user-group Trade-A. A product has a discounted price for customer group Trade-A. All works well when a visitor in the customer group logs into the front of the store and places an order, the discounted product price is used, however :
When the client receives a telephone order from a customer in customer group Trade-A and enters an order manually within the admin system, selecting the correct customer group for that telephone client the discounted price for the product is not used, but the products standard none discounted price.

danielkerr:
fixed in master

PR #6855, opened 2018-09-10
Fix the admin order add / edit customer group restriction.

danijelGombac:
#6854

### DIFF
--- a/upload/catalog/controller/api/customer.php
+++ b/upload/catalog/controller/api/customer.php
@@ -55,7 +55,7 @@ class ControllerApiCustomer extends Controller {
 			}
 
 			// Customer Group
-			if (is_array($this->config->get('config_customer_group_display')) && in_array($this->request->post['customer_group_id'], $this->config->get('config_customer_group_display'))) {
+			if (!empty($this->request->post['customer_group_id'])) {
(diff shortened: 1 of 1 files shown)


straightlight:
The (array) validation has already been added by Daniel prior on Github in order to resolved these issues. An example would be with this one: https://forum.opencart.com/viewtopic.php?f=201&t=206915#p734216

danijelGombac:
Before commenting take a look the changed code. There is no array validation.

straightlight:
The original code is about validating an array which you decided to replace with an not empty statement with the array which this array is identified right from the OC installation in install/opencart.sql file:

`(0, 'config', 'config_customer_group_display', '["1"]', 1),`

As you can see, it is JSONized. As explained, and properly replied on the above, Daniel did made those changes accordingly based on the arrays that should rather be validated as I was the one who did reported it last time on GitHub.

danijelGombac:
This restriction is stupid and belong to customer forms, not admin. And why is no restiriction in the admin customer edit page. You can select customer group, but can't use it.

straightlight:
With higher PHP versions, these validations have been mostly rectified. Therefore, this commit will now be closed.

PR #6867, opened 2018-09-13
Cookie popup

gerardadamczewski:


ADDCreative:
@gerardadamczewski Cookie popup code does not meet the requirements for gaining consent under the GDPR. Maybe OK for the ePrivacy Directive, but thats been about since 2009 and is due to be replaced with the ePrivacy Regulation which has supposedly has simpler rules on cookies. So it would seem strange to add it now.

Issue #6875, opened 2018-09-17
No more white screen - Add a set_exception_handler and display a message


dmitriyzhuk:
Hi Daniel 
How about adding a set_exception_handler as we have set_error_handler

in system/framework.php under set_error_handler

set_exception_handler(function ($e) use($log, $config) {
    if ($config->get('error_display')) {
        echo $e->getMessage() . ' in <b>' . $e->getFile() . '</b> on line <b>' . $e->getLine() . '</b><br>';
    }
    if ($config->get('error_log')) {
        $log->write($e->getMessage() . ' in ' . $e->getFile() . ' on line ' . $e->getLine());
    }
});

Many users have different php settings and often they just see a 500 error page or "white page" this is still the case even if they turn on error reporting since these exceptions are treated differently by php. 

for example 

Error: Could not load model ***/****! in /Users/dmitriyzhuk/Sites/opencart/latest/upload/system/engine/loader.php on line 93

this is a very common error since people delete an extension but may keep a modification that is requesting a model. they will not see this error, rather the white death screen.

Simple solution - add it to system/framework.php

straightlight:
The white screens are intended for store owners operating their stores in production mode, also when the display errors option from the admin - > systems - > settings - > settings - > server tab is set to disabled.

dmitriyzhuk:
Yeah, that doesn't work with Exceptions. Opencart does not handle exceptions so it just throws errors. so for example if a template or a model is not found, instead of showing a well-designed pages saying, we found an error or logging it into the log.txt file - it just shows a white screen.

 try it for your self.

danielkerr:
under system/config/default.php

change

$_['error_display']        = true;

to

$_['error_display']        = false;

PR #6878, opened 2018-09-18
Exception handler

gerardadamczewski:
https://github.com/opencart/opencart/issues/6875


ADDCreative:
The error handler is also in the following. Made need updating to include the exception handler.
admin/controller/startup/error.php
catalog/controller/startup/error.php

Issue #6880, opened 2018-09-18
str_split used in SMTP library is not multibyte UTF-8 safe


ADDCreative:
str_split is used in the SMTP library at  the line below. However it does not handle multibyte UTF-8 characters. If the contents of a email contain a multibyte UTF-8 character that happens to start just before a split, it will display corrupted in an email client.

https://github.com/opencart/opencart/blob/3.1.0.0_a1/upload/system/library/mail/smtp.php#L193

straightlight:
Take note that using preg_match_all uses multiples scalar indexes rather than a single indexed that is used prior with str_split. Which means, a single foreach statement may not suffice. The PREG_PATTERN_ORDER may also need to be added in the switch case of preg_match_all in this case.

danielkerr:
execpt mb_detect has to be installed.

PR #6881, opened 2018-09-19
replaced str_split with preg for utf8 support

gerardadamczewski:
https://github.com/opencart/opencart/issues/6880


danielkerr:
already addded

PR #6885, opened 2018-09-20
Grp product discount fix

gerardadamczewski:
https://github.com/opencart/opencart/issues/6854


straightlight:
I am not sure I understand how the api_id being involved while validating with isset would resolved the issue logically ...

gerardadamczewski:
My fault, that's why it's been restored with the next commit

straightlight:
The API backward compatibility has been modified on the master branch which may bring different results to what you're suggesting nowadays. Therefore, this commit will now be closed.

Issue #6882, opened 2018-09-19
Cart is cleared without proper check, by opening checkout/success 


zemguru:
URL link `?route=checkout/success`, opened outside order process, will always show default "Order placed" message, which - is not correct.

Even more - if You make an order, and reach last "confirmation step" (the moment, when order has been inserted into the database with `status_id=0` (Missing order) ), and then go to `?route=checkout/success` page somehow - cart content is deleted.

There are users, who leaves added products to cart for a long period. This accidental deletion should be fixed.

E.g. - check by session's  order status_id ( >0 ?, not in Missing orders).

zemguru:
Yeah ... will not be so easy, of course. 
Scenario: 
Create order -> Use Paypal payment -> Confirmation step creates Order with status 0, -> user is redirected to Paypal -> Paypal redirects back directly to checkout/success without any post/get additional parameters. but executes "callback" after few seconds/minute.

This would mean - if customer returns to the site BEFORE Paypal has posted `/?route=extension/payment/pp_standard/callback`, order will still be in status=0 :(

straightlight:
OC version unposted.

mangeshweb:
We placed a pull request regarding cart product remove at the checkout process.
#6887

zemguru:
> OC version unposted.

Any. And current master included.

> We placed a pull request regarding cart product remove at the checkout process.

Great! Still - if `checkout/success` is opened directly, it will show "Order created" message, which not correct.

mangeshweb:
I have fixed the issue of opening default order success page and removing the cart product on visiting the URL link?route=checkout/success. Now it will be redirected to the URL link ?route=checkout/checkout, Fixed Patch  #6882

danielkerr:
there is nbo reason a customer needs to visit the success page without being lead there!

this is not a bug! make sure you dont have your default order status to complete.

PR #6887, opened 2018-09-21
[Issue Fixed:#6882 Cart clear issue fixed.]

Webkul-Cart:
There was an issue regarding cart clear on checkout process. If we hit ?route=checkout/success this URL Link directly in the new tab, then cart product will remove.


straightlight:
As per the addressed commit states, as per Daniel, he would be correct to state it is not a bug. The events system can already take initiative on handling the checkout success activities. Therefore, this commit will now be closed.

Issue #6886, opened 2018-09-20
Errors seen after installation regarding images


infografik:
system/library/image.php line 169 & other...

Correction:

1. to line 140 (parameters of if) add:
      || empty($width) || empty($height)
2. in line 144 & 145 add // or delete this lines (unnecessary): 
//		$xpos = 0;
//		$ypos = 0;

After changes:
     /**
     * 
     *
     * @param	int	$width
     * @param	int	$height
     * @param	string	$default
     */
     public function resize($width = 0, $height = 0, $default = '') {
		if (!$this->width || !$this->height || empty($width) || empty($height)) {
			return;
		}

//		$xpos = 0;
//		$ypos = 0;
		$scale = 1;

straightlight:
If empty width and height are being conditioned at this point, then there would be no point on defining the width = 0 and height = 0 in the function parameters.

infografik:
If null, not if empty. Empty is value.

⁣-- 
pozdrawiam
  Marek Pawłowski ​

W dniu 22.09.2018, 15:16, o 15:16, użytkownik straightlight <notifications@github.com> napisał:

PR #6888, opened 2018-09-21
image tweak

gerardadamczewski:
https://github.com/opencart/opencart/issues/6886


zemguru:
What if negative values?

straightlight:
This is currently in reflection to use JS validator as per Daniel in the future. Therefore, this commit will now be closed.

PR #6889, opened 2018-09-21
Remove semicolon from end on lines in php.ini

ADDCreative:
php.ini files should not have a semicolon at the end on a line.

Can cause issues in some hosting configurations where the php.ini file is not processed directly by PHP. An example is cPanel with its MultiPHP INI Editor enabled. It will include the semicolon with the value leading to "charset `UTF-8;' not supported" errors.


jamesallsup:
Thanks @ADDCreative - ported back to maintenance branch too.

PR #6897, opened 2018-09-22
Solving race condition in session file handling

gbmaster:
When writing the session file on disk in the system/library/session/file.php, the fopen function in mode 'w' erases the file BEFORE acquiring the lock. This creates a race condition with the fopen in the read function, resulting in an empty session.

This resulted in annoying disconnections from the admin panel, but can be easily fixed by changing the fopen access mode from 'w' to 'c' and by emptying the file AFTER acquiring the lock by calling ftruncate.

### DIFF
--- a/upload/system/library/session/file.php
+++ b/upload/system/library/session/file.php
@@ -30,12 +30,12 @@ class File {
 	public function write($session_id, $data) {
 		$file = DIR_SESSION . 'sess_' . basename($session_id);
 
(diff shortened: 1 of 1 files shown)


gbmaster:
@jamesallsup @danielkerr 
Any chance this PR could be reviewed before the next release?

Thanks in advance :)

danielkerr:
did a blind merge. I hope you know what you are talking about

PR #6901, opened 2018-09-23
Fixed wishlist counter and empty wishlist label

amrikasir:
if you have a better way, please tell me..

### DIFF
--- a/upload/catalog/controller/account/account.php
+++ b/upload/catalog/controller/account/account.php
@@ -6,7 +6,17 @@ class ControllerAccountAccount extends Controller {
 
 			$this->response->redirect($this->url->link('account/login', 'language=' . $this->config->get('config_language')));
 		}
-
+		
+		//Load Model Wish List to access getTotalWishList()
+		$this->load->model('account/wishlist');
+		
+		//Load Wish List Language
+		$this->load->language('account/wishlist');
+		
+		//Overwrite var text_wishlist
+		$data['text_wishlist'] = sprintf($this->language->get('text_wishlist'), $this->model_account_wishlist->getTotalWishlist());
+		
+		//Overwrite lang variable with account
 		$this->load->language('account/account');
 
 		$this->document->setTitle($this->language->get('heading_title'));
--- a/upload/catalog/controller/account/wishlist.php
+++ b/upload/catalog/controller/account/wishlist.php
@@ -8,7 +8,9 @@ class ControllerAccountWishList extends Controller {
 		}
 
 		$this->load->language('account/wishlist');
-
+		//parse text_empty to view
+		$data['text_empty'] = $this->language->get('text_empty');
+		
 		$this->load->model('account/wishlist');
 
 		$this->load->model('catalog/product');


straightlight:
Fixed in the pre-release. Well done!

ADDCreative:
This isn't a fix at all. For a start you don't need to set $data['text_empty'] in catalog/controller/account/wishlist.php as it's already auto loaded. Now it's just being set twice.

Second the changes to catalog/controller/account/account.php now make the wishlist wording inconsistent to the other wording on the my account page. Should stick with "Modify your wish list" so it matched the same action based wording of your the other links. Also the way it has been done in this pull request is just wrong. It you are going to change the string you need to modify the line below, not load a different language file that happens to have a string similar to the one you want.

https://github.com/opencart/opencart/blob/ca02cc73b4995c2f0a42b07ae38de5fbf12eb14b/upload/catalog/language/en-gb/account/account.php#L15

@straightlight Even worse is the way it has been added to the 3.0.3.4 Pre-release. Now the My Account page has `<title>My Wish List</title>`, because you are loading two language files. I would recommend you undo these changes.

straightlight:
@ADDCreative: Which step are you referring between the two? Or is it all of it?

ADDCreative:
@straightlight All of it.

straightlight:
@ADDCreative: A sprintf function is being suggested with the OP's instructions for the account/account controller. Are you sure this is the case stating that the wishlist title will not show correctly based on these changes?

ADDCreative:
@straightlight It has to do with the fact that you are loading two language files in catalog/controller/account/account.php. Both of which set heading_title.

```php
		$this->load->language('account/account');
		
		$this->load->language('account/wishlist');
```

This means when the customer goes to route=account/account the title (in the browser tab) is now saying "My Wish List" when it should say "My Account".

If you do want to change the "Modify your wish list" link text (I don't believe you should as it won't match the other links, but that is just my opinion), then edit the line below and only load the account/account language file. You shouldn't load the language from another controller just because it has a string similar to what you want to use.
https://github.com/opencart/opencart/blob/ca02cc73b4995c2f0a42b07ae38de5fbf12eb14b/upload/catalog/language/en-gb/account/account.php#L15

straightlight:
Ok. Either the text_wishlist needs to originate from the same language file rather than bringing two different ones for the same language key is what you're saying. If the language key cannot be found in one of them, still, it needs to originate from one language file to avoid conflicts of initialization between the language files.

I'll look into the best option. Re-opening.

straightlight:
There seem to be bigger issues with the account/wishlist controller with the wishlistotal method. It doesn't use a sprintf and the related language file already has a %s . I think this is the real issue here. Will fix it for the next update.

straightlight:
Ok. I have updated the core-pre and release-pre versions by adjusting the conventional names. The previous ones even confused me while looking at it. This commit may have to remain opened with the hopes it's also going to be fixed on the master branch to avoid further confusions in the future.

ADDCreative:
Why change the 3.0.3.4 Pre-release when the change has not been approved in the mater branch? As I said it's not a bug fix. The original poster just wants to change the wording from 'Modify your wish list' to 'Wish List (0)', which makes less sense.

Also you haven't removed the following from the 3.0.3.4 Pre-release.
https://github.com/opencart/opencart/pull/6901/commits/ad96f11db0fa805abadd59fb8f106bcf7bcd0c6e

straightlight:
The problem with wishlist was larger than expected. The conventional names were confusing as I have already explained on my previous reply. As for the text_empty however, correct, it has now been removed. I will update shortly.

ADDCreative:
There was no problems with the wishlist, only the ones this pull request created.

Also, by changing ```text_wishlist``` to ```text_wishlist_total``` you have now broken compatibility with any 3.0.3.3 theme that overrides account/account.twig, for no reason whatsoever.

As i have said before if you are going to break theme compatibility, follow your own version number guidance.
https://github.com/opencart/opencart#versioning

PR #6905, opened 2018-09-25
Update account.twig

fallax010:
variable replaced in `catalog/language/*/account/account.php`


straightlight:
Might be considered the for features-pre.

danielkerr:
this doe not happen any more since i changed the system to have a language instance for each controller.

PR #6906, opened 2018-09-25
Update account.php

fallax010:
$_['text_wishlist'] is already in use in `catalog/language/*/common/header.php`


straightlight:
Might be considered the for features-pre.

Issue #6746, opened 2018-07-17
Admin design/seo_url unable to edit


howdu:
After creating seo url we're unable to edit it because it's checking for unique query/keyword but not excluding itself. 

Push is required but not represented in html with required class.

danielkerr:
i dont think this is relivatye any more. open abnother ticket if this issue still exsits.

PR #6908, opened 2018-09-26
[Fixed Issue #6746 : related to Edit the sel_url at admin/index.php?route=design/seo_url/edit from admin end.]

vivek-webkul:
While editing any entry for the seo_url.php, was showing already "query and keyword already in use!".


danielkerr:
added manually

PR #6927, opened 2018-10-03
Recaptch validating once & remove white space before php tag

howdu:


danielkerr:
supposed to be that way.

PR #6938, opened 2018-10-08
The index method of the ControllerCommonFooter accepts the parameter $data

iBet7o:
used with the event system.


straightlight:
This suggestion does not make sense. Therefore, this commit will now be closed.

PR #6948, opened 2018-10-12
Simplify UK VAT Zone as 3972 is not valid

ADDCreative:
Zone 3972 was included in the UK VAT Zone, but is not in the zone table.
As list without 3792 matches all zones, simpler to just use all zones.


straightlight:

ADDCreative:
Why has this been closed. I can see that zone 3972 has correctly been removed. But it's pointless to still have all the zones added separately to UK VAT Zone.

PR #6951, opened 2018-10-14
Tiny bug which causes the rating stars not to show

MustafaSabur:


jamesallsup:
Dev branch is not used!

PR #6959, opened 2018-10-18
Font Awesome 5.4.1 in admin

stalker780:
Why not upgrade Font Awesome to latest version? 
And give more freedom to extension developers who can use this font in OC3.1.

I have checked CSS files.
There were no icon classes renamed or deleted. But over 100 new icons were added.
So it is quite compatible with existing admin template.

While OC3.1 is released it will get even more outdated.


danielkerr:
to many new versions coming out for me to accept this. i will be doing bootstrap and font awesome update before i do a release.

Issue #6935, opened 2018-10-08
[v3.1.0.0a1 - CSS] - Ongoing issue with ebay_listing.twig file


straightlight:
**What version of OpenCart are you reporting this for?**

**Describe the bug**
Ongoing reported issue based on this report: https://forum.opencart.com/viewtopic.php?f=191&t=166012&start=140#p644199

**To Reproduce**
Simply visit the codes section of this namespace. They CSS key entry still contains the typo.

**Expected behavior**
Will output the right CSS content.

**Screenshots / Screen recordings**
No screenshots required on this post.

**Server / Test environment (please complete the following information):**
Affects all browsers.

**Additional context**
No additional content.

andrei-bogdan:
The twig `catalog/view/theme/default/template/extension/module/ebay_listing.twig` right now looks like this for me:
`<div class="col-lg-3 col-md-3s col-sm-6 col-12">`
You mean it should be changed to this?
`<div class="col-lg-3 col-md-3 col-sm-6 col-xs-12">`

danielkerr:
all openbay stuff has been removed and not supported by opencart.com.

PR #6960, opened 2018-10-19
Fix for issue #6935

bogdan-manea:
removed bootstrap typos


danielkerr:
this has been removed already

PR #6962, opened 2018-10-19
Use getLatestProducts in latest module

ADDCreative:


straightlight:
Might be considered in the features-pre version later on.

PR #6970, opened 2018-10-25
fix: Add missing php tag

nadim1992:


danielkerr:
not exactly part of opencart.

PR #6979, opened 2018-10-29
added currency code to the eway refund

gerardadamczewski:


ADDCreative:
Lines 85 and 155 might need checking, if currency is not always to 2 decimal places.

```php
$refund_data->Refund->TotalAmount = (int)(number_format($refund_amount, 2, '.', '') * 100);
```

Issue #6980, opened 2018-10-29
[Issue 3.1.0.0_b: The button Class is not changing on enabling the Cron Job.]


vivek-webkul:
**What version of OpenCart are you reporting this for?**
3.1.0.0_b
**Describe the bug**
While you are enabling the Cron Jobs at the admin end, the button class is not changing. 

What section does it affect?
Admin->Extensions->Cron Jobs

**To Reproduce**
1. Go to 'Admin->Extensions'
2. Click on 'the Enable button for the Cron job'.
3. See error: Button class remains to Success (green color).

**Expected behavior**
There must danger class while enabling the Cron job.

**Screenshots / Screen recordings**
![cron jobs](https://user-images.githubusercontent.com/28401696/47650399-2533b580-dba6-11e8-8fee-f5a995294915.jpg)

danielkerr:
No. I already fixed it before I saw ur commit

PR #6981, opened 2018-10-29
[Fixed Issue: #6980 : Cron job button class corrected.]

vivek-webkul:
An issue fixed while enabling the Cron job, then button class remains same.
cron jobs 1


danielkerr:
i think fixed already

PR #6986, opened 2018-10-30
Font awesome unicodes

avneesh-webkul:
Updated font family while using unicodes for font awesome in css.
It was creating issue at many points, Ex - Slideshow on the home page, accordion on the cart page.


danielkerr:
already gone past this version

PR #7005, opened 2018-11-05
Show actual PHP version for install view

Kolpikov:
Updated install view in order to display correct minimum required version of PHP.

Currently it shows that Opencart requires PHP 5.6+, but in upload/system/startup.php:6 declared minimum version to be not less then 7.0.0.
image


Kolpikov:
Already fixed: https://github.com/opencart/opencart/issues/7281

PR #7014, opened 2018-11-08
Fix for affiliate mail links

ADDCreative:


straightlight:
Already fixed in the pre-release. However, it's still appealing in the master branch. Good catch!

PR #7015, opened 2018-11-08
[Fixed #7011 issue for the checkout while login with returning custom…

### LINKED ISSUE #7011: [3.1.0.0_b] Checkout for Returning Customer login broken
**Situation**
Customer is going direct without login to the products and ordering a product. The he start the checkout process. He choosing a to checkout as returning customer. He enter his login data (User/Pass) and press Continue

**Problem**
An error message occur.
image

Webkul-Cart:
There is a if condition in the system/lebrary/cart/customer.php file in the login fucntion
  if (password_verify($password, $customer_query->row['password']) ) { 
Here if this condition will be executed then variable $new_password_hashed will never be registered and in the line 59 in the same fucntion
 if ($new_password_hashed) { 
this condition will  show warning for the undefined variable. Hence i made it as 
if (isset($new_password_hashed) && $new_password_hashed) { 
to fix the issue.


straightlight:
Already added in the master branch. It might also be added in the features-pre in the future. Therefore, this commit will now be closed.

PR #7018, opened 2018-11-08
[Fixed #7016 issue for the Get Request language variable]

### LINKED ISSUE #7016: while changing currency from the front-end GET request variable for the language is added twice in the URL
When we change currency from the header currency drop-down URL is like 
index.php?route=common/home&language=en-gb&language=en-gb 
where Get Request varaible language is added twice.
screenshot_4

Aqib-Webkul:
Fixed issue #7016 while changing currency from the front-end GET request variable for the language is added twice in the URL.


danielkerr:
already done

PR #7030, opened 2018-11-12
Update footer.php

osworx:
double


straightlight:
Confirmed. Issue found in the master branch. Good catch!

PR #7043, opened 2018-11-15
Fixing typo

ShawnCorrigan:
There is a typo in the Google validate function allowing users to bypass the Google Captcha

### DIFF
--- a/upload/catalog/controller/extension/captcha/google.php
+++ b/upload/catalog/controller/extension/captcha/google.php
@@ -17,7 +17,7 @@ class ControllerExtensionCaptchaGoogle extends Controller {
     }
 
     public function validate() {
-		if (empty($this->session->data['gcapcha'])) {
+		if (empty($this->session->data['gcaptcha'])) {
 			$this->load->language('extension/captcha/google');
 
 			if (!isset($this->request->post['g-recaptcha-response'])) {


ShawnCorrigan:
I didn't see anything that was majorly different between this logic and the regular Captcha logic. Fixing the typo seemed to have fixed the problem for me.  I will fix the second one.

ADDCreative:
You will probably find that fixing the second typo will bring the problem back.

ADDCreative:
That makes no sense. $this->session->data['gcaptcha'] is only used in those 2 places. If you correct both the logic will be exactly the same, regardless of the spelling.

ShawnCorrigan:
Oh you're right, that is odd. I will have to look further into it.

ShawnCorrigan:
I figured out that the original issue was. I didn't realize that $this->session->data was a wrapper for session variables (go figure). The issue was that the SESSION variable is never unset, and a CAPTCHA completion is only needed once per session. This would allow someone to complete the CAPTCHA once, and then proceed to perform any other action an unlimited amount of times in a session without completing the CAPTCHA again. This defeats the whole purpose of a CAPTCHA.

Changing one of the typos yet not the other made it so that it requires a CAPTCHA every time because it is checking if a variable is set that will never be set.

I should have cancelled this PR once I realized that the typo was in there twice, I didn't get a chance to look further into it. I'm not sure why this was merged, although in a sense it is more secure.

Can you give some insight into why you store whether the CAPCTHA was completed or not in a session? I'm thinking it might be so that subsequent requests can be made that would both need CAPTCHA verification,, but you were for some reason unable to store the CAPTCHA variable in both requests, Whatever the reason is, this should be looked into and fixed so that you are not storing this in the session.

danielkerr:
yes you are not supposed to do a catpcha more than once. other wise its tiresome.

ShawnCorrigan:
If this is the case than there should be a check in the function `index()` to see if it has been completed before showing the CAPTCHA.  Regardless, this is poor practice and defeats the point of a CAPTCHA. 

Google's CAPTCHA decides whether it's necessary for someone to complete the verification process of a CAPTCHA or whether just checking the box is sufficient, that decision should be left to them.

PR #7055, opened 2018-11-29
Update gdpr_export.php

osworx:
text_country is double


danielkerr:
added it manually. cheers

PR #7058, opened 2018-11-30
Update gdpr_export.php

osworx:
Either delete or export, but because we are here in the export file, I am sure you meant export ..


straightlight:

PR #7093, opened 2018-12-10
Duplicates

Gorobez:
Lines 163, 235 duplicates 
if (!$json) {


ADDCreative:
It's not a duplicate as ```$json``` could be set in the block above.

Gorobez:
If the $json variable does not in the first block (163-233), then it will not in the second block (235-342).
Therefore, lines 233-235 can be removed.

PR #7112, opened 2018-12-27
Improvement on Amazon Pay and Login with Amazon pay

simo-vt:
Suppress error messages triggered after the payment gets processed. This is to ensure the checkout flow does not get interrupted.


danielkerr:
this is not a good fix. you need to tell users to set the settings to error display to no.

PR #7115, opened 2018-12-28
changed session db engine

FarahmandM:
Fixed garbage collector when using db as session engine. The old session records will be deleted from session table. This will improve performance of stores.


danielkerr:
updated manually

PR #7136, opened 2019-01-08
Update customer_order.php

osworx:
1 variable is enough


straightlight:
Already fixed in the pre-release. However, it is still appealing in the master branch.

PR #7146, opened 2019-01-12
Paypal Standard State code 3.0.x.x branch

condor2:


ADDCreative:
This is only correct for a handful of countries. https://developer.paypal.com/docs/classic/api/state_codes/

The the other countries the zone name and not code should be use. See this PayPal example. https://www.paypal.com/uk/cgi-bin/webscr?cmd=_pdn_xclick_prepopulate_outside

A test similar to the one in the PayPal Express module should be used.
https://github.com/opencart/opencart/blob/3.0.3.1/upload/catalog/controller/extension/payment/pp_express.php#L1311-L1324

Although, the ISO code should be use not the database entry ID!

straightlight:
`Although, the ISO code should be use not the database entry ID!`

Correct. This is also being discussed on this commit: https://github.com/opencart/opencart/issues/7372 to categorize these ISO codes as per service providers compliance rather than referring to the local store's geographic settings aside from the customers registration profiles with the exception of looking up the customers' addresses during checkout or via the admin API.

Issue #7152, opened 2019-01-16
OC 3.1.0.0_b  Twig ignoring template variables added by view/*/before events


mhcwebdesign:
The OpenCart 3.1.0.0-b Twig rendering engine ignores template variables added by view/*/before events when added via lambda functions. E.g. something like this won't work:

```
<?php
class ControllerExtensionModuleGoogleTagManager extends Controller {

	public $lambda;

	public function callback(&$code) {
		$lambda = $this->lambda;
		$lambda($code);
	}

	// event handler for catalog/view/common/footer/before
	public function ViewCommonFooterBefore( &$route, &$data, &$template=null ) {
		// .....
		// Attach as template filter
		$template->addFilter('extension/module/google_tag_manager/ViewCommonFooterBefore-' . $this->config->get('config_store_id'), $this);
		// .....
		// event handler will run when template calls back this lambda function
		$this->lambda = function (&$code) use (&$route, &$data) {
			$data['new_var'] = 'whatever';
			// .....
			// do some changes to template code
		}
	}
?>
```
This basically limits the lambda function to changing the template code only, changing the $data array in there has no effect in the final Twig rendering.

This whole concept of using lambda functions is cumbersome, surely there are better approaches!

danielkerr:
seems to be the same issue

https://github.com/opencart/opencart/issues/7230

Issue #7151, opened 2019-01-16
OC 3.1.0.0_b Twig has no working auto-refresh


mhcwebdesign:
There is no working auto-refresh implemented in OpenCart's twig.

When a change in a template is done in an event handler's view/*/before lambda function, and the lambda function was changed in its logic  to the way it changes the template, it won't kick in in a subsequent rendering via Twig. 

I even tried 'auto_reload' => true for the Twig environment's configuration, to no avail.

This is bad, we shouldn't have to resort to manually refreshing template caches via the admin's Developer Settings.

Issue #7149, opened 2019-01-14
OC 3.1.0.0_b Twig doesn't accept changes from view/*/before events


mhcwebdesign:
After turning off caching for Theme and SASS in the admin Developer Settings the Twig rendering ignores all changes from lambda functions set by view/*/before events, including changes from the theme editor. In particular, all catalog/view/*/before handlers of priority 500 or higher ('event/theme' and others with template modifications) are ignored.

Is there any reason why this thing has to be so messy? Is there any reason why we can't use the Twig_Loader_Array instead of Twig_Loader_Filesystem? There are more bugs which I will report separately tomorrow.

danielkerr:
you have to attach the twig filter.

mhcwebdesign:
I did, using $template->addFilter(...) calls. And it works fine as long as caching is turned on in the admin Developer Settings for Theme and SASS. But it fails when caching is switched off.

mhcwebdesign:
Daniel, it's late at night here in the UK. But I'll post a working solution for this Twig issue, as well as for the other ones I posted about Twig (see @7152 and @7151), tomorrow morning.

mhcwebdesign:
I got a working solution for twig, still doing some more testing before posting it here.

danielkerr:
fixed here
https://github.com/opencart/opencart/pull/7153

PR #7153, opened 2019-01-17
OC 3.1.0.0-b: Twig fixes for issues #7152, #7151, #7149

mhcwebdesign:
Hi Daniel,

Please find attached pull request for issues #7152, #7151, #7149

It basically makes OC 3.1.0.0 backward compatible for view/*/before events.
And it solves the problems with Twig caching, including auto-refresh, cooperates nicely with the theme editor changes. There is no more need for cumbersome lambda functions. And last not least, it can now automatically detect whether a template is twig or php before rendering. Finally, the rendering time, therefore the overall page load time, is now faster, too.

### DIFF
--- a/upload/catalog/controller/event/theme.php
+++ b/upload/catalog/controller/event/theme.php
@@ -1,6 +1,5 @@
 <?php
 class ControllerEventTheme extends Controller {
-	public $lambda;
 
 	public function index(&$route, &$args, &$template) {
 		if (!$this->config->get('theme_' . $this->config->get('config_theme') . '_status')) {
@@ -14,33 +13,27 @@ class ControllerEventTheme extends Controller {
 			$directory = $this->config->get('config_theme');
 		}
 
-		if (is_file(DIR_TEMPLATE . $directory . '/template/' . $route . '.twig')) {
+		$this->load->model('design/theme');
+		$theme_info = $this->model_design_theme->getTheme($route, $directory);
+		if ($theme_info) {
+			$template = html_entity_decode($theme_info['code'], ENT_QUOTES, 'UTF-8');
 			$this->config->set('template_directory', $directory . '/template/');
+			$this->config->set('template_engine','twig');
+		} else if (is_file(DIR_TEMPLATE . $directory . '/template/' . $route . '.twig')) {
+			$this->config->set('template_directory', $directory . '/template/');
+			$this->config->set('template_engine','twig');
+		} else if (is_file(DIR_TEMPLATE . $directory . '/template/' . $route . '.tpl')) {
+			$this->config->set('template_directory', $directory . '/template/');
+			$this->config->set('template_engine','template');
 		} elseif (is_file(DIR_TEMPLATE . 'default/template/' . $route . '.twig')) {
 			$this->config->set('template_directory', 'default/template/');
+			$this->config->set('template_engine','twig');
+		} elseif (is_file(DIR_TEMPLATE . 'default/template/' . $route . '.tpl')) {
+			$this->config->set('template_directory', 'default/template/');
+			$this->config->set('template_engine','template');
+		} else {
+			exit("Error: Unable to find template for '$route'!");
 		}
-
-		// Attach to the template
-		$template->addFilter('theme-override-' . $this->config->get('config_store_id'), $this);
-
-		// If you want to modify the output of the template we add a
-		$this->lambda = function (&$code) use (&$route, &$args, &$directory) {
-			// If there is a theme override we should get it
-			$this->load->model('design/theme');
-
-			$theme_info = $this->model_design_theme->getTheme($route, $directory);
-
-			if ($theme_info) {
-				$code = html_entity_decode($theme_info['code'], ENT_QUOTES, 'UTF-8');
-			}
-		};
 	}
 
-	// Ridiculous we have to use these work around's because magic methods can not pass by reference!
-	public function callback(&$code) {
-		// Genius
-		$lambda = $this->lambda;
-
-		$lambda($code);
-	}
 }
--- a/upload/system/engine/loader.php
+++ b/upload/system/engine/loader.php
@@ -110,20 +110,23 @@ final class Loader {
 		// Keep the original trigger
 		$trigger = $route;
 
-		$template = new Template($this->registry->get('config')->get('template_engine'));
+		// Modified template contents. Not the output!
+		$template_code = null;
 
 		// Trigger the pre events
-		$result = $this->registry->get('event')->trigger('view/' . $trigger . '/before', array(&$route, &$data, &$template));
+		$result = $this->registry->get('event')->trigger('view/' . $trigger . '/before', array(&$route, &$data, &$template_code));
 
 		// Make sure its only the last event that returns an output if required.
 		if ($result && !$result instanceof Exception) {
 			$output = $result;
 		} else {
+			$template = new Template($this->registry->get('config')->get('template_engine'));
+
 			foreach ($data as $key => $value) {
 				$template->set($key, $value);
 			}
 
-			$output = $template->render($this->registry->get('config')->get('template_directory') . $route, $this->registry->get('config')->get('template_cache'));
+			$output = $template->render($this->registry->get('config')->get('template_directory') . $route, $this->registry->get('config')->get('template_cache'), $template_code);
 		}
 
 		// Trigger the post events
--- a/upload/system/library/template.php
+++ b/upload/system/library/template.php
@@ -32,15 +32,6 @@ class Template {
 	/**
 	 *
 	 *
-	 * @param    mixed $value
-	 */
-	public function addFilter($key, $value) {
-		$this->adaptor->addFilter($key, $value);
-	}
-
-	/**
-	 *
-	 *
 	 * @param    string $key
 	 * @param    mixed $value
 	 */
@@ -53,10 +44,11 @@ class Template {
 	 *
 	 * @param    string $template
 	 * @param    bool $cache
+	 * @param    string &$template_code (optional modified template code from pre-events)
 	 *
 	 * @return    string
 	 */
-	public function render($template, $cache = false) {
-		return $this->adaptor->render($template, $cache);
+	public function render($template, $cache = false, &$template_code=null) {
+		return $this->adaptor->render($template, $cache, $template_code);
 	}
 }
--- a/upload/system/library/template/template.php
+++ b/upload/system/library/template/template.php
@@ -1,68 +1,61 @@
 <?php
 namespace Template;
 final class Template {
-	protected $code;
-	protected $filters = array();
 	protected $data = array();
-
-	public function addFilter($key, $value) {
-		$this->filters[$key] = $value;
-	}
 
 	public function set($key, $value) {
 		$this->data[$key] = $value;
 	}
 
-	public function render($filename, $cache = true) {
-		$file = DIR_TEMPLATE . $filename . '.tpl';
-
-		if (is_file($file)) {
-			$this->code = file_get_contents($file);
-
-			foreach ($this->filters as $filter) {
-				$filter->callback($this->code);
-			}
-
-			ob_start();
-
-			if (!$cache && function_exists('eval')) {
-				extract($this->data);
-
-				echo eval('?>' . $this->code);
+	public function render($filename, $cache = true, $template_code=null) {
+		// render from modified template code if there
+		if ($template_code) {
+			$file = DIR_CACHE.'tpl/'.$filename.'.tpl';
+			if (is_file($file)) {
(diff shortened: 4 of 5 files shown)


straightlight:
For this file: https://github.com/opencart/opencart/pull/7153/commits/bbd109fe5cff11d80b92f66ff06b650e1c6e92df, 

This line: `$parts = explode('/', substr($dir,1));`

should be replaced with: `$parts = explode('/', rtrim(str_replace('\\', '/', $dir)));`

danielkerr:
what do u think of the template code changes. i need t update the twig part.

mhcwebdesign:
@danielkerr : Looks better now, we are getting there!

Just a minor change to the 'library/template.php' where the method 'render' should look like this:

```
	public function render($filename, $code = '') {
		return $this->adaptor->render($filename, $code);
	}
```
instead of
```
	public function render($filename, $code = '', $data = array()) {
		return $this->adaptor->render($filename, $code, $data);
	}
```
The data is already set at 'system/engine/loader.php' in the 'view' method, so no need to pass $data as an argument to the 'render' method.

BTW.: I know you are a fan of Twig, but it would be easy enough to add support the old PHP template engine, too, in the 'catalog/controller/event/theme.php', see my original pull request for this file!

straightlight:
In system/library/template/template.php file,

`file_put_content($file, $code, LOCK_EX);`

should read as:

`file_put_contents($file, $code, LOCK_EX);`

straightlight:
In system/library/template/twig.php file, the twig environment is loaded twice:

// 2. Initiate Twig Environment

		$twig = new \Twig_Environment($loader, $config);



		try {

			$twig = new \Twig_Environment($loader, $config);

danielkerr:
ok should be ok now.

osysltd:
I guess exactly after this merge include directives in TWIG templates stopped working
@danielkerr @mhcwebdesign would you please help to solve the issue?
see https://github.com/opencart/opencart/issues/7443

danielkerr:
Will check

mhcwebdesign:
@osysltd: Your problem is caused by an issue in an OpenCart extension, namely the 'yandex money', it is not an OpenCart bug. Your extension should only load the template in the controller, like this:

```
		...
		$kassa = $this->load->view('extension/payment/yandex_money/kassa', $data);
		...
```
and then refer to the 'kassa' variable in the template like this:
```
<div class="tab-pane{{ (lastActiveTab == 'tab-kassa' ? ' active' : '') }}" id="tab-kassa">
   {{ kassa }}
</div>
```
Ask the authors of this extension to do a proper rewrite of their payment extension!

osysltd:
Before this commit includes in TWIG were working fine.

straightlight:
@osysltd: In other words, what you are implying is since the implementation of the modified TWIG engine, none of your events are working or it's only from the extension this error message appears? Typically, if other events do work, in general, then the TWIG engine does respond as it should.

osysltd:
@straightlight What Im trying to say that include directives stopped working

https://twig.symfony.com/doc/2.x/tags/include.html

straightlight:
The OC core does not have any example with include calls which is why it may have never been considered if true case were found. However, including files from one-and-another is not in the best practice of programming when parsing TWIG files. Which is why, we use controllers to output the results into the TWIG files rather.

If the extension does not involve a license, could you provide some example from the controller method parsing the TWIG file and the affected TWIG file's area where the inclusion is acquired?

osysltd:
The extension what Im talking about is https://github.com/yandex-money/yandex-money-ycms-opencart3
And about the best practices for the programming, I guess its not the best practice to put different not related to each other pieces of code into one template.


https://github.com/yandex-money/yandex-money-ycms-opencart3/blob/967b7b62e1681519f8a153ac4e528a9a5449f924/src/upload/admin/controller/extension/payment/yandex_money.php#L337

loads 

https://github.com/yandex-money/yandex-money-ycms-opencart3/tree/master/src/upload/admin/view/template/extension/payment

straightlight:
`I guess its not the best practice to put different not related to each other pieces of code into one template.`

Correct, since OS is a CMS platform, controllers are the primary source of constructors where projects are being developed before being outputted on either TPL or TWIG files.

Therefore, by looking at the provided links above, what should of been used by this extension developer is an Ajax request with 'html' dataType or to retrieve the required data with 'json' dataType and to append the results to those DIV containers on the extension rather than including other TWIG files from this file: https://github.com/yandex-money/yandex-money-ycms-opencart3/blob/master/src/upload/admin/view/template/extension/payment/yandex_money.twig . Also take note that TWIG paths may also differ based on the OC Loader Engine that requires your extension to load.

osysltd:
Thanks for your kind recommendations but instead of rewriting bicycle I would rather ask to fix the issue with OpenCart

straightlight:
As said on the above, it is not an issue with OC but rather with the TWIG Engine. Besides, writing to a PHP  controller before outputting the results provides way more advantages than being restricted with the re-write functions in the TWIG files.

osysltd:
TWIG out of the box supports includes properly, includes are not working within OpenCart

straightlight:
Have it your way. If you can prove, however, with the original TWIG Engine does work with include versus the replacement, could you please post the results?

osysltd:
I would rather see your test results, have you done any before claiming?
> As said on the above, it is not an issue with OC but rather with the TWIG Engine.

Again, before patches includes have been working fine for ages, now they are not working despite supported functionality by TWIG engine. This is an OpenCart issue which has appeared just recently.

straightlight:
Could you post a screenshot or provide both URLs of what you're reporting about?

osysltd:
https://twig.symfony.com/doc/2.x/tags/include.html

https://github.com/yandex-money/yandex-money-ycms-opencart3/blob/967b7b62e1681519f8a153ac4e528a9a5449f924/src/upload/admin/view/template/extension/payment/yandex_money.twig#L46

```
Twig_Error_Loader: Template "extension/payment/yandex_money/kassa.twig" is not defined in "extension/payment/yandex_money.twig" at line 46. in C:\OSPanel\domains\oshp\storage\vendor\twig\twig\lib\Twig\Loader\Array.php on line 65
```

straightlight:
Are you able to reproduce the same issue with a core TWIG file rather than an extension file?

osysltd:
Yes, as soon as an include directive in place, the engine is not able to find template to which include is referring to.

straightlight:
Great. Please specify which file I should try on a core TWIG file to load another core TWIG file noticing if the same issue occurs on the modified version of the TWIG engine.

osysltd:
just try to include any twig into another, no matter which.

straightlight:
Even in the core TWIG engine, the include does not seem to work. As I just tried, it may be the reason why the include is never used. A good thing the suggested theme override engine is for PHP. At least, it's not as limitless on that end.

osysltd:
May I ask to fix the bug?

osysltd:
Dear Daniel, @danielkerr 
I can see that @mhcwebdesign switched loader from filesystem
```
$loader = new \Twig_Loader_Filesystem(DIR_TEMPLATE);
```

to an array loader

```
$loader = new \Twig_Loader_Array(array($filename . '.twig' => $code));
```

which in its turn requires to have all includes defined in the key - value pair: 
https://github.com/symfony/symfony/issues/5553#issuecomment-8726614
I guess we still need to figure out how to provide compatibility, @mhcwebdesign @danielkerr  any ideas?

mhcwebdesign:
@osysltd: I'll have to test the {% include %}. If it does indeed fail, then the only option I can think of would be to implement our own array loader. Or even better, we could use a ChainLoader which in turn uses the ArrayLoader first, and then a FilesystemLoader as a fallback.

osysltd:
@mhcwebdesign looking forward to implementation.

mhcwebdesign:
@osysltd I looked into it. In file system/library/template/twig.php replace the line

```
			$loader = new \Twig_Loader_Array(array($filename . '.twig' => $code));
```

with these lines:
```
			$loader1 = new \Twig_Loader_Array(array($filename . '.twig' => $code));
			$loader2 = new \Twig_Loader_Filesystem(array(DIR_TEMPLATE)); // to find further includes
			$loader = new \Twig_Loader_Chain(array($loader1, $loader2));
```
The second loader will process Twig's {% include %} instructions.

mhcwebdesign:
@danielkerr This Twig 2.6.2 version, even with its caching enabled, is slower than the older version used in OC 3.0.x releases. There is no advantage in using Twig when PHP can do the same job much faster!

IMHO the OpenCart core template files should be switched back to PHP templates!

danielkerr:
execpt handle missing vars and being able to store template files in the db without getting hacked.

mhcwebdesign:
PHP handles missing vars brilliantly, it lets you if there such an error, unlike Twig which silently ignores them. And you can store template files in the DB regardless of what template engine you use, though it doesn't make much sense anyway.

The fact is, the current implementation of Twig has a poor performance!

osysltd:
I fully support @mhcwebdesign cause majority of customization is done in twig template files on many sites, storing in DB doesnt make much sense and doesnt introduce simplicity while editing in IDE, also database is not for such purposes imho

straightlight:
The only downside on bringing the PHP language back into TPL files are the vast possibilities that can be implied into custom theme and almost to an infinity solution where forum supporters, and perhaps here, would feel probably harden to find the root cause as compared to the TWIG Engine where limited functions are being used to introduce the similarities of what the core originally presented with PHP prior to the TWIG files.

Granted, PHP allows to do way more but also allow custom functions / methods to be used without anyone to know other than those who have the proper knowledge to troubleshoot such issues. Not saying that the TWIG cache is quite friendly, in fact, it really ain't to be honest since I do feel as well the mass slowness of the 56k modem versus the fiber optic speed solutions. However, it would be to know how to optimize the cache engines despite if TWIG or PHP is being used.

danielkerr:
@osysltd storing of templates is required if you want the customer to have a template edityor in the admin. eval is very unsecure

danielkerr:
forget it. to many templates out there.

fallax010:
<file path="system/library/template/twig.php" error="skip" v="OC3.0.3.6">
	<operation error="skip" info="add support for include()">
		<search><![CDATA[
		$loader = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
		]]></search>
		<add position="replace"><![CDATA[
		$loader1 = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
		$loader2 = new \Twig\Loader\FilesystemLoader(array(DIR_TEMPLATE)); // to find further includes
		$loader  = new \Twig\Loader\ChainLoader(array($loader1, $loader2));
		]]></add>
	</operation>
</file>

fallax010:
```
<file path="system/library/template/twig.php" error="skip" v="OC3.0.3.6">
	<operation error="skip" info="add support for include()">
		<search><![CDATA[
		$loader = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
		]]></search>
		<add position="replace"><![CDATA[
		$loader1 = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
		$loader2 = new \Twig\Loader\FilesystemLoader(array(DIR_TEMPLATE)); // to find further includes
		$loader  = new \Twig\Loader\ChainLoader(array($loader1, $loader2));
		]]></add>
	</operation>
</file>
```

jethvavipul20:
> ```
> <file path="system/library/template/twig.php" error="skip" v="OC3.0.3.6">
> 	<operation error="skip" info="add support for include()">
> 		<search><![CDATA[
> 		$loader = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
> 		]]></search>
> 		<add position="replace"><![CDATA[
> 		$loader1 = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
> 		$loader2 = new \Twig\Loader\FilesystemLoader(array(DIR_TEMPLATE)); // to find further includes
> 		$loader  = new \Twig\Loader\ChainLoader(array($loader1, $loader2));
> 		]]></add>
> 	</operation>
> </file>
> ```

This solution is perfectly work in my opencart version 3.0.3.8 version. Thanks

mhcwebdesign:
@BuslikDrev : To quote Daniel's own words:

> The Twig syntax is good, but the implementation and the available methods is a joke!
> All the Symfony developer has done is create a garbage frame work putting 3rd party scripts into DI containers.
> The Twig syntax he ripped off from Jinja and Django templates then did a garbage implementation!
> 
> The fact that this system cache is just compiling php into more php code instead of html is a disgrace!
> 

That's why we created an alternative Twig engine which translates directly from Twig to the final HTML output, no interim PHP files or caches needed. It all run's at near-native PHP speeds. We are currently testing it on a few websites, including a live site, and hope to release it soon under a GPL license.

PR #7164, opened 2019-01-31
Fix bugs preventing square payment from working

falconkirtaran:
Fixes a SQL error in the square plugin that caused it to try to pull transactions starting at negative numbers in some cases, and a Javascript error whereby the confirm transaction buttons were bound to an element with the wrong ID.


falconkirtaran:
Superseded by #7315

PR #7180, opened 2019-02-15
fix wishlist text in account/account

pravdomil:
image


straightlight:
This solution has now been added in the pre-release. Issue also identified in the master branch. Nice catch!

danielkerr:
fixed ages ago

PR #7221, opened 2019-02-19
Fixing issues with unexpected token

svermigo:
Fix for error during uploading and instaling ocmod.zip files. 
Unexpected token < in json at position 0 error


straightlight:
An else statement should be added along with:

`if ($directories)` statements since, right now, it does validate directories but won't return an error from JSON when directories won't be found particularly. By identifying to validations, it would ease troubleshooting on the forum.

danielkerr:
i think this code is now redundant

PR #7255, opened 2019-03-10
All OC v3.x releases - pagination fix

straightlight:
Followed is a solution I created on the Opencart forum that will resolved the page=0 queries along with all other non-numeric ones: https://forum.opencart.com/viewtopic.php?f=24&t=210459#p748892


straightlight:
Closing this commit since an official solution with (int)$this->request->get['page'] has been posted instead.

PR #7264, opened 2019-03-11
fix duplicate setting key 'total_voucher_status' and 'total_voucher_s…

bkoopy:
…ort_order' causing unable to disable voucher extension


danielkerr:
just tested this and can not find any issue. voucher can be disabled.

PR #7266, opened 2019-03-13
Fixed if sending mail with UTF-8 encoded body, mail sending to die

engwalid1971:


straightlight:
This fix might also be for server-specifics since other OC users may not encounter any issues while sending emails with the email class.

engwalid1971:
For sending an Arabic, Japanese or other Unicode languages you must encoding it.

straightlight:
@danielkerr: Since email encoding are based on regional settings; to add:

- Encoding option in admin - > systems - > settings - > add / edit settings - > mail tab
- Use the parent call for $this->registry in system/library/mail.php file to validate the encoding method with an if statement. Else, default settings from the method as is.

straightlight:
Alternatively, the mail setting would need to be used such as this solved issue: https://forum.opencart.com/viewtopic.php?f=202&t=211530

engwalid1971:
What is the downside of my solution ?

engwalid1971:
What is the disadvantages  of my solution ?

engwalid1971:
The best solution is using 'quoted_printable_encode' function\, it will encode the unicode characters only not the whole message ( gmail do the same ), for example:
`<html><body><p>cómo estás</p></body></html>`
will be:
`<html><body><p>c=C3=B3mo est=C3=A1s</p></body></html>`
and
`<html><body><p>how are you</p></body></html>`
will be:
`<html><body><p>how are you</p></body></html>`
with no changes

- system/library/mail/mail.php
- system/library/mail/smtp.php
Replace:
```
    $message .= 'Content-Type: text/html; charset="utf-8"' . PHP_EOL;
    $message .= 'Content-Transfer-Encoding: 8bit' . PHP_EOL . PHP_EOL;
    $message .= $this->html . PHP_EOL;
```
With:
```
    $message .= 'Content-Type: text/html; charset="utf-8"' . PHP_EOL;
    $message .= 'Content-Transfer-Encoding: quoted-printable' . PHP_EOL . PHP_EOL;
    $message .= quoted_printable_encode($this->html) . PHP_EOL;
```

straightlight:
The only disadvantage by using those codes is that many users on the forum reported email encoding issues due to server restrictive-based on locale and charset Unicode settings. In order to work entirely, we'd need to define each locations where the locale settings would match the OC email class.

Your solution may work on your end. Unfortunately, it may not work as provided for other regions. While there are scripts like PHPMyAdmin that can successfully submit emails from several regions, it would require similar results with the OC email class at this point.

straightlight:
Further development has been done on this suggested fix. It should now be safe to use it as per engwalid1971 .

engwalid1971:
> @straightlight:
>> @danielkerr: Since email encoding are based on regional settings; to add:
>> 
>>     * Encoding option in admin - > systems - > settings - > add / edit settings - > mail tab
>>     * Use the parent call for $this->registry in system/library/mail.php file to validate the encoding method with an if statement. Else, default settings from the method as is.

My suggestion:
Add encoding option in admin - > systems - > settings - > add / edit settings - > mail tab
. 8bit
. Base64
. quoted-printable
* 8bit - The current/original code.
* Base64 - My modifications.
* quoted-printable - Encoding text/plain, text/html, from, reply-to and subject using quoted-printable, but the attachments encoding using base64.

straightlight:
Followed is an update about these modifications you made: https://forum.opencart.com/viewtopic.php?f=190&t=211599#p754444 . As per this user is saying on the topic, it has been handled by server-specific settings by using a trigger after trying your modifications.

engwalid1971:
SSL/TLS working with my modifications, also working with the current OC code.
SSL hostname must start with 'ssl://' also TLS hostname must start with 'tls://' in all cases.

You can test the current OC code with Gmail SMTP server,
SMTP Hostname: ssl://smtp.gmail.com
SMTP Port: 465
Username: example@gmail.com
It will be working.

Please note that, my modifications not modify the core of OC system, but change the encoding of message only and **solve the issues if sending an Unicode emails** ( some servers have a trouble for sending a Unicode messages ).

straightlight:
I guess there's no other choice than gathering each incident requests case-by-case noticing if this is an email connectivity or Unicode issue at this point.

straightlight:
It would seem this issue with charset versus Unicode has been resolved since PHP v7.2.4 and above: https://bugs.php.net/bug.php?id=62545 . Your solution should work without issues for any regions starting with this PHP version.

PR #7268, opened 2019-03-14
Change url to https to eliminate Document moved error

interbiznw:
Was getting these "Moved Permanently" Errors on all of the extension pages in the admin section, see below:
image

Fix: Change url to HTTPS in admin config.php


danielkerr:
I just did a mass search and replace for non https. thx for pointing this out.

PR #7277, opened 2019-03-17
Flexible featured module header

osysltd:
We have several featured modules across the page with different headers.
This change introduces flexibility for the module header.


ilcomp:
What about multilingual

osysltd:
Should be managed separately

straightlight:
This might be considered for the features-pre version.

PR #7297, opened 2019-03-29
Added initialization vector to Encryption class

garudacrafts:
This is a drop-in replacement that fixes the issue where openssl_encrypt emits an E_WARNING level error if an empty value is passed in via the iv parameter (https://www.php.net/manual/en/function.openssl-encrypt.php).

### DIFF
--- a/upload/system/library/encryption.php
+++ b/upload/system/library/encryption.php
@@ -11,6 +11,16 @@
 * Encryption class
 */
 final class Encryption {
+	private $cipher;
+	private $iv;
+	private $iv_length;
+
+    public function __construct() {
+		$this->cipher = 'aes-128-cbc';
+		$this->iv_length = openssl_cipher_iv_length($this->cipher);
+		$this->iv = openssl_random_pseudo_bytes($this->iv_length);
+    }
+	
 	/**
      * 
      *
@@ -20,7 +30,7 @@ final class Encryption {
 	 * @return	string
      */	
 	public function encrypt($key, $value) {
-		return strtr(base64_encode(openssl_encrypt($value, 'aes-128-cbc', hash('sha256', $key, true))), '+/=', '-_,');
+	    return strtr(base64_encode($this->iv . openssl_encrypt($value, $this->cipher, hash('sha256', $key, true), 0, $this->iv)), '+/=', '-_,');
 	}
 	
 	/**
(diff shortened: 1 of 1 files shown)


straightlight:
Is this vector validation method also compatible with PHP 7.3x?

straightlight:
This post addresses: https://forum.opencart.com/viewtopic.php?f=179&t=202581&start=20#p754235 . This methodology should be used instead with PHP Sodium: https://forum.opencart.com/viewtopic.php?p=737628#p737628 .

straightlight:
Further detailed explanation regarding the purpose of this proposed functionality: https://forum.opencart.com/viewtopic.php?f=179&t=202581&start=20#p754235

garudacrafts:
@straightlight I think I remember testing this with PHP 7.2. If there's a reason why 7.3 would cause a breaking change, I'm not aware.

garudacrafts:
@straightlight also, I see the comment about using the IV only once in the forum discussion at the link you provided.  give that, i suppose this pull request could be improved by removing the `$iv` class property and moving it to a local variable inside the `ecrypt()` method.

straightlight:
!@garudacrafts: What is being discussed on the address OC forum topic is that the authentication with the SSL handshake may not return any message event whenever an error case might be involved as compared to the php-sodium solution.  

In addition to the above statement, since your last evaluation was done with php v7.2x from the publication date of this commit and not with higher versions of PHP while the OC forum topic informs of testing the change with PHP v7.3.4, it means further testing has been made during the process as compared to your solution.

However, if you could modify your solution by ensuring the SSL handshake authentication can provide error response with the compression on a flexible query load level, both methods will surely be reviewed.

garudacrafts:
@straightlight I'm not so familiar with SSL handshakes.  Can such an error even be handled at the Application layer?  My projects implement SSL everywhere at the server level and it's my understanding requests would not be processed at the Application layer if a handshake error were to occur.  I've never experienced such a problem and never tested for it either.  Perhaps someone else with more experience could chime in and modify this pull request as you envision is necessary.

straightlight:
`Perhaps someone else with more experience could chime in and modify this pull request as you envision is necessary.`

That might be already done. Which is why, the address OC forum post has been provided above to apply this fix since it does provide authentication with error handling.

danielkerr:
fixed in latest commit

PR #7305, opened 2019-04-03
Fix max length of square sandbox token

falconkirtaran:
Square changed the format of the sandbox token to 64 characters, but the admin UI limits it to less than that.  Small patch to fix the limit.


falconkirtaran:
Superseded by #7315

PR #7306, opened 2019-04-03
Fixes for square breaking changes to oauth2 requirements

falconkirtaran:
Migrate square to use refresh tokens, as the old deprecated APIs to refresh without refresh tokens no longer work.  Add required grant type parameters for oauth2.  Fix SQL error that was preventing updating locations from square on fresh installs.

Without this patch, it's not possible to connect the square plugin to a square account right now, and locations won't update correctly even if it's currently connected.


falconkirtaran:
Superseded by #7315

PR #7315, opened 2019-04-05
Fixes for square payment extension

falconkirtaran:
Fixes multiple issues with the Square payment extension:

1. Sandbox token length has risen to 64 characters; amend the admin form to allow this length so that sandbox mode can be configured
2. Fix the name of the submit button in the checkout process so the corresponding event applies to it, so that checkout with square works
3. Update the token acquisition and refresh process to match Square's new refresh token requirements.  This will require clicking "connect" and signing into square once for existing users; new users cannot register using the old flow and the currently-used token renewal API is deprecated.
4. Add the PAYMENTS_WRITE scope so that charges and refunds can be made outside sandbox mode

### DIFF
--- a/upload/admin/controller/extension/payment/squareup.php
+++ b/upload/admin/controller/extension/payment/squareup.php
@@ -451,7 +451,7 @@ class ControllerExtensionPaymentSquareup extends Controller {
         $this->load->model('extension/payment/squareup');
 
         if (isset($this->request->get['page'])) {
-            $page = (int)$this->request->get['page'];
+            $page = max((int)$this->request->get['page'], 1);
         } else {
             $page = 1;
         }
@@ -462,7 +462,7 @@ class ControllerExtensionPaymentSquareup extends Controller {
         );
 
         $filter_data = array(
-            'start' => max(($page - 1), 0) * (int)$this->config->get('config_limit_admin'),
+            'start' => ($page - 1) * (int)$this->config->get('config_limit_admin'),
             'limit' => $this->config->get('config_limit_admin')
         );
 
--- a/upload/admin/controller/extension/payment/squareup.php
+++ b/upload/admin/controller/extension/payment/squareup.php
@@ -935,7 +935,7 @@ class ControllerExtensionPaymentSquareup extends Controller {
                 $this->error['sandbox_client_id'] = $this->language->get('error_sandbox_client_id');
             }
 
-            if (empty($this->request->post['payment_squareup_sandbox_token']) || strlen($this->request->post['payment_squareup_sandbox_token']) > 42) {
+            if (empty($this->request->post['payment_squareup_sandbox_token']) || strlen($this->request->post['payment_squareup_sandbox_token']) > 64) {
                 $this->error['sandbox_token'] = $this->language->get('error_sandbox_token');
             }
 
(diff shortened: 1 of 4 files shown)


danielkerr:
max(($page - 1), 0) needs removing an putting back to the way all other pages are set.

falconkirtaran:
Hey @danielkerr, the reason for the max is that mysql doesn't accept a range that has negative numbers.  Is there another way we're assuring the number is not negative?  I was seeing SQL errors in the log indicating that it was choking on a negative number in limit when doing many things, for example trying to refresh the token:

`Exception:  Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '-20, 20' at line 1<br />Error No: 1064<br />SELECT * FROM `oc_squareup_transaction` ORDER BY created_at DESC LIMIT -20, 20 in /var/www/opencart/upload/system/library/db/mysqli.php on line 43`

Is something else wrong here?  For example is the page supposed to be 1-indexed instead of 0-indexed?  Or is something else supposed to floor that value to 0?  MySQL syntax doesn't accept negative numbers there.

falconkirtaran:
Ok @danielkerr, I think I got the answer for myself.  How's my patch now?

danielkerr:
I'll look into it. I need time. Just doing s few things.

falconkirtaran:
No problem, and good luck with your current change.

bhmhrex1:
How do I get these files? I am new to github. I went to files changed, viewed each individual file and copied and pasted each code out of the file on here and into a new file on my desktop, then replaced on my ftp account. I am at a loss as if I did this correctly, it has not fixed my issue but given me the new following issue:

"An error occurred when trying to refresh the extension's connection to your Square account. Please verify your application credentials and try again."

Then if I click "reconnect" - it directs me to the square login, I sign in and then see this error:

"An error occurred
Invalid value for parameter redirect_uri"

bhmhrex1:
So, I went to each directory on the github and saved each file. I then uploaded to the respective directories on my server......Now my admin backend for my site directs to this github pull.........

falconkirtaran:
You might need to save your configuration again, but really I'm not sure.  I tested this with master, but I don't know about compatibility with released versions.  Not sure how the site could redirect to the github pull either, but I'm guessing you didn't download them in raw mode?  Really, the best way to get this fixed would be to have it included in the next version and update to that.

bhmhrex1:
Thanks for the quick response. I am not sure what you mean by raw mode, but this is what I did:

1. Went to files changed(4)
2. Followed the path for each file, then backed out to the folder the file was in.
3. Right clicked on the file and did "Save Link as"
4. Saved the file as (example) squareup.php, transaction.php etc
5. Went to the path on my file server for each respective changed file on here and replaced with the file you fixed.

Refreshed my back end and it directed me to this page. I tried it on multiple devices, too. Did the same thing each time.  I was kind of confused by that too. 

I am running OpenCart 3.0.3.2 with the included Square app in that version.

When you say "save configuration again", do you mean refreshing inside of opencart under extensions - modifications?

I greatly appreciate your help and the time you took out to fix this issue as it seems to have worked for a few!

straightlight:
Square payment extension module is no longer supported on the master branch nor on the pre-release version. Therefore, this commit will now be closed.

PR #7329, opened 2019-04-14
Add built-in plugin for NMI Collect.js inline

falconkirtaran:
Add built-in support for the USA merchant account platform NMI.  Support is done using their inline collect.js option, which uses iframes to gather card data, which should reduce the PCI compliance burden to something like SAQ A-EP rather than the SAQ-D required for authorize.net emulation with them.

### DIFF
--- a/upload/admin/controller/extension/payment/nmi_collectjs.php
+++ b/upload/admin/controller/extension/payment/nmi_collectjs.php
@@ -1,0 +1,162 @@
+<?php
+class ControllerExtensionPaymentNMICollectJS extends Controller {
+	private $error = array();
+
+	public function index() {
+		$this->load->language('extension/payment/nmi_collectjs');
+
+		$this->document->setTitle($this->language->get('heading_title'));
+
+		$this->load->model('setting/setting');
+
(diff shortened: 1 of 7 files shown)


falconkirtaran:
Howdy @danielkerr.  Is this kind of PR of interest to the project?  I'm happy to keep contributing things I use to opencart as long as they're welcome.

We've been using this in production on our store cryoflesh.com for about a month now with no issues.

danielkerr:
I'll look into it. I've been recently contacted by authorise.net

falconkirtaran:
I was looking at making an updated integration for them too possibly, in 
case they wanted to avoid the SIM and AIM flows. This is kind of 
orthogonal to them, but useful for NMI; NMI used to emulate 
authorize.net's AIM protocol but the new iframe stuff differs between 
NMI and authorize.net.

falconkirtaran:
I accidentally left TLS certificate checking off in my PR; that's fixed now.

I also sent one over for authorize.net (which was a lot more work; their API is a mess) in #7465 if you want to have a look, @danielkerr :)

danielkerr:
im removing all 3rd party extensions. u need to post it into the marketplace.

PR #7330, opened 2019-04-14
Fix instances of CWE-113 HTTP response splitting

falconkirtaran:
This issue fixes a vulnerability where in some cases it is possible to split HTTP responses, which proxies and clients will tend to interpret in such a way as to allow the attacker to set arbitrary headers or cause the server to appear to send arbitrary resources based on the request URI.  The fix here is to strip CR and LF out of the redirect value in the two instances where it's based on the query string used to invoke the controller.

### DIFF
--- a/upload/admin/controller/common/login.php
+++ b/upload/admin/controller/common/login.php
@@ -67,7 +67,7 @@ class ControllerCommonLogin extends Controller {
 				$url .= http_build_query($this->request->get);
 			}
 
-			$data['redirect'] = $this->url->link($route, $url);
+			$data['redirect'] = str_replace(array("\r", "\n"), "", $this->url->link($route, $url));
 		} else {
 			$data['redirect'] = '';
 		}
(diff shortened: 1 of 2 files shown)


ADDCreative:
Already has it in the response class.
https://github.com/opencart/opencart/blob/master/upload/system/library/response.php#L36

falconkirtaran:
@ADDCreative this issue happens when constructing links, not when sending 302.  For example, the payment one will give you this snippet if it doesn't filter:

```
    <input type="hidden" name="code" value=""/> <input type="hidden" name="redirect" value="https://opencart.falconk.rocks/index.php?route=common/home&amp;action=buy_now&amp;amp;products_id=22
Content-Length: 0

HTTP/1.0 200 OK
Content-Type: text/html
Content-Length: 18
```

I found this one with dynamic testing, not source review.

ADDCreative:
That wouldn't be in a response header.

falconkirtaran:
No, you're right, there's no path from there to the response header, so it's a false positive.  It still causes saint to whine, which causes my PCI ASV to make me do paperwork, but this may not be the best solution to that.

PR #7331, opened 2019-04-14
Update jquery and jquery-ui

falconkirtaran:
Updates jquery to 3.4.0 and jquery-ui to 1.12.1.  This is partly just to keep things up to date and partly because old versions of these cause PCI scanners to whine about:

CVE 2016-7103
CVE 2015-9251

It's way easier to update now than to prove that we're not affected by these issues and update later, and saves people having to fill forms with their PCI ASV.


cgcmart:
Great job!

danielkerr:
i update just before a release.

PR #7381, opened 2019-04-24
Fix support for session lifetimes other than 0, secure flag, and httponly flag

falconkirtaran:
Fix support for session lifetimes other than 0 for sessions set via admin and catalog (mirroring fix from daaa9b6113d51fcaf53fd2c11ede32abb51021b0).  Add support for secure and httponly flags to session cookie.

This corrects the same issue as the noted commit, where set_cookie got passed the raw value of session.cookie_lifetime, so admin and user session cookies were expiring at whatever that time was after the unix epoch instead of present (works fine for 0, not for other values).

### DIFF
--- a/upload/admin/controller/startup/startup.php
+++ b/upload/admin/controller/startup/startup.php
@@ -29,7 +29,7 @@ class ControllerStartupStartup extends Controller {
 
 		$this->session->start($session_id);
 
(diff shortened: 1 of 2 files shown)


danielkerr:
added it manually

falconkirtaran:
Cool, thanks for incorporating it.

PR #7397, opened 2019-05-01
SEO Improvement: Add Archive status to Products, so it will display Archived items on store when accessed directly

machadoug:
No one should ever disable or delete a product because it hinders SEO.
Removing product pages on your website (or changing the URL of them) that carry out authority and pass significant weight to other pages without 301 redirecting them or ensuring any authority/weight is maintained on a publicly accessible page on your website can, of course, result in a negative effect in how your website performs (ranks) in search.

Therefore URLs of archived products should always be accessible even though it should not be visible in Categories, Brand pages, search results,..., however OpenCart does not have an Archived product status.

In order to solve this issue, I've created a new product status called "Archived". We use that in Joomla and it works well. Archived items can be accessed via direct URL or even through Google or other Search Engines. This way we don't lose any traffic and it won't cause any negative impact on SEO.

We could implement as an extension, however because of the SEO implications we believe this should be part of the OpenCart's core, that's why we are pulling this request.

It's my first PR for this project. I hope to be able to help more in the future.


straightlight:
Take note that the archived SEOs must not reflect on archiving customer information due to the GDPR law since customers have the right to request personal data information to be removed from the stores.

machadoug:
The changes proposed are only to Archive Products.

straightlight:
Great! :)

danielkerr:
i know what u mean. products do not get blocked their status just goes to not in stock.

PR #7398, opened 2019-05-01
Remove code repetition

diegospm:
Hi guys,

I created two classes that enable drastically reduce lines of code, as well as avoid repetition.

For exemplify the realy power of the classes, I modified the controller of the product in administration. I tested it on localhost and it worked perfectly. In only this class, 340 lines of code were removed (saved).

The Breadcrumbs class has a standard function that set the information to the listing. For the forms, just set the text and route (if necessary).

The Provider class can receive parameters for be used after the get request and also generates the url link using the get request and the passed arguments. With this it is no longer necessary to restart the url string to pass to link function.


straightlight:
For this commit: https://github.com/opencart/opencart/pull/7398/commits/716e3610475f5524aa68e542f552c7dbc13b1fea

This line: $url .= "&{$key}={$param}"; may become problematic for some web hosters regarding assembled strings with & versus &amp ; keys. I would suggest to use either a str_replace or to create an array before the foreach statement and then use that array to pull the URL itself - then to use an array_map to replace the & / &amp ; characters since OC is already depending on that from the core.

diegospm:
> For this commit: 716e361
> This line: $url .= "&{$key}={$param}"; may become problematic for some web hosters regarding assembled strings with & versus &amp ; keys. I would suggest to use either a str_replace or to create an array before the foreach statement and then use that array to pull the URL itself - then to use an array_map to replace the & / &amp ; characters since OC is already depending on that from the core.

Hi @straightlight,

Thanks for the suggestion :)

I don't know if I understood correctly, but I added a replace for the special character.

diegospm:
If needed, we can add urlencode/html_entity_decode/ENT_QUOTES for non-numeric strings with an if condition in the loop

for this:
$url .= '&filter_model=' . urlencode(html_entity_decode($this->request->get['filter_model'], ENT_QUOTES, 'UTF-8'));

straightlight:
I would not suggest to use the html_entity_decode since your suggestion could also be used via remote APis. See if the following will work by replacing since it checks for non-numeric and being empty - it simply continues on anyhow:

`$url = '';
		
		foreach ($params as $key => $param) {
			if (!is_numeric($param) && empty($param)) {
				continue;
			}
			$url .= "&{$key}={$param}";
		}`

with:

`$url = http_build_query($params, '', '&');

diegospm:
Thanks again! @straightlight

Unfortunately (or not) http_build_query only discard null values, but your suggestions gave me a light! ;)

Done! If possible, could you test? I can only test in localhost ...

straightlight:
The replaced codes looks more positive than before by sanitizing the URL. I would leave it as is and await for future inputs regarding the new modifications.

diegospm:
Ok!

When you and the OC team would consider the appropriate time, I propose to the make changes in the pertinent classes :)

diegospm:
sorry guys, I had a problem with my git and I ended up sending unnecessary commits

danielkerr:
im going to spend a week going through issue and pull requests.

diegospm:
> im going to spend a week going through issue and pull requests.

Nice to know that. This pull it's a little messy, because my intention was to present the solution for analysis. I'll close this pull and open a new one with the modifications in all class

PR #7399, opened 2019-05-02
Some comments

engwalid1971:


straightlight:
Not the most problematic report but I will keep this topic on-hold.

danielkerr:
added manually.

PR #7402, opened 2019-05-02
Update developer.php

engwalid1971:


straightlight:
Confirmed. The directories from sass has been renamed into scss accordingly. However, the pre-release version still uses the old folder name of: 'sass' in it. It may have to be changed in the future in the core-pre or features-pre version.

PR #7405, opened 2019-05-02
Update sagepay_direct.twig

engwalid1971:


ADDCreative:
I don't believe this is needed, as due to the ```{% if existing_cards %}``` ```{% else %}``` the div with ```id="card-new"``` is only defined once.

engwalid1971:
> I don't believe this is needed, as due to the `{% if existing_cards %}` `{% else %}` the div with `id="card-new"` is only defined once.

Yes, you are right ...

PR #7408, opened 2019-05-05
[Master Branch - Feature] - catalog/controller/product/product.php

straightlight:
In that file, the category ID would be a great asset to use in the catalog/view/theme/default/template/product/product.twig file since the addToCart and similar JS functions are passing parameters to the catalog/view/javascript/common.js file. While having the product ID and the quantity, the category ID could also be used in custom themes when users developers their JS codes.

In catalog/controller/product/product.php file,

find:

`$data['product_id'] = (int)$this->request->get['product_id'];`

add below:

`$data['category_id'] = (!empty($category_id) && (int)$category_id > 0 ? (int)$category_id : 0);`


straightlight:
Or, another way, before that $data['product_id'] line, to add:

if (!empty($category_id)) {
    $data['category_id'] = (int)$category_id;
} else {
    $data['category_id'] = 0;
}

straightlight:
> line 329 should change to:
> `/* Support */`

Which file are you referring about from the master branch?

cgcmart:
@straightlight 
I reviewed Daniel's request merge opencart/upload/install/view/stylesheet/stylesheet.css

eka7a:
@straightlight This request conflicts with the maintenance branch.

PR #7409, opened 2019-05-06
Fixed if sending unicode mail, message not sending correctly

engwalid1971:


straightlight:
Should be merged with this commit: https://github.com/opencart/opencart/pull/7297

straightlight:
This has been discussed earlier. I guess we could try to fit this in the features-pre version later on and await users reviews.

PR #7414, opened 2019-05-08
[Master Branch] - catalog/model/checkout/recurring.php

straightlight:
Below the editReference method, the getProfile should be added as well:

public function getProfile($product_id, $recurring_id) {
		$query = $this->db->query("SELECT * FROM " . DB_PREFIX . "recurring r JOIN " . DB_PREFIX . "order_recurring or ON (or.recurring_id = r.recurring_id AND or.product_id = '" . (int)$product_id . "') WHERE or.recurring_id = '" . (int)$recurring_id . "' AND `or`.`status` = '1' AND pr.customer_group_id = '" . (int)$this->config->get('config_customer_group_id') . "'");

		return $query->row;
	}


straightlight:
Recurring payments have been fixed on the pre-release version. Since then, no other issues have been reported. Therefore, I will close my own commit.

PR #7416, opened 2019-05-08
Update googleshopping

stasdrvn:
There is a parametr called "sale_price_effective_date"  in google feed, which is two dates range (start and end) of the special price is active. If there won't be end or start date, so the feed won't be created.


straightlight:
This extension/extension module is no longer supported in the master branch nor in the pre-release version. Therefore, this commit will now be closed.

PR #7417, opened 2019-05-08
[Master Branch] - catalog/model/catalog/product.php

straightlight:
In the getProfile method, the: `status = '1'` should be replaced with: `r.status = '1'`

Granted, there are no status field in the product_recurring table. However, it should ease the development for extensions for whomever requires the status to be pulled from their extensions due to the suggestion of this commit: #7414


straightlight:
This issue has now been fixed in the pre-release version.

PR #7436, opened 2019-05-15
Customer change default address

eka7a:
Update session address values when customer changes default address

### DIFF
--- a/upload/catalog/model/account/address.php
+++ b/upload/catalog/model/account/address.php
@@ -7,6 +7,10 @@ class ModelAccountAddress extends Model {
 
 		if (!empty($data['default'])) {
 			$this->db->query("UPDATE " . DB_PREFIX . "customer SET address_id = '" . (int)$address_id . "' WHERE customer_id = '" . (int)$customer_id . "'");
+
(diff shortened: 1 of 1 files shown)


straightlight:
While the main idea is a great one, the only downside to this solution is when the controllers may call these models without redirections in order for the $this->session->data to take effect since it does not require site redirection to add or edit an address.

straightlight:
In addition. your code implies on creating a session data but there are no locations to unset this variable as it may be problematic with GDPR compliance. In order to ensure the variable is unset and the right location to do it would be under the catalog/controller/checkout/success.php file.

Find:

`unset($this->session->data['payment_methods']);`

add below:

if (!empty($this->session->data[$this->config->get('config_tax_customer') . '_address']) {
    unset ($this->session->data[$this->config->get('config_tax_customer') . '_address']);
}

This should resolved the issue.

eka7a:
Even if this code is not this data is already being created according to default address when customer login but the data in session does not change when default address is changed. https://github.com/opencart/opencart/blob/6188738c1a22dc8b3e873122a06eb9862331c204/upload/catalog/controller/account/login.php#L21

straightlight:
This might be considered in the features-pre version.

PR #7440, opened 2019-05-17
Fix https://forum.opencart.com/viewtopic.php?t=191696. Remove

budimanjojo:
Fix https://forum.opencart.com/viewtopic.php?t=191696. Remove FROM_EXCESS_BASE64 and SUBJ_EXCESS_BASE64 warning on Spamassasin and other spam filters.

Also, add Message-ID to the header to remove MISSING_MID warning on
Spamassasin and other spam filters. Message-ID format is using the format from https://gist.github.com/Robbert/4967898

### DIFF
--- a/upload/system/library/mail/mail.php
+++ b/upload/system/library/mail/mail.php
@@ -11,9 +11,9 @@ class Mail {
 		$boundary = '----=_NextPart_' . md5(time());
 
 		$header  = 'MIME-Version: 1.0' . PHP_EOL;
-    $header .= 'Date: ' . date('D, d M Y H:i:s O') . PHP_EOL;
-    
-    if (mb_check_encoding($this->sender, 'ASCII')) {
+		$header .= 'Date: ' . date('D, d M Y H:i:s O') . PHP_EOL;
+
+		if (mb_check_encoding($this->sender, 'ASCII')) {
 			$header .= 'From: ' . $this->sender . ' <' . $this->from . '>' . PHP_EOL;
 		} else {
 			$header .= 'From: =?UTF-8?B?' . base64_encode($this->sender) . '?= <' . $this->from . '>' . PHP_EOL;
@@ -32,8 +32,8 @@ class Mail {
 				$header .= 'Reply-To: =?UTF-8?B?' . base64_encode($this->reply_to) . '?= <' . $this->reply_to . '>' . PHP_EOL;
 			}
 		}
-		
-    $header .= 'Message-ID: <' . base_convert(microtime(), 10, 36) . '.' . base_convert(bin2hex(openssl_random_pseudo_bytes(8)), 16, 36) . substr($this->from, strrpos($this->from, '@')) . '>' . PHP_EOL;
+
+		$header .= 'Message-ID: <' . base_convert(microtime(), 10, 36) . '.' . base_convert(bin2hex(openssl_random_pseudo_bytes(8)), 16, 36) . substr($this->from, strrpos($this->from, '@')) . '>' . PHP_EOL;
 		$header .= 'Return-Path: ' . $this->from . PHP_EOL;
 		$header .= 'X-Mailer: PHP/' . phpversion() . PHP_EOL;
 		$header .= 'Content-Type: multipart/mixed; boundary="' . $boundary . '"' . PHP_EOL . PHP_EOL;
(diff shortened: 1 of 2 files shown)


straightlight:
A solution has already been posted here: https://github.com/opencart/opencart/pull/7409

straightlight:
However, I don't know why the subject title of your topic mentions to 'remove'.

budimanjojo:
With the default smtp headers, my emails ended up in Spam folder everytime, so I use Rspamd (similar to Spamassasin) to analyze the message problem. And this is what I get.
MISSING_MID (2.5)
FROM_EXCESS_BASE64 (1.5)
SUBJ_EXCESS_BASE64 (1.5)
CTYPE_MIXED_BOGUS (1)
That gives me a 6.5 score (score higher that 4 will ended up marked as Spam by most mail server, which is not good).

MISSING_MID is because the mail doesn't contain Message ID in the header (https://wiki.apache.org/spamassassin/Rules/MISSING_MID)

FROM_EXCESS_BASE64 happens when From header is encoded with base64 while it is not needed (when the store name is in ASCII encoding), so when my store name is MyStore, we shouldn't use base64_encode, but when my store name is MyS®ore, we need to use base64_encode (https://wiki.apache.org/spamassassin/Rules/FROM_EXCESS_BASE64)

SUBJ_EXCESS_BASE64 is the same as FROM_EXCESS_BASE64 but this is the Subject header instead of From. 

This is the related problem I have: https://forum.opencart.com/viewtopic.php?t=191696

And this is the Message-ID format that I use: https://gist.github.com/Robbert/4967898
and the related problem:
https://forum.opencart.com/viewtopic.php?t=136317

straightlight:
`I use Rspamd (similar to Spamassasin) to analyze the message problem. And this is what I get.
MISSING_MID (2.5)
FROM_EXCESS_BASE64 (1.5)
SUBJ_EXCESS_BASE64 (1.5)
CTYPE_MIXED_BOGUS (1)`

Which brings back the analysis from my previous reply, these results are based on a tool that you are using specifically on your server which does not necessarily reflect via other servers where they don't use spamassasins or Rspamd.

For instance, if using the commit solution: 7409 suggested on the above, does the same result returns from Rspamd?

budimanjojo:
No, this is not about using Rspamd or not. This is about mail sent from Opencart is more likely to end in spam folder because of the headers. Not just servers with spamassasin or rspamd installed, but also other mail server like Gmail and such. Because all Spam filters work by scanning the message headers to detect whether the mail is Spam or not. If you send an email using Opencart to email tester like https://www.mail-tester.com/, it will say your mail is a Spam because it's missing Message-ID, using excessive base64 etc.

The commit 7409 is fixing mail message not displaying correctly, not the mail header. The mail header is displaying correctly with or without my patch, but spam filters are marking the message as spam. And this patch is about the mail not being in Spam folder, not mail not displaying correctly. This patch will parse this when the mail is sent from MyStore when a customer order something:
```
From: MyStore <mystore@domain.com>
Subject: MyStore - Order 2
Message-ID: <kakajhf7e837s.38d7jekke@domain.com>
```
And will parse this when customer order something to MyS®ore:
```
From: =?UTF-8?B?VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw==?= <mystore@domain.com>
Subject: =?UTF-8?B?VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw==?=
Message-ID: <kakajhf7e837s.38d7jekke@domain.com>
```

straightlight:
Then, the patch that you described may work for server-specifics, not as a general solution. This issue has already been reported on the issues tab before but still being considered to be false / positive due to mail server settings specifications.

However, if the mail header should be modified to fit otherwise due to major reports, then greater mount of OC users would of reported it already and the mail class would of been modified since a long time. Since we know already, it's not the case other than OC users, in general, that are improperly configuring their mail form in the OC admin, there's no issue with the mail class other than what's been reported from commit: 7409 at this time.

Also, which PHP version are you using?

budimanjojo:
Okay, I guest I'll just apply this on my own server then.
I'm using PHP 7.3

straightlight:
Closure notes: Server-specifics instructions. Closing commit.

kenkit:
Your pull request is what works for me opencart 3.0.3  with a few touches, pliz fix to also remove the warning. And add the following 
```php
$cleanNumber = preg_replace( '/[^0-9]/', '', microtime(false));
base_convert($cleanNumber, 10, 36)
```

kenkit:
@budimanjojo

PR #7451, opened 2019-05-22
Conditionally include Moment.js on product page

a-v-l:
Only add Moment.js scripts (js & css) to product page if option type 'date' or 'datetime' or 'time' is set. This will save 401,5 KB of potentially unnecessary code to be loaded and parsed.

### DIFF
--- a/upload/catalog/controller/product/product.php
+++ b/upload/catalog/controller/product/product.php
@@ -222,11 +222,6 @@ class ControllerProductProduct extends Controller {
 			$this->document->addScript('catalog/view/javascript/jquery/magnific/jquery.magnific-popup.min.js');
 			$this->document->addStyle('catalog/view/javascript/jquery/magnific/magnific-popup.css');
 
-			$this->document->addScript('catalog/view/javascript/jquery/datetimepicker/moment/moment.min.js');
-			$this->document->addScript('catalog/view/javascript/jquery/datetimepicker/moment/moment-with-locales.min.js');
-			$this->document->addScript('catalog/view/javascript/jquery/datetimepicker/bootstrap-datetimepicker.min.js');
-			$this->document->addStyle('catalog/view/javascript/jquery/datetimepicker/bootstrap-datetimepicker.min.css');
-
 			$data['heading_title'] = $product_info['name'];
 
 			$data['text_minimum'] = sprintf($this->language->get('text_minimum'), $product_info['minimum']);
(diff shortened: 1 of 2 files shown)


a-v-l:
There are several other catalog pages with Moment.js being included by default. Mostly where custom fields are potentially involved.

Since performance on those pages is not as important as on the product pages I didn't include them with this commit. But of course it would be desirable to also only load Moment.js resources if needed.

ilcomp:
We need to transfer all styles and scripts to the theme event, except common.js.

a-v-l:
@ilcomp: Sorry, which _theme event_ are you referring to? If you need Moment.js on the product page all the time in a certain theme then you can easily add it via `install.xml`.

I just think that in the vast majority of cases a date/time picker on the product page is not required. And loading the scripts all the time is a huge extra payload on all opencart shop product pages out there.

ilcomp:
```
class ControllerEventThemeDefault extends Controller {
    public function product($route, &$data) {
        $this->document->addScript('catalog/view/javascript/jquery/magnific/jquery.magnific-popup.min.js');
        $this->document->addStyle('catalog/view/javascript/jquery/magnific/magnific-popup.css');
        
        $this->document->addScript('catalog/view/javascript/jquery/datetimepicker/moment/moment.min.js');
        $this->document->addScript('catalog/view/javascript/jquery/datetimepicker/moment/moment-with-locales.min.js');
        $this->document->addScript('catalog/view/javascript/jquery/datetimepicker/bootstrap-datetimepicker.min.js');
        $this->document->addStyle('catalog/view/javascript/jquery/datetimepicker/bootstrap-datetimepicker.min.css');
    }
}
```
If you define a product template before the header and footer, and then glue them together, then you can add scripts depending on the data transferred to the template.
I have improvements, but the problem with time.

a-v-l:
@ilcomp: Sorry, but I really don't understand your problem. Are you sure you are talking about a clean opencart install? I can't [find the code snippet](https://github.com/opencart/opencart/search?q=ControllerEventThemeDefault&unscoped_q=ControllerEventThemeDefault) you provided in this repository. 

If you need help with a custom theme please post in the [opencart community forum](https://forum.opencart.com/).

My pull request is about *removing* Moment.js from the default product page. Because currently it _is_ included even if it is not used.

danijelGombac:
@a-v-l  If I understood correctly. Different themes may use different widgets. Move magnific-popup and datetimepicker from product controller to theme event controller.

danielkerr:
not required

PR #7456, opened 2019-05-25
Update map.php - Fix for empty map

MadMcJack:
If you have multiple complete statuses then you would get a statement like "WHERE o.order_status_id IN('5,3')" which gives 0 results because of the quotes. If you leave the int() declaration before the implode() function you get "WHERE o.order_status_id IN(5)".

Removing both the quotes and the (int) declaration before the implode() function gives a working map again.

### DIFF
--- a/upload/admin/model/extension/dashboard/map.php
+++ b/upload/admin/model/extension/dashboard/map.php
@@ -10,7 +10,7 @@ class ModelExtensionDashboardMap extends Model {
 		}
 		
 		if ($implode) {
-			$query = $this->db->query("SELECT COUNT(*) AS total, SUM(o.total) AS amount, c.iso_code_2 FROM `" . DB_PREFIX . "order` o LEFT JOIN `" . DB_PREFIX . "country` c ON (o.payment_country_id = c.country_id) WHERE o.order_status_id IN('" . (int)implode(',', $implode) . "') GROUP BY o.payment_country_id");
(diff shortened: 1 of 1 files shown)


straightlight:
The single quotes after the opening parentheses and the one before the closing parentheses should be no different, however. As for the int in front of the implode, good catch!

MadMcJack:
Thanks for the reply. If I check the SQL result the parenthesis could be removed or should be set on the individual status ID's. In my pull request I did not bother with the parenthesis around the individual items because the implode items are already set as integers.

`$implode[] = (int)$order_status_id;`

So it looks as the most logical commit to me... but I am open to corrections.

```
SELECT COUNT(*) AS total, SUM(o.total) AS amount, c.iso_code_2 FROM `oc_order` o LEFT JOIN `oc_country` c ON (o.payment_country_id = c.country_id) WHERE o.order_status_id IN('5,3') GROUP BY o.payment_country_id;

Results: 0

SELECT COUNT(*) AS total, SUM(o.total) AS amount, c.iso_code_2 FROM `oc_order` o LEFT JOIN `oc_country` c ON (o.payment_country_id = c.country_id) WHERE o.order_status_id IN('5','3') GROUP BY o.payment_country_id;

Results: 24

SELECT COUNT(*) AS total, SUM(o.total) AS amount, c.iso_code_2 FROM `oc_order` o LEFT JOIN `oc_country` c ON (o.payment_country_id = c.country_id) WHERE o.order_status_id IN(5,3) GROUP BY o.payment_country_id;

Results: 24
```

PR #7465, opened 2019-05-30
Create authorize.net accept hosted payment extension

falconkirtaran:
It's an extension that enables the authorize.net accept suite hosted iframe method, which NMI asserts should be SAQ A (might be SAQ A-EP in the future, who knows?).  This provides a migration path from AIM and SIM, which are both deprecated.

### DIFF
--- a/upload/catalog/controller/extension/payment/authorizenet_accept_hosted.php
+++ b/upload/catalog/controller/extension/payment/authorizenet_accept_hosted.php
@@ -122,8 +122,9 @@ class ControllerExtensionPaymentAuthorizeNetAcceptHosted extends Controller {
 		} elseif (!$response_raw) {
 			$this->log->write('AUTHORIZE.NET ACCEPT HOSTED CURL ERROR: No response');
 		} else {
-			if (substr($response_raw, 0, 3) === "\xef\xbb\xbf")
+			if (substr($response_raw, 0, 3) === "\xef\xbb\xbf") {
 				$response_raw = substr($response_raw, 3); //result of RFC-violating inclusion of BOM by authorize.net; remove if they fix their bug
+			}
 
 			$response = json_decode($response_raw);
 			if ($response === null) $data['error'] = $this->language->get('error_no_form');
@@ -133,8 +134,9 @@ class ControllerExtensionPaymentAuthorizeNetAcceptHosted extends Controller {
 					$this->log->write('Authorize.net accept hosted returned error while getting payment form: ' .
 						$response->messages->resultCode . ': ' . $response->messages->message[0]->code . ' ' . $response->messages->message[0]->text);
 				}
-				else
+				else {
 					$data['token'] = $response->token;
+				}
 			}
 		}
 		$data['token_responder'] = 'https://accept.authorize.net/payment/payment';
@@ -212,8 +214,9 @@ class ControllerExtensionPaymentAuthorizeNetAcceptHosted extends Controller {
 			$this->response->redirect($this->url->link('checkout/failure', 'language=' . $this->config->get('config_language')));
 			return;
 		} else {
-			if (substr($tx_resp_raw, 0, 3) === "\xef\xbb\xbf")
+			if (substr($tx_resp_raw, 0, 3) === "\xef\xbb\xbf") {
 				$tx_resp_raw = substr($tx_resp_raw, 3); //result of RFC-violating inclusion of BOM by authorize.net; remove if they fix their bug
+			}
 
 			$tx_resp = json_decode($tx_resp_raw);
 			if ($tx_resp === null || !property_exists($tx_resp, 'transaction') || $tx_resp->transaction === null) {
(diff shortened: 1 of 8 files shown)


straightlight:
The UTF-8 Encoding functions should be removed since remote providers are already validating the requests on their end with 99% of the RFC standards already since php.ini can already use the UTF-8 parameter from PCI servers. Granted, that 1% leftover may go to the Unicode where you indicated a violation RFC in the comment.

As for this line:

`$message .= 'Masked Account Number: ' . $txn->payment->creditCard->cardNumber . "\n";`

it should be replaced with:

`$message .= 'Masked Card Number: ' . $txn->payment->creditCard->cardNumber . "\n";`

Optional: Since OC does not really use property_exists functions and objects but rather the use of json_decode with true case to convert the results into an array, it would be the regular way of presenting an extension module (but, again, optional).

As for the filenames, as compared to the ones being used from the core, they are a bit too long. Acronyms could rather be used like the ones already built-in such as: SIM and AIM at the end of the filename before the php extension.

Other than that, it may be a practical module to have.

straightlight:
In addition, for instance, the catalog/controller/extension/payment/too_long_file_to_write.php contains IF and ELSE statement with and without brackets which may be a bit confusing for OC forum users when reporting issues with their incident requests and might not be as obvious to troubleshoot on a quickly basis. If this idea would ever be implemented, I would suggest using structured and clean codes.

falconkirtaran:
I'm a little confused about your comment about UTF-8.  The other end of that connection, authorize.net's API servers ("PCI server" isn't a thing in this integration), currently always includes the UTF-8 BOM in violation of the spec, and PHP's JSON implementation rejects it.  I included a conditional there in case they fix their implementation at some point in the future.  Removing the BOM is absolutely required for this integration to work.  I don't think there's a php.ini setting to accept the spurious BOM.

The name of the method is "Accept Hosted"; there's no canonical acronym.  I suppose the names could be changed if there is a technical reason to do so.  Is there?

I've updated my code to call the masked PAN a "masked card number" since that is indeed more clear, and changed "account type" to "card account type".  I also added curly braces to conditionals that are broken across two lines for greater clarity.

straightlight:
`("PCI server" isn't a thing in this integration)`

This part has nothing to do with confusion. It is simply to know that PCI is still a thing despite if it's hosted by the merchant or by service providers. It's just that, this time, hosted checkout is hosted by the service provider rather than the merchant taking full responsibility of it since customers are being redirected to authorize.net to complete their transactions online.

`I don't think there's a php.ini setting to accept the spurious BOM.`

I beg to differ on that one. See this post on the OC forum: https://forum.opencart.com/viewtopic.php?f=202&t=209654&p=745233#p745286 . As you can see, the UTF-8 setting can be used by following this suggestion as to use the html_entity_decode function is no longer a necessity for servers supporting UTF-8 encoding. Besides, OC is focused on UTF-8 settings, including database by default.

`I suppose the names could be changed if there is a technical reason to do so. Is there?`

Not only for technical reasons.

1 - Using snake size conventions can be quite frustrating for users
2 - Windows IIS servers may not be as tolerant as Linux servers to accept long file names as this size which was good idea for authorizenet_sim and authorizenet_aim to have as convention names

`I included a conditional there in case they fix their implementation at some point in the future.`

Yes, that 1% of Unicode check seem to be an exception as said on my above reply. A bit odd that the API programmers on their end did not validated those Unicode yet …

`I've updated my code to call the masked PAN a "masked card number" since that is indeed more clear, and changed "account type" to "card account type". I also added curly braces to conditionals that are broken across two lines for greater clarity.`

Great!

straightlight:
The authorize.net extension payment module is no longer supported in the master branch nor in the pre-release version. Therefore, this commit will now be closed.

upload/catalog/controller/extension/payment/authorizenet_accept_hosted.php
```diff
@@ -0,0 +1,259 @@
...
+		}
+		$data['token_responder'] = 'https://accept.authorize.net/payment/payment';
+		if ($this->config->get('payment_authorizenet_accept_hosted_testmode')) $data['token_responder'] = 'https://test.authorize.net/payment/payment';
+
+		return $this->load->view('extension/payment/authorizenet_accept_hosted', $data);
+	}
+
+	public function callback() {
+		$data = array();
+		$this->response->setOutput($this->load->view('extension/payment/authorizenet_accept_hosted_callback', $data));
```
bloodyKnuckles, line 150: I needed to insert at line 150: 

    header("Content-Security-Policy: frame-ancestors 'self' https://test.authorize.net");

Otherwise the browser denied loading of the callback URL by the offsite page, loaded in the `iframe` in file upload/catalog/view/theme/default/template/extension/payment/authorizenet_accept_hosted.twig, line 5. 

The "Pay" and "Cancel" buttons did not appear on the page until after adding the `Content-Security-Policy: frame-ancestors` header.

upload/catalog/controller/extension/payment/authorizenet_accept_hosted.php
```diff
@@ -0,0 +1,259 @@
...
+			}
+		}
+		$data['token_responder'] = 'https://accept.authorize.net/payment/payment';
+		if ($this->config->get('payment_authorizenet_accept_hosted_testmode')) $data['token_responder'] = 'https://test.authorize.net/payment/payment';
+
+		return $this->load->view('extension/payment/authorizenet_accept_hosted', $data);
+	}
+
+	public function callback() {
+		$data = array();
```
bloodyKnuckles, line 149: I needed to insert at line 150—below this line:

header("Content-Security-Policy: frame-ancestors 'self' https://test.authorize.net https://secure.authorize.net");

Otherwise the browser (Firefox) denied loading of the callback URL by the offsite page, loaded in the iframe in file upload/catalog/view/theme/default/template/extension/payment/authorizenet_accept_hosted.twig, line 5.

The "Pay" and "Cancel" buttons did not appear on the page until after adding the Content-Security-Policy: frame-ancestors header.

PR #7485, opened 2019-06-17
Fixed Admin end recurring column sort issue

WebkulOpencart:
Fixed Admin end recurring column sort issue


WebkulOpencart:
This issue is still presented. Please look into this

WebkulOpencart:
@jeffreyeng-opencart Please look into this as well. This issue is still exist in master.
https://github.com/opencart/opencart/blob/master/upload/admin/controller/catalog/recurring.php#L250

PR #7488, opened 2019-06-18
Fixed Authorize.net SIM Module

TheRealDJ:
Fixed SHA512 hash calc (changed payment_authorizenet_sim_key to payment_authorizenet_sim_hash in hash calc) to submit transaction.

Reverted callback hash check to v3.0.3.2 SHA512 hash check.

Changed callback redirects to javascript href.location=... The issue is that Authorize.net doesn't redirect back to the store, it posts to the store, gets the result, and displays the result but the browser URL shows the authorize.net URL. Because of this, cookies could not be read, and the cart would not empty after a successful transaction. Replacing redirects with javascript href.location=... fixes this issue.


straightlight:
Not sure why the .htaccess file needs to be there. However, to all users who wants to try these replacement codes, ensure to uninstall the extension before making modifications. Then, re-install the extension to stay on the safe-side.

TheRealDJ:
I accidentally deleted the .htaccess file (mv .htaccess /var/www/html/ instead of cp ...). I just copied it back to it wasn't missing. Sorry ... this if my first pull request ever.

straightlight:
What I mean is why we need to use the .htaccess file at all along with these replacement codes?

TheRealDJ:
It is not on the list of changed files. I forked the repo, made my changes and commit'd. Then I realized that my checked in changes included deleting the .htaccess file because I had accidentally removed it from my local branch. When I saw this, I re-added the .htaccess to my branch to avoid a merge deleting the .htaccess file in the master.

If I need to re-do something, let me know (but I may need help).

straightlight:
This extension payment module is no longer supported in the master branch and on the pre-release version. Therefore, this commit will now be closed.

PR #7490, opened 2019-06-22
Remove code repetition - refactoring

diegospm:
Hi guys and @danielkerr,

As I mentioned in the pull #7398, I am sending this pull with an initial refactoring for the catalog. I took advantage to correct some indentations in the library classes and also add some type validations in the parameters.

I also refactored some codes that could be passed to an array and iterate with loop.

The refactoring has gone a little deeper than I expected and I know it will hamper review, but in the future I believe it will be easier to maintain the project.

I will be sending in the next few days the refactoring for the other classes.

I tested in localhost and it worked fine, but I did not perform performance test (for having increased some loops) ...

### DIFF
--- a/upload/admin/controller/catalog/attribute.php
+++ b/upload/admin/controller/catalog/attribute.php
@@ -24,21 +24,7 @@ class ControllerCatalogAttribute extends Controller {
 
 			$this->session->data['success'] = $this->language->get('text_success');
 
-			$url = '';
-
-			if (isset($this->request->get['sort'])) {
-				$url .= '&sort=' . $this->request->get['sort'];
-			}
-
-			if (isset($this->request->get['order'])) {
-				$url .= '&order=' . $this->request->get['order'];
-			}
-
-			if (isset($this->request->get['page'])) {
-				$url .= '&page=' . $this->request->get['page'];
-			}
-
-			$this->response->redirect($this->url->link('catalog/attribute', 'user_token=' . $this->session->data['user_token'] . $url));
+			$this->response->redirect($this->provider->link('catalog/attribute'));
 		}
 
 		$this->getForm();
@@ -56,21 +42,9 @@ class ControllerCatalogAttribute extends Controller {
 
 			$this->session->data['success'] = $this->language->get('text_success');
 
-			$url = '';
-
-			if (isset($this->request->get['sort'])) {
-				$url .= '&sort=' . $this->request->get['sort'];
-			}
-
-			if (isset($this->request->get['order'])) {
-				$url .= '&order=' . $this->request->get['order'];
-			}
-
-			if (isset($this->request->get['page'])) {
-				$url .= '&page=' . $this->request->get['page'];
-			}
-
-			$this->response->redirect($this->url->link('catalog/attribute', 'user_token=' . $this->session->data['user_token'] . $url));
+			$this->provider->detach('attribute_id');
+			
+			$this->response->redirect($this->provider->link('catalog/attribute'));
 		}
 
 		$this->getForm();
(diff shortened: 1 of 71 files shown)


straightlight:
There's no need to create a new commit about the same pull request and from the same OP. In addition, your code suggests to remove:

`$data['breadcrumbs'] = array();`

This could become problematic with IIS servers in Windows as each variable types must be pre-defined first as compared to other OSes.

diegospm:
I decided to close the previous pull and open this one because I made many modifications, including the classes I created. That pull was to present the idea of the solution.

About the variable, it is predefined in the Breadcrumbs class.

As I said, we need to test, in my localhost it worked fine, but his remark was well placed ...

straightlight:
`About the variable, it is predefined in the Breadcrumbs class.`

We still need to be careful about pre-defined variables, still with the `$this->load->controller`. By taking an example with the controller pagination class, the `$this->request->get['page']` required to be sanitized before being referenced to the target controller. Otherwise, an error message was being returned to the browser.

Initiating a breadcrumb array as such could also reproduce the same issue even on most Linux / UNIX servers.

diegospm:
I created the breadcrumbs class as an object and defined the private variable of the array type. I believe we will have no problem with the issue you raised. :)

It is no longer necessary to define the breadcrumbs variable in the scope of the controller classes, because just pass the language and route parameters to the set method, which will add in the array of the object.

I also added the setDefaults method, which will add in the array the route to the lists.

When we request the render method, the class will return the impression in the view with the breadcrumbs array data.

About the pagination, I've also modified it and includes in the provider class the variable page and route so they can be accessed from any class.

danielkerr:
not a chance.

diegospm:
> not a chance.

I did not understand

gitforwebwork:
@danielkerr not very diplomatic. While he may have gone a bit overboard, there are some good ideas in the PR. Maybe worth while going though it and giving him some hints what could have a chance.

diegospm:
I know that @danielkerr is a little conservative and resistant to some changes, but I think that just as there was a lot of progress removing the languages from the controller files (codes reduction), we can make good progress with this suggestion.

What does not make sense to me is to use this:

`$data['pagination'] = $this->load->controller('common/pagination', array(
    'total' => $product_total,
    'page'  => $page,
    'limit' => $this->config->get('config_limit_admin'),
    'url'   => $this->url->link('catalog/product', 'user_token=' . $this->session->data['user_token'] . $url . '&page={page}')
));

$data['results'] = sprintf($this->language->get('text_pagination'), ($product_total) ? (($page - 1) * $this->config->get('config_limit_admin')) + 1 : 0, ((($page - 1) * $this->config->get('config_limit_admin')) > ($product_total - $this->config->get('config_limit_admin'))) ? $product_total : ((($page - 1) * $this->config->get('config_limit_admin')) + $this->config->get('config_limit_admin')), $product_total, ceil($product_total / $this->config->get('config_limit_admin')));`

When could use this:

`$data['pagination'] = $this->load->controller('common/pagination', $product_total);

$data['results'] = $this->load->controller('common/pagination/results', $product_total);`

Then as @gitforwebwork cited, maybe it was necessary to look a little deeper and point out what is bad (to be corrected) and what can be harnessed.

straightlight:
@diegospm: Yes, I do agree that simplified calls should be used when using $this->load->controller .

diegospm:
@straightlight: Yes, but not only for the load method, the central idea of ​​this PR is give more simplicity on controller classes as a whole.

straightlight:
Yes, I am aware. I have discussed this lately with my teammates. I hope the same could be done for the custom fields as this feature, especially, can get very extensive to code while all we need, despite of the extensions, would be the common/custom_field . It would be a cleaner way to use the calls rather than using the big pile of codes altogether.

diegospm:
Yes I agree. There are some classes in OC that use the same data that could be called from a common class, as I did with breadcrumbs and results.

Also, as it is in PR, it is not necessary to restart the url string to clear the get request, just replace the value of the key with a new one, as I did in the PR (includes the contribution you gave in the other PR)....

The OC, which is very simple and easy to maintain, would look even better with this solution.

straightlight:
Regarding the PR, I may not provide much input on this one since the way it's been originally coded does work with the URLs and the most the forum users are mostly complaining about are about the SEO URL structures which, obviously, are user-based and not specifically issues related with OC except the marketing strategies being used out-of-the-box.

As for the common folder, I was referring to repetitive codes being used from the core and not from re-write URL sanitizing methods as there already repetitive codes, for instance, by loading the payment, shipping and total extension modules from the controller. These could all be added in the common folder pretty much since $this->load->controller can already catch each of the arrays / objects on anyhow.

diegospm:
Yes, I understood about the use of the common folder.

I mentioned the url string in another context, but on the same topic (remove code repetition).

engwalid1971:
> When could use this:
> 
> $data['pagination'] = $this->load->controller('common/pagination', $product_total);
> $data['results'] = $this->load->controller('common/pagination/results', $product_total);

@diegospm
It must be:
```
$data['pagination'] = $this->load->controller('common/pagination', $product_total, $page, 'catalog/product');
$data['results'] = $this->load->controller('common/pagination/results', $product_total, $page);
```

straightlight:
Correct. The product total needs to be defined based on the page number relying on the depth level where the browser is currently addressed to.

diegospm:
I'm removed a page variable from the argument because I set it in the scope of the target class.

$this->request->get['page'] is being stored in the $page variable of the Provider object. So I access each value with $this->provider->page

I left the optional third argument to set the limit, as it may be necessary to pass a value other than 'config_limit_admin', as in the filemanager class.

I also opted to leave the pagination and results keys separate to be compatible with the event system and ocmod if someone wants to change these variables individually.

Please, see these two classes:

https://github.com/opencart/opencart/blob/2a3506e275f4c1899c2dbaab7e1fa31432863452/upload/system/library/provider.php

https://github.com/opencart/opencart/blob/81901ac7e6eb47f73705cd901438f0528be2db7c/upload/admin/controller/common/pagination.php

PR #7497, opened 2019-06-28
add the modification priority point

WebkulOpencart:
add the modification priority point in the Default Opencart

### DIFF
--- a/upload/admin/controller/marketplace/install.php
+++ b/upload/admin/controller/marketplace/install.php
@@ -316,6 +316,14 @@ class ControllerMarketplaceInstall extends Controller {
 						} else {
 							$link = '';
 						}
+
+						$priority = $dom->getElementsByTagName('priority')->item(0);
+
+						if($priority) {
+							$priority = $priority->nodeValue;
+						} else {
+							$priority = 0;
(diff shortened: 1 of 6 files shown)


straightlight:
This has been suggested before, however. This could become problematic if users installed multiple extensions where priorities can only be defined for each installed extensions since most extensions may also require to work together. By setting a sort order on the list, the XML files first loading the XML tags may still not know the proper priority to make the modifications to its destination at a certain time where another extension may still require the same tactic to accomplish similar goals to the stores.

Which is why, I suggested to Daniel on another commit the use and implementation of the $this->load->controller('author / shaX code')->get('extension_name')(maybe with:->sort($array) additionally?) to ensure the loaded tasks do not confront with other extension developers and their packages in the future.

WebkulOpencart:
@straightlight , I already implemented this point and tested from my end as well. This is working fine. you can also check.

straightlight:
There were massive modifications made on the master branch regarding these activities. There are simply too many variances that could fit into the new design nowadays since the new design is still on the development stage on the master branch. Therefore, this commit will now be closed.

PR #7499, opened 2019-06-29
add post_max_size directive

gnovaro:
post_max_size is related with upload_max_size, always suggest bigger than upload. When you make and upload with post you send the file/s and text fields for that reason post_max_size must be bigger


straightlight:
This setting provides no guarantee for all users since it is defined per server-specifics. It would be best to contact your host to define the maximum definition that could be set before setting any values on that parameter.

straightlight:
Closure note: Server-specifics solution. Not related to master branch. Please address these solutions on the OC forum.

Issue #7501, opened 2019-07-01
Bad SQL in admin/model/sale/order.php


amanaloaie:
**What version of OpenCart are you reporting this for?**
3.0.3.2

**Describe the bug**
On line 391 there is an extra "o.":
$sql .= " AND CONCAT(firstname, ' ', o.lastname) LIKE '%" . $this->db->escape($data['filter_customer']) . "%'";

It should be:
$sql .= " AND CONCAT(firstname, ' ', lastname) LIKE '%" . $this->db->escape($data['filter_customer']) . "%'";

straightlight:
Confirmed. Also found in master branch. Good catch!

PR #7502, opened 2019-07-02
Fixed issue for the sale/order model wrong alias : issue #7501

Webkul-Cart:
Fixed issue for the admin side sale/order model wrong alias used in the filter clause: issue #7501


qasim02:
very effective and for more visit here. [SoftSoldier](http://softsoldier.com/)

Issue #7506, opened 2019-07-04
Why there is need of remove first cart and add same product in cart if guest logged In??


WebkulOpencart:
**What version of OpenCart are you reporting this for?**
3.1.0.0_b

**Describe the bug**
I don't think there is any need to remove the first cart with the customer id = 0, that's why it takes time to log in to the customer if there are more than 15-20 product in the cart.

**To Reproduce**
1. Go to 'system/library/cart/cart.php'

**Expected behavior**

We can easily update the customer_id in the cart table using session_id because session_id remain the same for as a guest and if customer logged in from the checkout.

straightlight:
The downside to this is once the time interval of 1 hour, default, expires. The session ID will be obviously un-retrievable afterwards which means the update statement will fail at this point.

PR #7507, opened 2019-07-04
remove extra code to remove cart and recreate the cart with same value

WebkulOpencart:
Pull request for this issue [https://github.com/opencart/opencart/issues/7506](https://github.com/opencart/opencart/issues/7506)

### DIFF
--- a/upload/system/library/cart/cart.php
+++ b/upload/system/library/cart/cart.php
@@ -18,15 +18,7 @@ class Cart {
 			// We want to change the session ID on all the old items in the customers cart
 			$this->db->query("UPDATE " . DB_PREFIX . "cart SET session_id = '" . $this->db->escape($this->session->getId()) . "' WHERE api_id = '0' AND customer_id = '" . (int)$this->customer->getId() . "'");
 
-			// Once the customer is logged in we want to update the customers cart
-			$cart_query = $this->db->query("SELECT * FROM " . DB_PREFIX . "cart WHERE api_id = '0' AND customer_id = '0' AND session_id = '" . $this->db->escape($this->session->getId()) . "'");
-
-			foreach ($cart_query->rows as $cart) {
-				$this->db->query("DELETE FROM " . DB_PREFIX . "cart WHERE cart_id = '" . (int)$cart['cart_id'] . "'");
-
-				// The advantage of using $this->add is that it will check if the products already exist and increase the quantity if necessary.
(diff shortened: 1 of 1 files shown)


danijelGombac:
And if that the customer already has a product in the cart, this could creates duplicates of products in cart.

WebkulOpencart:
@danijelGombac No, It will update the customer_id on the basis of session_id and customer_id = '0'
session_id always different for different customer.

danijelGombac:
@WebkulOpencart 

You did not understand me.

Example:

I logged in yesterday and add product A in the cart and then logged out.

Today i add product A to the cart like guest and then logged in. Your code update only the customer_id.

Now i have duplicate product in the cart instead two pieces.

MrPlow254:
@waqasjawed Your logic is flawed. The update query would have to be modified to update the quantity of the product id. If you are going to simplify code it needs to retain the original function. You need to test if this is even faster. I doubt it will be that much quicker since this only removes 1 select statement. The what $this-> add does is calls a select statement for the unique cart id and an insert/update statement for the same unique cart id.

```
$this->db->query("UPDATE " . DB_PREFIX . "cart C1
INNER JOIN (SELECT SUM(quantity) AS quantity, product_id FROM session_id = '" . $this->db->escape($this->session->getId()) . "' GROUP BY product_id) C2
ON C1.product_id = C2.product_id
SET customer_id = '" . (int)$this->customer->getId() . "', C1.quantity = C2.quantity, api_id = '" . (isset($this->session->data['api_id']) ? (int)$this->session->data['api_id'] : 0), `option` = '" . $this->db->escape(json_encode($option)) . "', recurring_id = '" . (int)true . "' WHERE session_id = '" . $this->db->escape($this->session->getId()) . "' AND customer_id='" . (int)$this->customer->getId() . "'");

$this->db->query("DELETE FROM " . DB_PREFIX . "cart WHERE api_id = '" . (isset($this->session->data['api_id']) ? (int)$this->session->data['api_id'] : 0) . "' AND customer_id = '" . (int)$this->customer->getId() . "' AND session_id = '" . $this->db->escape($this->session->getId()) . "'");
```

WebkulOpencart:
@MrPlow254 Thanks for the suggestion.
@danijelGombac Yes, you are right, I am looking into this.

straightlight:
Closure notes: Other technical opinions have been shared on the commit for better solutions. This commit will now be closed.

Issue #7512, opened 2019-07-10
Opencart installation error by console "PHP Notice:  Use of undefined constant DIR_STORAGE"


prabhatkumaroc082:
**What version of OpenCart are you reporting this for?**
3.1.0.0_b

**To Reproduce**
1. Go to Opencart 'install/' directory
2. Open console inside this dir
3. try to execute the following command without options like db_username, etc
php cli_install.php install
PHP Notice:  Use of undefined constant DIR_STORAGE - assumed 'DIR_STORAGE' in /path_to_opencart/system/startup.php on line 70

**Expected behavior**
"DIR_STORAGE" constant should be checked as it is defined or not

prabhatkumaroc082:
I have created PR for the same issue #7513

straightlight:
Confirmed bug.

PR #7513, opened 2019-07-10
fixed opencart installation bug by console "PHP Notice:  Use of undef…

prabhatkumaroc082:
…ined constant DIR_STORAGE"


straightlight:
This could only happen if users are using an old version of OC and wants to upgrade to a more recent release. However, line comparison in config.php files should be integrated to see if the some defined constants are missing before applying the upgrade. When so, the required lines should be automatically added  by the install models prior to complete the upgrade.

prabhatkumaroc082:
Hello @straightlight 
Thanks for the response. I am considering this issue in case of fresh OC installation by using the command line so that "DIR_STORAGE" constant is not found. Also, I have listed all steps in this issue #7512

straightlight:
For fresh installs, according to install/cli_install.php file, the DIR_STORAGE is already defined in the string before the config.php and admin/config.php files are being created.

straightlight:
For the upgrade, install/model/upgrade/1009.php file already validates this constant to be added in both config files.

prabhatkumaroc082:
@straightlight 
After considering all of your comments, I want to say that please have a look again at install/cli_install.php file so first, some constants are defined like DIR_APPLICATION, DIR_SYSTEM, etc except DIR_STORAGE and after that startup file is included using require_once() method. Now just go to line 70 of system/startup.php then you will see DIR_STORAGE is being used in if condition which is not defined yet so that it is generating an error.

straightlight:
Correct. Confirmed. system/startup.php file is being loaded in cli_install.php prior to either a fresh installation or upgrade process is being initiated. The DIR_STORAGE may not yet exists for most of the users who are using old OC versions and wants to upgrade to a higher version. Good catch!

prabhatkumaroc082:
Thanks @straightlight

prabhatkumaroc082:
@jeffreyeng-opencart Please look into this.

PR #7516, opened 2019-07-10
Return copied product

stasdrvn:
In this case developers would be able to set an event on this method.

### DIFF
--- a/upload/admin/model/catalog/product.php
+++ b/upload/admin/model/catalog/product.php
@@ -351,7 +351,7 @@ class ModelCatalogProduct extends Model {
 			$data['product_special'] = $this->getProductSpecials($product_id);
 			$data['product_store'] = $this->getProductStores($product_id);
 
-			$this->addProduct($data);
+			return $this->addProduct($data);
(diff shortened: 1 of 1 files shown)


artxneo:
Good idea!

straightlight:
While this request might be a good idea, the only problematic is while a product may be copied, it does not include the order ID while the sales return feature requires the order ID.

stasdrvn:
May I ask you how do orders connect with the cloned product?

straightlight:
They're not and this is what my previous reply indicates. Which is why, to work with returned products, it requires an order ID.

straightlight:
As confused it may sound to you, just look in the codes and you'll see. admn/model/sale/return.php file for instance.

straightlight:
For the copied product itself, when gathering the new product ID you probably mean, then yes, an event could be created.

stasdrvn:
Majestic!

straightlight:
The returned object from the master branch has been updated since with its own function. Therefore, this commit will now be closed.

PR #7520, opened 2019-07-14
http to https google sitemap urls

condor2:


straightlight:
I did noticed some other files that may need to be changed from http to https as well. Hope they will be changed before the next release.

straightlight:
catalog/controller/extension/feed/google_base.php file may also require this change from:

`http://base.google.com/ns/1.0`

to:

`https://base.google.com/ns/1.0`

ADDCreative:
These are namespaces not links. The documentation for both sitemaps and Google Merchant Center XML feeds still refer to http://. A namespace starting with https:// is a different namespace to one starting with http://.

straightlight:
Kind of odd since when using the http protocol, instead of https, it says on the browser that the http is not secured and the https is secured but the same XML schema appears on the page.

ADDCreative:
They are namespaces, not something you would ever need to put in a browser. Look at the specification on comparison. https://www.w3.org/TR/xml-names/#NSNameComparison Any difference in the string would indicate a different namespace. It will probably work, although you may find the odd validator that will reject it a the namespace is not what it was expecting.

straightlight:
A comparison between the same namespace with different protocols ... budget cuts for sure ... :/

PR #7526, opened 2019-07-18
fixed free checkout payment method issue

WebkulOpencart:
fixed free checkout payment method issue


falconkirtaran:
No, that's contrary to the intent.  Why check greater than zero here?  The entire purpose of the free checkout module is to provide a way to complete the transaction when nothing is owed.  You can't use any of the other payment modules to pay $0.  This should be closed.

Issue #7495, opened 2019-06-26
Admin - Save action being triggered twice


AlexTaylor:
**What version of OpenCart are you reporting this for?**
3.0.1.1

**Describe the bug**
When using the administration in Chrome on Windows to add or edit practically anything (Categories, Products, Customers etc), you will see in the browser console that when you hit the save button the request to save is being sent twice. The first one gets cancelled by the browser when the second one makes a successful trip to the server and exits the page. However, the first request has a very high chance of also making it to the server before it gets cancelled, so depending on what is being saved it could result in duplicates occurring in the database.

The cause of this is pretty simple to find. Look in `/upload/admin/view/javascript/common.js` line 27 for the save button click event:

`$('button[type=\'submit\']').on('click', function() {`
`    $("form[id*='form-']").submit();`
`});`

The simple fix for this is to simply cancel the original event that buttons are assigned if the target form exists, so this code becomes:

`$('button[type=\'submit\']').on('click', function(e) {`
`    if ($("form[id*='form-']").length > 0) {`
`        e.preventDefault();`
`        $("form[id*='form-']").submit();`
`    }`
`});`

**Screenshots / Screen recordings**
Here is a screenshot of the duplicate requests occurring on the Categories admin. You need to Preserve the network log to capture this as the page reload resets the log.

![image](https://user-images.githubusercontent.com/683820/60218471-6086a100-9867-11e9-8fbb-7de1c9cb3dc0.png)

straightlight:
This code has also originally been found on the master branch ... once Daniel sees this commit and respond, we'll see from there.

straightlight:
We seem to have a feedback on this forum post about it: https://forum.opencart.com/viewtopic.php?f=202&t=212638#p759002 .

vaiil:
Maybe just delete this code? It is pollyfill for old IE (6-9: https://caniuse.com/#feat=form-submit-attributes).

lifis:
https://bugs.chromium.org/p/chromium/issues/detail?id=977882

I had a workaround to use firefox, but I use chrome for everything so I have removed it for now:
`	//Form Submit for IE Browser
	$('button[type=\'submit\']').on('click', function() {
		$('form[id*=\'form-\']').submit();
	});`

as I will never use IE 6-9

dyslexicdrakness:
[...]

PR #7528, opened 2019-07-19
Prevent double submit on save button

rensreinders:
We got SQL duplicate key errors. After some examination, we found out that the dit form got submit twice. This fixes block the default HTML behavior and only uses the javascript submit().

### DIFF
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -25,8 +25,10 @@ function getURLVar(key) {
 $(document).ready(function() {
 	//Form Submit for IE Browser
 	$('button[type=\'submit\']').on('click', function(e) {
-		e.preventDefault(); // prevent the native submit behavior of the browser which results in double submits
-		$('form[id*=\'form-\']').submit();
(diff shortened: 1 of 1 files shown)


condor2:
https://github.com/opencart/opencart/issues/7495

rensreinders:
No clue, but latest release still has the bug so to see. All forms with id="form*" in the entire admin get submitted twice, lame...

CodeInKlingon:
No clue how to communicate this problem and get a code commit through but I have narrowed down the problem. 
See this chrome bug thread: https://bugs.chromium.org/p/chromium/issues/detail?id=977882
The issue is in the common.js file. 
https://github.com/opencart/opencart/compare/master...CodeInKlingon:patch-1

CodeInKlingon:
You can see more about this issue here: https://bugs.chromium.org/p/chromium/issues/detail?id=977882

danielkerr:
its been added

	//Form Submit for IE Browser
	if (isIE()) {
		$('button[type=\'submit\']').on('click', function() {
			$('form[id*=\'form-\']').submit();
		});

PR #7537, opened 2019-08-03
Update featured.php

faca5:
Variable $product_info has been changed to $result. Variable $result is used in all other files. Keep it compatible.

### DIFF
--- a/upload/catalog/controller/extension/module/featured.php
+++ b/upload/catalog/controller/extension/module/featured.php
@@ -2,76 +2,61 @@
 class ControllerExtensionModuleFeatured extends Controller {
 	public function index($setting) {
 		$this->load->language('extension/module/featured');
-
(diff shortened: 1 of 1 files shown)


straightlight:
No longer needed since the ```$this->load->controller``` now takes the lead to load the product/thumb from the master branch. However, ```if ($data['products']) {``` is not needed either since the TWIG engine can already take care of the IF statements prior to load the needed arrays or objects.

faca5:
Hello.

Just rename variable name from "product_info " to "result".
It will match all other files like latest.php etc.

Thank you.

straightlight:
No longer relevant with the master branch. Nowadays, the master branch uses the $this->load->controller('product/thumb', $product_data) to handle the arrays altogether from each required controllers. Therefore, this commit will now be closed.

PR #7538, opened 2019-08-03
Update divido_calculator.php

faca5:
Variable $product_info has been changed to $result. Variable $result is used in all other files. Keep it compatible.


straightlight:
No longer among the master branch. Deprecated extension. Therefore, this commit will now be closed.

PR #7540, opened 2019-08-07
Make phone number part of the link in #top in header

stancl:
This is the only item whose text is not part of the link.

### DIFF
--- a/upload/catalog/view/theme/default/template/common/header.twig
+++ b/upload/catalog/view/theme/default/template/common/header.twig
@@ -47,7 +47,7 @@
     </div>
     <div class="nav float-right">
       <ul class="list-inline">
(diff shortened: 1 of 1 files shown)


straightlight:
Not really sure how this commit is useful since clicking the contact link even without the phone number would still work ...

stancl:
All links except this one are clickable. That's inconsistent.

straightlight:
I believe this has been reported on the forum before as I am still not convinced this would be significant feature to have since a custom theme could say otherwise or an activated phone number as a link could also be used for something else. On either case, the link itself still redirects to the contact page.

According to the last conversation on the forum, it was apparently voluntary to have the phone number as an active link. Besides, it is still possible to set a small modal window to use this phone number for other reasons when we roll-over on the element or text.

straightlight:
Based on the notes I provided on the above, this commit will now be closed.

Issue #7162, opened 2019-01-28
OC 3.0.3.1 Softaculous Upgrade (Registration Error) & Fresh Install Inconsistancies


gallnomad:
**What version of OpenCart are you reporting this for?**
This is regarding a clean install of OC 3.0.3.1 from Softaculous and an upgrade from 3.0.2.0 to 3.0.3.1 in Softaculous.

**Describe the bug**
When upgrading from 3.0.2.0 to 3.0.3.1 in Softaculous the extensions now shows Advertising as a module. When doing a clean install of 3.0.3.1 in Softaculous this module does not appear, you can see in both of the screen captures attached.

The second issue which also seems to be related to an upgrade from 3.0.2.0 to 3.0.3.1 in softaculous is that there is a strange entry in the extensions list heading_title (0) this does not appear in the fresh install of 3.0.3.1 in Softaculous. Also on the upgraded version when google capture is used the registration screen goes to a 500 error, it does register the account but does not return to the correct screen. This does not happen on the clean installs, only on the upgrade.

What section does it affect?

**To Reproduce**
Doing a fresh install of OC 3.0.3.1
Doing an upgrade of 3.0.2.0 to 3.0.3.1

**Expected behavior**
I am not sure what is to be expected I assume that I am supposed to see the Advertising Module on the clean installs which is not showing. I also assume that I am not supposed to see heaing_title (0) on the upgrade and expect that the registration on the upgrade using captcha should not go to a 500 error.

**Screenshots / Screen recordings**
Clean install 3.0.3.1
![install soft to new version](https://user-images.githubusercontent.com/47094628/51812425-08f9fe80-22e4-11e9-805d-4873032feea0.png)

Upgrade from 3.0.2.0 to 3.0.3.1
![oc upgrade](https://user-images.githubusercontent.com/47094628/51812426-0c8d8580-22e4-11e9-867f-ad414a1eff60.png)

**Server / Test environment (please complete the following information):**
All websites and servers are running the same version so I don't think it is an issue there or the problem would be happening on all sites.

 - Web Server
 - Centos 7
 - PHP version 7.2.7
 - Apache version Easy Apache 4
 - Browser(s) tested with chrome
 - Version current

**Additional context**
[...]

PR #7570, opened 2019-08-30
Missing translation file: promotion.php

osworx:
Also missing in 3.0.3.2

### DIFF
--- a/upload/admin/language/en-gb/extension/extension/promotion.php
+++ b/upload/admin/language/en-gb/extension/extension/promotion.php
@@ -1,0 +1,3 @@
+<?php
+// Heading Title
+$_['heading_title']	= 'Promotion';


engwalid1971:
The file is not needed at all !

osworx:
Then remove that file: https://github.com/opencart/opencart/blob/master/upload/admin/controller/extension/extension/promotion.php
Otherwise in any 3.0.3.2 installation at many clients I see the error in the dropdown (extension):
`heading_title(0)`
I (and we) know that it useless, but do that also the customers?

engwalid1971:
> Otherwise in any 3.0.3.2 installation at many clients I see the error in the dropdown (extension):
> `heading_title(0)`

**The same issue:**
https://github.com/opencart/opencart/issues/7162
**The solution:**
https://github.com/opencart/opencart/commit/15b592ebbcc4bade488f11b3c99e37917c61dd16

osworx:
"The Solution" was a commit on the 29 Jan 2019
The release of v3.0.3.2 was April 09, 2019

Sounds good ..?

**Question:**
Why is the message still displayed when there was a fix?
And I am sure it will be also in 3.0.3.3 (whenever this will be published - IF).

engwalid1971:
The release of 3.0.3.2 is the same as 3.0.3.1, but only includes fixes for Google Shopping.
The release of 3.0.3.1 was Jan 07, 2019

Screenshot from 2019-08-31 13-52-28

osworx:
Cool down baby ...

I know exactly when is what and how published.
So, either delete the controller or add the language file.
Simple solution and hopefully for the 3.0.3.3 release - if ever.

PR #7574, opened 2019-09-02
Session work done#2

engwalid1971:
- Added clean() function
- Auto clean the expired sessions


danielkerr:
i think this is redundant.

PR #7591, opened 2019-09-09
Fix CORS OPTIONS issue

Shinrai:
Fixes issues when dealing with Custom API creations and handling CORS correctly. Issue is that headers will never be sent on an OPTIONS request as the output SHOULD be '' thus the headers portion is never fired and thus no CORS headers are sent.

### DIFF
--- a/upload/system/library/response.php
+++ b/upload/system/library/response.php
@@ -13,7 +13,7 @@
 class Response {
 	private $headers = array();
 	private $level = 0;
-	private $output;
+	private $output = false;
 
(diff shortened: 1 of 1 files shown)


ilcomp:
Null will not do?

straightlight:
Not happening. Therefore, this commit will now be closed.

Shinrai:
What exactly is not happening? If you attempt to send out a OPTIONS header page there is no output. Therefore the page never loads the header response.

PR #7593, opened 2019-09-10
store new address in session if new user

giulianoriccio:
If a new user with no address adds a new address using the account panel it will not be stored in session and the tax calculation doesn't consider the new and only address that the user has.

This fix stores the new address if no "address_id" is stored.


straightlight:
Not bad. Might be considered for the features-pre version later.

PR #7594, opened 2019-09-10
Fix tax calculation for new users with no address

giulianoriccio:
Based on "config_tax_customer" the $this->session->data['payment_address'] or $this->session->data['shipping_address'] will be set to "false" if a new user with no address set logs in.

Replacing "isset" (which returns "true" even if the value is "false") with "!empty" allows the software to temporarily use the store locale settings.

### DIFF
--- a/upload/catalog/controller/startup/startup.php
+++ b/upload/catalog/controller/startup/startup.php
@@ -222,13 +222,13 @@ class ControllerStartupStartup extends Controller {
 		// Tax
 		$this->registry->set('tax', new Cart\Tax($this->registry));
 
-		if (isset($this->session->data['shipping_address'])) {
+		if (!empty($this->session->data['shipping_address'])) {
 			$this->tax->setShippingAddress($this->session->data['shipping_address']['country_id'], $this->session->data['shipping_address']['zone_id']);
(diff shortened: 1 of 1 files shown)


straightlight:
Will not rectify with PHP v7.4 compatibility. Therefore, this commit will now be closed.

giulianoriccio:
> 
> 
> Will not rectify with PHP v7.4 compatibility. Therefore, this commit will now be closed.

I'm sorry for the late reply, but can you explain what does this mean?
Is there any behaviour change in 7.4 I'm not aware of?

straightlight:
> > Will not rectify with PHP v7.4 compatibility. Therefore, this commit will now be closed.
> 
> I'm sorry for the late reply, but can you explain what does this mean?
> Is there any behaviour change in 7.4 I'm not aware of?

Yes, the proposed change on my behalf has been integrated into the master branch officially since recently:

https://github.com/opencart/opencart/blob/master/upload/catalog/controller/startup/application.php#L70

straightlight:
Also take note that the extensions folder might also involve these validations on the payment, shipping or total folder. If you're also using 3rd party extensions, I would also suggest to look into these validation types.

giulianoriccio:
understood, thank you

Issue #7606, opened 2019-09-13
Displaying query error when we delete the recurring profile.


akhil4016:
![Screenshot_469](https://user-images.githubusercontent.com/54854052/64874709-3922c180-d669-11e9-8d6f-8efd2babdbd8.png)

![Screenshot_470](https://user-images.githubusercontent.com/54854052/64874728-42139300-d669-11e9-9d87-dd2a84151f9d.png)

**What version of OpenCart are you reporting this for?
3.1.0.0_b

**Describe the bug**
When we delete recurring profile from Catalog/Recurring Profiled page, It's displaying the query error.

PR #7607, opened 2019-09-13
Resolved query error on Recurring Profile page.

akhil4016:
A single quote (') is missing in delete recurring query due to that some query error display at the admin front.
Displaying query error when we delete the recurring profile. #7606


danielkerr:
added manually and also had to added in a few pother places

PR #7608, opened 2019-09-13
issue fixed for unchecked box for recaptcha v2

mehdiRma10:
Forms are being sent even if the recaptcha v2 box is not checked, solved when the unnecessary "empty" function is deleted from the condition here.


straightlight:
No longer supported on the master branch. Therefore, this commit will now be closed.

PR #7611, opened 2019-09-17
Fixed console error for the datetimepicker js library

Aqib-Webkul:
Everywhere in admin where datetimepicker is used error is showing in Console when we click on  it.
Screenshot_2


Aqib-Webkul:
@jeffreyeng-opencart Please look into this.

PR #7614, opened 2019-09-19
Added workaround for undefined GLOB_BRACE on non GNU systems

tim-geier:
The GLOB_BRACE flag is not available on all systems, see: https://www.php.net/manual/en/function.glob.php -> Notes
I removed the flag where possible or split the glob command in multiple parts where needed.
Behavior on systems with GLOB_BRACE available is unchanged. Systems missing GLOB_BRACE functionality have to execute more glob commands, one for each pattern in braces, and the results have to be merged into a single result array.
Tested on systems with and without GLOB_BRACE functionality, working as expected.
Needed for a few webhosters which are running Solaris (e.g. Strato).

### DIFF
--- a/upload/admin/controller/common/filemanager.php
+++ b/upload/admin/controller/common/filemanager.php
@@ -63,7 +63,18 @@ class ControllerCommonFileManager extends Controller {
 
 		$data['images'] = array();
 
-		$files = glob($directory . $filter_name . '*.{jpg,jpeg,png,gif,JPG,JPEG,PNG,GIF}', GLOB_BRACE);
+		if (defined('GLOB_BRACE')) {
+				$files = glob($directory . $filter_name . '*.{jpg,jpeg,png,gif,JPG,JPEG,PNG,GIF}', GLOB_BRACE);
+			} else {
+				$files = glob($directory . $filter_name . '*.jpg');
+				$files += glob($directory . $filter_name . '*.jpeg');
+				$files += glob($directory . $filter_name . '*.png');
+				$files += glob($directory . $filter_name . '*.gif');
+				$files += glob($directory . $filter_name . '*.JPG');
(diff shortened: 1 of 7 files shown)


straightlight:
Server-specifics. Therefore, this commit will now be closed.

qroft:
> Server-specifics. Therefore, this commit will now be closed.

But does this mean, that it will be implemented or not in one of the next branches?
Sorry for this maybe silly question, but i am still noobish about all those GitHub procedures.

osworx:
> 
> 
> Server-specifics. Therefore, this commit will now be closed.

What a poor answer!
Not the server has to be compatible, the software.
With such a reponse, you kick out all users using Strato (e.g.) as Hoster.

I could understand if OpenCart decides to use php 7.4 as minimum, but having a mixture of old and new libraries is not what I expect from a serious source.

qroft:
What i wonder is why you don´t simply use the IF ELSE solution someone has added in here?
AFAIK and understood, glob_brace is nice, but not mandatory - am i correct?

At the moment i am struggeling with the thumbnails not showing up on my image manager.
This and the Glob_Brace thingy are two things that i have been confrontated with for months and it would be nice if you could explain why the IF THEN ELSE solution is not a good one.

By the way: yesterday i tried out Shopware 6.
After uploading trillions of files for hours, the first thing that pops up is a beautiful page with the info that Shopware only works with MySql 5.7 and not mine, that is 5.6

The weird: Shopware asks you to update to v6 as soon as possible, but MySQL 5.7 is widely not used by many.
This said, people are asked to move back to Shopware 5 :)
Nice move Karen!

StephanRichter:
Still an issue with the newest version. 

GLOB_BRACE is not available in alpine linux, which will prevent MANY docker users from using opencart.

> Server-specifics. Therefore, this commit will now be closed.

This is a "f*ck you" to so many people…

PR #7624, opened 2019-09-27
removed errant L that was causing install to fail

LucasBSmith:
errant lowercase L on line 1837 was causing the installation process to fail with MYSQL error 1064


straightlight:
No longer relevant due to the new SEO URL update from the master branch. Thefore, this commit will now be closed.

PR #7626, opened 2019-09-28
when you remove the filter, it remains "&filter="

BuslikDrev:


upload/catalog/view/theme/default/template/extension/module/filter.twig
```diff
@@ -37,6 +37,10 @@ $('#button-filter').on('click', function() {
 		filter.push(this.value);
 	});
 
-	location = '{{ action }}&filter=' + filter.join(',');
+	if (filter != 0) {
+		location = '{{ action }}&filter=' + filter.join(',');
+    	} else if (filter = []) {
+		location = '{{ action }}';
+    	}
```
mesili: Please let me suggest this version :
```suggestion
	location = '{{ action }}';
	if (filter.length) {
		location += '&filter=' + filter.join(',');
    	}
```
Why ? 
1 - Checking the length is better than type casting (`filter != 0`)
2 - Line 42, you are modifying filter by assigning a new array to it
3 - Avoiding code repetition (l`ocation = ...`)

You could even go further with a ternary operator but I reckon it would be too much.

ps : github seems to suggest on the wrong line number (44 instead of 40)

ADDCreative: @mesili That's not going to work as the new document will be loaded on assignment of location. https://developer.mozilla.org/en-US/docs/Web/API/Window/location

PR #7667, opened 2019-10-29
Fixed get heading_title captcha in the store settings

WebkulOpencart:
Fixed get heading_title captcha in the store settings


WebkulOpencart:
@jeffreyeng-opencart Please check this.

PR #7668, opened 2019-10-30
fixed datetimepicker CSS issue

dilipkumardev:
fixed datetimepicker CSS issue


jeffreyeng-opencart:
@dilipkumardev: You also need to update the min.css file as well. Currently by default, it is using bootstrap-datetimepicker.min.css. So you need to update and commit bootstrap-datetimepicker.min.css as well.

dilipkumardev:
@jeffreyeng-opencart Now I have updated bootstrap-datetimepicker.min.css file as well

PR #7705, opened 2019-11-19
Several fixes

Grumvol:
Several fixes


straightlight:
While you may address the good point for the catalog.php file, the framework.php file should not be using a dynamic way to add headers especially due to the mount of extensions being used out there for security reasons.

PR #7721, opened 2019-11-28
Fixed design layout issue at the currency form page.

webkul-dilip:
Fixed design layout issue at the currency form page.


webkul-dilip:
@jeffreyeng-opencart same design issue at currency page, please check

PR #7727, opened 2019-11-30
Update language

Grumvol:
First commit.
Unnecessary condition. The request contains this condition.
"SELECT * FROM " . DB_PREFIX . "language WHERE status = '1' ORDER BY sort_order, name"

Second commit.
There is no longer such a field 'image' (left from the old version).


cgcmart:
conflicts with opencart/upload//admin/model/localisation/language.php

Grumvol:
@cgcmart added fix

PR #7733, opened 2019-12-04
fixed the blank space issue in the admin end catalog section

WebkulOpencart:
fixed the blank space issue in the admin end catalog section

### DIFF
--- a/upload/admin/controller/catalog/attribute.php
+++ b/upload/admin/controller/catalog/attribute.php
@@ -350,7 +350,7 @@ class ControllerCatalogAttribute extends Controller {
 		}
 
 		foreach ($this->request->post['attribute_description'] as $language_id => $value) {
-			if ((utf8_strlen($value['name']) < 1) || (utf8_strlen($value['name']) > 64)) {
+			if (empty(trim($value['name'])) || (utf8_strlen($value['name']) < 1) || (utf8_strlen($value['name']) > 64)) {
 				$this->error['name'][$language_id] = $this->language->get('error_name');
 			}
 		}
--- a/upload/admin/controller/catalog/attribute_group.php
+++ b/upload/admin/controller/catalog/attribute_group.php
@@ -330,7 +330,7 @@ class ControllerCatalogAttributeGroup extends Controller {
 		}
 
 		foreach ($this->request->post['attribute_group_description'] as $language_id => $value) {
-			if ((utf8_strlen($value['name']) < 1) || (utf8_strlen($value['name']) > 64)) {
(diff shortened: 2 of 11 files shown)


ADDCreative:
@jeffreyeng-opencart I don't believe this is correct. If you are going to trim the validation you need to also trim when inserting into the database.

For example the name column in the attribute_description table is varchar(64). If the input was over 64 with whitespace at the front, it may pass the now pass validation because of the added trim. However, the full input including the whitespace is still being written to the database, which would lead to the end being truncated.

ADDCreative:
@jeffreyeng-opencart Ignore my last comment as the example was wrong. However, it does apply to the the review author and a few other inputs. There are being trimmed before validation, but not before insertion into the database.

https://github.com/opencart/opencart/blob/aaadf0a56e4368938bcc1a1591a42651062bb885/upload/admin/controller/catalog/review.php#L547

As the column for the review author is varchar(64), if the input had whitespace at the start it would pass validation, but still be over 64 in length when inserted into the database.

WebkulOpencart:
@ADDCreative We trim the review author in both the condition for min length and max length so according to you if we add space at the start it will not passed the validation because trim remove the white space from both start and end of string.
We don't need to trim before inserting into the database because database trim itself the string according to its length. If column length is varchar(64) than only 64 length string will be inserted into the database.

ADDCreative:
@WebkulOpencart If you had review author that had white space at the start and then a 64 character name. This would pass your validation, but when written to the database the last character would be lost.

Similarly, if you have a download description name of ```' A very long download name to demonstrate the problem - episode 1'``` (note space at the start), you would lose the number off the end.

You may also have an issue using empty(), if someone wants to set something to ```'0'```. Which might be the case  for the filter description name or the option value description name.

Grumvol:
@ADDCreative function trim() does not change the original string

Grumvol:
@ADDCreative you're right. I added a PR #7737  to do it right

ADDCreative:
@jeffreyeng-opencart There is still the same issue with your commit https://github.com/opencart/opencart/commit/30b0a42fe755583b67cfebac87e87ea8e860df6b.

You don't need to trim when checking for greater than. Only less than. As I said here https://github.com/opencart/opencart/pull/7733#issuecomment-562189305, you could lose information when written to the database.

You would probably need the validation to be generally in this format, with different minimum and maximum values as needed.
```php
if ((utf8_strlen(trim($value['name'])) < 1) || (utf8_strlen($value['name']) > 64)) {
```

PR #7797, opened 2020-01-02
Update order.php

GautamMKGarg:
Mails are getting sent even if notify was set to false. Now I am checking notify before sending mail.

### DIFF
--- a/upload/catalog/controller/mail/order.php
+++ b/upload/catalog/controller/mail/order.php
@@ -266,20 +266,22 @@ class ControllerMailOrder extends Controller {
 			$from = $this->config->get('config_email');
 		}
 
-		$mail = new Mail($this->config->get('config_mail_engine'));
-		$mail->parameter = $this->config->get('config_mail_parameter');
-		$mail->smtp_hostname = $this->config->get('config_mail_smtp_hostname');
-		$mail->smtp_username = $this->config->get('config_mail_smtp_username');
-		$mail->smtp_password = html_entity_decode($this->config->get('config_mail_smtp_password'), ENT_QUOTES, 'UTF-8');
-		$mail->smtp_port = $this->config->get('config_mail_smtp_port');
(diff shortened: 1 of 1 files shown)


GautamMKGarg:
Hello Team,

Is there any update on this? This pull request is pending since long. Do you need more clarification before approving the pull request?

ADDCreative:
Notify is already checked before sending, at least for edit. If you don't want to send a email for add then you shouldn't call the function at all.

https://github.com/opencart/opencart/blob/2cc1b4f0c4cada4213423e7d8419f96cf0d38b21/upload/catalog/controller/mail/order.php#L38-L40

GautamMKGarg:
I am payment gateway plugin developer, I am using this function to save the history of order
$this->model_checkout_order->addOrderHistory
This function is sending order emails to customers even if email notification is off for orders. But in actual, it should not send a notification email to the customer if order emails is off in settings, but it is not working.
Am I doing something wrong?

danijelGombac:
The setting option is for admin email notification, not customer.

GautamMKGarg:
ok, from where should I choose wether to send email notification to customer or not?

danijelGombac:
I think you should remove or disable the "mail_order_add" event.

ADDCreative:
As long as the fourth parameter of $this->model_checkout_order->addOrderHistory is false or not supplied, no email will be sent when changing the order history.

To ensure that a email is not sent on setting the first status. Add  ```&& $notify``` to:
https://github.com/opencart/opencart/blob/2cc1b4f0c4cada4213423e7d8419f96cf0d38b21/upload/catalog/controller/mail/order.php#L33

However, doing this would cause problems for some payment modules where the comment is intended for the admin only, but the email is still required to be sent. One such example.
https://github.com/opencart/opencart/blob/4ba536198a4638b6d3902a7c720ce8503b0dc394/upload/catalog/controller/extension/payment/authorizenet_aim.php#L151-L156

PR #7802, opened 2020-01-03
Update filemanager.php

Grumvol:
mistakenly specified key 'images'


danielkerr:
this has already been added

PR #7805, opened 2020-01-07
getProductOptionValues not used

Grumvol:


straightlight:
Confirmed. We have different variances for the product options in the master branch but this method name itself is not used. Nice catch!

danielkerr:
already been removed

PR #7816, opened 2020-01-09
ControllerStartupSeoUrl::rewrite() Method with Table seo_url

bodo-hugo-barwich:
I made the Method `ControllerStartupSeoUrl::rewrite()` work with the entries of the Table `seo_url`


straightlight:
Not sure why a bunch of echo lines should be left commented out as opposed to the original file. Removing the extra space has its positive side of views but adding troubleshooting lines as a suggestion would not be a great example for store owners running their Business on a live site.

bodo-hugo-barwich:
Thank you for taking you time to look into this issue.

Formatting Issues have been corrected easily.

The value of this development lies in Reenabling the "SEO URL" Functionality that has been disabled some months ago.

Still I am not completely satisfied by this development since it produces hundreds of SQL Requests for each Web Page with the Model `design/seo_url` but it works already.

straightlight:
The SEO URL has been re-written in the master branch. Therefore, this commit will now be closed.

PR #7817, opened 2020-01-10
Update Bootstrap to version 4.4.1 in admin panel

Grumvol:
scssphp version updated to 0.8.4 (for correct compilation scss)
only the scss/_variables.scss is changed in bootstrap source files
styles breadcrumb and card moved from bootstrap source files to stylesheet.css


danielkerr:
already gone past this version

PR #7824, opened 2020-01-17
Fix model (checkbox replaced by radio)

Grumvol:
Need to change the code in the model, since now the radio is used and not the checkbox.


straightlight:
I'm not sure I would have to agree with the removal of isset validations as opposed to put the (int) as a stand-alone since the top / bottom or any side keys are not mandatory and can still be shown in the set_error_handler whenever the E_ALL ~ E_NOTICE are being used to identify those missing keys. Besides, this has happened a lot in the past with older OC versions.

However, the side keys should be sanitized and validated on the controller before calling each models, in this case.

Grumvol:
the value will always be 0 or 1, because the radio input is used

straightlight:
True. However, this sanitizing and validation would not be any different than the page query on the URL that needs to be sanitized from the controller in the mean time. Therefore, validating from the controller before adding either value, 0 or 1, on the database should be added rather than validating those options directly from the model.

straightlight:
Since no update was provided on the subject, this commit will now be closed.

PR #7825, opened 2020-01-17
Delete Deprecated

haxcop:
Update Delete() for del as is deprecated in redis latest version

PR #7826, opened 2020-01-17
update cache delete for redis

haxcop:
updated delete() for del() to work correctly with latest redis version

### DIFF
--- a/upload/system/library/cache.php
+++ b/upload/system/library/cache.php
@@ -58,7 +58,7 @@ class Cache {
      *
      * @param	string	$key	The cache key
      */
-	public function delete($key) {
+	public function del($key) {
 		return $this->adaptor->delete($key);
 	}
 }


straightlight:
It has never been reported where the delete() method in the library cache handler creates problematic issues by keeping the same conventional name or none that has been linked from from this namespace. 

haxcop:
@straightlight when one person reports and error bug or improvement, the right way to understand the report is to test it yourself accordingly to the report, there's no need for someone else to report the same issue to be sure this is happening. Otherwise, the purpose of Testing and reporting here for all of us is broken as you just rely on others to face the same issue, but you have to also know and understand the fact that other may not be using the same version of Redis as I used to face the problem, your duty before closing this is to understand the issue and replicate it before saying is wrong.

ADDCreative:
@haxcop You have changed the wrong delete. You need to change this one.
https://github.com/opencart/opencart/blob/0fdfc437c9be81a712e8cd6358612c3bb7f0c356/upload/system/library/cache/redis.php#L31

@straightlight This will need changing, although this pull request is wrong. PhpRedis delete method is deprecated. See: https://github.com/phpredis/phpredis/#del-delete-unlink

> Note: delete is an alias for del and will be removed in future versions of phpredis.

straightlight:
I re-opened the case: #7825 . While the #7826 may have a wrong pull request, there's no need to re-address the issue.

Issue #7834, opened 2020-01-23
Suggestion: customer lastName should also be added in contact us page


prabhatkumaroc082:
**What version of OpenCart are you reporting this for?**
3.1.0.0_b

**Describe the bug**
Open contact us page and log in by any customer
Then you will see only first name of logged in user as the input box value

chongshengdz:
you can easily added yourself.
contact-us page should make it as simple as possible.
if too complicated, customer may close the window.

straightlight:
Duplicated commit: https://github.com/opencart/opencart/pull/7835

PR #7835, opened 2020-01-23
added customer lastname in contact us page if login

prabhatkumaroc082:
added description here #7834


straightlight:
I don't see how beneficial that would be. Once the store owners receives the email, a guest user would simply require not to login in order to hide his first name and last name as opposed to logged in customers.

prabhatkumaroc082:
Hello @straightlight 
As this was the suggestion but I don't think it will create an issue with the customer. 
It's for the logged-in customer, not for the guest as we are offering the full name of a logged-in customer but it is totally up to the user that he wants to have the full name in the contact us form or something else.

If you don't think that it is beneficial then kindly close to this issue because there is no need to communicate for this little suggestion.

straightlight:
Commit closed as per user.

PR #7846, opened 2020-01-29
Correct ISO-Codes for German States according ISO 3166-2

osworx:
See: https://en.wikipedia.org/wiki/ISO_3166-2:DE


straightlight:
Better to use an extension to handle compliance such ways since the installation does not allow store owners to choose their countries unless completely installing Opencart - then, to go to the admin systems settings to switch the country ID into another country for each stores. Otherwise, it may impact too many stores and their installed extensions.

PR #7847, opened 2020-01-29
Add POST with Content-Type: application/json

ilcomp:
Add POST request support with Content-Type: application/json


straightlight:
This proposed code is insufficient. It validates whenever there's an array but when it's a JSON string output, it is already tracked from the controller/startup folder and other controllers using the custom fields with the ```json_decode($var, true)``` when pull from the database. Using an array validation in this case may not be useful since it would also interfere with JSON decoding processes.

straightlight:
Not useful since remote APIs can still rely on a connector to pull JSON encoding response. Therefore, this commit will now be closed.

PR #7848, opened 2020-02-03
Replace file name empty’s space of image [Admin area]

ILDaviz:
it's best to see a URL with no spaces for uploaded images. I require this change to automate the replacement of spaces with "_".
example: image 1.jpg to image_1.jpg and same thing for folder names.
Sorry for my bad english 💃


yvonn:
Change replateEmptySpaces into replaceEmptySpaces for better reading

opencartstore:
Hi,

I suggest:

`private function replaceEmptySpaces(string $image_url) {`

and:
`return str_replace(' ', '-', $image_url);`

It is better to use "-" than "_"

ILDaviz:
@opencartstore ok, I added the changes!

straightlight:
Not very useful in this case because even the core uses underscores as opposed to dashes where the suggested code on this commit does not seem to consider. Therefore, this commit will now be closed.

ILDaviz:
Ok 🙌🏻

PR #7852, opened 2020-02-10
Added ability to crop image and to get link

ilcomp:
Also, when setting zero sizes, the image is recalculated or returned.
When creating a link to an image, it is cached.
`link` - get an external link to the image (cached)
`resize` - resize image
`crop` - crop image
`getUrl` - get an external link to the image (not cached)

### DIFF
--- a/upload/admin/model/tool/image.php
+++ b/upload/admin/model/tool/image.php
@@ -1,5 +1,28 @@
 <?php
 class ModelToolImage extends Model {
+	public function link($filename) {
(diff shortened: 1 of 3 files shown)


straightlight:
The cache and image folder are both being used as user-defined rather than using the defined constants from the config.php and admin/config.php in this example. Take note that different results might be expected for store owners moving their storage folder outside of their public_html folder when these folders being hard-coded.

straightlight:
Might be a useful one for the features-pre in the future.

PR #7860, opened 2020-02-23
Image directory set from config

straightlight:
Instead of using a static image path name for the uploads or emails to show the logo or when using the file manager, there's a way to automatically define the image path name from DIR_IMAGE.

In catalog/controller/startup/startup.php file,

find:

``$this->registry->set('currency', new Cart\Currency($this->registry));``

add below:

`// Image directory`
		`$parts = explode('/', rtrim(str_replace('\\', '/', DIR_IMAGE)));`

		`$end_parts = array_filter($parts);`

		`$dir_image = end($end_parts);`
		
		`$this->config->set('config_image_directory', $dir_image);`

Then, we could use: `$this->config->get('config_image_directory')` instead of: `image/` . The same codes could also be used on the admin-end.


ilcomp:
I suggest using two constants: DIR_PUBLIC_IMAGE (for cache), DIR_IMAGE (for protection against direct access)

straightlight:
Yes, that is also a good suggestion. The storage folder is already used to upload images. However, the idea from the above is especially about the use of `$data['logo']` for instance where the image static path is being defined rather than fetching the dynamic one from the constant.

PR #7862, opened 2020-02-25
Important: New features/security for session library

v5mvidal:
Optimizes and ensures the security of session management, in addition to adding protection against: session hijacking and session fixation.

**Do not discard our contribution if you encounter any problems. Let us know that we will make the necessary adjustments.**

> Please, I ask you to test before accepting, as we carry out exhaustive tests, but nothing is perfect.

**New features:**

- 32 character for session_id.
- Session cookie management.
- Cookies HTTPOnly for default.
- Cookies Secure when HTTPS is detected.
- Cookie SameSite when PHP 7.3 or higher.
- Support for garbage collector with delay in execution.
- Support to check if the new session_id is already in use (generate new and unique).
- Support to regenerate session_id when: login, password change and data editing.
- Support to clear the session and invalidate the cookie when logoff.

### DIFF
--- a/upload/admin/controller/common/login.php
+++ b/upload/admin/controller/common/login.php
@@ -13,7 +13,10 @@ class ControllerCommonLogin extends Controller {
 
 		if (($this->request->server['REQUEST_METHOD'] == 'POST') && $this->validate()) {
 			$this->session->data['user_token'] = token(32);
-			
+
+			// Regenerate session_id
+			$this->session->regenerate_session_id();
+
 			if (isset($this->request->post['redirect']) && (strpos($this->request->post['redirect'], HTTP_SERVER) === 0)) {
 				$this->response->redirect($this->request->post['redirect'] . '&user_token=' . $this->session->data['user_token']);
 			} else {
--- a/upload/admin/controller/common/logout.php
+++ b/upload/admin/controller/common/logout.php
@@ -5,6 +5,10 @@ class ControllerCommonLogout extends Controller {
 
 		unset($this->session->data['user_token']);
 
+		// Cleanup session
+		$this->session->destroy();
+		$this->session->forget();
+
 		$this->response->redirect($this->url->link('common/login'));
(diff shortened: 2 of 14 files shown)


opencartstore:
> Optimizes and ensures the security of session management, in addition to adding protection against: session hijacking and session fixation.
> 
> **Do not discard our contribution if you encounter any problems. Let us know that we will make the necessary adjustments.**
> 
> > Please, I ask you to test before accepting, as we carry out exhaustive tests, but nothing is perfect.
> 
> **New features:**
> 
> * 32 character for session_id.
> * Session cookie management.
> * Cookies HTTPOnly for default.
> * Cookies Secure when HTTPS is detected.
> * Cookie SameSite when PHP 7.3 or higher.
> * Support for garbage collector with delay in execution.
> * Support to check if the new session_id is already in use (generate new and unique).
> * Support to regenerate session_id when: login, password change and data editing.
> * Support to clear the session and invalidate the cookie when logoff.

The second parameter of the constructor must become mandatory, otherwise the code will break if some smart guy instantiates the Session object without passing the Registry object.

Please, change:
`public function __construct($adaptor, $registry = '') {`

For this:
`public function __construct(string $adaptor, Registry $registry) {`

The conditional `if ($registry)` can then be removed

v5mvidal:
The constructor was kept in the original pattern to facilitate acceptance by Daniel, after the pull is approved, he can make the change if he deems it necessary.

We are happy that you have evaluated. =)

opencartstore:
I hope Daniel accepts or at least looks at his PR.

This was being needed in the OC.

straightlight:
Too repetitive. The:

```$this->load->controller('common/session');```

could possibly return all these calls from one controller rather than handling all these since we might never know what to expect with 3rd party extensions playing with search lines that may also affect those sessions once we see the results.

cgcmart:
#8133
#8141

PR #7882, opened 2020-03-31
update session.php

saeed1:
using $this->start(); instead of   exit('Error: Invalid session ID!');  can be much user friendly.


straightlight:
Not suggested since validating the session string to a specific way must not be considered otherwise in the original while your suggestion imposes the fact that a new session should still be started despite of the session string being validated in the first place.

danielkerr:
no because we need validation its a session id

PR #7883, opened 2020-04-02
Always display the name

devincentiis:
I propose this change to display the product name even when you move to another tab. The same could be done on the other forms.

### DIFF
--- a/upload/admin/view/template/catalog/product_form.twig
+++ b/upload/admin/view/template/catalog/product_form.twig
@@ -20,7 +20,13 @@
       </div>
     {% endif %}
     <div class="card">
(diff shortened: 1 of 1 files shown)


ilcomp:
Use js for pull up the field "name"

straightlight:
Using the product_description at the {{ text_form }} position is above the form element which doesn't make sense without the use of an ID as Opencart originally uses to pull the form ID. Therefore, this commit will now be closed. 

ilcomp would be correct. Better to use jQuery to accomplish this than using hardcoded HTML to show the language names. Therefore, this commit will now be c;losed.

PR #7895, opened 2020-04-19
Uninstall all extension files

optimlab:
When you delete an extension, files may remain in the folder system/config/ .


danielkerr:
already fixed

PR #7898, opened 2020-04-21
Remove duplicates

FedkaEv:


straightlight:
Confirmed. Fixed in the pre-release version.

PR #7900, opened 2020-04-21
Add Italy jqvmap

devincentiis:
Add Italy jqvmap with all new provinces
Aggiungo jqvmap dell'Italia con tutte le nuove province


danielkerr:
i updated to the latest jqvmap. u will need o do a new pull request

PR #7922, opened 2020-05-04
Update multiple security issues

dushanDissanayake:
FILTER_VALIDATE_EMAIL is replaced with a proper custom regex to validate emails at the registration.

### DIFF
--- a/upload/admin/view/javascript/ckeditor/plugins/codemirror/js/codemirror.js
+++ b/upload/admin/view/javascript/ckeditor/plugins/codemirror/js/codemirror.js
@@ -1667,7 +1667,7 @@ function extractLineClasses(type, output) {
     var prop = lineClass[1] ? "bgClass" : "textClass"
     if (output[prop] == null)
       { output[prop] = lineClass[2] }
-    else if (!(new RegExp("(?:^|\s)" + lineClass[2] + "(?:$|\s)")).test(output[prop]))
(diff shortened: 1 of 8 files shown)


cgcmart:
What is the reason of the change?
[https://stackoverflow.com/questions/13719821/email-validation-using-regular-expression-in-php](url)

dushanDissanayake:
> What is the reason of the change?
> 
> [https://stackoverflow.com/questions/13719821/email-validation-using-regular-expression-in-php](url)
> 
> > Using regular expressions to validate email addresses is not recommended as it is certainly not pretty, especially if you don't want to exclude somebody who has a valid email address that is correct according to RFC 822 grammar.

But the usage of FILTER_VALIDATE_EMAIL allows users to register with some emails which cannot even exist because opencart does not verify email addresses.
> Ex:- j!o*h+n$d%=o#e^@gmail.com is not a valid email in gmail domain. But this web application allows to create an account using this mail.

cgcmart:
The fundamental problem is spambot attacking register form. The solution is implementing a smart spambot trap to stop spambot.

ADDCreative:
Why prevent the use of TLS 1.3?

straightlight:
This solution is user-defined. It will not guarantee its succession on preventing SPAM bots, in this case. Therefore, this commit will now be closed since vendors scripts are also being suggested. Please provide such solutions to the officials so to know their point of view in the future before officially posting such solutions in this namespace.

PR #7936, opened 2020-05-13
Update bank_transfer.php

eka7a:
Missing value.


straightlight:
Now fixed on the pre-release version.

PR #7941, opened 2020-05-22
Product Options Sort Order

eka7a:
The order of options changes every time the product is edited


straightlight:
Might be a good idea.

straightlight:
Added in the features-pre release.

PR #7945, opened 2020-05-23
update api.php

redochka:
typo in function name


cgcmart:
There are more lines on upload/admin/controller/sale/order.php

straightlight:
Not sure if this is related to SessOnId or SessionId. Will have to wait for confirmation on that one until this conventional name can get fixed in the pre-release version.

danielkerr:
added manually

PR #7965, opened 2020-05-28
Custom field is checked although it is not required.

eka7a:
Custom field is checked although it is not required.

It is already defined before foreach (custom_field_customer_group and custom_field_required).
No need to define for null array value again.


straightlight:
Should be:

```if (isset($custom_field_customer_group['required']) && $custom_field_customer_group['required'] && isset($custom_field_customer_group['customer_group_id'])) {```

eka7a:
I fixed

PR #7975, opened 2020-06-01
Featured module - rename variables

stalker780:
Make product variable names same as in other product modules (bestseller, special, latest, etc).
This may be important for OCMODs.

Actually, all code, assigning $product_data = array(), could be moved to controller 'product/thumb'.
So all product lists in (category, search, manufacturer, special, bestsellers, etc.) should look like:
```
    $data['products'][] = $this->load->controller('product/thumb', $result);
}
```


stalker780:
Closed by #7976

PR #7976, opened 2020-06-02
Product thumb - reduce repeated code

stalker780:
Product thumb is a great idea and a blow of fresh air in opencart code structure.
But we can reduce more repeated code


eka7a:
Hi,
Each module has its own image size setting. This way, you cancel module image size settings. Module settings must be included in the array.

straightlight:
Not suggested to use. eka7a would be correct, each settings are handled by variables and only some are left with the $result calling a key specifically from the $product_data array. Therefore, this commit will now be closed.

PR #7980, opened 2020-06-04
No download status id setting in settings

eka7a:


straightlight:
Confirmed. Issue found in the master branch and the pre-release. Thanks for confirming this. It has now been removed from the pre-release version.

PR #7985, opened 2020-06-06
Tax Rules Order By Priority

eka7a:


straightlight:
Might be a good idea.

straightlight:
Added in the features-pre.

PR #7991, opened 2020-06-10
If customer name or ip address does not equivalent to filter value

eka7a:
If customer name or ip address does not equivalent to the filter value an empty result appears. It is more useful to list the results without typing an equivalent name.


straightlight:
There might also be multiple extensions using filtering different ways as well. It would be harden to consider this solution without affecting anything else at this time. Therefore, this commit will now be closed.

PR #7995, opened 2020-06-13
Customer approval- wrong variables

condor2:


condor2:
Also I think all filter_name variables from customer controller should be changed.

PR #8009, opened 2020-06-16
Fix admin order_form view template

norzon:
Line 747: Added the missing `else` value for the input's value attribute for the custom field of type `text`


straightlight:
Confirmed. However, this commit could of been posted in the issues tab. Thanks for posting this.

PR #8010, opened 2020-06-16
Product thumb - reduce repeated code

stalker780:
Update for #7976
Another try.
Now optional $url and $setting variables can also be passed to thumb controller.
Image uses a template default size if it is not set in $setting array.

### DIFF
--- a/upload/catalog/controller/extension/module/bestseller.php
+++ b/upload/catalog/controller/extension/module/bestseller.php
@@ -13,44 +13,7 @@ class ControllerExtensionModuleBestSeller extends Controller {
 
 		if ($results) {
 			foreach ($results as $result) {
-				if ($result['image']) {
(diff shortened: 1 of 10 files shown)


straightlight:
By considering this, all variables being conditioned in the $product_data will be disregarded. Not quite a valuable solution as compared to the $product_data being passed to the product/thumb.

straightlight:
Since the commit #7976 also indicates the facts, this commit will now be closed.

stalker780:
Life must be a pain. 

1. Just create 9 events, instead of 1, if you want to modify product_data of product thumb.

2. You need just gues if there is any custom extension using product thumb. And create event for it.

3. This will never happen. Nobody ever modifies product thumb.

*Sarcasm mode off *

Issue #8016, opened 2020-06-17
incorrect sent count in marketing mail success message


githubertus:
**What version of OpenCart are you reporting this for?**
3.0.3.4-core-pre
Having successfully sent 2 mails to 2 customers - I'm getting 
_Your message has been successfully sent to 0 of 2 recipients!_

Looks like computing of $start in admin\controller\marketing\contact.php line 190
needs some fixing

straightlight:
This issue seem to be an on-going report from users. If this issue still can be reproduced on the master branch, then I'll re-open the commit.

githubertus:
Hmm, I don't know how you would expect any difference - the master branch code is almost identical.
You simply can't use the same var acting as a record offset pointer (0, 10, 20, ...) on one hand, and a counter of actually retrieved records (sent mails) on the other.

straightlight:
I'll re-open this commit but, at this time, this issue won't be fixed on the pre-release until a definite solution can be found for this.

githubertus:
i think something like
```
`$sent = count($emails) + $start;
$json['success'] = sprintf($this->language->get('text_sent'), $sent, $email_total);
`
would already do
```

straightlight:
Added to features-pre. Therefore, this commit will now be closed.

PR #8018, opened 2020-06-18
https://github.com/opencart/opencart/issues/8016

sdareios:


straightlight:
```$json['next'] = str_replace('&amp;', '&', $this->url->link('marketing/contact/send', 'user_token=' . $this->session->data['user_token'] . '&page=' . ($page + 1), true));```

This won't work for the master branch. Are you posting this for the pre-release?

sdareios:
> `$json['next'] = str_replace('&amp;', '&', $this->url->link('marketing/contact/send', 'user_token=' . $this->session->data['user_token'] . '&page=' . ($page + 1), true));`
> 
> This won't work for the master branch. Are you posting this for the pre-release?

Why doesn't it work?

straightlight:
> > `$json['next'] = str_replace('&amp;', '&', $this->url->link('marketing/contact/send', 'user_token=' . $this->session->data['user_token'] . '&page=' . ($page + 1), true));`
> > This won't work for the master branch. Are you posting this for the pre-release?
> 
> Why doesn't it work?

Because you're using a true case where true cases with $this->url->link are no longer significant in the master branch.

sdareios:
> > > `$json['next'] = str_replace('&amp;', '&', $this->url->link('marketing/contact/send', 'user_token=' . $this->session->data['user_token'] . '&page=' . ($page + 1), true));`
> > > This won't work for the master branch. Are you posting this for the pre-release?
> > 
> > 
> > Why doesn't it work?
> 
> Because you're using a true case where true cases with $this->url->link are no longer significant in the master branch.

Pre-release is this 3.0.3.4? I did according to what is said here - https://github.com/opencart/opencart/wiki/Creating-a-pull-request. Where should I make a change?

straightlight:
Your first post states an empty description as it did not indicate which OC version you were posting this solution about. When creating a comment, please use the provided template for OC developers to understand the changes and the purpose of the change.

sdareios:
Ok. Understand

straightlight:
The suggested code is too manual. It could be added in the features-pre release with a $this->load->controller instead of using all these starts and ends manually as range. Besides, the common/pagination already uses this sort of ways throughout the master branch and the core-pre release.

straightlight:
Ok, I will try it with this code. See the features-pre release if it's working as intended.

sdareios:
Yes, it works in the features-pre release.

sdareios:
For some reason, everything is very confusing for me. It is unclear what is in the main branch and what code should we use today.

straightlight:
core-pre is about handling all (or most) of the bug-fixes from previous OC 3.x versions and the features-pre is about features that were added in the master branch but could have a fit in future OC stable releases for Opencart.com in the future and being integrated one step at a time as opposed to the core where the core all bug-fixes that can be found are being added more in realtime.

Issue #7946, opened 2020-05-24
OCSESSID - Session Cookie will be rejected as samesite none or invalid without secure attribute


RichardThomas1:
**What version of OpenCart are you reporting this for?**
3.0.3.3 and 2.3.0.2

**Describe the bug**
Cookie “OCSESSID” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To learn more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Cookies

**To Reproduce**
Firefox 76.0.1 console error message.

Although the roll-out is temporarily held on account of Covid-19, Chrome is also enforcing this:
https://blog.chromium.org/2020/04/temporarily-rolling-back-samesite.html

**Expected behavior**
Expect continue working once enforcement is rolled-out across all browsers.
https://developer.mozilla.org/docs/Web/HTTP/Cookies
This states: Browsers are migrating to have cookies default to SameSite=Lax. If a cookie is needed to be sent cross-origin, opt out of the SameSite restriction by using the None directive. The None directive requires that the Secure attribute also be used.

**Additional context**
Concerned for both V3.0.3.3 and same on live stores V2.3.0.2: Cookie “default” will be soon rejected.

cgcmart:
Both Edge and Firefox throw out `Error: invalid session id!` when navigate demo store after master branch installation completed.

danielkerr:
Cookie “OCSESSID” will be soon rejected because it has the “sameSite” attribute set to “none” or an invalid value, without the “secure” attribute. To know more about the “sameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

RichardThomas1:
How do you prevent this on V2.3.0.2?

danielkerr:
im checking now

danielkerr:
need to replace all set cookies like:

setcookie('language', $code, time() + 60 * 60 * 24 * 30, '/', $this->request->server['HTTP_HOST']);

to 

$option = array(
	'expires'  => time() + 60 * 60 * 24 * 30,
	'path'     => ini_get('session.cookie_path'),
	'domain'   => $this->request->server['HTTP_HOST'],
	'secure'   => 0,
	'httponly' => 0,
	'SameSite' => 'None'
);

setcookie('language', $code, $option);

RichardThomas1:
[...]

PR #8019, opened 2020-06-18
Using $this when not in object

osworx:
See https://github.com/opencart/opencart/issues/7946#issuecomment-645580525


ADDCreative:
Shouldn't be setting the domain anyway.

danielkerr:
i already fixed

PR #8037, opened 2020-06-22
Added error warning for download, attribute, option

Hkr32:


straightlight:
Added in the pre-release. Thanks for posting this.

PR #8038, opened 2020-06-23
Use getTotalReviewsAwaitingApproval #4752

ADDCreative:


straightlight:
Correct, from the controller, that is a valid point. Now added in the pre-release.

straightlight:
Also noticed it's in the master branch. Re-opening.

PR #8040, opened 2020-06-24
The model load file has been moved out of array.

eka7a:


straightlight:
Ok, fixed.

Also found this issue on the same controller:

```if ($voucher_theme_info && is_file(DIR_IMAGE . $voucher_theme_info['image'])) {
								$data['image'] = HTTP_CATALOG . 'image/' . $voucher_theme_info['image'];
							} else {
								$data['image'] = '';
							}```

for:

```if ($voucher_theme_info && is_file(DIR_IMAGE . html_entity_decode($voucher_theme_info['image'], ENT_QUOTES, 'UTF-8'))) {
								$data['image'] = HTTP_CATALOG . 'image/' . html_entity_decode($voucher_theme_info['image'], ENT_QUOTES, 'UTF-8');
							} else {
								$data['image'] = '';
							}```

This will be fixed as well.

ADDCreative:
@straightlight Please don't add and html_entity_decode for setting the $data['image']. It will be use in an HTML context so will need to have special characters encoded. Where else have you added html_entity_decode? As not only could they be unnecessary they may also cause security issues.

straightlight:
I have escalated this commit. Originally, it used to be (string) in front of the $data['image'] as I don't see how more secured that can be at this time. We'll see what our next option might need to be.

straightlight:
Perhaps the use of a helper function to pass all the parameters for the entities similar to the oc_setcookies function ...

danielkerr:
im sure file names need to be decoded also.

PR #8053, opened 2020-07-02
Paypal Commerce Platform updated with the latest files

stasdrvn:


ADDCreative:
The bit below is still wrong. You need to number_format after the ```$handling_total = -$rebate; ```. Otherwise you will be still lbe passing a float to PayPal when it should be a formatted string.

https://github.com/opencart/opencart/blob/27c7b2661c6bfa9a44c11348528de5bf0d9b827e/upload/catalog/controller/extension/payment/paypal.php#L216-L222

Alos best to remove the following. If you are still having to mess around with the precision. It means you are not formatting in the right places.
https://github.com/opencart/opencart/blob/27c7b2661c6bfa9a44c11348528de5bf0d9b827e/upload/catalog/controller/extension/payment/paypal.php#L8-L11

PR #8058, opened 2020-07-03
Only list affiliate customer

eka7a:


ivent:
Hello Good idea, There was a link in marketing / affiliate that did this I don't know why it was removed much more practical for users, everything is more complex to identify.

eka7a:
@ivent If no changes are made by @danielkerr, marketing/affiliate page will come back in new version.

danielkerr:
Please do a mock of how it should look

ivent:
I don't know if I explained it well but in the old version 2.0 there was a link in the main menu marketing / affiliate where we could follow all the affiliate approvals etc ..., I don't know why this was hidden in versions 3.x I know it's still possible see this item on clients but now it is very complex I would like the way it used to be when the types were separated and I'm just asking to return as before.

PR #8060, opened 2020-07-04
Added category name when editing

Hkr32:


straightlight:
For the $current_language_id, the following could also be used:

```$current_language_id = $this->model_localisation_language->getLanguageByCode($this->config->get('config_language_id'));```

**Comment by Hkr32** [comment]
`$this->config->get('config_language_id')`
Is this the id of the admin or front language?
I was looking for admin language.

**Comment by Hkr32** [comment]
Thank. I'll know.

**Comment by chulcha** [comment]
```
		if (isset($this->request->get['category_id')) {
			if (isset($data['category_description'][$this->config->get('config_language_id')]['name'])) {
				$data['text_form'] .= ': ' . $data['category_description'][$this->config->get('config_language_id')]['name'];
			}
			
		}

```

**Comment by Hkr32** [comment]
@straightlight, maybe you should do it all the same as suggested above, in a separate block?

**Comment by straightlight** [comment]
I would not suggest the use of hardcoded text. Besides, it is not good practice. This could also be accomplished with the sprintf function as Opencart mostly use with the language definitions. As for the ['name'] key, I would also suggest the use of html_entity_decode with ENT_QUOTES and UTF-8 since multiple keys are being called in the suggestion above.

**Comment by Hkr32** [comment]
Then maybe it’s better not to stick to $ data ['text_form'], but to make a separate variable and just display it after the colon in the template?

**Comment by chulcha** [comment]
> 
> 
> I would not suggest the use of hardcoded text. Besides, it is not good practice. This could also be accomplished with the sprintf function as Opencart mostly use with the language definitions. As for the ['name'] key, I would also suggest the use of html_entity_decode with ENT_QUOTES and UTF-8 since multiple keys are being called in the suggestion above.

This code is not language specific

UTF8?
See  getList()
```

			$data['categories'][] = array(
				'category_id' => $result['category_id'],
				'name'        => $result['name'],
				'sort_order'  => $result['sort_order'],
				'edit'        => $this->url->link('catalog/category/edit', 'user_token=' . $this->session->data['user_token'] . '&category_id=' . $result['category_id'] . $url, true),
				'delete'      => $this->url->link('catalog/category/delete', 'user_token=' . $this->session->data['user_token'] . '&category_id=' . $result['category_id'] . $url, true)
			);
		}
```
Where html_entity_decode with ENT_QUOTES and UTF-8

Hkr32:
@straightlight, why use html_entity_decode? Still displays tags as text on the front?
oc_c1
oc_c2

straightlight:
@Hkr32: It's still wrong. The language definition should be used, not hardcoded. $this->language->get('text_x') where 'x' should be the key name followed by its prefix based on your example.

straightlight:
> @straightlight, why use html_entity_decode? Still displays tags as text on the front?
> oc_c1
> oc_c2

As per the screenshot, the html_entity_decode has been said to be used under the admin based on the folder path with the category controller you have provided as the same for this commit report. It has never been said to use it as per your screenshot shows under the catalog category controller (whether it's for an extension module or from the product/category).

Hkr32:
> @Hkr32: It's still wrong. The language definition should be used, not hardcoded. $this->language->get('text_x') where 'x' should be the key name followed by its prefix based on your example.
can I have an example?

Hkr32:
> As per the screenshot, the html_entity_decode has been said to be used under the admin based on the folder path with the category controller you have provided as the same for this commit report. It has never been said to use it as per your screenshot shows under the catalog category controller (whether it's for an extension module or from the product/category).

why then use html_entity_decode?

chulcha:
> 
> 
> @Hkr32: It's still wrong. The language definition should be used, not hardcoded. $this->language->get('text_x') where 'x' should be the key name followed by its prefix based on your example.
Which $ this->language-> get ('text x')
Where can I get it if it is not in use?

example
$data['text_form'] .= ': ' . $data['category_description'][$this->config->get('config_language_id')]['name'];
Where languge prefix?

straightlight:
> > @Hkr32: It's still wrong. The language definition should be used, not hardcoded. $this->language->get('text_x') where 'x' should be the key name followed by its prefix based on your example.
> > Which $ this->language-> get ('text x')
> > Where can I get it if it is not in use?
> 
> example
> $data['text_form'] .= ': ' . $data['category_description'][$this->config->get('config_language_id')]['name'];
> Where languge prefix?

If you're looking for examples, please use the Opencart forum. In here, code changes must be understood.

chulcha:
[](https://github.com/opencart/opencart/blob/df64ae13dcaf7d603433e7d2d8dc288c57e4322b/upload/catalog/controller/product/category.php#L97)
Where  html_entity_decode?
or
strip_tags

chulcha:
And here is the forum?
You were offered a code that improves UX/UI

Hkr32:
@straightlight, the code that I suggested, it does not suit you?

straightlight:
This is not the forum. The Opencart forum is located at: https://forum.opencart.com .

ADDCreative:
You do not need html_entity_decode. Adding it will create a stored XSS vulnerability, where an admin with restricted privileges could add a script to a category name, that could be executed when other admins view that category.

straightlight:
> You do not need html_entity_decode. Adding it will create a stored XSS vulnerability, where an admin with restricted privileges could add a script to a category name, that could be executed when other admins view that category.

If that being the case, then the issue is larger than simply using html_entity_decode function. While Opencart still uses that function in many other locations, the .user.ini file also calls for html entity decode which propagates throughout the platform already. I took a look at the latest PHP changelogs from php.net just now as it seem there's a bug-fix in there that even prevents .user.ini file to propagate .user.ini file to the root path in the PHP v7.3x series (even in the PHP v7.4x series unfortunately).

In addition, Opencart also uses the strip_tags function where a bug-fix in the PHP v7.3x series has also been posted in the logs. Those two might need to be reviewed in Opencart especially for users that are using below versions than those addressed with the bug-fixes.

ADDCreative:
Whether anything needs to be passed through html_entity_decode has nothing to do with any settings in .user.ini. Therefore the lack of propagating of .user.ini on Windows, would also not matter.

If you are referring to.
> Fixed bug #79099 (OOB read in php_strip_tags_ex). (CVE-2020-7059)

The description suggests that it only applies when fgetss() is used, which is not used by OpenCart.

danielkerr:
not needed

PR #8072, opened 2020-07-17
Fix PayPal Standard when session has SameSite=Lax

ADDCreative:
When new web browsers start to change the default of SameSite to Lax for cookies (https://www.chromium.org/updates/same-site). PayPal needs to return to the store with a GET and not a POST or the session will be lost.

Changing 'rm' to 1 will force PayPal to return using a GET. https://www.paypal.com/ee/smarthelp/article/how-do-i-use-the-rm-variable-for-website-payments-ts1011


cartbinder:
Only twig file change will work?
Because change would also be needed in the callback function of Paypal right?

ADDCreative:
No changes to the callback function are needed. The PayPal IPN callback always uses a POST, but does not require a session.

PR #8087, opened 2020-07-20
Updated product-list view

engwalid1971:
Screenshot from 2020-07-20 13-25-32


eka7a:
@engwalid1971 This fixes the error but if product-list class is added to modules, the module views change when clicked list button.

Issue #8086, opened 2020-07-20
Free Checkout Order Status Saved but not selected


qahmad81:
**What version of OpenCart are you reporting this for?**
checked in last ver 3.0.3.5

**Describe the bug**
When I save Free Checkout it saves all data with (Order Status), but when I reload the form (Order Status) not select on the status with I saved

**To Reproduce**
1. Go to Extensions > Extensions
2. Choose the extension type to payments
3. Edit Free Checkout and select order status to any.
4. Back to the Form and you see order status not on status selected

**Bug Reason**
on the controller when filling the (order status) field, forget (payment) prefix on the field name

**Bug Location**
admin/controller/extension/payment/free_checkout.php
Line: 47
```
if (isset($this->request->post['free_checkout_order_status_id'])) {
	$data['free_checkout_order_status_id'] = $this->request->post['free_checkout_order_status_id'];
} else {
	$data['free_checkout_order_status_id'] = $this->config->get('free_checkout_order_status_id');
}

```
**Bug Fix**
```
if (isset($this->request->post['payment_free_checkout_order_status_id'])) {
	$data['payment_free_checkout_order_status_id'] = $this->request->post['payment_free_checkout_order_status_id'];
} else {
	$data['payment_free_checkout_order_status_id'] = $this->config->get('payment_free_checkout_order_status_id');
}
```

**Bug OCMOD Correction**
add this XML on any ocmod you have
```
	<file path="admin/controller/extension/payment/free_checkout.php">
		<operation>
			<search><![CDATA[if (isset($this->request->post['free_checkout_order_status_id'])) {]]></search>
			<add position="replace" offset="4"><![CDATA[
	//  fix start
	if (isset($this->request->post['payment_free_checkout_order_status_id'])) {
		$data['payment_free_checkout_order_status_id'] = $this->request->post['payment_free_checkout_order_status_id'];
	} else {
		$data['payment_free_checkout_order_status_id'] = $this->config->get('payment_free_checkout_order_status_id');
	}
	//  fix end
			]]></add>
		</operation>
    </file>

```

PR #8089, opened 2020-07-20
[3.0.x.x_Maintenance] Added more fixes

condor2:


kenfai:
@condor2 checkout issue #8086 too, related to your fixes on payment variables.

ADDCreative:
@condor2 This one is wrong. https://github.com/opencart/opencart/commit/4de67d1d4b948e58f702feb4bdfe00988965c3df#diff-eec1d6d7b86e5c5709f103796ce30967 It's a workaround for a bug in Microsoft Mail that would ignore the Reply-To field. Hopefully Microsoft have fixed it by now. The problem is the workaround can cause problems with mail servers that are correctly configured with SPF and DMARC policies.

PR #8104, opened 2020-07-21
Update account.php

AwesomeKhan:


danielkerr:
my acount details would be for the edit age

PR #8105, opened 2020-07-22
[3.0.x.x] Added include not working in opencart fix

condor2:
https://github.com/opencart/opencart/pull/7153#issuecomment-500143157


condor2:
Same bug in Master. Not fixed yet.

PR #8106, opened 2020-07-22
[3.0.x.x] Added '{% include %}' not working fix

condor2:
https://github.com/opencart/opencart/pull/7153#issuecomment-500143157

### DIFF
--- a/upload/system/library/template/twig.php
+++ b/upload/system/library/template/twig.php
@@ -28,7 +28,9 @@ final class Twig {
 		);
 
 		try {
-			$loader = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
+			$loader1 = new \Twig\Loader\ArrayLoader(array($filename . '.twig' => $code));
+			$loader2 = new \Twig\Loader\FilesystemLoader(array(DIR_TEMPLATE)); // to find further includes
+			$loader = new \Twig\Loader\ChainLoader(array($loader1, $loader2));
 
 			$twig = new \Twig\Environment($loader, $config);
 


straightlight:
Not broken.

condor2:
Not broken what?

straightlight:
https://github.com/opencart/opencart/pull/8108#issuecomment-662584496

condor2:
You have tested?...

condor2:
See....
Screenshot_2020-07-22_20-32-52

Error

`Fatal error: Uncaught Twig\Error\LoaderError: Template "test.twig" is not defined in "default/template/product/product.twig" at line 8.`

straightlight:
This is what I am saying. Originally, this is not built-in. Daniel did confirmed it would not happen on Github before. Therefore, this is not broken but rather restricted since Opencart relies on the controller files to output the TWIG files from the TWIG Engine. It is intended that way.

condor2:
Yes, worked before when was twig 1.x because was another code in twig file from system library/template.

straightlight:
It's not a fix. It's been intentionally restricted to maintain the CMS originally from the controllers.

condor2:
I think is bad ideea...

Daniel will decide..

straightlight:
The reason why this restriction is in-place is due to the extensions that could create problems where store owners may not understand or know where additional routes might be originating from in their custom themes. Rather, using a controller can help identify the issue more quickly.

straightlight:
Daniel has already made the decision here: https://github.com/opencart/opencart/pull/7153#issuecomment-500318268 . Yet, you distributed an alternate way on the same commit, underneath, that has also been declined in the past from other commits since 2016.

condor2:
Is like a car shop owner stay in the front of the car and tell you to enter in that car wearing a suit but you wear jeans... :))

straightlight:
> Is like a car shop owner stay in the front of the car and tell you to enter in that car wearing a suit but you wear jeans... :))

Which is why, when fixing your car at the dealers, additional products are being delivered on the passenger's seat in a sealed bag by the time customer service confirms your payment and going back to your car on their parking lot. Therefore, this commit will now be closed.

devheavan:
> I think is a bad ideea...
> 
> Daniel will decide..

think again!
i already decided which layout to use ... why do i need to redefine ???

devheavan:
we require selected layout works with embed/include twig functions too! not only with root request twig files

mhcwebdesign:
At the moment OC4 uses the **FilesystemLoader** by default which would allow for include-tags. However, in case a view/before event handler modifies the template file, the modified template file will be in a string buffer, and the **system/library/template/twig.php** will then use the **ArrayLoader**, the latter won't work for include-tags. This should be fine since template modifications via view/before event handlers are usually only done for OC core template files which don't have include-tags.

There is a real danger that so-called theme designers will make terrible use of unneeded include-tags, block-tags, embed-tags, or worse of all, use-tags, ending up with terrible non-standard compliant web themes again for OC4.

I have already seen discussions of where someone came up with the breadcrumb example as a case for using the include-tag. Just to give one an idea how quickly it can become hard to manage, take a look at this simple twig discussion: https://nystudio107.com/blog/twig-processing-order-and-scope . Imagine more complex cases with embed-tags, use-tags, etc. 

It's easy to drift away from the simple MVC design of OC, especially for the web theme designers, look at the vast number of rubbish web themes on the OpenCart marketplace, or the themeforest.net etc.

devheavan:
simply admit that you can't handle it. it is okay

devheavan:
stay in via before stage

mhcwebdesign:
The real issue here is we are talking here about a bug in the Symfony implementation of Twig. Using an ArrayLoader should not prevent the usage of e.g. include-tags in the first place. The Symfony implemtation of Twig is of poor quality in general, no proper LALR(1) parser design here, no direct translation of Twig to the final HTML in its rendering function (as pointed out by Daniel). It has a poor performance, and an non-intuitive syntax, and a number of redundant functions or tags doing the same thing. Even its expression parser doesn't always use the normal operator precedence. Not to talk about using strange names for its binary operators.

PR #8110, opened 2020-07-22
[3.0.x.x_M] Updated Whos Online footer code

condor2:


straightlight:
system/startup already takes care of that.

PR #8115, opened 2020-07-23
Opencart Version in admin

straightlight:
Opencart admin in version should be v3.1.0.0b . Sorry, messed up the other commit in Pull Requests in opencart v3.x.x.x section. Can't seem to remove it.


cartbinder (changes requested):
When creating a new order, there is an error:
Exception: Error: Could not call registry key model_sale_order! in /system/engine/controller.php on line 25

Fix:

File name:
admin/controller/sale/order.php

Line no: 966
$totals = $this->model_sale_order->getTotals($order_id);

Solution:
if($order_id) {
$totals = $this->model_sale_order->getTotals($order_id);

PR #8117, opened 2020-07-23
GDPR abbreviates to General Data Protection Regulation, not General Data Personal Request

mehov:
Source: https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=celex%3A32016R0679

### DIFF
--- a/upload/admin/language/en-gb/mail/gdpr_export.php
+++ b/upload/admin/language/en-gb/mail/gdpr_export.php
@@ -1,6 +1,6 @@
 <?php
 // Text
-$_['text_subject']    = '%s - General Data Personal Request';
+$_['text_subject']    = '%s - General Data Protection Regulation';
 $_['text_request']    = 'Export Personal Data';
 $_['text_hello']      = 'Hello <strong>%s</strong>,';
 $_['text_user']       = 'User';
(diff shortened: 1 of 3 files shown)


ADDCreative:
Why change this? It makes no sense to have emails sent with the subject General Data Protection Regulation. The emails are about a customer's data request. If anything the subjects should be something like 'GDPR data request' to match the other GDPR emails.

https://github.com/opencart/opencart/blob/436b9042fed792311c3346101bf65e212eda8f27/upload/admin/language/en-gb/mail/gdpr_approve.php#L3

https://github.com/opencart/opencart/blob/436b9042fed792311c3346101bf65e212eda8f27/upload/admin/language/en-gb/mail/gdpr_delete.php#L3

mehov:
@ADDCreative 

> The emails are about a customer's data request.

These were my original thoughts too. But then the text should have said exactly that, without "General Data". 

The presence of "General Data" shows it was kind of meant to look familiar. But having the wrong GDPR abbreviation expansion looks sloppy and clueless, so I suggested to fix it.

> If anything the subjects should be something like 'GDPR data request' to match the other GDPR emails.

I agree, makes sense

PR #8121, opened 2020-07-24
Add ability to filter custom fields by location

mehov:


straightlight:
My repository does use location with custom fields. It might also be a good idea to integrate the other way around as well for filtering.

PR #8126, opened 2020-07-25
Fixed installation issue

qwerty541:
I got the following error while installation:
```
Exception: Error: Variable 'sql_mode' can't be set to the value of 'MYSQL40'
Error No: 1231
SET @@session.sql_mode = 'MYSQL40' in D:\OpenServer\domains\exampleshop.local\system\library\db\mysqli.php on line 43
```
After removing these lines all works without problems. In case of questions, I use PHP 7.3 and MySQL 8.0


engwalid1971:
As of MySQL 5.7.22, MYSQL40 is deprecated.

https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html

pro100Hub:
@danielkerr 3.0.3.7 not fixed :(
/system/library/db/mysqli.php on line 41

Open_Server_XEQRROUB14

qwerty541:
Outdated

PR #8127, opened 2020-07-25
Updated Latest Module

engwalid1971:
Tested with 11,000 products, the MySql CPU resources is 4.7%, after this modification the CPU resources down to 2%


cartbinder:
Wow... Nice finding !!

cgcmart:
particularly useful whenever bulk import products

chulcha:
This is a hack
From a sql point of view this is not correct.

ADDCreative:
I agree with @chulcha, this should be reverted. What if the products have been imported into the database directly or via a feed. The date added isn't guaranteed to match the product_id order.

Also the query is cached.
https://github.com/opencart/opencart/blob/82f72f849a57f28a9569465b9116060bfa076e41/upload/catalog/model/catalog/product.php#L266-L269

If there is a performance issue, it's probably that the cache is cleared on every order history update.
https://github.com/opencart/opencart/blob/82f72f849a57f28a9569465b9116060bfa076e41/upload/catalog/model/checkout/order.php#L394

PR #8130, opened 2020-07-25
Correct version

qwerty541:


ADDCreative:
The master branch is not 3.0.3.6.

qwerty541:
> The master branch is not 3.0.3.6.

Got it, but I have two questions:

1. In the files which I wanted to make changes different versions are written to the constant `VERSION`. It's a mistake, right? Perhaps, string `define('VERSION', '3.1.0.0_b');` inside file `upload/index.php` should be replaced with `define('VERSION', '3.0.3.2');`?

2. How can I in the future understand which version on master?

cgcmart:
If you want to match admin and catalog versions, master version should be
`define('VERSION', '3.1.0.0_b');`
All releases of admin and catalog versions shouldn't correct to match (for example, admin uses bs 4, but catalog stays at bs 3).
All marketplace extensions do not support master branch. Bootstrap version is main factor.

PR #8139, opened 2020-07-28
Added missing Sass compiler in config/catalog.php

kenfai:
### Issue:
`.scss` files in theme stylesheets `upload/catalog/view/theme/*/stylesheet/` does not compile into `.css` during startup process.

### Solution:
Added missing startup pre_action Sass compiler `'startup/sass'` in config/catalog.php

### DIFF
--- a/upload/system/config/catalog.php
+++ b/upload/system/config/catalog.php
@@ -34,6 +34,7 @@ $_['action_pre_action']  = array(
 	'startup/startup',
 	'startup/error',
 	'startup/event',
+	'startup/sass',
 	'startup/maintenance',
 	'startup/seo_url'
 );


cgcmart:
I believe sass is dead, deprecated, instead scss.

kenfai:
It is indeed using an Scss compiler by https://github.com/scssphp/scssphp:
https://github.com/opencart/opencart/blob/3.0.x.x_Maintenance/upload/catalog/controller/startup/sass.php

'Sass' is just being used as the controller's name. Perhaps all instances of the 'Sass' name throughout the codebase need to be updated?

ilcomp:
Add an event to a header so that you don't start this process every time

PR #8150, opened 2020-07-30
Fix Missing product_id in Order Returns

kenfai:
### Issue:
When submitting Order Returns from customer account Order History on `route=account/return/add` page, the `product_id` is not recorded into `oc_return` table.

_Database table screenshot:_
Missing product_id return

### Solution:
[See code changes]


upload/catalog/controller/account/return.php
```diff
@@ -324,6 +324,14 @@ public function add() {
 			$data['order_id'] = '';
 		}
 
+		if (isset($this->request->get['product_id'])) {
```
eka7a, line 327: Must be a post request.

PR #8154, opened 2020-07-30
fix: add extra modules to the dashboard

valdeirpsr:
**01 - Extensions (Settings)**
01 - Extensions

**02 - Before modification**
02 - Before

**03 - After modification**
03 - After

### DIFF
--- a/upload/admin/controller/common/dashboard.php
+++ b/upload/admin/controller/common/dashboard.php
@@ -75,6 +75,8 @@ class ControllerCommonDashboard extends Controller {
 				$column = array();
 			}
 		}
+		
+		$data['rows'][] = $column;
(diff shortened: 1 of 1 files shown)


cgcmart:
Please reverse this commit that changes dashboard view into many rows and many columns, a stepladder like:
1st row 2 cols
2nd row 3 cols
3rd row 4 cols
4th row 5 cols

eka7a:
@cgcmart 
I did not encounter any problems in my tests. Test dashboard does not worked when code is revert.
test

cgcmart:
It seems my repository is different, don't have `$data['rows'][] = $column;`  at line 79, outside foreach.

PR #8168, opened 2020-08-02
Update redis.php

Grumvol:
The cache object does not have a "del" method.


condor2:
I see a commit here where delete is deprecated
https://github.com/opencart/opencart/commit/6e2913c29ada4e34393bc53506d8f463862ba3e3

Grumvol:
We are calling the delete method on the adapter, not the library

Grumvol:
Although most likely this is my mistake

PR #8169, opened 2020-08-02
Update db.php

Grumvol:
1.  Call gc in the constructor (this method is not called anywhere on the site)
2. Invalid comparison in the condition. Can be tested with the code below.

        $probability = 1 / 100;
        $count = 0;
        $total = 0;

        while ($count < 100) {
            $total++;

            $result = mt_rand() / mt_getrandmax();

            if ($result < $probability) {
                $count++;
            }
        }

        $this->log->write('Desired probability - ' . $probability * 100 . '%');
        $this->log->write('Real probability - ' . round($count / $total * 100, 2) . '%');

### DIFF
--- a/upload/system/library/session/db.php
+++ b/upload/system/library/session/db.php
@@ -15,6 +15,8 @@ final class DB {
 		$this->db = $registry->get('db');
 		
 		$this->expire = ini_get('session.gc_maxlifetime');
+		
+		$this->gc();
 	}
 	
 	public function read($session_id) {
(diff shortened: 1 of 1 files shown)


ADDCreative:
I think the garbage collection should consistent between the db and file session adaptors. db uses gc() (which is never called) and file is using __destruct().

https://github.com/opencart/opencart/blob/9f5dd7ae96fd8d382f3f545f27083b9049952ea8/upload/system/library/session/db.php#L44-L64

https://github.com/opencart/opencart/blob/9f5dd7ae96fd8d382f3f545f27083b9049952ea8/upload/system/library/session/file.php#L56-L82

Grumvol:
I can cancel this request and send a new request by replacing gc with __destruct

Issue #8081, opened 2020-07-19
session


engwalid1971:
upload/system/library/session/db.php
upload/system/library/session/file.php
**Replace:**
```
		if ($gc_divisor) {
			$gc_divisor = $gc_divisor;
		} else {
			$gc_divisor = 1;
		}
```
**With:**
```
		if (!$gc_divisor) {
			$gc_divisor = 1;
		}
```

WebkulOpencart:
Please mention the OC version

engwalid1971:
already fixed this issue

PR #8174, opened 2020-08-03
Update db.php

Grumvol:
1. The gc method is not called anywhere, so you can rename it to __destruct (same method name as in file library/session/file.php)
2. Invalid comparison in the condition

### DIFF
--- a/upload/system/library/session/db.php
+++ b/upload/system/library/session/db.php
@@ -41,7 +41,7 @@ final class DB {
 		return true;
 	}
 	
-	public function gc() {
+	public function __destruct() {
 		$gc_divisor = (int)ini_get('session.gc_divisor');
 
 		if ($gc_divisor) {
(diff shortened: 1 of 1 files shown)


cgcmart:
#8081

cgcmart:
comment on system/library/session/db.php line 44:
`public function __destruct() {`

This causes Json dysfunction db syntax errors, Php Warnings:

> PHP Warning:  mysqli::query(): Couldn't fetch mysqli in /system/library/db/mysqli.php on line 21
> PHP Warning:  DB\MySQLi::query(): Couldn't fetch mysqli in /system/library/db/mysqli.php on line 23
> PHP Warning:  mysqli::real_escape_string(): Couldn't fetch mysqli in /system/library/db/mysqli.php on line 48

Grumvol:
I also suggested another way #8169

cgcmart:
your comment on system/library/session/db.php line 59 was merged into master branch

Issue #8043, opened 2020-06-25
Empty GET request causes errors and creates duplicates with different canonical URLs


optimlab:
**Opencart 3.x.x.x**

**Empty GET request causes errors and creates duplicates with different canonical URLs**
If the URL has empty:
1. [`page=`](https://demo.opencart.com/index.php?route=product/category&path=20&page=)
2. [`limit=`](https://demo.opencart.com/index.php?route=product/category&path=20&limit=)
Duplicates with different canonical URLs:
`<link href="https://demo.opencart.com/index.php?route=product/category&amp;path=20" rel="canonical" />`
`<link href="https://demo.opencart.com/index.php?route=product/category&amp;path=20&amp;page=" rel="canonical" />` - **This url is indexed!**

**Fix Instruction**
1. Disable Display Errors.
2. Add in **robots.txt**:
```
Disallow: /*?page=$
Disallow: /*&page=$
Disallow: /*?sort=
Disallow: /*&sort=
Disallow: /*?order=
Disallow: /*&order=
Disallow: /*?limit=
Disallow: /*&limit=
Disallow: /*?filter_name=
Disallow: /*&filter_name=
Disallow: /*?filter_sub_category=
Disallow: /*&filter_sub_category=
Disallow: /*?filter_description=
Disallow: /*&filter_description=
```
**Fix in in future versions Opencart**
Search:
[`if (isset($this->request->get['page'])) {`](https://github.com/opencart/opencart/blob/master/upload/catalog/controller/product/category.php#L30)
Replace:
`if (!empty($this->request->get['page'])) {`

etc..

optimlab:
If you do, we will do it to the end.

```
User-agent: *
Disallow: /*route=account/
Disallow: /*route=affiliate/
Disallow: /*route=checkout/
Disallow: /*route=product/search
Disallow: /index.php?route=product/product*&manufacturer_id=
Disallow: /admin
Disallow: /catalog
Disallow: /system
Disallow: /*?page=$
Disallow: /*&page=$
Disallow: /*?sort=
Disallow: /*&sort=
Disallow: /*?order=
Disallow: /*&order=
Disallow: /*?limit=
Disallow: /*&limit=
Disallow: /*?filter_name=
Disallow: /*&filter_name=
Disallow: /*?filter_sub_category=
Disallow: /*&filter_sub_category=
Disallow: /*?filter_description=
Disallow: /*&filter_description=
Disallow: /*?tracking=
Disallow: /*&tracking=
[...]

PR #8181, opened 2020-08-06
[3.0.x.x_Maintenance] Empty GET request causes errors and creates duplicates with different canonical URLs

condor2:
https://github.com/opencart/opencart/commit/8f0c7237a79ed7fa7fa6d6abf72938006f7930a6

### DIFF
--- a/upload/robots.txt
+++ b/upload/robots.txt
@@ -1,0 +1,15 @@
+
+Disallow: /*?page=$
+Disallow: /*&page=$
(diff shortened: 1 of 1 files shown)


ADDCreative:
As commented here https://github.com/opencart/opencart/issues/8043#issuecomment-662372780. This doesn't actually fix the root cause of the errors and may cause issues with already indexed content.

PR #8193, opened 2020-08-07
Update menu.twig

chongshengdz:
fix all category in one line and won't start a new line


danielkerr:
needs to use the theme css

danielkerr:
this would break many things

PR #8198, opened 2020-08-11
Feature: permission to view order statistics

Grumvol:


eka7a:
It would be better to check the `report/statistics` access permission.

Grumvol:
Yes you are right

PR #8199, opened 2020-08-11
Feature: permission to change cache

Grumvol:


eka7a:
It would be better to check the `common/developer` access permission.

PR #8211, opened 2020-08-11
Fix: old breakpoint and class xs

Grumvol:


eka7a:
After this pull, filters overlap on mobile devices on marketplace list.

marketplace

Grumvol:
I was wrong, now I will change

PR #8216, opened 2020-08-11
Fix: removed unnecessary calls trim

Grumvol:
Remove spaces in the class Request

	public function clean($data) {
		if (is_array($data)) {
			foreach ($data as $key => $value) {
				unset($data[$key]);

				$data[$this->clean($key)] = $this->clean($value);
			}
		} else {
			$data = trim(htmlspecialchars($data, ENT_COMPAT, 'UTF-8'));
		}

		return $data;
	}

### DIFF
--- a/upload/admin/controller/catalog/attribute.php
+++ b/upload/admin/controller/catalog/attribute.php
@@ -180,7 +180,7 @@ class ControllerCatalogAttribute extends Controller {
 				'edit'            => $this->url->link('catalog/attribute/edit', 'user_token=' . $this->session->data['user_token'] . '&attribute_id=' . $result['attribute_id'] . $url)
 			);
 		}
-		
+
 		if (isset($this->error['warning'])) {
 			$data['error_warning'] = $this->error['warning'];
 		} else {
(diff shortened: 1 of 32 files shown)


condor2:
Using trim in request will broke currency. If you add a space before symbol and save and edit again, the space is removed.

So you can't use something like this:

`4 Euro`

condor2:
When you remove trim from controllers. You can save using only white spaces. So is needed another method of warning.
Screenshot_2020-08-11_08-46-25

Grumvol:
You cannot specify a space as the name of an attribute group, there is validation

condor2:
There is no validation if you remove trim from controllers...
I added only spaces. Branch is cloned yesterday.

Grumvol:
It's strange. Trim is used in the clean method of the Request class. Check file system/library/request.php, there should be a line

    $data = trim(htmlspecialchars($data, ENT_COMPAT, 'UTF-8'));

condor2:
I know about trim in request but need to be removed untill currency symbol space is fixed.

Grumvol:
The space before the currency symbol needs to be implemented in a different way. Storing a space at the beginning of a line in the database is not a good idea.

PR #8225, opened 2020-08-15
Feature countries on off

Hkr32:
Ability to disable and enable multiple countries on the list page
feature-countries

### DIFF
--- a/upload/admin/controller/localisation/country.php
+++ b/upload/admin/controller/localisation/country.php
@@ -112,71 +112,71 @@ class ControllerLocalisationCountry extends Controller {
 
 	public function enable() {
 		$this->load->language('localisation/country');
-		
+
 		$this->document->setTitle($this->language->get('heading_title'));
-		
+
 		$this->load->model('localisation/country');
-		
+
 		if (isset($this->request->post['selected']) && $this->validateEnable()) {
 			$ids = implode(",", $this->request->post['selected']);
 			$this->model_localisation_country->enableCountries($ids);
-			
+
 			$this->session->data['success'] = $this->language->get('text_success');
-			
+
 			$url = '';
-			
+
 			if (isset($this->request->get['sort'])) {
 				$url .= '&sort=' . $this->request->get['sort'];
 			}
-			
+
 			if (isset($this->request->get['order'])) {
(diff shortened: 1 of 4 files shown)


upload/admin/controller/localisation/country.php
```diff
@@ -216,6 +289,7 @@ protected function getList() {
 		$data['sort_name'] = $this->url->link('localisation/country', 'user_token=' . $this->session->data['user_token'] . '&sort=name' . $url);
 		$data['sort_iso_code_2'] = $this->url->link('localisation/country', 'user_token=' . $this->session->data['user_token'] . '&sort=iso_code_2' . $url);
 		$data['sort_iso_code_3'] = $this->url->link('localisation/country', 'user_token=' . $this->session->data['user_token'] . '&sort=iso_code_3' . $url);
+		$data['sort_status'] = $this->url->link('localisation/country', 'user_token=' . $this->session->data['user_token'] . '&sort=status' . $url, true);
```
straightlight, line 292: This line won't work in the master branch: https://github.com/opencart/opencart/pull/8225/commits/a33476ab80cc10a7ec609360894fbce9b221e220#diff-ae87ef052a1157e79231def36b367e57R292 .

Hkr32:
Thanks! I have corrected. Now everything is ok?

straightlight:
> @straightlight, Thanks! I have corrected. Now everything is ok?

Not really. Your suggestion is merchant-based but not customer-based nor affiliate-based. Both startup/startup controllers needs to be edited as well since the config_country_id and config_zone_id are still applied from the default store settings. As for the payment and shipping extensions, the geo zones are also configured for each contries and zones. By disabling those options, neither party will be able to identify the root cause when customers will buy the goods from the checkout page.

In addition, disabling a geographic area needs to be informed to customers from the stores (information page). Without relying on payment service providers' API responses, it would then mean that merchants would take full responsibilities to inform customers as opposed to their service providers while, normally, service providers do collect the information by informing the merchants and customers regarding a transaction that occurred.

While there are already extensions on the Opencart.com Marketplace providing the possibility to restrict payment and shipping methods, it would be better to restrict than disable at this point.

danielkerr:
already added

Issue #7932, opened 2020-05-10
Error in FedEx Shipping module which causes FedEx not to display during checkout


D-whitten:
Opencart 3.x

Stock FedEx Shipping module has an error around line 104 in admin/view/template/extension/shipping/fedex.twig which causes FedEx shipping not to display during checkout

`   <input type="checkbox" name="fedex_service[]"`

needs to be changed to:

`   <input type="checkbox" name="shipping_fedex_service[]"`

danielkerr:
was removed in master branch

PR #8228, opened 2020-08-15
Fix missing shipping_ in FedEx

ADDCreative:


ADDCreative:
https://github.com/opencart/opencart/issues/7932

PR #8232, opened 2020-08-16
Fix: removed class *-12 (Mobile First)

Grumvol:


danielkerr:
rto old to merge and i and also altered the product list sizes already

PR #8252, opened 2020-08-17
Order Mail Language Fix

GautamMKGarg:
The language Editor option was not working for order emails. Hence fixed it.

### DIFF
--- a/upload/catalog/controller/mail/order.php
+++ b/upload/catalog/controller/mail/order.php
@@ -57,34 +57,32 @@ class ControllerMailOrder extends Controller {
 		}
 
 		// Load the language for any mails that might be required to be sent out
-		$language = new Language($order_info['language_code']);
-		$language->load($order_info['language_code']);
-		$language->load('mail/order_add');
+		$this->load->language('mail/order_add');
 
 		// HTML Mail
-		$data['title'] = sprintf($language->get('text_subject'), $order_info['store_name'], $order_info['order_id']);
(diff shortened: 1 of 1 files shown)


eka7a:
Your change is wrong. Order email language code must be called from the order db table. This change you made cancel this feature.

I tested the order mail system and didn't see any errors. Please check your mail settings are correct and your mail settings in event list. If your event list is out of date, update it.
https://github.com/opencart/opencart/blob/e24580b0a930ede34e29434a916d87c0066638db/upload/install/opencart.sql#L730-L807

GautamMKGarg:
I never mentioned that email was not working. I mentioned that the language editor option was not working for order emails.
In the admin panel, there is an option of "Language editor" in the Design section. I tried creating entry with the "mail/order_add" route and "text_greeting" key and it was not working. Then I did the following changes and it started working.

You can also refer other email controllers, in all of them $this->load->language is used to load the language, where as in order.php file, new Language is getting used to load the language, which is causing the issue and it is ignoring the changes I am doing in Language editor page. Kindly refer below to URLs for your reference.
https://github.com/opencart/opencart/blob/b8b9b4ea5607f61a65ce836176af58e9305980b6/upload/catalog/controller/mail/order.php#L61
https://github.com/opencart/opencart/blob/b8b9b4ea5607f61a65ce836176af58e9305980b6/upload/catalog/controller/mail/affiliate.php#L5

PR #8255, opened 2020-08-18
Update slideshow.php

MadMcJack:
Use minified version (also resolves missing js/maps/swiper.jquery.js.map error in javascript console)


danielkerr:
removed ages ago

PR #8264, opened 2020-08-24
Fix: removed unused method getTotalCategoriesByCategoryId

Grumvol:


rafaeldph:
It's not used in base code, but extensions could need it and it would be fine if OpenCart include that method in base code.

PR #8265, opened 2020-08-24
Refactor cart query

rafaeldph:
It makes only one product query on cart->getProducts() and adds it to an array where the key is the product_id and the value is the product query. Works the same way as before.


straightlight:
Since the catalog/model/catalog/product.php can also be used to keep the queries from getProducts() of the cart library, it would be more simplified rather than using $cart['product_id'] in the process.

rafaeldph:
Models are not loaded from libraries. That would be a major change.

straightlight:
Too complexed. Queries should still be pulled from the catalog model instead.

PR #8313, opened 2020-09-19
Added get_path method in system/helper/general.

straightlight:


straightlight:
Perhaps the DirectoryIterator should be used to compare the entered parameters as used in this install model upgrade file already does: upload/install/model/upgrade/upgrade_1006.php . A helper file should be specifically be used to handle those requests whether it's a directory or a file by two separate function names.

eka7a:
Please check conflicts

straightlight:
Not much I can do for conflicts if I am not posting on my own branch. As for the DIR_IMAGE, that is correct, which is why I posted this here: https://github.com/opencart/opencart/issues/8441#issuecomment-695815726 .

PR #8319, opened 2020-09-19
Calling models from library cart\cart - system/library/cart/cart.php file.

straightlight:


eka7a:
This also calls many unnecessary db tables.

PR #8424, opened 2020-09-20
Added integer to GET address ID request - catalog/controller/account/address.php file.

straightlight:


straightlight:
Closing commit. These validations are still required to avoid security issues.

PR #8442, opened 2020-09-21
change if else to ternary in guest checkout

stickpro:
An unpleasant bug that adds spaces in value, which is why the placeholder and additional user fields are not displayed. Just replaced with ternary operator


straightlight:
Yes, good point. There are other locations using ternary in the TWIG files on anyhow. Using IF and ternary on different locations could be confusing for store owners.

PR #8476, opened 2020-09-23
Added backward escaped quotes - admin/model/localisation/weight_class.php file.

straightlight:


upload/admin/model/localisation/weight_class.php
```diff
@@ -47,7 +47,7 @@ public function getWeightClasses($data = []) {
 			if (isset($data['sort']) && in_array($data['sort'], $sort_data)) {
 				$sql .= " ORDER BY " . $data['sort'];
 			} else {
-				$sql .= " ORDER BY title";
+				$sql .= " ORDER BY `title`";
```
straightlight, line 50: This field should include the field's prefix related to the title being pulled.

PR #8478, opened 2020-09-23
Added backward escaped quotes - admin/model/marketing/affiliate.php file.

straightlight:


upload/admin/model/marketing/affiliate.php
```diff
@@ -66,7 +66,7 @@ public function getAffiliates($data = []) {
 		if (isset($data['sort']) && in_array($data['sort'], $sort_data)) {
 			$sql .= " ORDER BY " . $data['sort'];
 		} else {
-			$sql .= " ORDER BY name";
+			$sql .= " ORDER BY `name`";
```
straightlight, line 69: May require the relative field prefix name.

PR #8512, opened 2020-09-24
Added backward escaped quotes - catalog/model/checkout/order.php file.

straightlight:


straightlight:
These methods already follows the same direction as this PR: https://github.com/opencart/opencart/pull/8319/commits/1073c43cb2ab99acbbd54cc0d231d6503d8025d3 .

PR #8539, opened 2020-09-25
Added backward escaped quotes - extension/opencart/admin/model/dashboard/chart.php file.

straightlight:


straightlight:
OR statement should be used rather than IN statement in the WHERE clauses.

PR #8540, opened 2020-09-25
Added backward escaped quotes - extension/opencart/admin/model/dashboard/map.php

straightlight:


straightlight:
OR statement should be used instead of IN statement in the WHERE clause.

PR #8564, opened 2020-09-25
extension/opencart/catalog/model/total/voucher.php - extension/opencart/catalog/model/total/voucher.php file.

straightlight:


straightlight:
OR statement should be used instead of IN statement in WHERE clause.

straightlight:
Task completed.

PR #8570, opened 2020-09-27
This file should be deleted from the master - admin/controller/event/debug.php file.

straightlight:


danielkerr:
its for testing. in case anyone wants to use it.

PR #8608, opened 2020-09-30
I have noticed that there are no zone areas for Guernsey and Jersey, could that be the problem? 

lakha:
Guernsey and Jersey Country dont have Region/State list so its giving error on Royal Mail how can i solve this issue?

its not working for this country only other country if i will select then shipping working fine for royal mail. And i added new Region/State Luis Alves then also not working for Jersey Country why its not working if having Region/State Select?

Waiting for your reply
shipping_issue


ADDCreative:
@lakha Please close this pull request as you are not making any code changes.

The issue is likely to be that the Royal Mail module doesn't check for ISO codes GG or JE, only GB.
https://github.com/opencart/opencart/blob/961c6f59b0fade37f7866724e0e0de9ec829b6cd/upload/catalog/model/extension/shipping/royal_mail.php#L264

You are probably best asking on the forums. https://forum.opencart.com/

PR #8630, opened 2020-10-04
Fixed a bug where after config files were created in step 2 of installa…

Boorinio:
…tion, controller would always redirect to upgrade form.

### DIFF
--- a/upload/install/controller/startup/upgrade.php
+++ b/upload/install/controller/startup/upgrade.php
@@ -9,9 +9,9 @@ class Upgrade extends \Opencart\System\Engine\Controller {
 		}
 		
 		if (isset($this->request->get['route'])) {
-			if (($this->request->get['route'] == 'install/step_4') || (substr($this->request->get['route'], 0, 8) == 'upgrade/') || (substr($this->request->get['route'], 0, 10) == '3rd_party/')) {
-				$upgrade = false;
-			}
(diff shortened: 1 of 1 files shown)


straightlight:
Perhaps utf-8 functions with substr should also be used above to see if that would replaced the alternative solution ...

straightlight:
The closing brackets of methods don't need to contain extra spaces. As for the config pagination, it may still be questionable due to the limitations of the autocomplete results. As for the arrays, that would be right.

Boorinio:
> Perhaps utf-8 functions with substr should also be used above to see if that would replaced the alternative solution ...

is there a reason for utf substr functions since the urls of installation are always English characters?

straightlight:
While it is correct that only the en-gb is installed by default, there are also language packs where the install folder are implicit.

osworx:
> is there a reason for utf substr functions since the urls of installation are always English characters?
Not correct!
For example German, can be used during the installation to install OpenCart in German. It is a second package and needed for German speaking people don't speaking English.
Note: the World exists of more not only English ..

Boorinio:
Correct me if I am wrong but even at other languages the controller urls do not change. The folder remains "install" and the controller remains as step_1 etc. You only add more folders in language folder just like catalog or admin works.

PR #8642, opened 2020-10-06
Warning: A non-numeric value encountered php7.3

WebkulOpencart:
'Warning: A non-numeric value encountered php7.3 when generating the shipping note form admin end for the order


ADDCreative:
@WebkulOpencart Please revert this change. It was already fixed with https://github.com/opencart/opencart/commit/ebabafc6fa9bcf12da603475ab2bd009077e3b37. Your change is actually wrong and will now only use the weight of the last option not all of them.

WebkulOpencart:
@ADDCreative I reverted the changes,

Issue #8611, opened 2020-10-01
Fedex Error


baltechies:
**What version of OpenCart are you reporting this for?**
3.0.3.6

**Describe the bug**
<b>Warning</b>: in_array() expects parameter 2 to be array, null given in <b>/var/www/html/catalog/model/extension/shipping/fedex.php</b> on line <b>180</b

What section does it affect?
Shipping methods not working.

**To Reproduce**
1. Setup shipping method
2. Add product to cart and Get Quote.

**Expected behavior**
Should be estimated cost for shipping methods

WebkulOpencart:
issues resolved so clo

PR #8643, opened 2020-10-06
resolved Fedex Error #8611

webkul-dilip:
https://github.com/opencart/opencart/issues/8611 resolved the issue

### DIFF
--- a/upload/catalog/model/extension/shipping/fedex.php
+++ b/upload/catalog/model/extension/shipping/fedex.php
@@ -157,7 +157,7 @@ class ModelExtensionShippingFedex extends Model {
 			curl_setopt($curl, CURLOPT_SSL_VERIFYPEER, false);
 
 			$response = curl_exec($curl);
-			
+
 			curl_close($curl);
 
 			$dom = new DOMDocument('1.0', 'UTF-8');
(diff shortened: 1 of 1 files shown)


webkul-dilip:
Resolved the issue #8611

upload/catalog/model/extension/shipping/fedex.php
```diff
@@ -177,11 +177,11 @@ function getQuote($address) {
 				foreach ($rate_reply_details as $rate_reply_detail) {
 					$code = strtolower($rate_reply_detail->getElementsByTagName('ServiceType')->item(0)->nodeValue);
 
-					if (in_array(strtoupper($code), $this->config->get('shipping_fedex_service'))) {
+						if (is_array($this->config->get('shipping_fedex_service')) && in_array(strtoupper($code), $this->config->get('shipping_fedex_service'))) {
```
straightlight, line 180: ```if (!empty($this->config->get('shipping_fedex_service')) && is_array($this->config->get('shipping_fedex_service')) && in_array(strtoupper($code), $this->config->get('shipping_fedex_service'))) {```

webkul-dilip: Code added

straightlight: Take note that is_array does not need to be used since the new restructuration of array instantiation in PHP 7 as well as the conversion in Opencart to no longer use (array) but [] instead. By removing the is_array validation, you could still validate the $this->config->get('shipping_fedex_service') as an array in the in_array function instead. This will shorten the codes a bit.

PR #8648, opened 2020-10-06
Update stream_socket_enable_crypto(..) when php >= 7.4.x  used

osworx:
using SMP and tls:// results in en error:
PHP Warning:  stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:
error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed in ../system/library/mail.php on line xxx (xx = line depends on used OpenCart Version - this is taken from a 2.x installation).

Regarding php.net: https://www.php.net/manual/de/function.stream-socket-enable-crypto.php#119122
the usage of STREAM_CRYPTO_METHOD_SSLv23_CLIENT instead of STREAM_CRYPTO_METHOD_TLS_CLIENT should be used.

Tested already and is working

### DIFF
--- a/upload/system/library/mail/smtp.php
+++ b/upload/system/library/mail/smtp.php
@@ -126,7 +126,7 @@ class Smtp {
 
 				$this->handleReply($handle, 220, 'Error: STARTTLS not accepted from server!');
 
-				stream_socket_enable_crypto($handle, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
+				stream_socket_enable_crypto($handle, true, STREAM_CRYPTO_METHOD_SSLv23_CLIENT);
 			}
 
 			if (!empty($this->smtp_username) && !empty($this->smtp_password)) {


straightlight:
I am typing 7.4 in the search of the browser and I cannot find any relative match for it. Changing this case may also rely on the service provider you are using as other services may not provide the same options by default since php.net also addresses:

```Returns TRUE on success, FALSE if negotiation has failed or 0 if there isn't enough data and you should try again (only for non-blocking sockets).```

osworx:
Typing 7.4 and search for it is not enough.
You must try it by yourself and then you will see that Emails are not going out when SMTP and the prefix tls:// is used.

As stated here:

> Now, be careful because since PHP 5.6.7, STREAM_CRYPTO_METHOD_TLS_CLIENT (same for _SERVER) no longer means any tls version but tls 1.0 only (for "backward compatibility"...).

osworx:
Can be closed, needs further invastigation - sorry.

Asif-Mahmood:
have fixed this issue in pull request #10085  . It was due to the fact that PHP was using TLS 1.0 by default which is to man in the middle attack and blocked by many mail hosts.

PR #8661, opened 2020-10-09
fixed store location listing issues

prabhatkumaroc082:
1) Admin panel added store locations are not listing but saving in DB
2) locations are not listing in front-end


eka7a:
All locations will be listed with your change on contact page. Only selected locations should be listed on setting page.

straightlight:
> All locations will be listed with your change on contact page. Only selected locations should be listed on setting page.

The scenario of this controller is that $data['location'][] key term is being used as singular. While specific locations needs to be loaded, the array is still being incremented into the loop. Extracting the singular term array from the target TWIG file will only bring more confusion as opposed to use: ```{{ for (location in locations) }}```

prabhatkumaroc082:
Hello @straightlight 
yes, no need for any frontend changes for this because locations will be selected by the store owner. admin side listing has a problem. $data['location'] must be changed to $data['locations']. added a new commit to remove all frontend changes.
Thanks,

PR #8664, opened 2020-10-09
Fix wrong MySQLi connect error string

ADDCreative:


condor2:
You miss this line

`if (!$this->connection->errno) {`

condor2:
This line should not be changed?

`if (!$this->connection->errno) {`

Or first command is mysqli_errno and second mysqli_connect_error?

ADDCreative:
No, connect_error and connect_errno are for connection errors. error and errno are for query errors.

PR #8670, opened 2020-10-09
getApiByToken command fix

condor2:


straightlight:
Yes, correct. Good catch.

PR #8672, opened 2020-10-09
Same website URL for IP check

condor2:


ADDCreative:
Should be ```https://``` not ```//```.

condor2:
When you use // will auto detect if is https or http so in the end url will be http:// or https://depending on the case.

ADDCreative:
I know exactly what it does and that is actually the reason for not using ```//```, It's a external link. You should not base the protocol of an external link on the local protocol.

condor2:
Changed...
https://github.com/opencart/opencart/pull/8682

PR #8678, opened 2020-10-10
[3.0.x.x_] Fixed Api upgrade error from 2.3.0.2

condor2:
When you try to upgrade the store from Opencart 2.3.0.2 to 3.0.3.6 in database is created automatically username field taken from opencart.sql.
And when upgrade controller read command from this file will give duplicate for username.

With this fix first username (added by opencart.sql) will be deleted then name will be renamed to username.

### DIFF
--- a/upload/install/model/upgrade/1009.php
+++ b/upload/install/model/upgrade/1009.php
@@ -66,11 +66,14 @@ class ModelUpgrade1009 extends Model {
 			
 			$this->db->query("DROP TABLE `" . DB_PREFIX . "affiliate_transaction`");
 		}
-	
+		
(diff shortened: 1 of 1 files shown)


condor2:
https://forum.opencart.com/viewtopic.php?f=199&t=219496

straightlight:
This process was already eliminated on the master branch a few months back. It was said on the commit's comment page by Daniel that users needs to figure it out by themselves on this one.

condor2:
This is for 3.0.x.x branch..and right now I don't care about Master Branch.
Why users need to figure out...when this fix work just fine...

PR #8679, opened 2020-10-10
[3.0.x.x_Maintenance] common/home seo url fix

condor2:
https://forum.opencart.com/viewtopic.php?p=794263#p794263

### DIFF
--- a/upload/catalog/controller/startup/seo_url.php
+++ b/upload/catalog/controller/startup/seo_url.php
@@ -84,6 +84,10 @@ class ControllerStartupSeoUrl extends Controller {
 
 						unset($data[$key]);
 					}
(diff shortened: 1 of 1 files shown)


ADDCreative:
Line spacing is wrong. See: https://github.com/opencart/opencart/wiki/Coding-Standards#new-lines

Anyway a better way may be changing the following, as path will never be valid for ```common/home``` and should be ignored.

Form.
```php
				} elseif ($key == 'path') {
```
To.
```php
				} elseif ($data['route'] == 'common/home') {
					$url .= '/';

					break;
				} elseif ($key == 'path') {
```

condor2:
I have changed

PR #8682, opened 2020-10-10
Added https protocol

condor2:


straightlight:
There are more than a single entry about this provider. There's also an older provider being used which also needs to be renamed into this one along to the addition of HTTPS. That would entirely fix the issue.

condor2:
Yes. The other entry is in online
https://github.com/opencart/opencart/blob/master/upload/admin/view/template/report/online.twig#L54

PR #8684, opened 2020-10-10
Fixed a bug where a db connection would close before session.

Boorinio:
The error

Or clearer

 
What's going on: 
On a previous commit in mysqli deconstructors were ditched in favor of register_shutdown_function. The problem is register_shutdown_function runs the functions in the order they were registered. There is one more register_shutdown_function in session (db) which causes in this case, to first close the db connection and then try to write the session in db causing the error.

This pull request adds close function in DB parent class and register the register_shutdown_function after session is loaded.


Boorinio:
Bump! This bug still exists, someone commented register shutdown function and replaced (again) with destructor in mysqli class. Problem is mysqli destructor is never called (that's why I think it was removed in the first place)

PR #8689, opened 2020-10-12
Update composer.json php minimum version

gnovaro:
According to the release the minimum php version now is 7.3
https://github.com/opencart/opencart/releases/tag/3.0.3.5


gnovaro:
Hi @danielkerr I think this is a simple change.

PR #8699, opened 2020-10-14
Added COVID-19 notes once the upgrade completes - install/upgrade/en-gb/upgrade/upgrade.php file.

straightlight:
Due to the COVID-19 situation, I thought to bring the importance to remind store owners to expand their OC admin > catalog > information > Terms and Agreements form by informing their customers due to restrictions that may be implied throughout geographic areas when the goods are being delivered.


straightlight:
The reason why this sentence has been added is because, out-of-the-box, store owners nor customers are being typically informed while either saling or purchashing the goods during checkout whereas terms and conditions needs to be check marked by customers in order to proceed to the next step. I think it is more than fair to state that due to the situation we're currently facing with the level of severity of the pandemic that this short note should definitely be added in the upgrade process while, unfortunately, there are no distinctive ways to add such notifications in the fresh install without altering the oc_information_description table where language barriers could also be involved in the process.

While I understand that many websites already do have a link on the site header, it does not disclose nor inform any types of restrictions where customers already needs to agree during checkout, or from the footer links, based on geographical restrictions settings that may be implied by payment or shipping providers.

Now, as for the sentence itself, if you believe it doesn't make sense, I am opened for suggestions on how else this should be addressed as long as it is conformed on the subject based on the commit I have created.

Issue #8669, opened 2020-10-09
[Master Branch] Undefined index: store_id admin/model/customer/customer.php on line 5


condor2:
**What version of OpenCart are you reporting this for?**
Master Branch

**Describe the bug**
A warning message ins Error Log

**To Reproduce**
I have added a customer account from Admin.

straightlight:
I have no idea why, between line 382 and line 388, the store array is not located outside of the foreach statement. We don't need to loop a reset array of the first store value.

prabhatkumaroc082:
@straightlight 
line 382 and 388 are for customer listing which is never used. please check admin/controller/customer/customer.php line 660, it's an array for store dropdown for the multistore case only so that store_id value missing as per twig code.

it could be the fix #8706

straightlight:
@prabhatkumaroc082 : That fix wouldn't do it. The store ID should be validated among the controller, not the model. The store ID of 0 can also be problematic with multi-stores for people using extensions, that would also create a problem. As for the language_id, it wouldn't be right either since the language ID of the customer can't simply be the current language ID of the store based on what the admin user is currently using from the browser. We had enough to worry about for the orders based on the relative languages especially from the internal APIs until it was apparently fixed on the master branch. We don't want to create additional problems with customer / affiliate accounts.

prabhatkumaroc082:
@straightlight 
**The store ID of 0 can also be problematic with multi-stores for people using extensions**

multi store dropdown is already there on twig
{% if multistore > 0 %}
                  <div class="form-group">
                   <label class="col-sm-2 control-label" for="input-store">{{ entry_store }}</label>
                   <div class="col-sm-10">
                     <select name="store_id" id="input-store" class="form-control">
                       {% for store in stores %}
                         {% if store.store_id == store_id %}
                           <option value="{{ store.store_id }}" selected="selected">{{ store.name }}</option>
                         {% else %}
                           <option value="{{ store.store_id }}">{{ store.name }}</option>
                         {% endif %}
                       {% endfor %}
                     </select>
                    </div>
                  </div>
		        {% endif %}

else part could be just added here instead of model changes for store_id 0 in PR 
{% else %}
<input type="hidden" name="store_id" value="0" />
{% endif %}

PR #8706, opened 2020-10-15
[Master Branch] added missing values for customer model

prabhatkumaroc082:
#8669 fixes
errors:

**Notice: Undefined index: store_id in line 5**
**Error: Field 'language_id' doesn't have a default value**

1) if admin doesn't have multistore then admin/view/template/customer/customer_form.twig line 40 not having any value
2) language_id is also missing in the admin/model/customer/customer.php model for both add/edit case


danielkerr:
theres no reason they should be missing.

PR #8714, opened 2020-10-15
Fix PHP Notice Trying to access array offset on value of type null

gnovaro:
PHP Notice:  Trying to access array offset on value of type null in /admin/controller/marketplace/marketplace.php on line 1011
PHP Notice:  Trying to access array offset on value of type null in /admin/controller/marketplace/marketplace.php on line 1013

### DIFF
--- a/upload/admin/controller/marketplace/marketplace.php
+++ b/upload/admin/controller/marketplace/marketplace.php
@@ -30,7 +30,7 @@ class Marketplace extends \Opencart\System\Engine\Controller {
 			$filter_rating = '';
 		}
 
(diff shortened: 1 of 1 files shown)


upload/admin/controller/marketplace/marketplace.php
```diff
@@ -30,7 +30,7 @@ public function index() {
 			$filter_rating = '';
 		}
 
-		if (isset($this->request->get['filter_member_type'])) {
+		if (isset($this->request->get['filter_member_type'])) {c
```
condor2, line 33: remove that **c**

Issue #7450, opened 2019-05-21
catalog/model/tool/image.php - resize vs case sensitivity on Windows


SelfMan:
Problem is present in version 3.0.3.1 and also in the current trunk (2019-05-21)

The first check does have case sensitivity problems where the whole method exits and no images are present on the site.
		if (!is_file(DIR_IMAGE . $filename) || substr(str_replace('\\', '/', realpath(DIR_IMAGE . $filename)), 0, strlen(DIR_IMAGE)) != str_replace('\\', '/', DIR_IMAGE)) {
			return;
		}
The issue can be easily reproduced. Remember windows is not case sensitive.
The site is located at D:\WWW\SomeSite.local but the DIR_IMAGE in config.php is set to
define('DIR_IMAGE', 'D:/www/somesite.local/image/');

The realpath() wont match the DIR_IMAGE because of case sensitivity even though on windows it does not matter.

The result of the problem is that when you load the site, there will be no error message nor a entry in the error log file, but the product images will be missing and the File manager will misbehave too.

 - Local development
 - Windows 7
 - PHP version 7.1.5
 - Apache version 2.4
 - Browser independant

straightlight:
See if this fix resolves the issue: https://forum.opencart.com/viewtopic.php?f=190&t=164152#p625218 . Some tweaking to be done for v3.x releases but the case sensitivity fix should be included.

SelfMan:
Hi strightlight,
That might solve some problems with new installations/new files but it does not help when you are upgrading a older OC installation and all the references are already in the database.
In my case it all happened after taking a installation from a linux server to my local developemnt enviroment, which is windows based. I was wondering why I dont see any images till I found the issue.

ADDCreative:
There is a similar issue with Windows path. https://github.com/opencart/opencart/issues/7080

Changing the test to the code below should hep.
```php
if (!is_file(DIR_IMAGE . $filename) || str_replace('\\', '/', realpath(DIR_IMAGE) . '/' . $filename) !== str_replace('\\', '/', realpath(DIR_IMAGE . $filename))) {
```

SelfMan:
@ADDCreative the solution you provided seem to be working just fine. So far I've seen, there are no side effects. Thank you.
P.S.> i've seen the #7080 but though that ithe whole problem is about the realpath() not returning the proper value on that particular server and I didn't read it to the end. My fault.

Issue #7080, opened 2018-12-06
Image path replaces


pekka2:
admin/model/tool/image.php line 4:

`if (!is_file(DIR_IMAGE . $filename) || substr(str_replace('\\', '/', realpath(DIR_IMAGE . $filename)), 0, strlen(DIR_IMAGE)) != str_replace('\\', '/', DIR_IMAGE)) {`

These replace scripts 

`str_replace('\\', '/', realpath(DIR_IMAGE . $filename))`

`str_replace('\\', '/', DIR_IMAGE)`

are completely unnecessary(?)

Path `define('DIR_IMAGE', 'C:\MAMP\htdocs\3020/image/') ` work successfully. But current replace scripts at files image.php and filemanager.php is not work all servers.

MrPlow254:
In Linux, the path separator is /. In Windows, it is either \ or /. So to be universal it needs to use forward slashes.

pekka2:
There is no question in this chain, what is universal script, but works with Opencart scripts.
Latest Openacrt version, where image path scripts work all servers, was Opencart 2.2

MrPlow254:
@pekka2 Can you rephrase that please? I am not following what the issue is.

ADDCreative:
The path checks have probably been added for a good reason. It's probably best not to remove them.

Your issue could be caused by some sort of symbolic link, which cause realpath a different path to your config. See https://github.com/copona/copona/issues/128

You could try something like.
```php
if (!is_file(DIR_IMAGE . $filename) || str_replace('\\', '/', realpath(DIR_IMAGE) . '/' . $filename) !== str_replace('\\'. '/', realpath(DIR_IMAGE . $filename))) {
```

pekka2:
> It's probably best not to remove them.

I am removed it few server, where it has not worked :)

pekka2:
DIR_IMAGE is Ok, realpath is not tested.

ADDCreative:
Are you able to test ```realpath(DIR_IMAGE);```?

pekka2:
Path is sulfur, path begins at document root. Missing diectories '/sites/domain.nl/'

osworx:
@pekka2 
1. it is sometimes very hard to understand your english
2. it would be more useful if you would write more about the used (and failed) environement (which server, which php-version, etc.)
3, if explicite asked, a straight answer would help more than answering scriptic .. (e.g. what is DIR_IMAGE set)
thx!

ADDCreative:
[...]

PR #8718, opened 2020-10-16
Added utf8 to strlen - admin/controller/design/theme.php file.

straightlight:


ADDCreative:
This is wrong. substr will expect the length in single-byte characters not multi-byte characters.

straightlight:
Tell me about it. The original function that checks for our server paths everywhere in OC uses realpath with utf8_strlen starting with substr instead of utf8_substr . Imagine how can one find this confusing!

Which is why, I mentioned to this commit to create a get_path helper function to avoid confusion like these in the future, since - clearly - we don't need 50000 of different ways to track a folder. IT'S JUST A FOLDER !

https://github.com/opencart/opencart/commit/d7a8fa40a45966bdcb63e4211ef6e82f721fc5bf

ADDCreative:
I would stick to using substr and strlen for the path checks.

straightlight:
I wouldn't.

ADDCreative:
Well at least close or correct this pull request as mixing them will break things if the path did contain a multi-byte character.

eka7a:
https://github.com/opencart/opencart/blob/65d7040c4bf30a70980ffda44a3ca1ad14b66660/upload/system/helper/general.php#L146-L148

I guess it can be more useful this way.

```
function get_path($source, $dir, $utf8 = false) {
	if ($utf8) {
		return utf8_substr(str_replace('\\', '/', realpath($source)), 0, utf8_strlen($dir));
	} else {
		return substr(str_replace('\\', '/', realpath($source)), 0, strlen($dir));
	}
}
```

straightlight:
UTF-8 should be set to true by default rather than false since OC is mainly developed on UTF-8 formatting. Otherwise, a bunch of true case might be showing during validations as opposed to false case.

ADDCreative:
You just need the following. The result will be the same whether the strings use UTF-8 or not. Test it for yourselves if you don't understand. Using utf8_substr and utf8_strlen will just be less efficient.

```php
function get_path($source, $dir) {
	return substr(str_replace('\\', '/', realpath($source)), 0, strlen($dir));
}
```

What is wrong is this pull request which mixes substr with utf8_strlen. This will break things if the path did contain a multi-byte character.

straightlight:
```What is wrong is this pull request which mixes substr with utf8_strlen.```

I would have to differ on this part, since the original codes already contains utf8 in them as opposed to the ended function of the statement which does not have a utf8 in it while it could or shouldn't.

**Comment by ADDCreative** [comment]
I'll try to explain it for you. Take your first change in this pull request.
```php
if (substr(str_replace('\\', '/', realpath(DIR_EXTENSION . $extension)), 0, utf8_strlen(DIR_EXTENSION)) != DIR_EXTENSION) {
	$safe = false;
}
```

Now, if DIR_EXTENSION was a path that did contain a multi-byte character such as ```/ápath/ext/```. ```utf8_strlen(DIR_EXTENSION)``` would return 11 and not 12. This would mean that ```substr``` would return ```/ápath/ext```, which wouldn't match the comparison with DIR_EXTENSION and therefore incorrectly set ```$safe``` to false.

Sticking with ```strlen(DIR_EXTENSION)```, which would return 12. This would mean that ```substr``` would return ```/ápath/ext/```, which would correctly match the comparison with DIR_EXTENSION.

If you have seen any other places in the code where ```substr``` is used with ```utf8_strlen``` then these are also likely wrong and would need fixing.

straightlight:
```If you have seen any other places in the code where substr is used with utf8_strlen then these are also likely wrong and would need fixing.```

This is what I am saying as well, mixes of those 50000 ways of tracking a folder in Opencart are being used which is why we need to put a stop to this and build an official helper function (or may be two if exceptional cases would appeal) to keep a more flexible folder tracking mechanism.

**Comment by ADDCreative** [comment]
If that is the case then please close this pull request. As I have explained it's incorrect and will just cause more confusion.

If you need a helper function to check paths then something clearer and more robust like the following might be a good idea.
```php
function is_path_safe($base, $path) {
	$real_base = realpath($base);
	$real_path = realpath($base . $path);

	if (!$real_base || !$real_path) {
		return false;
	}

	if (str_replace('\\', '/', $real_base . '/' . $path) !== str_replace('\\', '/', $real_path)) {
		return false;
	}

	return true;
}
```

Then use it as follows.
```php
if (!is_path_safe(DIR_EXTENSION, $extension) {
	$safe = false;
}
```

It would also help with issues https://github.com/opencart/opencart/issues/7080 and https://github.com/opencart/opencart/issues/7450.

straightlight:
While there's no need to close the commit before confirmation of the best option, the solution above provides a way to validate the folder structure but may also prevent forum supporters to track down the root cause for OP users on the OC forum. Therefore, eka7a's solution would be the right one in this case, since the error triggers still needs to show errors whenever OP users are committing one or if there's a 'true' encountered issue unexpected from the core.

ADDCreative:
You don't understand. Please read my https://github.com/opencart/opencart/pull/8718#issuecomment-728902719 again. This explains why this pull request will not work. If you you still don't understand, I would be happy to try and explain it more clearly for you,

straightlight:
I've seen that post. It will prevent the error trigger to launch whenever there's an invalid path as well. We need those triggers. The first solution is enough in case there's an invalid path. If it needs to be done with UTF-8, we'll see from there.

ADDCreative:
You haven't read that post at all or you haven't understood it. It doesn't suggest any changes just points out the problem with this pull request. I'll posts the contents again below. Please take time to read and understand.

I'll try to explain it for you. Take your first change in this pull request.
```php
if (substr(str_replace('\\', '/', realpath(DIR_EXTENSION . $extension)), 0, utf8_strlen(DIR_EXTENSION)) != DIR_EXTENSION) {
	$safe = false;
}
```

Now, if DIR_EXTENSION was a path that did contain a multi-byte character such as ```/ápath/ext/```. ```utf8_strlen(DIR_EXTENSION)``` would return 11 and not 12. This would mean that ```substr``` would return ```/ápath/ext```, which wouldn't match the comparison with DIR_EXTENSION and therefore incorrectly set ```$safe``` to false.

Sticking with ```strlen(DIR_EXTENSION)```, which would return 12. This would mean that ```substr``` would return ```/ápath/ext/```, which would correctly match the comparison with DIR_EXTENSION.

If you have seen any other places in the code where ```substr``` is used with ```utf8_strlen``` then these are also likely wrong and would need fixing.

straightlight:
Yeah, yeah, I read it. It's already mixed in the core and which is why I submitted a pull request about it so to remove the mix. I am already aware of it. Helper function with or without UTF-8 needs to be determined.

ADDCreative:
Then please look at this pull request again and close as it adds more mixes not removes them.

straightlight:
No need to close it. I am not requesting a mix but a solution. Just let it go, instead. It will reduce your own headaches eventually.

ADDCreative:
Look at you pull request again. You are requesting a mix.

PR #8729, opened 2020-10-16
Added html_entity_decode on string filters - admin/controller/marketing/affiliate.php file.

straightlight:

PR #8738, opened 2020-10-16
Added html_entity_decode to customer and Order St - admin/controller/sale/order.php file.

straightlight:


ADDCreative:
This add XSS vulnerabilities.

straightlight:
If so, then this doesn't explain the bunch of urlencode with html_entity_decode when using those filter_customer inputs.

straightlight:
Also adding the fact that filter_customer inputs don't all use the html_entity_decode . If the server can already handle those from php.ini, or .user.ini file, then those functions shouldn't be there other than encoding the inputs alone.

ADDCreative:
It would be safe it you were using urlencode after, but you're not. filter_order_status does not need decoding anyway, as it's just a list of numbers.

Also it has absolutely nothing to do with any setting in php.ini or .user.ini.

straightlight:
Regarding the urlencode, yes, I am aware. However, by adding it after, when making the call to the models, obviously, the input would need to be url decoded before querying the request on anyhow which would be more headache for developers to, either, troubleshoot the core or their extensions.

filter_order_status is an entered string. filter_order_status_id is a selected number with the exception of 'selected' array key which is also sanitized as (array) in front already on the master branch.

ADDCreative:
An demonstration of the XSS vulnerability added by this pull request would be.

```/admin/index.php?route=sale/order&user_token=xxx&filter_order_status="><script>alert(123)</script>```

straightlight:
If the case, then using filters should not be used by GET method throughout a URL, at least not in the admin, but rather as POST method since these are caused by the fact HTML forms are being used and using GET methods might not be the safest path. Besides, filters should be added as an event since it would avoid the mount of repetitive lookups within the controllers as the events would be a unique point of location to edit those filters and propose further PRs when necessary in the future.

ADDCreative:
No, the issue is caused by you adding html_entity_decode in places where you shouldn't. This is not the first time you have suggested it. Which, along with your response, gives the impression you are not understanding where and why it should be used.

You do NOT want to html_entity_decode any request data that is going to be output in an HTML context, unless you are encoding in another safe way.

straightlight:
Regardless the mount of times, other controllers also implies these functions with the filters. If you don't agree with those functions being put in, since it is inconclusive at this time, simply create new PRs and request for those functions to be removed.

ADDCreative:
Nothing is inconclusive, you have just not understood. There is no problem with the existing code, just this and the other https://github.com/opencart/opencart/pull/8729 pull request by you,

straightlight:
Simply create new PRs for all controllers containing this function where you'd want them removed..

ADDCreative:
As I said everything in OK as it is. You just need to close this and https://github.com/opencart/opencart/pull/8729.

straightlight:
There are also original codes that contains the html_entity_decode with these variable names for the same purpose. Until being removed, your theory remains unconfirmed. If you ever decide to create your own pull requests about them, feel free to do so.

ADDCreative:
A simple test for you to confirm that this pull request adds XSS vulnerabilities.

1. Implement the changes in this pull request.
2. Login to admin.
3. Go to Sales -> Orders.
3. Add ```&filter_order_status="><script>alert(123)</script>``` to the end of the URL in your web browser's address bar and go to that new URL.

You will see an alert box pop up, thus proving the XSS vulnerability.

If you did the same with out making the changes you suggest then there will be no pop up and no XSS vulnerability.

straightlight:
Then, post your suggestions on your own pull requests. All users can anyways. No need to flood your replies on my commits.

ADDCreative:
Nothing needs changing, the original code is correct as it is. It's your pull requests that add the XSS vulnerabilities. Which is why you could save everybody some time and just close them.

Just because you have seen a variable encode or decode somewhere else doesn't mean it should be done everywhere. You need the considered the context in which it is used. Other wise you will end up with vulnerabilities in the code.

danielkerr:
yep u are gpoijg over board.

PR #8744, opened 2020-10-16
Added url_encode on filter_customer - admin/controller/sale/recurring.php file.

straightlight:


ADDCreative:
Have you tested this? As the customer name would of been encoded before storing in the database, decoding before the query will not return the correct results if the filter contains special characters.

straightlight:
Already answered here: https://github.com/opencart/opencart/pull/8738#issuecomment-709997188 .

ADDCreative:
That is a completely different problem.

straightlight:
Let's not forget that the model also contains (string) in front of their variables when validating string types on database queries. I have already addressed this issue to Daniel before but those (string) types seem to remain there unfortunately.

ADDCreative:
Typecasting to (string) will not make any difference and is nothing to do with the problem.

PR #8750, opened 2020-10-16
Added trim to city - catalog/controller/api/payment.php file.

straightlight:


ADDCreative:
Again, trim is either unnecessary or the trim before the greater than comparison could lead to loss of the end of the string that had leading spaces.

Alos, not related to the pull request, bu that line in general, is 32 the correct size for city?
https://github.com/opencart/opencart/blob/4fc9627cf813e11e09b2baf3bc2ad9f1141a0b43/upload/system/helper/db_schema.php#L45-L47

straightlight:
It's still inconclusive regarding the trim. Functions are still in there versus other controllers that aren't while the system/library/request.php already has the trim. We just don't know which side of the story is gonna be sticked on.

As for the city, indeed, it's not related. Create a new PR about it if you believe the size should be different.

PR #8752, opened 2020-10-16
Added trim - catalog/controller/api/voucher.php file.

straightlight:


ADDCreative:
This is wrong for the same reason as https://github.com/opencart/opencart/pull/8754#issuecomment-710438129.

straightlight:
No relative. This part still inconclusive.

PR #8754, opened 2020-10-16
Added trim on email field - catalog/controller/checkout/guest.php file.

straightlight:


ADDCreative:
This is wrong. Either the request is already trimmed, in which case adding trim serves no purpose.

Or if you are not going to trim the request before hand, you could get the following issue.

Say the customer's email address is the following, which happens to be 96 characters..
```a+stupidly+long+email+address@to.demonstrate.the.issue.with.trim.before.greater.then.example.com```

Now say they pasted it into the form (well it's too long to type), but had a space at the front so that it is now 97 characters long including the leading space.

It will pass your validation as the trim function will return 96 characters. However, the string in not trimmed before writing to the database. The email column in the database is only 96 characters.
https://github.com/opencart/opencart/blob/9cd75293b78afd7104c82c07cdf2480899784ae9/upload/system/helper/db_schema.php#L3223-L3224

As a result the database will contain the following (note the leading space and the missing m on the end), which in incorrect.
```" a+stupidly+long+email+address@to.demonstrate.the.issue.with.trim.before.greater.then.example.co"```

straightlight:
Ok, valid point here. Closing.

osworx:
Okay this discussion is closed.
But there is another open issue which should be solved.
The maximum lenght of an email address on OpenCart should not exceed 96 characters (that's why I am writing this here).

**But the maximum - and valid length - of an email address can be 254 characters**!

Which means, OpenCart does not except any email address which is longer than 96 characters ..
While most of the internet users will not exceed those 96 characters, but what with those who do?
Do we not want them as customer?

straightlight:
> Okay this discussion is closed.
> But there is another open issue which should be solved.
> The maximum lenght of an email address on OpenCart should not exceed 96 characters (that's why I am writing this here).
> 
> **But the maximum - and valid length - of an email address can be 254 characters**!
> 
> Which means, OpenCart does not except any email address which is longer than 96 characters ..
> While most of the internet users will not exceed those 96 characters, but what with those who do?
> Do we not want them as customer?
> 
> If someone want to read about that: https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address
> 
> p.s.: the minimum can be 3 characters.

May be the varchar and validateForm should be set from a range of 3 characters to 128 characters instead of a limit set of 96 characters since domains like co.uk or many more domain prefix names could appeal before the root domain name in the email addresses.

osworx:
@straightlight I see you have not read the comments at stackoverflow.
Who wants 128 chars?
And nobody is telling stories about domains like co.uk .gov.xx(s) or similiar.
Better read that here before answering wrong the next time: https://www.rfc-editor.org/errata_search.php?rfc=3696

When discussing, then correct.

straightlight:
It's not that I haven't read the URL you have addressed on the above, it's rather the fact that what's being recommended in there by StackOverFlow does not include the fact that OC only uses VARCHAR of 96 and regardless how we'd be trying to improve this validation at this point, store owners and customers would still be stuck at 96 without extending the database field.

PR #8755, opened 2020-10-16
Added validation on unset guest - catalog/controller/checkout/login.php file.

straightlight:


ADDCreative:
I have to ask. What is the purpose of this change?

straightlight:
Simple. The customer needs to login before going though the next steps. Unsetting those session super globals on-the-fly provides no absolute guarantee that these session keys have been created prior to go through these steps. Which is why, better to validate especially due to the fact that events can now be used.

ADDCreative:
If the session wasn't created, the following if statement would be true and therefore never reach the unset of the guest session data.

https://github.com/opencart/opencart/blob/e6337c94d9e5e52b4d04c93b2b981d8a2fc1e450/upload/catalog/controller/checkout/login.php#L29-L31

PR #8756, opened 2020-10-16
Added trim on city - catalog/controller/checkout/payment_address.php file.

straightlight:


ADDCreative:
The request is already trimmed.
https://github.com/opencart/opencart/blob/cd7ab5306519bc4ec1c6866670e145942f88cc3d/upload/system/library/request.php#L47

Trimming before the greater then comparison is wrong anyway. See my comment in. 
https://github.com/opencart/opencart/pull/7733

straightlight:
The commit you're addressing is still under review. It's not official yet.

ADDCreative:
That doesn't make your commit correct.

straightlight:
If I could put this commit under review as well I would so to see which review is rather considered in this case.

ADDCreative:
Why not just close it as it wrong either way.

straightlight:
Because the commit you've addressed is still not official either.

ADDCreative:
Which commit is not official?

straightlight:
Both.

ADDCreative:
https://github.com/opencart/opencart/pull/7737 looks official to me.

You can read my comment as to why the other connit and your pull request are wrong.
https://github.com/opencart/opencart/pull/7733#issuecomment-562137753
https://github.com/opencart/opencart/pull/7733#issuecomment-562189305

straightlight:
#7733 is still under review on anyhow as said above. The same goes for this commit here since we're awaiting for confirmation.

condor2:
#7737 broke spaces on currency symbol. So you can't use something like this yet... **20 USD**

straightlight:
Here we go, a case already. Best to wait for the approvals.

osworx:
> 
> 
> #7737 broke spaces on currency symbol. So you can't use yet something like this... **20 USD**

And especially with the currency symbols it IS important to have a (or more) space() - no matter if before or after set/defined.

straightlight:
@osworx: In American continent, the currency symbol from financial institutions do show without spaces. I did, however, addressed the issue before on this repository regarding the fact that this may also be reliable by payment service providers, this may become problematic for some areas on earth where people are selling the goods from since one could say one thing and the other one says the other while both could be right as both could be wrong politically based on what the order invoice will provide to customers by payment service providers, either over email or from a webpage, showing the currency symbol either stick to the decimal or spaced. Which is why, the geo zones also needs to be dealt with payment service providers. Not only for shipping options in order to deal specifically with those issues.

danijelGombac:
I added a new radio option in currency table. When checked yes, the space is added automatically.

straightlight:
That only remains merchant-based though.

osworx:
> 
> 
> @osworx: In American continent, the currency symbol from financial institutions do show without spaces. 

Sorry, but America is not the only state on this beautiful planet (though some think that ..)!
Nevermind, we are talking here not about "financial institutions" or "payment providers" etc.
The space is only for the visual output in the shop.
And here (in Europe) you can write it before or after ..

straightlight:
@osworx: I was not referring to the left and right symbol location but rather based on when the currency symbol is stick or spaced which was the original subject implied above.

straightlight:
I'll figure out something once the currency URL link fixes I posted yesterday will be approved. It simply doesn't make sense having to manually deal with all these 'one says the other says' requests.

PR #8757, opened 2020-10-16
Added $language as private instance - system/library/cart/currency.php file.

straightlight:


straightlight:
Changed my mind. Not needed since the cart library only has __construct with all defined objects and database object can still propagate throughout other methods in the constructor.

PR #8763, opened 2020-10-17
Removed 'is_null' from Product model

condor2:


condor2:
@danielkerr change the code because is still there.

PR #8764, opened 2020-10-17
Removed 'is_null' from Customer model

condor2:


straightlight:
Yeah, I wondered why those two !is_null were there originally. I would suggest replacing both entire lines from the two commits you posted to: !empty . This will shorten the line codes as , typically, this is what the lines means anyways.

straightlight:
Besides, the !empty / empty function implicits the isset.

condor2:
That two !is_null are from 3.0.x.x migrated from 2.x.x.

And if you mean to replace these lines 

`if (isset($data['filter_newsletter']) && $data['filter_newsletter'] !== '') {`

are more lines like that

straightlight:
Leftovers probably. Yes, those are the lines I meant and should all be replaced by !empty instead of using isset and !is_null since, as mentioned above, !empty already includes those lookups on the server-end side in anyhow.

condor2:
@danielkerr change the code because is still there.

danielkerr:
added manually

PR #8770, opened 2020-10-17
Added localisation_zone_history in oc_event - install/opencart.sql file.

straightlight:


straightlight:
The last link modification is a suggestion. However, since the localisation_currency_history event can also be disabled, store owners also have the ability to edit their countries from the dropdown list from each languages in the admin > localisation > languages if previous country IDs can no longer be found in the admin > localisation > countries when store owners would ever decide to remove them as a last result..

straightlight:
Closing as per this post: https://github.com/opencart/opencart/issues/8798#issuecomment-712877569 .

PR #8781, opened 2020-10-18
For security purposes, reduced CHMOD 0777 to 0755 - admin/controller/common/filemanager.php file.

straightlight:


straightlight:
Based on this discussion: https://forum.opencart.com/viewtopic.php?f=202&t=220546&p=802714#p802714 .

PR #8783, opened 2020-10-18
Replacing the js code of filters with forms

ilcomp:


straightlight:
Reverting the HTML form inputs away from JS and setting user_token as POST instead of GET will prevent admin/controller/common controllers to compare the token strings since $this->session->data['user_token'] is originally captured from GET method throughout the URL. Therefore, I would not suggest this modification.

ilcomp:
Get method is default value for form
https://www.w3.org/TR/html401/interact/forms.html#h-17.3

PR #8784, opened 2020-10-18
Used currency from localisation - system/library/cart/currency.php file.

straightlight:


straightlight:
Closing. Will output a better query.

PR #8785, opened 2020-10-18
Added backward escaped quotes - catalog/model/localisation/currency.php file.

straightlight:


straightlight:
First link is valid for core change.

PR #8797, opened 2020-10-18
Fixed array keys - admin/model/localisation/order_status.php file.

straightlight:


danielkerr:
does not matter for deleting

PR #8809, opened 2020-10-21
LARAVEL ORM IN OPENCART

CloudVisionBulgaria:
Hello, this feature will help you to use laravel illuminate database (ORM) requests without coding mysql syntax.

### DIFF
--- a/upload/composer.json
+++ b/upload/composer.json
@@ -1,0 +1,5 @@
+{
+    "require": {
+        "illuminate/database": "^8.11"
+    }
+}
--- a/upload/composer.lock
+++ b/upload/composer.lock
@@ -1,0 +1,1364 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "This file is @generated automatically"
+    ],
+    "content-hash": "830a714b6130a08770963d763aa7e31a",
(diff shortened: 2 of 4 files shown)


straightlight:
If it'd required automatic calls to load a DB library, I wouldn't rely on that.

CloudVisionBulgaria:
> 
> 
> If it'd required automatic calls to load a DB library, I wouldn't rely on that.

Why? This will call existing methods of laravel capsule with an instance?

CloudVisionBulgaria:
> 
> 
> > If it'd required automatic calls to load a DB library, I wouldn't rely on that.
> 
> Why? This will call existing methods of laravel capsule with an instance?

CloudVisionBulgaria:
> 
> 
> If it'd required automatic calls to load a DB library, I wouldn't rely on that.

I can hardcore the existing methods but i don't have a reason to do it.. cause it will make automaticly instances.

straightlight:
That's the point on my reply above. Using automatic instances as opposed to the namespace calls already handled from the core would not be the best option since Opencart already handles databases user-defined from both config.php files since day 1.

straightlight:
Besides, the action engine is already preventing controllers and models to load magic methods. This wouldn't provide the right example for extension developers to follow this path since they'd be prevented to do so on anyhow.

danielkerr:
opencart already has its own lightweight framework. it does not require LARAVEL

Issue #8704, opened 2020-10-15
Suggestion: when applying special symbols in reward then showing successful message, there should display error message


virendra695:
**What version of OpenCart are you reporting this for?**
Test Environment--
-- opencart version: Version 3.0.3.6
--Branch: 3.0.x.x_Maintenance

**Describe the bug**
A clear and concise description of what the bug is when applying the special symbol in the reward then still showing the success message.

**To Reproduce**
1. Open the checkout page
2. Enter the special symbols !@#$%^&*() in the reward field
3. press on Appy button

1. Go to 'checkout page'
2. Enter "special Symbol !@#$%^&*()"
3. Click on 'Apply button'
4. still success message will display 

**Expected behavior**
-- There should display the error message

**Screenshots / Screen recordings**
![Screenshot (2)](https://user-images.githubusercontent.com/51983890/96097396-861d3b00-0eee-11eb-94b0-72f690fe69b6.png)

**Server / Test environment (please complete the following information):**
 - Operating system: Ubuntu 16.04 LTS
 - PHP version: 7.4.11
 - Browser(s) tested with: Google Chrome (Version 85.0.4183.121 (Official Build) (64-bit))
 - Opencart Version: 3.0.3.6

ADDCreative:
It's not just those special symbols. If you enter a string it will pass the empty check then do a string comparison with the points values. It will be any string starting with a character with a decimal value below 49. For example, typing ```.0``` will also give the success message. Which could happen if you were trying to type ```30``` but hit the . key below 3.

The line that checks is here.
https://github.com/opencart/opencart/blob/7bfca481664fa2cde0ba11776bf23c569abeba9e/upload/extension/opencart/catalog/controller/total/reward.php#L47

Perhaps to catch all, assuming reward can be a float, which I'm not sure it should be.
```php
if (!isset($this->request->post['reward']) || !is_numeric($this->request->post['reward']) || ($this->request->post['reward'] <= 0)) {
```

Or for integers replace the is_numeric.
```php
!ctype_digit((string)$this->request->post['reward'])
```

ADDCreative:
@WebkulOpencart This also looks to affect the master branch.
https://github.com/opencart/opencart/blob/ab11174fbfb818f3e047c2525142fb503844051f/upload/extension/opencart/catalog/controller/total/reward.php#L47

I would also check if the reward can be a decimal of has to be a integer. The use of abs with no typecast suggests that it could be decimal.
[...]

PR #8813, opened 2020-10-22
Resolved reward validation issue

webkul-dilip:
Added suggestion https://github.com/opencart/opencart/issues/8704

### DIFF
--- a/upload/catalog/controller/extension/total/reward.php
+++ b/upload/catalog/controller/extension/total/reward.php
@@ -43,7 +43,7 @@ class ControllerExtensionTotalReward extends Controller {
 			}
 		}
 
-		if (empty($this->request->post['reward'])) {
+		if (!isset($this->request->post['reward']) || !is_numeric($this->request->post['reward']) || ($this->request->post['reward'] <= 0)) {
 			$json['error'] = $this->language->get('error_reward');
 		}
 
@@ -63,7 +63,7 @@ class ControllerExtensionTotalReward extends Controller {
 			if (isset($this->request->post['redirect'])) {
 				$json['redirect'] = $this->url->link($this->request->post['redirect']);
 			} else {
-				$json['redirect'] = $this->url->link('checkout/cart');	
+				$json['redirect'] = $this->url->link('checkout/cart');
 			}
 		}
 


straightlight:
For the core, this fix does make sense. The only differences will be with developed extensions where other var types may come along.

webkul-dilip:
> For the core, this fix does make sense. The only differences will be with developed extensions where other var types may come along.

>> Okay, We will change the code

straightlight:
Idealistically speaking, however, based on your suggestion, a shorter code could be used to validate the voucher POST:

```if (!isset($this->request->post['reward']) || !is_numeric($this->request->post['reward']) || ($this->request->post['reward'] <= 0)) {```

for:

```if (!isset($this->request->post['reward']) || !filter_var($this->request->post['reward'], FILTER_VALIDATE_INT)) {```

But, again, that won't resolve extension validations to maintain the example of what the core would already demonstrate in the first place unfortunately.

**Comment by webkul-dilip** [comment]
if (!isset($this->request->post['reward']) || !filter_var($this->request->post['reward'], FILTER_VALIDATE_INT)) {
In the case of a float number like 20.5, it will show a warning message.  In this case, it will only check value is an integer or not.

**Comment by webkul-dilip** [comment]
if (!isset($this->request->post['reward']) || !filter_var($this->request->post['reward'], FILTER_VALIDATE_INT)) {

ok, I will change

**Comment by webkul-dilip** [comment]
https://github.com/opencart/opencart/pull/8831 @straightlight, please check

**Comment by straightlight** [comment]
There has been previous replies regarding float use instead of int. Perhaps replacing FILTER_VALIDATE_INT with FILTER_VALIDATE_FLOAT could do it for the core and extensions altogether. The only problem will be with extensions directly calling a string type from service providers but that is not good practice anyways. All service providers string types should be associated with database INTs at this point prior to compare the data on the browser.

**Comment by webkul-dilip** [comment]
As I checked in the database all places use Int for reward point, so FILTER_VALIDATE_INT will fine for this case. OR we need to change in the database plase check https://github.com/opencart/opencart/blob/ab11174fbfb818f3e047c2525142fb503844051f/upload/system/helper/db_schema.php#L1623-L1626

**Comment by straightlight** [comment]
We don't necessarily need to change the database for this, since all associated types should be validated with INT types on the database as mentioned above.

**Comment by ADDCreative** [comment]
@webkul-dilip You still need to check for zero and possibly negative.

So maybe something like.
```php
if (!isset($this->request->post['reward']) || !filter_var($this->request->post['reward'], FILTER_VALIDATE_INT) || ($this->request->post['reward'] <= 0)) {
```
Or got back to using empty() but with FILTER_VALIDATE_INT if you are happy with negative values.

straightlight:
Yes, correct. However, empty() would also check for other variable types, not only INT types.

straightlight:
Another possibility to validate INTs by range with only positive values: https://stackoverflow.com/questions/4844916/best-way-to-check-for-positive-integer-php .

ADDCreative:
> Yes, correct. However, empty() would also check for other variable types, not only INT types.

Which is why I said "empty() but WITH FILTER_VALIDATE_INT".

straightlight:
The $filter_options should do it since it does not require other functions but simply a switch case within the same function of filter_var .

PR #8827, opened 2020-10-26
fix load href

BuslikDrev:
There is no point in loading like this:
https://demo.opencart.com/index.php?route=common/menu

I also suggest moving "Cookie, Currency, Language, Menu, Search" to the "Extension" folder
And make it possible to set the position from the admin panel.
I would also like to set the width of the layouts (col-sm -...) from the admin panel.

### DIFF
--- a/upload/catalog/controller/common/cart.php
+++ b/upload/catalog/controller/common/cart.php
@@ -1,7 +1,7 @@
 <?php
 namespace Opencart\Application\Controller\Common;
 class Cart extends \Opencart\System\Engine\Controller {
-	public function index($setting) {
+	public function index() {
 		$this->load->language('common/cart');
 
 		// Totals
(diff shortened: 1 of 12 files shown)


chulcha:
ArgumentCountError: Too few arguments to function Opencart\Application\Controller\Common\Search::index()

index.php?route=common/search

```
namespace Opencart\Application\Controller\Common;
class Search extends \Opencart\System\Engine\Controller {
	public function index($setting) {
```

BuslikDrev:
> ArgumentCountError: Too few arguments to function Opencart\Application\Controller\Common\Search::index()
> 
> index.php?route=common/search
> 
> ```
> namespace Opencart\Application\Controller\Common;
> class Search extends \Opencart\System\Engine\Controller {
> 	public function index($setting) {
> ```

/index.php?route=extension/module/banner

Let it be @danielkerr fix

PR #8645, opened 2020-10-06
Suggestion add .editorconfig

gnovaro:
EditorConfig helps maintain consistent coding styles for multiple developers working on the same project across various editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles. EditorConfig files are easily readable and they work nicely with version control systems.

https://editorconfig.org/

PR #8840, opened 2020-11-03
Already triggered. Removed white spaces as well - admin/view/template/marketplace/extension.twig file.

straightlight:


gnovaro:
i suggest use this  .editorconfig #8645

PR #8846, opened 2020-11-07
Removed comment part + the uncommented array - admin/controller/common/header.php file.

straightlight:


straightlight:
This commented block doesn't make sense based on the $data['notifications'] that hasn't been commented out among the below part in the code and where it also doesn't make sense is calling a model route under setting while the $this->url->link points to the tool path.

PR #8853, opened 2020-11-08
More modifications on affiliates - admin/controller/report/customer_order.php file.

straightlight:


danielkerr:
i think already added

danielkerr:
commission should not be in the customer order report

PR #8873, opened 2020-11-12
marketplace api for stored credit cards

stasdrvn:
Added functionality to purchase from admin marketplace by saved credir card. PIN was removed.


straightlight:
While the rests of the lines might look ok, idealistically speaking, I am not sure this one should be in as such:

```$_['text_no_cards']		 = 'No payment methods have been added to your account on www.opencart.com!';```

stasdrvn:
Updated. Issue appeared due to tabulation

straightlight:
While the tabulation itself may be the case, I was rather referring to the sentence itself.

stasdrvn:
Get it from opencart.com. If Daniel decides to change it, so will change it with no problem.

straightlight:
$_['text_no_cards'] does not exist in OC v3.0.3.6 nor the master branch nor should it.

PR #8874, opened 2020-11-13
Drop support for php 7.1, curl required extension

gnovaro:
Php drop official support for 7.2 in 1 Jan 2021
https://www.php.net/supported-versions.php


ADDCreative:
Minimum should be 7.3 to match check at startup.
https://github.com/opencart/opencart/blob/65d7040c4bf30a70980ffda44a3ca1ad14b66660/upload/system/startup.php#L6-L8

gnovaro:
Updated!

PR #8876, opened 2020-11-14
Switched two InnoDB to MyISAM - system/helper/db_schema.php file.

straightlight:


danielkerr:
mysam should not be used for sessions

PR #8877, opened 2020-11-14
Added validation to $args[0] - catalog/controller/event/activity.php file.

straightlight:


danielkerr:
why would it throw an error for $args[0]

PR #8878, opened 2020-11-14
Added is_array in config_mail_alert - catalog/controller/mail/affiliate.php file.

straightlight:


danielkerr:
it forces to array

PR #8880, opened 2020-11-14
Feature new image formats

Hkr32:
I think it's time to add modern image formats. If you missed something, please give me a hint and I will try to fix it.


danielkerr:
already added

PR #8908, opened 2020-11-17
Allow longer module code

stalker780:
Cause now it also includes extension code prefix

What was `my_beautiful_module` now will look like `my_awesome_extension.my_beautiful_module`


straightlight:
Good point ...

Issue #8787, opened 2020-10-18
[Master Branch] - admin/controller/common/reset.php


yvonn:
[Master Branch] - admin/controller/common/reset.php
at line 62 and 67 change 

`);`

into this

`];`

straightlight:
More locations found than a single spot in that file. Confirmed, however.

gnovaro:
We can close this

PR #8918, opened 2020-11-20
Fix syntax error reset.php #8787

gnovaro:
Line   admin/controller/common/reset.php
 ------ --------------------------------------------------------
  62     Syntax error, unexpected ')', expecting ']' on line 62
  67     Syntax error, unexpected ')', expecting ']' on line 67


straightlight:
Confirmed.

PR #8920, opened 2020-11-21
OC4 Stop reconstructing controllers on each call

stalker780:
Fix controllers calls:
1. Allows storing variables inside controller classes.
2. Also most 3rd party extensions call lots of init functions on their controllers __construct(). This may cause significant performance loss which is hard to debug.
3. Event system would make things only worse, calling more controller functions and reconstructing controller on each call.

### DIFF
--- a/upload/system/engine/action.php
+++ b/upload/system/engine/action.php
@@ -62,7 +62,13 @@ class Action {
 
 		// Initialize the class
 		if (class_exists($this->class)) {
-			$controller = new $this->class($registry);
+			$controller_route = 'controller_' . str_replace(array('/', '|'), '_', (string)$this->route);
+			if (!$registry->has($controller_route)) {
+				$controller = new $this->class($registry);
+				$registry->set($controller_route, $controller);
+			} else {
+				$controller = $registry->get($controller_route);
+			}
 		} else {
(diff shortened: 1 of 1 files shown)


straightlight:
So what do you suggest? To only stick to core controllers as an alternative to use extensions then?

stalker780:
I have pulled my suggestion.
It checks if controller was constructed previously and takes it from registry with all it's variables.

straightlight:
That also means that all codes at the same time, including 3rd party, would be loaded into the Engine.

```As far as I remember we discussed this here with Daniel since OC 2.x versions. But this behavior is still there in core.```

That might be because the suggestion wasn't considered at the time.

stalker780:
> That also means that all codes at the same time, including 3rd party, would be loaded into the Engine.

There is no problem loading all models and libraries, so why loading controllers should be an issue?
Isn't their multiple reconstruction a bigger problem for CPU and disk?

mhcwebdesign:
Different controller instances, even of the same controller class, may have different variables. It won't be a straight-forward shot with a simple core engine change.

straightlight:
Agreed.

danielkerr:
not required

PR #8922, opened 2020-11-21
Replaced manually queries with models - admin/model/customer/customer.php file.

straightlight:


straightlight:
The same should be handled with admin/model/sale/order/getOrder method.

TheCartpenter:
Note: The total alias have now been renamed accordingly on the master branch since.

PR #8926, opened 2020-11-23
Fix customer group API

danijelGombac:
Add error for customer group and fix wrong custom field validation

### DIFF
--- a/upload/catalog/controller/api/customer.php
+++ b/upload/catalog/controller/api/customer.php
@@ -56,23 +56,27 @@ class Customer extends \Opencart\System\Engine\Controller {
 			}
 
 			// Customer Group
-			if (is_array($this->config->get('config_customer_group_display')) && in_array($this->request->post['customer_group_id'], $this->config->get('config_customer_group_display'))) {
+			if ($this->customer->isLogged() && ($this->customer->getGroupId() == $this->request->post['customer_group_id'])) {
+				$customer_group_id = $this->request->post['customer_group_id'];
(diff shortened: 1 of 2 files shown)


TheCartpenter:
The $this->config->get('config_customer_group_id') is registered on startup and on the admin > systems > settings > add / edit settings page. Removing this validation, as part of the store lookup, would contradict those settings in the mean time. In order to avoid alteration to those lookups, especially for multi-stores, better to validate with the $this->load->model('setting/setting') and to pull the editSetting with the order info's store ID value. If that cannot be validated, then to remain with the default store ID assigned from the admin settings.

danijelGombac:
Then the entire validation is wrong.

When you select the customer group in admin order edit page, the custom field inputs changes for that group.

Then you send that fields with customer_group_id to the api controller and without errors changes in diferent customer_group_id  and select wrong validation criteria.

```
if (is_array($this->config->get('config_customer_group_display')) && in_array($this->request->post['customer_group_id'], $this->config->get('config_customer_group_display'))) {
      $customer_group_id = $this->request->post['customer_group_id'];
} else {
      $customer_group_id = $this->config->get('config_customer_group_id');
}

// Custom field validation
$this->load->model('account/custom_field');

$custom_fields = $this->model_account_custom_field->getCustomFields($customer_group_id);
```

TheCartpenter:
They're not entirely wrong. The getSetting simply needs to be pulled to ensure that the store_id can indeed be found with the specific order prior to load the related custom fields prior to focus on the default store ID as I do notice that the default store ID is being used too many times rather than lookup through the settings in the process.

TheCartpenter:
I believe this is what you're looking for to accomplish: https://github.com/opencart/opencart/pull/9226 and https://github.com/opencart/opencart/pull/9227 .

TheCartpenter:
Also this one: https://github.com/opencart/opencart/pull/9228 .

PR #8928, opened 2020-11-24
Text Loading at wrong button

osworx:


danielkerr:
already removed text-loading

PR #8929, opened 2020-11-24
Text Loading .. at wrong button

osworx:


eka7a:
You removed it from one and added it to the other but both buttons should have data-loading-text.

osworx:
> 
> 
> You removed it from one and added it to the other but both buttons should have data-loading-text.

I know ... 

But, to the position (button) I have moved has NO translation, only "loading .."
So, some of the users do not speak English nor is there the transaed text.
Second (2.), the translation should be assigned to the button I have changed, not at the "Delete" button > there loads nothing.

Instead if all (buttons) would be correct, at the "Delete" button the text should be "Deleting item .." (or something else) and at the "Edit" the text I assigned to.

That's why I have removed (or moved) the text ..
Do with the commit whatever you want if you think it is not correct.

PR #8938, opened 2020-11-28
Fix FontAwesome5 left menu icons - p2

stalker780:
Since FontAwesome5 may contain not only fas class, it is preferred to pass both classes in controller.

3d party extensions may use far, fab, etc. classes in left_column menu.

Sorry for two separate pull requests. Web version of GitHub seems does not allow pulling several files in one request

### DIFF
--- a/upload/admin/view/template/common/column_left.twig
+++ b/upload/admin/view/template/common/column_left.twig
@@ -5,9 +5,9 @@
     {% for menu in menus %}
       <li id="{{ menu.id }}">
         {% if menu.href %}
-          <a href="{{ menu.href }}"><i class="fas {{ menu.icon }}"></i> {{ menu.name }}</a>
+          <a href="{{ menu.href }}"><i class="{{ menu.icon }}"></i> {{ menu.name }}</a>
         {% else %}
-          <a href="#collapse{{ i }}" data-toggle="collapse" class="parent collapsed"><i class="fa {{ menu.icon }}"></i> {{ menu.name }}</a>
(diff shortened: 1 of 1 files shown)


upload/admin/view/template/common/column_left.twig
```diff
@@ -5,9 +5,9 @@
     {% for menu in menus %}
       <li id="{{ menu.id }}">
         {% if menu.href %}
-          <a href="{{ menu.href }}"><i class="fas {{ menu.icon }}"></i> {{ menu.name }}</a>
+          <a href="{{ menu.href }}"><i class="{{ menu.icon }}"></i> {{ menu.name }}</a>
         {% else %}
-          <a href="#collapse{{ i }}" data-toggle="collapse" class="parent collapsed"><i class="fa {{ menu.icon }}"></i> {{ menu.name }}</a>
+          <a href="#collapse{{ i }}" data-toggle="collapse" class="parent collapsed"><i class="{{ menu.icon }}"></i> {{ menu.name }}</a>
```
straightlight, line 10: I don't see any fa class on that line ... I do see fas like line 8 though.

stalker780: FontAwesome5 have no fa class. So this was a typo, left from old OC versions. 
https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use

My proposal is to pass icon classes in column_left.php. So that in the future it is possible to use other FontAwesome5 classes like far, fab and others which are available in free version. 
https://github.com/opencart/opencart/pull/8937

These icons can be used when adding new menu branches for extensions using events system. Otherwise all icons are limited only with fas class, which is wrong and will not allow to use hundreds of icons from other classes.

PR #8945, opened 2020-12-01
fix removePermission

BuslikDrev:
PHP Warning:  array_diff(): Argument #1 is not an array in

В лучшем случае нужно весь документ переделать.

			$this->load->model('user/user_group');

			foreach ($this->model_user_user_group->getUserGroups() as $result) {
				$this->model_user_user_group->removePermission($result['user_group_id'], 'access', 'extension/module/mymodule');
				$this->model_user_user_group->removePermission($result['user_group_id'], 'modify', 'extension/module/mymodule');
			}

### DIFF
--- a/upload/admin/model/user/user_group.php
+++ b/upload/admin/model/user/user_group.php
@@ -3,7 +3,7 @@ namespace Opencart\Application\Model\User;
 class UserGroup extends \Opencart\System\Engine\Model {
 	public function addUserGroup($data) {
 		$this->db->query("INSERT INTO `" . DB_PREFIX . "user_group` SET `name` = '" . $this->db->escape((string)$data['name']) . "', `permission` = '" . (isset($data['permission']) ? $this->db->escape(json_encode($data['permission'])) : '') . "'");
-	
+
 		return $this->db->getLastId();
 	}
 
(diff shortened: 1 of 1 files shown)


upload/admin/model/user/user_group.php
```diff
@@ -75,10 +75,12 @@ public function addPermission($user_group_id, $type, $route) {
 	public function removePermission($user_group_id, $type, $route) {
 		$user_group_query = $this->db->query("SELECT DISTINCT * FROM `" . DB_PREFIX . "user_group` WHERE `user_group_id` = '" . (int)$user_group_id . "'");
 
-		if ($user_group_query->num_rows) {
+		if ($user_group_query->num_rows && $user_group_query->row['permission']) {
```
straightlight, line 78: The $user_group_query->row['permission'] field is a JSON string. There is no point on validating that with a simple IF statement. By adding a validation on the type field, this would also prevent error triggers to be outputted as to provide limited support originating from the installed extensions. I would not suggest this.

straightlight:
If you're looking for a way to validate JSON strings, I would suggest to do so from SQL such as this e.g, for instance: https://stackoverflow.com/questions/49053341/mysql-select-where-json-field-property-has-value?answertab=active#tab-top .

BuslikDrev (commented):
> Если вы ищете способ проверки строк JSON, я бы предложил сделать это из SQL, например, например: [https://stackoverflow.com/questions/49053341/mysql-select-where-json-field -property-has-value? answertab = active # tab-top](https://stackoverflow.com/questions/49053341/mysql-select-where-json-field-property-has-value?answertab=active#tab-top) .

Why didn't you add your proposal to the code OpenCart?

BuslikDrev:
> If you're looking for a way to validate JSON strings, I would suggest to do so from SQL such as this e.g, for instance: https://stackoverflow.com/questions/49053341/mysql-select-where-json-field-property-has-value?answertab=active#tab-top .

Why didn't you add your proposal to the code OpenCart?

straightlight:
It's not extra. In fact, it's a simplified lookup as opposed to set extra validations of JSON string that won't track the exact permission as opposed to SQL queries I suggested to use on the above.

PR #8949, opened 2020-12-05
Added value for oc_event.description

Hkr32:
Added value for oc_event.description
Changed not_null value to false

Error during installation


cgcmart (approved):

danielkerr:
not required since no description added to any. this is more so extension devs can give the user a description of whaty the even t is for.

PR #8951, opened 2020-12-05
common.js popover product images fix AND modification.php log  FILE name in modification loop

paolobiavati:
the LOG line with the "FILE: " indication is written at beginning of each file processing, and also when a file is already in the modification array with a note saying 'sub modification'.
I think is useful when you have an error in the log.


paolobiavati:
common.js. popover in the products images
If there is a popover already open and you click on an image **below**, the old popover is closed and a new one is shown, but in the new one the buttons don't work, because the events are binded in the old popover. 
I add a setTimeout to delay the binding

https://user-images.githubusercontent.com/2340162/103411385-9648cc00-4b67-11eb-993c-75e4f0b3a6d7.mp4

Issue #8808, opened 2020-10-21
Suggestion: admin customer history, reward points and transactions listing limit should be used by setting


prabhatkumaroc082:
**What version of OpenCart are you reporting this for?**
Version 3.0.3.6, master branch 4.0.0.0_b

**Describe the suggestion**
Admin side customer history, reward points, and transaction pagination limit should use by this setting https://prnt.sc/v3keum
for now, 10 limits are set so we can add this setting count https://prnt.sc/v3keum

**To Reproduce**
1. Go to admin->customers->customers
2. Edit any customer
3. check customer history, reward points, and transaction listing which is set to 10 limit

PR #8973, opened 2020-12-15
added admin setting limit to customer history,reward,transaction and ip listing

prabhatkumaroc082:


prabhatkumaroc082:
fix for #8808

PR #8980, opened 2020-12-18
Update seo_url.php

danijelGombac:


chulcha:
What for?
Hope the commit won't be accepted

mhcwebdesign:
What are your changes for? What issue or problem are you trying to solve?

stalker780:
This is an attempt to fix seo_url in OC4.

I have discribed this earlier in https://github.com/opencart/opencart/issues/8923#issuecomment-739525181

But in my opinion it is not a good idea to move Action construction to seo_url controller.
We need to wait for a fix from Daniel.

danijelGombac:
@stalker780 

catalog/controller/startup/maintenance.php has also Action construction in controller and in admin login and permission. Why seo_url should different?

I fixed seo_url in my store, but this changes should be done by Daniel, because is too important feature. This is why i closed all.

PR #8981, opened 2020-12-18
fix error request post

BuslikDrev:
fix errors when sending a request without filling data

image


ADDCreative:
You don't need to use ```isset``` twice.

BuslikDrev:
> You don't need to use `isset` twice.

I don't need anything at all, I just suggest a solution. You can offer yours.

PR #8982, opened 2020-12-18
Update file.php

PhamToan:
sometime it is deleted by another process. We need to check file exist before deleting

### DIFF
--- a/upload/system/library/session/file.php
+++ b/upload/system/library/session/file.php
@@ -74,7 +74,7 @@ class File {
 			$files = glob(DIR_SESSION . 'sess_*');
 
 			foreach ($files as $file) {
-				if (filemtime($file) < $expire) {
+				if (is_file($file) && filemtime($file) < $expire) {
 					unlink($file);
 				}
 			}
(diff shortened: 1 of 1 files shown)


ADDCreative:
If the file is deleted by another process then using is_file will probably not help as it take its information from the [file status cache](https://www.php.net/manual/en/function.clearstatcache.php). There is also the chance that the other process deletes the file after the is_file.

PhamToan:
> If the file is deleted by another process then using is_file will probably not help as it take its information from the [file status cache](https://www.php.net/manual/en/function.clearstatcache.php). There is also the chance that the other process deletes the file after the is_file.

I have used this way for fixing on my site. it is really help ... the time shows on error less (5 days without error since from fixed)

PhamToan (commented):
better function

ADDCreative:
I would of thought the original is_file would be better. It will be slightly faster and will not match if a directory. It would also match the checks in the read and destroy methods.

TheCartpenter:
is_file lookups are more persistent in the core than file_exists . I am not sure how more relevant it would be to track a file with file_exists as opposed with is_file function in this case.

danielkerr:
already added

PR #8986, opened 2020-12-19
updated composer.json and /vendor

prabhatkumaroc082:


prabhatkumaroc082:
@straightlight please check it

PR #9007, opened 2020-12-23
fix issue 9006

### LINKED ISSUE #9006: [Master Branch] Error: Field 'description' doesn't have a default value
**What version of OpenCart are you reporting this for?**
4.0.0.0.b master branch

**Describe the bug**
Installer bug in step3 when populating the 'oc_event' table

**How to fix**
Update the install/opencart.sql to include this missing field 'description' when populating the 'oc_event' table.

stickpro:
Closes #9006


TheCartpenter:
Correct. The description field is already set in the db_schema helper file on the master branch and does not use a default empty field. This needs to be added.

PR #9018, opened 2020-12-28
Revert "fixed 0 special price issue on product page"

mathsolutions:
Reverts opencart/opencart#8985
The merged fix is wrong, the price shown is always 0 (zero) even if no 0 special price is set.


mathsolutions:
Bug already solved

PR #9024, opened 2020-12-29
Typo Event language

condor2:


cgcmart:
the event

condor2:
events....are more....and the message is displayed when a restricted user doesn't have access/modify permissions to access events 

But I know....language structure is broken.... No singular or plural....when %s variable is used.

PR #9040, opened 2020-12-31
Added missing language variable-Affiliate

condor2:


eka7a:
Language variables (column_ip, column_store etc) that you removed are used in affiliate_report.twig file.

PR #9047, opened 2021-01-03
Removed duplicated language variable-Startup

condor2:


condor2:
Maybe first column_action should be changed in column_code_action and in language....Code Action...

Issue #9052, opened 2021-01-03
[Master Branch] Cart library update may not use this->data properly to cache results with empty cart (optimalization)


vladaman:
I found following line. Given the **Cart is empty with 0 items** it will still perform multiple queries to the DB and get nothing. I think it should use null value instead of empty array [];

https://github.com/opencart/opencart/blob/964c9d857ad6a243b2c8c5679148bab130a87912/upload/system/library/cart/cart.php#L34

May improve #9049 with empty cart

danielkerr:
it performs one query if no products that's it!

i think its a case of "i want to add my own little piece of code because i'm going to use it in my projects from now on. so ill make an issue up and hope hes to stupid to check!"

wasting my time!

Issue #9050, opened 2021-01-03
[Master Branch] Extension class attempts to load Opencart module 40 times


vladaman:
Following query for extensions causes the loader to be called 40 times for **opencart module** (_since opencart modules are present 40 times in SQL table_):
https://github.com/opencart/opencart/blob/964c9d857ad6a243b2c8c5679148bab130a87912/upload/catalog/model/setting/extension.php#L5
and later the DB result is used in: https://github.com/opencart/opencart/blob/964c9d857ad6a243b2c8c5679148bab130a87912/upload/catalog/controller/startup/extension.php#L8
```php
// catalog/controller/startup/extension.php
namespace Opencart\Application\Controller\Startup;
class Extension extends \Opencart\System\Engine\Controller {
	public function index() {
		// Add extension paths from the DB
		$this->load->model('setting/extension');
		$results = $this->model_setting_extension->getExtensions();
		foreach ($results as $result) { // looping 40x
```

**Proposed solution:**
Replace SQL query with:
```sql
SELECT DISTINCT extension FROM `" . DB_PREFIX . "extension`
```

danielkerr:
it adds the same stuff to the autoloader but does not load stuff 40 times!

Issue #9049, opened 2021-01-03
[Master Branch] - too many duplicated queries to DB - performance issue


vladaman:
While investigating performance of Opencart I made a list of queries which are performed on homepage. There are quite few  duplicated DB queries. Removing those would probably lead to easy performance gain.

See lines with **!!** indicating duplicates.
```
   SET time_zone = '+00:00'
   SELECT * FROM `oc_store` WHERE REPLACE(`url`, 'www.', '') = 'http://localhost:8091/'
   SELECT * FROM `oc_setting` WHERE `store_id` = '0' OR `store_id` = '0' ORDER BY `store_id` ASC
!! SET time_zone = '+00:00'
   SELECT `data` FROM `oc_session` WHERE `session_id` = '4ab90625d63f6cf87267062e94' AND `expire` > '2021-01-03 17:58:16'
   SELECT * FROM `oc_currency`
   SELECT tr1.`tax_class_id`, tr2.`tax_rate_id`, tr2.`name`, tr2.`rate`, tr2.`type`, tr1.`priority` FROM `oc_tax_rule` tr1 LEFT JOIN `oc_tax_rate` tr2 ON (tr1.`tax_rate_id` = tr2.`tax_rate_id`) INNER JOIN `oc_tax_rate_to_customer_group` tr2cg ON (tr2.`tax_rate_id` = tr2cg.`tax_rate_id`) LEFT JOIN `oc_zone_to_geo_zone` z2gz ON (tr2.`geo_zone_id` = z2gz.`geo_zone_id`) LEFT JOIN `oc_geo_zone` gz ON (tr2.`geo_zone_id` = gz.`geo_zone_id`) WHERE tr1.`based` = 'shipping' AND tr2cg.`customer_group_id` = '1' AND z2gz.`country_id` = '56' AND (z2gz.`zone_id` = '0' OR z2gz.`zone_id` = '891') ORDER BY tr1.`priority` ASC
[...]

PR #9053, opened 2021-01-03
Loading and caching extensions (optimalization)

vladaman:
Following changes improves performance 
Related: #9052 #9050 #9049


memeq1:
i think you should check if cache get is false

danielkerr:
some of this has been added already. maybe add delete cache on the uninstall of an extension

PR #9060, opened 2021-01-05
[Feature Request] Added DB Cache Support

vladaman:
Added generic function to DB class to allow us use cached queries. This also requires to move initialization of cache above DB initialization in framework.php


danielkerr:
dumb way of doing things. i can do a lot better.

Issue #7762, opened 2019-12-15
Need to update the front end to use confirm before deleting the address ...


engwalid1971:
... because the customer may be delete  the addresses by a mistake.

PR #9070, opened 2021-01-07
fixed popup on deleting the address from address book

swetasom:
fixed popup on deleting the address from address book


swetasom:
https://github.com/opencart/opencart/issues/7762

PR #9100, opened 2021-01-10
Update statistics.php

SaifAlSubhi:
I think this should be 'text_return'
I may be wrong.


osworx:
Why?
Return is singular, Returns plural

SaifAlSubhi:
> Why?
> Return is singular, Returns plural

because this what I see in php & twig
`text_return`

PR #9101, opened 2021-01-10
Added padding for footer

Hkr32:
before
image

after
image


Hkr32:
@eka7a, conflict fixed

Issue #9069, opened 2021-01-06
OC4 module/install function is never executed


stalker780:
https://github.com/opencart/opencart/blob/3c17f113cdce4e9f58f2ef2a38c06bc41083ce6f/upload/admin/controller/extension/module.php#L32
is never executed because controller route to install() function is built based on oc_extension table when index.php is called.
But when index.php is called this module is not installed yet.

So routes structure must be rebuilt or route to new module must be added to autoloader before calling install function.

Same for other extensions types: theme, feed, captcha, payment, shipping, etc.

One of solutions is to call startup/extension after installation and before calling module's install function:
`$this->load->controller('startup/extension');`

SaifAlSubhi:
Why if you installed "extension.ocmod.zip"; then "install.xml" file is not copied into "/extension/" folder which contains modifications even if the module is listed in admin/extension panel?

danielkerr:
because the install xml is alreadyed added to the db!

SaifAlSubhi:
> because the install xml is alreadyed added to the db!

I see, thanks Daniel.

danielkerr:
the extension paths are added on the startup. it works fine. just tested it. wasting my time. i think u are trying to build an extension directly in the extension folder. you need to setup key files first before installing and working from there.

stalker780:
@danielkerr 

No! I am not wasting your time! Perhaps I need to give you steps to reproduce.
Otherwise you will get tons of issues here, that modules are not installed correctly.

It's best to check with empty DB or with a module you never installed in this DB before. 
**Main requirement to reproduce issue is: module must be absent in oc_extension table**

1. Goto Installer and upload our dummy_ext.ocmod.zip

2. Click [+] to unzip and Install extension

3. Goto Extensions->Modules and see dummy_ext_heading_title

4. Click [+] install for this module. **After this step, module is finally added  forever in oc_extension table and it's path can be used for language and startup load. BUT NOT DURING THIS AJAX CALL, ONLY FOR FUTURE PHP CALLS, WHEN STARTUP WILL AUTOLOAD IT FROM oc_extension** 

5. F5 to refresh a page and select Feeds and then Modules again (to call select.onchange event, stupid ajax selectbox makes me sick)

6. Now, when Modules list is refreshed you can see DUMMY EXT language title in modules list

7. Check Error Logs. Nothing there. However module has install() and uninstall() functions with logging.

8. Now uninstall a module and you will get a line in Error Logs for uninstall()

[...]

PR #9103, opened 2021-01-10
OC4 fix admin extension startup

stalker780:
Fix language and install() function for uninstalled modules

https://github.com/opencart/opencart/issues/9069#issuecomment-757287925

If extension have no modules installed, it is absent in oc_extension table, so it's route will be never autoloaded at startup.

This breaks language titles and call to $this->controller->load(route)->install() function on module installation.

A fix is to take extensions codes from oc_extension_install table.

### DIFF
--- a/upload/admin/controller/startup/extension.php
+++ b/upload/admin/controller/startup/extension.php
@@ -5,24 +5,24 @@ class Extension extends \Opencart\System\Engine\Controller {
 		// Add extension paths from the DB
 		$this->load->model('setting/extension');
 
-		$results = $this->model_setting_extension->getExtensions();
+		$results = $this->model_setting_extension->getInstalls();
 
 		foreach ($results as $result) {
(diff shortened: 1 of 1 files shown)


chulcha:
You can't do this

startup/extension.php - is runtime 

You can't do this
it is in it that the necessary paths for active extensions are added.

For installation methods, your code needs

stalker780:
> You can't do this
> 
> startup/extension.php - is runtime
> 
> You can't do this
> it is in it that the necessary paths for active extensions are added.
> 
> For installation methods, your code needs

Oh yes, I can :)

Pay attention, my fix does absolutely the same - iterates extension_codes.

It only takes these codes from a different table.

TheCartpenter:
Another problematic with this pull request is the fact that the extension_download_id can also be filtered and the admin extension startup controller does not require this filter. Therefore, getExtensions with the use of the extension array key would remain the right one.

PR #9112, opened 2021-01-12
Update template.php

SaifAlSubhi:
I think .tpl should be .twig


ADDCreative:
Twig templates are handled by the system/library/template/twig.php adaptor.

PR #9132, opened 2021-01-17
Fix 'policy' cookie

SaifAlSubhi:
The cookies confirmation will pop up each time the web page changed. This fixed it for me.

### DIFF
--- a/upload/catalog/controller/common/cookie.php
+++ b/upload/catalog/controller/common/cookie.php
@@ -34,7 +34,7 @@ class Cookie extends \Opencart\System\Engine\Controller {
 
 			$option = [
 				'expires'  => time() + 60 * 60 * 24 * 365,
-				'path'     => !empty($_SERVER['PHP_SELF']) ? dirname($_SERVER['PHP_SELF']) . '/' : '',
+				'path'     => !empty($_SERVER['PHP_SELF']) ? dirname($_SERVER['PHP_SELF']) : '',
 				'SameSite' => 'Strict'
 			];
 


SaifAlSubhi:
I think there are other instances where ". '/' " needs to be removed from 'path'

cgcmart:
> 
> 
> The cookies confirmation will pop up each time the web page changed. This fixed it for me.

It looks like you need to re-install the last updated master.

SaifAlSubhi:
> > The cookies confirmation will pop up each time the web page changed. This fixed it for me.
> 
> It looks like you need to re-install the last updated master.
> pop up happens on chrome browser and edge browser, either "click here" or press "Yes, that's fine!" does not move the pop up away at all.
> It seems firefox browser ok.

Yes, I keeps poping up in Chrome, Edge and Safari ... removing '/' from path seems to fix it.

cgcmart:
here is the [explanation](https://web.dev/samesite-cookies-explained/):

> Changes to the default behavior without SameSite #

While the SameSite attribute is widely supported, it has unfortunately not been widely adopted by developers. The open default of sending cookies everywhere means all use cases work but leaves the user vulnerable to CSRF and unintentional information leakage. To encourage developers to state their intent and provide users with a safer experience, the IETF proposal, Incrementally Better Cookies lays out two key changes:

    Cookies without a SameSite attribute will be treated as SameSite=Lax.
    Cookies with SameSite=None must also specify Secure, meaning they require a secure context.

> Chrome implements this default behavior as of version 84. Firefox has them available to test as of Firefox 69 and will make them default behaviors in the future. To test these behaviors in Firefox, open about:config and set network.cookie.sameSite.laxByDefault. Edge also plans to change its default behaviors.

SaifAlSubhi:
Thank you for the explanation.
I think the cookies are set with a path // (double /) and the path should be / (single /)

Issue #9122, opened 2021-01-15
Invoice has wrong ID in header sale/order/invoice


faca5:
Hello.

Latest version v3.0.3.6

Invoice have wrong ID in header. It shows ORDER_ID. It should be INVOICE_ID.

Example:
https://demo.opencart.com/admin/index.php?route=sale/order/invoice&user_token=opo1IvZiucdoczcWj0dCiJdiarrcNNXR&order_id=10832

Wrong:
Invoice #ORDER_ID

Correct:
Invoice #INVOICE_ID

Thank you.

condor2:
I think correct is:
Order #ORDER_ID

Because can't be:
Invoice #INVOICE_ID

when Invoice isn't generated.

faca5:
Then it should be only "INVOICE" without ID.

condor2:
Or that....

If isn't generated and click Print invoice to show:

"Invoice"

and when generated and print invoice to show:
"Invoice #INVOICE_ID"

just un idea :)

condor2:
Or Print invoice to be disabled if invoice isn't generated and after is generated Print invoice to be enabled and header to be 

"Invoice #INVOICE_ID"

faca5:
Invoice must be compatible with law.

Invoice can be upgraded with extra text in middle/footer to provide other law data.

osworx:
> 
> 
> Invoice must be compatible with law.
> 


That's good .. an Invoice is a document which shall not be edited/manipulated after it is generated.
But with OpenCart you can edit any Invoice after generation, simply edit the database table.
With that in mind, this so called "Invoice" will never be accepted by the authorities and therefore cannot be used as "Invoice".

If you really want a correct Invoice, create a PDF and store that encrypted.

yvonn:
you can replace this in order_invoice.twig
delete:
`<h1>{{ text_invoice }} #{{ order.order_id }}</h1>`
add:
`<h1>{% if order.invoice_no %}{{ text_invoice }} {{ order.invoice_no }}{% else %}{{ text_order }} {{ order.order_id }}{% endif %}</h1>`

If invoice number is generated text on invoice is Invoice + invoice number else Order + #Order ID

yvonn:
![2021-01-16_15-07-45](https://user-images.githubusercontent.com/737137/104814099-56dbda00-580d-11eb-85a5-a6e959e5e9c7.jpg)

yvonn:
[...]

PR #9134, opened 2021-01-18
fixed the wrong invoice header

swetasom:
fixed https://github.com/opencart/opencart/issues/9122

### DIFF
--- a/upload/admin/view/template/sale/order_invoice.twig
+++ b/upload/admin/view/template/sale/order_invoice.twig
@@ -14,7 +14,7 @@
 <div class="container">
   {% for order in orders %}
   <div style="page-break-after: always;">
-    <h1>{{ text_invoice }} #{{ order.order_id }}</h1>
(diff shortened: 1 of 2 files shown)


upload/admin/view/template/sale/order_invoice.twig
```diff
@@ -14,7 +14,7 @@
 <div class="container">
   {% for order in orders %}
   <div style="page-break-after: always;">
-    <h1>{{ text_invoice }} #{{ order.order_id }}</h1>
+    <h1>{% if order.invoice_no %}{{ text_invoice }} #{{ order.invoice_no }}{% else %}{{ text_order }}{% endif %}</h1>
```
osworx, line 17: No in every country the **#** sign is used!
If someone need it, better to do it via the text - otherwise it has to be deleted by an OCMod or Event!

yvonn: Not every country. If you want to use # for Order AND Invoice:
`<h1>{% if order.invoice_no %}{{ text_invoice }} #{{ order.invoice_no }}{% else %}{{ text_order }} #{{ order.order_id }}{% endif %}</h1>`

PR #9140, opened 2021-01-18
Typo voucher language

condor2:


danielkerr:
my grammer is correct you can not recieve from. its a gift sent from

PR #9156, opened 2021-01-21
Typo language ecb.php

condor2:


danielkerr:
should be Fixer is that is the name of the server.

condor2:
oh, ok

PR #9181, opened 2021-01-23
add missing 'text_category' to menu

SaifAlSubhi:
'text_category' is missing


condor2:
Better remove All and leave only Categories.

PR #9183, opened 2021-01-23
Added missing help text for pagination

condor2:


condor2:
If is not a mistake, I think this should be changed

**Determines how many catalog items** in **'Determines how many elements items**

PR #9184, opened 2021-01-23
text_edit cannot be found - catalog/language/en-gb/account/address.php file.

TheCartpenter:


TheCartpenter:
Found in controller but the output should be handled differently.

PR #9213, opened 2021-01-29
Typo 'store_url' value in mail/transaction controller

condor2:


TheCartpenter:
I am not sure how this is a typo. The startup folder still registers the config_url with the Opencart registry. However, I do notice that constants can be confusing with the use of OC objects since constants have more accessibility throughout the methods as opposed to the OC registry object.

condor2:
If you have two store this line is used to show store URL on mail

$store_url = $store_info['store_url'];

If you have only one store , this line is used
$store_url = HTTP_CATALOG;

Also was fixed.....
https://github.com/opencart/opencart/blob/master/upload/admin/controller/mail/transaction.php#L45

TheCartpenter:
It wasn't fixed, what you're suggesting to change from the original line is still there. However, since the $customer_info array is pulled to validate the store, the getSetting method should be pulled to ensure that the customer's store ID is still valid as opposed to the original store ID. Otherwise, not only the config_url from the $this->config->get could be wrong.

TheCartpenter:
Ok, I see what you mean.

In admin/controller/mail/reward.php file, line 42 should also read: url instead of store_url: https://github.com/opencart/opencart/pull/9230/commits/6c86aca973d9d150be9d88654fcd4a55dfa7ff22

TheCartpenter:
On the same file you're reporting as well, the store_url key in the store_info must be renamed to url.

TheCartpenter:
In fact, catalog/controller/mail/order.php file shows two results of:

```$store_url = HTTP_SERVER;```

instead of using HTTP_CATALOG . Each controllers should use this constant from the master branch instead.

TheCartpenter:
catalog/controller/mail/transaction.php file also should use HTTP_SERVER and $store_info['url'] instead of $store_info['store_url'] .

PR #9216, opened 2021-01-30
Typo design/layout model

condor2:


danijelGombac:
@condor2  I think that should be like this:

``WHERE `route` LIKE '" . $this->db->escape($route) . "' AND `store_id` = '" . (int)$this->config->get('config_store_id') . "'``

condor2:
Is weird if you use '' instead of `` are modules are not displayed...

condor2:
Or maybe should be...

``route` = '" . $this->db->escape($route) . "'`

cgcmart:
> 
> 
> Yes, I think is as you say.



> 
> 
> @condor2 I think that should be like this:
> 
> `` WHERE `route` LIKE '" . $this->db->escape($route) . "' AND `store_id` = '" . (int)$this->config->get('config_store_id') . "' ``

SQL syntax error

condor2:
That was an example.
Isn't entire command.

cgcmart:
> 
> 
> [2ebedf7](https://github.com/opencart/opencart/commit/2ebedf798ff17e483cd4de148f2799eb35e0ebc1)

SQL syntax error

condor2:
SQL syntax error where?
Work for me but I think better to be changed to

route = '" . $this->db->escape($route) . "'

like is in other files

cgcmart:
2021-01-30 15:33:13 - Exception:  Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '0' ORDER BY `route` DESC LIMIT 1' at line 1<br />Error No: 1064<br />SELECT * FROM `oc_layout_route` WHERE `route` LIKE 'common/home` AND `store_id` = '0' ORDER BY `route` DESC LIMIT 1 in .../system/library/db/mysqli.php on line 48

condor2:
Tested on 4.0 and 3.0.3.6 and there is no error.
Also tested SQL command  no issue
layout

Sure you have copied right?

cgcmart:
I have installed the the latest updated master version. Your pull is 1 line change from the origin. I restored the origin, the error is gone.

condor2:
You have cloned my fork?

Because I cloned master branch repo (from here) replaced that command...and no error.

danijelGombac:
I have website with changes and for me there is no error.

PR #9226, opened 2021-02-05
Validating store_id and customer_group_id - catalog/controller/api/customer.php file.

TheCartpenter:


eka7a:
Is there a need for this? I am not sure.
config_customer_group_id is already assigned in API calls.
Perhaps similar, config_store_id value can be set.

https://github.com/opencart/opencart/blob/3f76c28dee2982429eb682dde75e41fbf9170fe6/upload/catalog/controller/startup/application.php#L17-L21

TheCartpenter:
There absolutely is since we're talking about an API which can also be instantiated externally to gather the store ID. In addition, admin/view/template/sale/order_form.twig file contains the store_id as a GET request where the catalog/controller/api/customer.php file does not validate which means the store ID is being left in the dark and as decorative purposes that should not be. Which is why, by using the getSetting (or getValue in this case may work as well), we validate the store ID as opposed to the one being stored with the order especially from the commit I have posted about the catalog/controller/api/order.php file where the original affected store ID shows the default store ID as opposed to the selected store ID from the order_form in the admin which still confuses matter on ensuring the right store ID is selected during, either, the creation or the edition of the order from the admin while using the API as opposed to the checkout page from the catalog-end.

TheCartpenter:
In addition to the above, the $this->session->data['customer'] super global array is being validated afterwards, which includes the customer_group_id in the array. Which is why, both must be validated in the first place prior to create the customer super global rather than simply validate a super global as a dried result.

TheCartpenter:
Closed without closure notes.

PR #9230, opened 2021-02-05
Replaced store_url with url on line 42 - upload/admin/controller/mail/reward.php file.

TheCartpenter:


TheCartpenter:
getStore should be replaced with getSetting in order to avoid future confusion between the url and store_url key since the getSetting pulls the config_url in both IF statements whenever the case while the same key name gets pulled by different array variable names. There seem to be multiple controllers handling this confusion while the getSetting could simplify this issue.

PR #9235, opened 2021-02-06
Removed unused code localisation/country.php

condor2:
I don't see where filters are used when you edit a country. (country_form)


condor2:
And I think are more controllers like that.

PR #9246, opened 2021-02-08
Update confirm.php

Yeleup:
an unnecessary order is created every time you click

### DIFF
--- a/upload/catalog/controller/checkout/confirm.php
+++ b/upload/catalog/controller/checkout/confirm.php
@@ -312,7 +312,9 @@ class Confirm extends \Opencart\System\Engine\Controller {
 
 			$this->load->model('checkout/order');
 
-			$this->session->data['order_id'] = $this->model_checkout_order->addOrder($order_data);
+			if (!isset($this->session->data['order_id'])) {
+          $this->session->data['order_id'] = $this->model_checkout_order->addOrder($order_data);
+      }
 
 			$this->load->model('tool/upload');
 


TheCartpenter:
I don't recommend this change.

Yeleup:
instead of creating a new order every time, isn't it better to give from the session


And I don't see that it can break

TheCartpenter:
```instead of creating a new order every time, isn't it better to give from the session```

No, it's not. It's the way it must be since extensions could alter its course without it when creating an order.

```And I don't see that it can break```

Exactly, you don't see it until a specific extension will be altering it's course due to validations rather than insisting on pulling the order ID upon new created orders. This has been suggested a long time ago now and this concept cannot change because of it.

xtensions:
This change is not recommended because of 2 reasons
1) A user can go to external payment method and it can fail or cancelled by user. When user re tries payment, it will use the same order id. Not all payment gateways accept same order id more than once, and payment gateway will throw them an error. And user will not be able to complete checkout without logging out or changing the payment method.
2) Instead of using 
`
if (!isset($this->session->data['order_id'])) {
          $this->session->data['order_id'] = $this->model_checkout_order->addOrder($order_data);
      }
`
the better would be 
`if (!isset($this->session->data['order_id'])) {
          $this->session->data['order_id'] = $this->model_checkout_order->addOrder($order_data);
  }else{
      $this->session->data['order_id'] = $this->model_checkout_order->editOrder($this->session->data['order_id'],$order_data);
      }`
      Reason: If a user changes payment method and in between add/remove a product in cart, change shipping address or any edit, that will not get saved in Order Table. And this will bring huge errors. Also not editOrder does not have language and currency edit.
      
      So this change should be avoided.

TheCartpenter:
Even with the alternative suggested change, that still won't prevent problematic with the order IDs due to the Opencart API where the store IDs are not being looked up since the store ID is associated with the order ID in the end. Which means, neither alternative should be suggested.

PR #9254, opened 2021-02-11
Added default type module in extensions

Hkr32:


TheCartpenter:
Yes, that would be needed.

TheCartpenter:
I take it back. The module parameter does not need to be there since the extensions from the Marketplace are already taking the lead on the master branch over the extension/module folder and the extension/extension/module.php file.

danielkerr:
maybe add something like this later for all extension types.

PR #9255, opened 2021-02-11
Added actualization url after change type extensions

Hkr32:


TheCartpenter:
I am sure there are other ways than using a .replace from a type query in order to gather the extension code ... besides, the type query is validated in the IF statement of the JS code but the extension code is not before gathering the selected extension code in the process.

Hkr32:
Fixed updating the link after loading the page

TheCartpenter:
This part still brings concerns regarding the .data('code') affected but not validated prior to add the url: https://github.com/opencart/opencart/pull/9255/commits/36eb5b07e9d0a2d4302ac74fbcc65e53cf272ca6#diff-7cfcd971ecd94e6868744e8bda99eba0aed13e3d7a60656861da594b75c49e1eR47-R55 .

- The let url = ''; could be changed to: var url = ''; . Also ensure to leave a line space after the variable created
- Intend is currently set to 4 on the TWIG files in your pull request. Opencart uses intend of 2

Hkr32:
> * Indent is currently set to 4 on the TWIG files in your pull request. Opencart uses indent of 2

in original 4
https://github.com/opencart/opencart/blob/17d7d4774d68a7e08ee9531b953067936dc79a18/upload/admin/view/template/marketplace/extension.twig#L43-L184



> * The let url = ''; could be changed to: var url = '';

var in 2021? As you say...

TheCartpenter:
The .data('code') is still not validated from the database with Ajax prior to add the value to the URL before pushing its content from the select name="type" .

TheCartpenter:
In addition, $(this).val() has been replaced over the $(this).attr('href') which, in your scenario, removes the user_token from the URL.

Hkr32:
> In addition, $(this).val() has been replaced over the $(this).attr('href') which, in your scenario, removes the user_token from the URL.

where?

TheCartpenter:
Correct for the extension.twig file using indent of 4. Indent of 2 should of been used from the master branch.

As for the $(this)val(): https://github.com/opencart/opencart/blob/1e730b640af8e16ae61f869f3cd62c64166075c3/upload/admin/view/template/marketplace/extension.twig#L61 .

Replaced on your behalf since: https://github.com/opencart/opencart/pull/9255/commits/36eb5b07e9d0a2d4302ac74fbcc65e53cf272ca6#diff-7cfcd971ecd94e6868744e8bda99eba0aed13e3d7a60656861da594b75c49e1eR59 .

TheCartpenter:
Correction: Not 'href' but 'value' in the attr.

TheCartpenter:
However, the .replace with the URL from that JS function is not needed. These replacements can be done via Ajax Response in the success method with .html or .replaceWith which Opencart already uses from the core.

Hkr32:
> Replaced on your behalf since:

I haven't changed anything. I just turned around to wait for the page to load and added a function with its call

Hkr32:
> However, the .replace with the URL from that JS function is not needed. These replacements can be done via Ajax Response in the success method with .html or .replaceWith which Opencart already uses from the core.

have an example?

TheCartpenter:
> > Replaced on your behalf since:
> 
> I haven't changed anything. I just turned around to wait for the page to load and added a function with its call

Exactly. Again, that function that replaces a URL is not needed.

PR #9257, opened 2021-02-13
Use valid command in bash code block in INSTALL.md

likemusic:
Also small fix in markdown.


danielkerr:
i think needs to be 777

PR #9269, opened 2021-02-14
[Maintenance v3.x] - Changed indent to 4 - admin/index.php file.

TheCartpenter:


TheCartpenter:
No way ... who the hell set all (or most) of the indent of 2 on the controller files in the master branch???

TheCartpenter:
Indent of 8 on the maintenance branch instead of 4???

ADDCreative:
Indent should be a tab in PHP files. See: https://github.com/opencart/opencart/wiki/Coding-Standards#indentation

Whoever did pull request https://github.com/opencart/opencart/pull/9189/files didn't read the coding standards. It's probably that which is messing with your editor.

TheCartpenter:
It doesn't explain why the master and maintenance branch uses two different indents altogether whereas the .editorconfig file is also not being used on the maintenance branch as opposed to the master branch.

ADDCreative:
Both the master and maintenance branches are using one tab as indent. They are looking different to you because the master has the .editorconfig file and the maintenance doesn't. The maintenance branch will have the tab size as the editor default, 8 spaces in the case of the GitHub viewer/editor. Whereas the master branch has the .editorconfig file which has the indent_size set to 2. This will make any editor that reads .editorconfig and the GitHub viewer/editor to display the tabs as 2 spaces.

ADDCreative:
@WebkulOpencart This commit is wrong. It removes the correct tab indents and replaces them with 4 spaces, which wrong according to the coding standards. https://github.com/opencart/opencart/wiki/Coding-Standards#indentation

Please revert it.

WebkulOpencart:
@ADDCreative in my code editor it was showing correctly so I merged, Now, i reverted

PR #9280, opened 2021-02-14
Update .editorconfig to match Coding Standards

ADDCreative:
https://github.com/opencart/opencart/wiki/Coding-Standards#indentation

Always going to have issues due to the mixed indentation in the .twig files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,5 +8,9 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+indent_style = tab
+indent_size = 4
+
+[*.twig]
 indent_style = space
 indent_size = 2


stalker780:
Agree about mixed indentation in one file

Also I still don't understand why not use same indentation both in templates and php files. 
Why create rules/standards that are impossible to use in 99% code editors. So everybody just ignores them when modifying OC sources for decade 🤦‍♂️

TheCartpenter:
> Agree about mixed indentation in one file
> 
> Also I still don't understand why not use same indentation both in templates and php files.
> Why create rules/standards that are impossible to use in 99% code editors. So everybody just ignores them when modifying OC sources for decade 🤦‍♂️

It is a bit questionable on that side.

Hkr32:
you need to use either tabs or spaces!
It's high time to update the standards!
better to use spaces throughout the project!!!

TheCartpenter:
So make up your minds folks. Is it by space or by tabs?

ADDCreative:
> So make up your minds folks. Is it by space or by tabs?

It's clearly described here and has been that way for years.
https://github.com/opencart/opencart/wiki/Coding-Standards#indentation

Yes, it incompatible with .editorconfig, but until @danielkerr changes it you'll have to live with it. I suspect in the end .editorconfig with get removed.

I have to agree that all files should have the same single indentation method.

TheCartpenter:
In the mean time, I have completed the indent hierarchy with the indent of 2 with the admin and catalog TWIG files with the exception of the extension folder (to be completed). Once these changes made, at least, the less work needed to be done to switch all over the same indentation method.

Hkr32:
> Yes, it incompatible with .editorconfig, but until @danielkerr changes it you'll have to live with it. I suspect in the end .editorconfig with get removed.

Well, if we have a community, we can discuss everything and find a solution

Hkr32:
> In the mean time, I have completed the indent hierarchy with the indent of 2 with the admin and catalog TWIG files with the exception of the extension folder (to be completed). Once these changes made, at least, the less work needed to be done to switch all over the same indentation method.

What job? Press ctrl+s in the editor for the plugin to fix missing files in all files? Are you in a notebook that you do it with your hands each line?

Hkr32:
да уж...

PR #9379, opened 2021-02-24
Update cart.php

kurtulusoz:
When the discount and the campaign are applied together, the price does not work correctly. I revised the code and installed it.


kurtitasarim:
Cart module price discount and special in use together problem fixed.

TheCartpenter:
I don't agree with this change since specials can be mixed with discounts.

kurtitasarim:
When both a discount and a campaign are applied to the product, the basket does not calculate properly. you can try it as a test

Kurtuluş ÖZ 
Junior Developer ~ Linux Admin
www.kurtulusoz.com.tr ~ www.kurtitasarim.com

TheCartpenter:
If this issue is based on cart calculations, then mixing specials with discounts in the lookup is not the right course of action. More information is needed on how supporters can reproduce the issue you're addressing about.

PR #9383, opened 2021-02-25
update APPLICATION name

SaifAlSubhi:
You will get an error
Error: Could not load config Catalog! in opencart/system/engine/config.php
and
Error: Could not load config Admin! in opencart/system/engine/config.php

### DIFF
--- a/upload/install/controller/install/step_3.php
+++ b/upload/install/controller/install/step_3.php
@@ -15,7 +15,7 @@ class Step3 extends \Opencart\System\Engine\Controller {
 			$output  = '<?php' . "\n";
 
 			$output .= '// APPLICATION' . "\n";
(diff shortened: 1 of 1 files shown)


SaifAlSubhi:
> You will get an error
> Error: Could not load config Catalog! in opencart/system/engine/config.php
> and
> Error: Could not load config Admin! in opencart/system/engine/config.php

if 'catalog' is used then you get another error:
Exception: Error: Could not load model Opencart\catalog\Model\Extension\Opencart\Total\SubTotal! in /opencart/system/engine/loader.php on line 125

SaifAlSubhi (commented):
There is an error when using 'catalog' and another error when using 'Catalog'

PR #9394, opened 2021-03-01
Update tax.php

danijelGombac:


danielkerr:
not sure where the error would show up. better to force bool if string

danijelGombac:
@danielkerr 

https://github.com/opencart/opencart/blob/1e3604443a34b3507138543693d172830dff569c/upload/catalog/controller/product/product.php#L338

PR #9395, opened 2021-03-02
Generate time zones without setting default

ADDCreative:


seo-bro (commented):
This is only a partial solution, you need to fix it globally.
Since the time of mysql base and server php may differ

ADDCreative:
@seo-bro That would be a totally separate issue. This is just to stop unnecessarily calling date_default_timezone_set when generating the list of selectable time zones for the settings page.

The PHP and database time zones are set in the framework.
https://github.com/opencart/opencart/blob/d9fd6c0975a40dc07debc906fbe43780b1e51e68/upload/system/framework.php#L24-L25
https://github.com/opencart/opencart/blob/d9fd6c0975a40dc07debc906fbe43780b1e51e68/upload/system/framework.php#L120-L121

PR #9411, opened 2021-03-05
Develope

cahedral:
FIX: If all languages are deactivated, the Catalog fault. Now Catalog always set defatult in case error and in admin is forbiden to disable all languagues


cahedral:
Error in request

PR #9433, opened 2021-03-15
Added IIS compatibility for DOCUMENT_ROOT - admin/controller/common/security.php file.

TheCartpenter:


TheCartpenter:
Closing since this would create an array while a string has been affected.

PR #9437, opened 2021-03-15
Removed duplicated array_shift - admin/controller/design/translation.php file.

TheCartpenter:


ADDCreative:
@eka7a I believe the second array_shift is intentional. Both `$part[0]` and `$part[1]` have been used, so need removing.

TheCartpenter:
If so, then we'd need a comment to identify that's the case. Otherwise, all it shows in there is a duplicated line function.

PR #9449, opened 2021-03-15
Made various changes - admin/controller/marketing/coupon.php file.

TheCartpenter:


TheCartpenter:
There are forum users who recently claimed that even though the coupon status would be changed to: 'Active' from the admin form, the coupon either remains on the same status or even though if it's active, the coupon still does not respond. With this fix, this might be a possible solution due to server-specifics, or PHP versions, where Boolean types might be unrecognizable as opposed to integers.

TheCartpenter:
Closing. This commit may not be needed.

PR #9468, opened 2021-03-20
OC4 - Fix seo_url model

stalker780:
Broken &language parameter
`http://website.loc/macbook?language=en-gb`
instead of
`http://website.loc/en-gb/laptop-notebook/macbook`

Fix returning empty string instead of cached keywords


stalker780 (commented):
clear code

PR #9475, opened 2021-03-23
Update token generator

ADDCreative:
token by default was generating 64 random bytes and then only using 16 of them. No need to check or use openssl_random_pseudo_bytes as random_bytes is a core PHP function.


TheCartpenter:
By removing the SSL crypto random bytes functions ... ?

ADDCreative:
> By removing the SSL crypto random bytes functions ... ?

Yes, random_bytes is a core PHP function and will always be available in PHP 7 and above. No point generating the token twice. openssl_random_pseudo_bytes was only there for PHP 5.x which isn't supported by OpenCart anymore.

Issue #9363, opened 2021-02-18
Product appears in all url variations


Farmshop:
**What version of OpenCart are you reporting this for?**
3.0.3.3
**Describe the bug**
If SEO is enabled, the SEO URLs of the categories can be changed in any combination in the title bar of the browser, and the product page will always be displayed.

**To Reproduce**
1. Nanavigate to 'a product trough its relevant categories for example: /demo/tablet/samsung-galaxy-tab-10-1'
2. Change 'the tablet to monitor in the url like this: /demo/monitor/samsung-galaxy-tab-10-1'
3. See error 'Product page still opens'

**Expected behavior**
I just expected a message like this: The page you were looking for could not be found.

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

**Server / Test environment (please complete the following information):**
 - Deployed to a web server.
 - Linux CentOS
 - PHP version 7.3.17
 - Apache 2.2
 - Browser(s) tested with Chrome
 - Version 88.0.4324.182

**Additional context**

WebkulOpencart:
@Farmshop I found the issue and checking the same.

PR #9478, opened 2021-03-24
fixed the product page URL issue

WebkulOpencart:
fix [https://github.com/opencart/opencart/issues/9363](https://github.com/opencart/opencart/issues/9363)

### DIFF
--- a/upload/catalog/model/catalog/product.php
+++ b/upload/catalog/model/catalog/product.php
@@ -538,10 +538,7 @@ class ModelCatalogProduct extends Model {
 	}
 
 	public function checkProductCategory($product_id, $category_ids) {
-
-		$category_id = implode("', '", $category_ids);
-
-		$query = $this->db->query("SELECT * FROM " . DB_PREFIX . "product_to_category WHERE product_id = '" . (int)$product_id . "' AND category_id IN('$category_id')");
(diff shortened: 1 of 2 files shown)


upload/catalog/model/catalog/product.php
```diff
@@ -536,4 +536,13 @@ public function getTotalProductSpecials() {
...
 		}
 	}
+
+	public function checkProductCategory($product_id, $category_ids) {
+
+		$category_id = implode("', '", $category_ids);
+
+		$query = $this->db->query("SELECT * FROM " . DB_PREFIX . "product_to_category WHERE product_id = '" . (int)$product_id . "' AND category_id IN('$category_id')");
+
+		return $query->row;
```
TheCartpenter, line 546: ```return $query->rows;```

As for the IN statement, I would suggest the use of catalog/model/account/download.php file the way the $implode with AND, OR statement are being used originally.

WebkulOpencart: I am checking for the same.

TheCartpenter: Not sure what that means but it seem you have a few 'for the same' in your replies there. Does this mean you'll make the same or you believe it's the same?

upload/catalog/model/catalog/product.php
```diff
@@ -536,4 +536,13 @@ public function getTotalProductSpecials() {
 			return 0;
 		}
 	}
+
+	public function checkProductCategory($product_id, $category_ids) {
+
```
TheCartpenter, line 541: Remove empty space. No models have spaced areas between the public function and the next line.

PR #9479, opened 2021-03-24
Replaced string js to html template

Hkr32:
It is also worth taking out all the js-code in a separate file and connecting it on the page
It is also worth moving the template into a separate file
@danielkerr, what do you think?

### DIFF
--- a/upload/admin/view/template/catalog/product_form.twig
+++ b/upload/admin/view/template/catalog/product_form.twig
@@ -2258,6 +2258,10 @@
   </tr>
 </template>
 
+<template id="template-form-upload">
+  <form enctype="multipart/form-data" id="form-upload" class="hidden"><input type="file" name="file"></form>
+</template>
+
 <script type="text/javascript">
 $(window).on('load', function () {
   $('textarea[data-toggle="ckeditor"]').ckeditor();
@@ -2801,8 +2805,8 @@ $(window).on('load', function () {
     let element = this;
 
     $('#form-upload').remove();
-    // @todo
-    $('body').prepend('<form enctype="multipart/form-data" id="form-upload" style="display: none;"><input type="file" name="file" /></form>');
+    let form_upload = document.querySelector('#template-form-upload').content.querySelector('form').cloneNode(true);
+    $('body').prepend(form_upload);
 
     $('#form-upload input[name="file"]').trigger('click');
 
--- a/upload/admin/view/template/catalog/product_form.twig
+++ b/upload/admin/view/template/catalog/product_form.twig
@@ -1,5 +1,6 @@
-{{ header }}{{ column_left }}
-<div id="content">
+{{ header }}
+{{ column_left }}
+<div id="content" data-user_token="{{ user_token }}" data-master_id="{{ master_id }}">
   <div class="page-header">
     <div class="container-fluid">
       <div class="float-right">
@@ -780,7 +781,8 @@
                 <label class="col-sm-2 col-form-label">{{ entry_manufacturer }}</label>
                 <div class="col-sm-10">
                   <div class="input-group">
-                    <input type="text" name="manufacturer" value="{{ manufacturer }}" placeholder="{{ entry_manufacturer }}" id="input-manufacturer" class="form-control"/> <input type="hidden" name="manufacturer_id" value="{{ manufacturer_id }}"/>
+                    <input type="text" name="manufacturer" value="{{ manufacturer }}" placeholder="{{ entry_manufacturer }}" id="input-manufacturer" class="form-control" data-text_none="{{ text_none }}" />
+                    <input type="hidden" name="manufacturer_id" value="{{ manufacturer_id }}"/>
                     {% if master_id %}
                       <div class="input-group-append">
                         <div class="input-group-text">
@@ -1003,12 +1005,12 @@
                               <textarea name="product_attribute[{{ attribute_row }}][product_attribute_description][{{ language.language_id }}][text]" rows="5" placeholder="{{ entry_text }}" class="form-control">{{ product_attribute.product_attribute_description[language.language_id] ? product_attribute.product_attribute_description[language.language_id].text }}</textarea>
                             </div>
                           {% endfor %}</td>
-                        <td class="text-right"><button type="button" onclick="$('#attribute-row{{ attribute_row }}').remove();" data-toggle="tooltip" title="{{ button_remove }}" class="btn btn-danger"><i class="fas fa-minus-circle"></i></button></td>
+                        <td class="text-right"><button type="button" data-toggle="tooltip" title="{{ button_remove }}" class="btn btn-danger"><i class="fas fa-minus-circle"></i></button></td>
                       </tr>
                       {% set attribute_row = attribute_row + 1 %}
                     {% endfor %}
                   </tbody>
-                  <tfoot>
+                  <tfoot data-attribute_row="{{ attribute_row }}">
                     <tr>
                       <td colspan="2"></td>
                       <td class="text-right"><button type="button" id="button-attribute" data-toggle="tooltip" title="{{ button_attribute_add }}" class="btn btn-primary"><i class="fas fa-plus-circle"></i></button></td>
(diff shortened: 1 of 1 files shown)


TheCartpenter:
Way too complicated to manage. Uses indent of 2 instead of 4, empty extra lines and stick together codes are found in your suggestion. Also, the use of <template id is not a common way being regularly handled in Opencart.

Hkr32:
> Way too complicated to manage

If it's hard for you, don't manage... and go study

> Uses indent of 2 instead of 4

See .editorconfig... it's not me doing it, the plugin does it

> Also, the use of <template id is not a common way being regularly handled in Opencart.

then it's time to start doing the right thing, and not how it was when it was php 4

osworx:
> then it's time to start doing the right thing, and not how it was when it was php 4

When saying "doing the right thing" and " .. was php 4 ..", then do it correct!
And delete such stupid things like
`<script type="text/javascript"><!--`
and use:
`<script>`

As well:
`//--></script>`
and use:
`</script>`

Then instead of `var user_token` use `let user_token` or `const user_token` (depending on script and usage).

Also things like 
`$('input[name=\'product_attribute[' + row + '][name]\']').autocomplete({` 
are completely nonsens, better: 
`$('input[name="product_attribute[' + row + '][name]"]').autocomplete({`

Also get rif of that repeated 
```
//--></script>
<script type="text/javascript"><!--
```

And such could be done also better:
`html = '<div id="modal-option" class="modal fade">';
html += '  <div class="modal-dialog">';`

And how many times you will do such: `var user_token = $('#content').data('user_token');`

Everthing found in this pull request.
You may see, doing things better is more ..

As you already said: **then it's time to start doing the right thing**

Hkr32:
I agree with you. 
To use const/let instead of var, you need to change the standards in opencart... And they don't want to do this

> And how many times you will do such
forgot to move everything to one place

> And such could be done also better:
apparently missed, I will revise today

osworx:
> To use const/let instead of var, you need to change the standards in opencart... And they don't want to do this

Okay, fair argument.
But - what IS standard in OpenCart?
We have 2021 .. and I use in all of my extension only const and let for example.
Don't think that Daniel will not have it - or?

Hkr32:
> But - what IS standard in OpenCart?

https://github.com/opencart/opencart/wiki/Coding-Standards

I used a const/let in one PR, someone wrote to me that I needed to use a var, therefore I did not change it, but in general I am for a const/let)
In general, there is a lot of things there is high time to update, now the PSR and choose either spaces or tabs in all project

osworx:
Thx, I know that "rules".
But they say nothing about Javascript and how to initialize variables and so on.
So, I would say: go ahead with "modern" and "actual" coding like we do in 2021 and not as it would be 1998, php 4, no PSR and if you would be from India, Pakistan and so on ..

TheCartpenter:
Talking about code evolution is one thing but if people are going to start talking about it then, at least, to respect the stick codes versus empty line of codes the way Opencart already provides originally while suggesting new codes.

Hkr32:
@osworx, fixed, is there anything else you haven't done?

TheCartpenter:
Still contains stick codes around lines 2600x

osworx:
> 
> 
> Still contains stick codes around lines 2600x

Beside this, there is much more work in front!
For example, repeated language strings eveywhere.
Repeated code (e.g. breadcrumbs) everywhere.
And so on.

While as it seems 4.x is on a better way to release once a "beta" package, before the final release this should be considered to simplifier.
But has to be discuseed with Daniel before, otherwise any comma of work is too much.

Hkr32:
> But has to be discuseed with Daniel before

So how do you do it? Discussions would be open in the repository, one could ask there. A lot of PRs have been done, they also need to be edited, the code style should be brought to a modern look. But in PR, no one writes what is wrong and what to redo, basically only one pseudo developer here goes and writes to everyone about spaces and that's it (

osworx:
Basically when I develop my Extensions, following ES6 (js) and single functions (php - whee possible).
Try to avoid repetetions as much as possible.
So I have several (sub) functions for different tasks, e.g. here the breadcrumbs:
```php
private function getBreadcrumbs( &$data ) {
		$data['breadcrumbs'] = [
			[
				'text' => $this->language->get( 'text_home' ),
				'href' => $this->url->link( 'common/dashboard', 'user_token=' . $this->session->data['user_token'], true )
			],
			[
				'text' => $this->language->get( 'text_extension' ),
				'href' => $this->url->link( 'extension/extension&type=payment', 'user_token=' . $this->session->data['user_token'], true )
			],
			[
				'text' => $this->language->get( 'text_title' ),
				'href' => $this->url->link( $this->_route, 'user_token=' . $this->session->data['user_token'], true )
			]
		];
	}
```

And call them with: 
```php 
$this->getBreadCrumbs( $data );
```

`$this->_route`is also an internal variable used at many places inside my scripts.

I know that this is not the "OC Standard" (when looking at the core code), but cleaner and simplier to read Extensions are - for me - much more important.
And it makes the development much cleaner, faster and easier - when you have more than 160 extensions this will save you a lot of time!
Of course I could add - like some other do - a "global" library, had this at many OC 2.x extensions but decided to get back to internal functions.

Have read all discussions here, also that about the indents and spaces.
Wasting time at the end (while, if the OC standards would be clean, modern and easy to follow), all devs should follow them.
At the end, if someone has one more or less space - who cares?

To summarize: this all are "basic" questions and tasks which can be confirmed only by Daniel when the code should cleaner, packages therefore (much) smaller and modern (following latest recoomendations and specifications.

stalker780:
@osworx 

Yes. Events system requires centralized data processing. More libraries and controllers needed to exclude repeated code.
product_thumb, breadcumbs, and many more.

Also it would be good to use automated model/controller/library loaders inside registry.

There must be no string loaders for controllers and models. Libraries must have routes too.

So functions can be called:
`$this->controller_product_category($params);`
or 
`$this->controller_product_product->review($params);`

Same for models. Remove repeated `$this->load->model('catalog/product')` from code;
Just call `$this->model_catalog_product->getProducts($params)`

Same for libraries.
`$this->library_cart->getProducts($params);`
`$this->library_response->output();`
```
$image_class = $this->library_image($params);
$image_class->resize($params;)
```


Where:
1. system/engine/controller.php->__get() parses called route, 
2. checks if this controller/model/library exists,
3. includes/loads it once, runs constructor once,
4. calls before and after events for called function.
5. or throws exception if this controller/model/library/function does not exist or cannot be called.


This will make code much more easy to read and to debug.
In my opinion, calling functions, using their string names like $this->load->controller('myroute/mycontroller|myfunction', $params) is unnatural in object programming.

But I am afraid, this will never happen :(

TheCartpenter:
Yes, this was already covered with Daniel before. The problem are the new instances loaded with the paths and methods altogether. For instance, let's say we have: new Cart\User . All these 'new' calls could be used with an object from the new Action from the action engine instead.

mhcwebdesign:
@stalker780: Just curious: 

Without a **$this->load->model(...)** how do you intend to create instances of model classes?

And how do you intend to map something like **$this->controller_xxx_yyy_zzz** to the correct folder? It could be **controller/xxx/yyy_zzz.php**, or **controller/xxx_yyy/zzz.php**, or **controller/xxx_yyy_zzz.php**.

stalker780:
> @stalker780: Just curious:
> 
> Without a **$this->load->model(...)** how do you intend to create instances of model classes?
> 
> And how do you intend to map something like **$this->controller_xxx_yyy_zzz** to the correct folder? It could be **controller/xxx/yyy_zzz.php**, or **controller/xxx_yyy/zzz.php**, or **controller/xxx_yyy_zzz.php**.

1. Loader will still exist, but it must be called once on first access to model methods/properties, imagine it as before_event.
2. How it works with models now? Same for controller_ and library_ prefixes.

Basically, my proposal is to use models structure also for controllers and libraries. But add automatic loader on first call.

TheCartpenter:
@stalker780: The problematic with automation being loaded as the first instance are with the extensions. How will developers know the first instance being loaded until the last if those declared instances are already in the loads versus the new ones especially with the controllers? Initially, such objects should only be used with libraries as opposed to other types to ensure no extensions gets impacted in their distributions since, in this case, regardless of the Events being used with /before or /after, multiple identical calls per methods calling other controllers, models or libraries could be in the ways of users running the stores.

eka7a:
It's been too long and there's a lot of conflicts. That's why I'm closing it. Maybe it can be resubmitted, but I don't think @danielkerr   would want that change.

PR #9480, opened 2021-03-25
Update product.php

Yeleup:
Events getProduct are not triggered if the method is called inside the class

### DIFF
--- a/upload/catalog/model/catalog/product.php
+++ b/upload/catalog/model/catalog/product.php
@@ -169,7 +169,7 @@ class Product extends \Opencart\System\Engine\Model {
 		foreach ($query->rows as $result) {
 			// for never get one more time with same product id
 			if (!isset($product_data[$result['product_id']])) {
-				$product_data[$result['product_id']] = $this->getProduct($result['product_id']);
+				$product_data[$result['product_id']] = $this->model_catalog_product->getProduct($result['product_id']);
 			}
 		}
 
@@ -220,7 +220,7 @@ class Product extends \Opencart\System\Engine\Model {
 		$query = $this->db->query($sql);
 
 		foreach ($query->rows as $result) {
-			$product_data[$result['product_id']] = $this->getProduct($result['product_id']);
+			$product_data[$result['product_id']] = $this->model_catalog_product->getProduct($result['product_id']);
 		}
 
 		return $product_data;
@@ -233,7 +233,7 @@ class Product extends \Opencart\System\Engine\Model {
 			$query = $this->db->query("SELECT p.`product_id` FROM `" . DB_PREFIX . "product` p LEFT JOIN `" . DB_PREFIX . "product_to_store` p2s ON (p.`product_id` = p2s.`product_id`) WHERE p.`status` = '1' AND p.`date_available` <= NOW() AND p2s.`store_id` = '" . (int)$this->config->get('config_store_id') . "' ORDER BY p.`product_id` DESC LIMIT " . (int)$limit);
 
 			foreach ($query->rows as $result) {
-				$product_data[$result['product_id']] = $this->getProduct($result['product_id']);
+				$product_data[$result['product_id']] = $this->model_catalog_product->getProduct($result['product_id']);
 			}
 
 			$this->cache->set('product.latest.' . (int)$this->config->get('config_language_id') . '.' . (int)$this->config->get('config_store_id') . '.' . $this->config->get('config_customer_group_id') . '.' . (int)$limit, $product_data);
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
This is not a bug. Rather, it's a design decision.

stalker780:
Events fire only on
$this->load->view
$this->load->controller
$this->load->library
$this->model_catalog_product->myFunction()

Why still need $this-load-model('mymodel') everywhere in the code. Why not autoload them on calling model functions?

There is still no events solution for libraries. For example you cannot change a result of $this->cart->getProducts()

Well, this design just shows, that there are lots of situations, where you cannot create complicated extensions/themes without OCMod/VQMod.

Events look good. But OC code must be changed, making absolutely every function triggering an event. 
Otherwise there are lots of developers deadends, where an only way is to cut through the core sources.

mhcwebdesign:
Well, the current limitation with events can be a blessing in disguise. it will stop the many poor-quality extensions from creeping up for OpenCart 4, especially for web themes.

Having said that, I am planning to port our Override Engine to OpenCart 4. It will use a factory design pattern and object-oriented class extensions for modifications.

stalker780:
> Well, the current limitation with events can be a blessing in disguise. it will stop the many poor-quality extensions from creeping up for OpenCart 4, especially for web themes.
> 
> Having said that, I am planning to port our Override Engine to OpenCart 4. It will use a factory design pattern and object-oriented class extensions for modifications.

That's the problem. Without having core mechanism for modifications, every developer will invent his own bicycle.

And again it will be a Hell of extensions and core modifications on the market.

stalker780:
I don't know why, but it seems Daniel does not care, how developers will work with OC4.

Though I believe extensions Marketplace brings him a lot of gold.
Why not help his peasants to bring him even more gold?
Just place a mine closer :)

mhcwebdesign:
You'll never accomplish the required core modification capabilities with event handlers alone, even if event triggers were added to OpenCart. Only the ability to do proper class extensions, or perhaps a decorator design pattern, will be able to accomplish this. VQmod/OCmod were a recipe for disaster.

However, I do agree that OpenCart should make use of the loader->library for the instantiation of the library singleton classes.

stalker780:
I don't see any disaster with VQMod used for a decade. On the contrary, it made OC even more popular and advanced.

Shitcode extensions will always be there. No matter what code structure Daniel makes.
All CMS have 80% of shitty extensions. And they still live. Look at Joomla, Wordpress, Drupal, etc.

This only shows that entry point for developing OC extensions is very low and this CMS is very popular, this results in a fact that even unqualified developers can do something more or less workable in OpenCart. I don't think this is bad. It's just life.

mhcwebdesign:
Yeah, I had the same discussion with Qphoria 10 years ago when he invented VQmod. XML is a markup language suitable for document structures and protocols, it's not intended to be as a programming language. It's as bad as XLST. It's time to go back to proper design patterns.

I'll write up a proposal for having the singleton classes to be instantiated by the loader->library which has event triggers.

stalker780:
Agree, XML is not a best container for PHP, HTML, JS codes.

But, good extension does not need hundreds lines of code to be stored inside XML.
Just 1-3 lines to inject your controller/model call in a needed place.

I saw a 2 megabyte XML once. That was a mess. Facepalm

Still VQMod is a most powerful tool for core modifications. And I think it will be used, until Daniel can propose anything instead.
OCMod was even worse, with all it's caches, manual refreshes, sources in DB. Yuck!

chulcha:
> OCMod was even worse, with all it's caches, manual refreshes, sources in DB

This is a plus 
You don't need a runtime to modify 

The topic here is that
this->method != $this->nodel->method 


And you can talk about what is better, and what is best, as much as you like, never having heard Daniel's opinion.
And how he plans to combine the development of different developers

TheCartpenter:
> Well, the current limitation with events can be a blessing in disguise. it will stop the many poor-quality extensions from creeping up for OpenCart 4, especially for web themes.
> 
> Having said that, I am planning to port our Override Engine to OpenCart 4. It will use a factory design pattern and object-oriented class extensions for modifications.

@danielker: Is this really happening to integrate a 3rd party engine with the OC Registry?

mhcwebdesign:
@TheCartpenter: 3rd party core engines like the Override Engine, while making use and fully supporting the registry, as well as being fully compatible with all the other OpenCart features, will still be a replacement core engine, and we plan to make it freely available under a GPL license on e.g. the OpenCart marketplace, just like we did it for previous OpenCart releases.

TheCartpenter:
@mhcwebdesign: Which is why, on the above, I asked for confirmation with the term ... 'we' for that to happen.

PR #9485, opened 2021-03-27
security js XSS

BuslikDrev:


ADDCreative:
This one also need changing. https://github.com/opencart/opencart/blob/a0a152758ddeaa4e8d884a30429727319a7dc1ea/upload/admin/controller/startup/session.php#L26

Also 'secure' should not be based on the request. Needs session_secure adding to the config. Or maybe set if HTTP_SERVER starts with https.

SameSite=Strict will also cause problems with some payment gateways. As the session will be lost on return. Even SameSite=Lax will cause problems with gateways that return using POST. session_samesite needs adding to the config and using ```'SameSite' => $config->get('session_samesite')```

SaifAlSubhi:
Hi,
what about this?
https://github.com/opencart/opencart/blob/58bae8e0adcbeac55243375f6cfd62bfa5e346bb/upload/php.ini#L13
keep it off?

PR #9540, opened 2021-04-02
Fixed public save() method - admin/controller/user/user_permission.php file

TheCartpenter:


TheCartpenter:
Task completed. However, I am not sure why $json['error']['x'] is used if the APIs handled with JSON can only support a single key at a time.

stalker780:
Stop spamming!

Why are you creating hundreds of requests?

Why not pull multiple files in one request?

TheCartpenter:
Because the commit title lengths are too limited in size.

ADDCreative:
> Because the commit title lengths are too limited in size.

That's why there is a description field.

Doing single file pull requests is just wasting everybody else's time.

TheCartpenter:
It if it'd be that simple, then the description would also show the filename directly on the headline. AFAIK, the subject line is what displays the headline, not from the description. The description is shown once you click on the headline and redirected to the next page only.

ADDCreative:
You don't need the filename in the subject. It's clear from the pull request what files it changes. It's clear you are just using the GitHub online editor. Please be nice to everybody else and use something like https://desktop.github.com/ and do a pull request that makes changes to multiple files.

TheCartpenter:
Of course, I am using the online editor. It is the simplest way to suggest changes to the repository. While one might believe it would be easy to track each files without enough space to put the notes in, it would then be less easier to track each files from a single commit in this case.

However, thanks for your concern but I am simply considering to keep the direct route to publicize to post the pull requests rather than taking all the different roads for a single commit to be pulled.

ADDCreative:
It may be easier for you. But you are making it harder for the developer. Now they have the hassle of reviewing and accepting 20 odd pull request when the should only need to accept one. You don't need to track single files.

TheCartpenter:
Well, on the other around, it might be harder for the developers that do want to contribute while dealing with multiple files into a single commit. Whichever's the easiest path for contributors, the better it is.

ADDCreative:
It's quicker and easier to do in one commit if your know how.

TheCartpenter:
All over Google, not an issue. Just too many process for one commit.

TheCartpenter:
Now fixed from this commit: https://github.com/opencart/opencart/commit/91011ff59f44cde0b87d7903857bc57cdacf6f4d . Therefore, this commit will now be closed.

Hkr32:
Ты так и не понял, что все что ты делаешь это никому не нужная и крайне бесполезная работа... Только спам в почте от тебя... Найди себе другую песочницу для своих побед и не мешай людям работать!

PR #9552, opened 2021-04-05
Fixes "Error: MAIL FROM not accepted from server!" with SMTP.

chetanmadaan:
This fixes the error "Error: MAIL FROM not accepted from server!" when using SMTP with Amazon SES, Sendgrid or any other provider where the username to authenticate with the SMTP server is not as same as the from email address.

I understand you guys may not want to use external libraries in your code, but this is where time tested stuff like PHPMailer comes very handy.

### DIFF
--- a/upload/system/library/mail/smtp.php
+++ b/upload/system/library/mail/smtp.php
@@ -169,9 +169,9 @@ class Smtp {
 			}
 
 			if ($this->verp) {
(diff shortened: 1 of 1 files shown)


ADDCreative:
@WebkulOpencart Why fix in on branch and no the other?

WebkulOpencart:
@ADDCreative We only manage the Maintenance branch and master branch PR will be merged by Daniel.

By mistake, the master branch PR was closed as they have the same subject.

PR #9563, opened 2021-04-06
Removed empty space on line 97 - upload/admin/controller/common/reset.php file

TheCartpenter:


TheCartpenter:
File no longer exists. Closing.

PR #9654, opened 2021-05-12
Multiple slashes - .htaccess.txt

ILDaviz:
Is it a good idea edit .htaccess and add this adjustment?
Removes multiple leading slashes directly after the domain name / TLD
Remove multiple slashes anywhere in the rest of the path
Is this adjustment a possible problem for google or seo?

Exemple:
https://demo.opencart.com/////////
or
https://demo.opencart.com/////////index.php?route=product///////product&product_id=40

(Big thanks for this software! <3)


danielkerr:
not sure this matters

PR #9655, opened 2021-05-12
OC4: Proposal to add extension code before module

stalker780:
Add extension code to module code settings.

This will make module code unique in oc_setting table.

Different developers can use same module codes for their extensions.
Like blog, filter, banner, slider, shipping, coupon, etc. And rewrite module settings for each other.

This pr is just a concept for @danielkerr. This requires modification of all default opencart extensions templates and controllers.

oc_module table already contain extension code. I think this should be done for oc_setting too.

### DIFF
--- a/upload/admin/controller/extension/module.php
+++ b/upload/admin/controller/extension/module.php
@@ -70,7 +70,7 @@ class Module extends \Opencart\System\Engine\Controller {
 				if ($module_data) {
 					$status = '';
 				} else {
-					$status = $this->config->get('module_' . $code . '_status') ? $this->language->get('text_enabled') : $this->language->get('text_disabled');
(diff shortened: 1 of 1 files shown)


danielkerr:
its fine. maybe get them to add soem checks on install

BuslikDrev:
> its fine. maybe get them to add soem checks on install

Make extension code everywhere like you did in modules: "opencart.bestseller".
Example:
$this->model_setting_setting->editSetting('opencart.module_category', $this->request->post);
Otherwise, it makes no sense to put extensions in a separate folder.
It seemed convenient to put extensions in a separate folder, but in the end it is only convenient for those who want to steal an extension from a client.

PR #9657, opened 2021-05-13
Free shipping based on country/postcode

makoehr:
Enable free shipping for a number of combinations of country/postcodes.
This allows a flexible way to provide free shipping options.

Currently, the list of country/postcodes has to be added manually to the corresponding table in the database.


TheCartpenter:
Checkout can already collect entered information prior to the order table without the need to store postcodes. In addition, the suggestion above is not GDPR / CCRA compliant.

PR #9668, opened 2021-05-19
fixed issue 9597

webkul-dilip:
fixed issue 9597


danijelGombac:
Why not 

`public function query($sql, $params = []) {`

TheCartpenter:
Because the maintenance branch, compared to master branch, don't have these types of arrays converted in the core yet since the rest of the platform still uses array() at this time.

PR #9686, opened 2021-05-30
Voided and Cancelled on Returns

haxcop:
Voided and Cancelled should be part of a Returned orders Too...

### DIFF
--- a/upload/install/opencart.sql
+++ b/upload/install/opencart.sql
@@ -3190,8 +3190,8 @@ CREATE TABLE `oc_return_action` (
 INSERT INTO `oc_return_action` (`return_action_id`, `language_id`, `name`) VALUES
 (1, 1, 'Refunded'),
 (2, 1, 'Credit Issued'),
-(3, 1, 'Replacement Sent');
-
+(3, 1, 'Replacement Sent'),
+(4, 1, 'Voided');
 -----------------------------------------------------------
 
 --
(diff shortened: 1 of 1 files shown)


WebkulOpencart:
@haxcop Admin can manually created these return status as per his need so no need to add those field in opencart.sql file. so I am closing this PR

haxcop:
@WebkulOpencart on which branch you said this it? because at least on 3.0.3.6  there's no option I can add new return status, also even if exist such option and I missed, These two additions to the newly created installation should exist by default, If an item is returned to the store, the store should have the option to decline the return once received if it doesn't meet his policy/criteria. 

So Return status 
and 
Return Actions
 should contain VOIDED and CANCELLED

See 

Current Opencart fresh install V 3.0.3.7

Screen Shot 2021-07-08 at 11 59 45 AM
Screen Shot 2021-07-08 at 11 59 35 AM

With the proposed changes

Screen Shot 2021-07-08 at 12 02 21 PM
Screen Shot 2021-07-08 at 12 02 13 PM

WebkulOpencart:
@haxcop  You can get these options inside admin -> system -> localisation -> returns

Please check this file for reference [https://github.com/opencart/opencart/blob/3.0.x.x_Maintenance/upload/admin/controller/localisation/return_status.php](https://github.com/opencart/opencart/blob/3.0.x.x_Maintenance/upload/admin/controller/localisation/return_status.php)

haxcop:
@WebkulOpencart Thanks for the quick reply. Yes I can see that, but my point here still valid as this is an **improvement** of the UX/UI and **doesn't affect anything**, this is a missing part of a shop/customer journey on any store

WebkulOpencart:
@haxcop we need the @danielkerr confirmation to add these options.

haxcop:
@WebkulOpencart Thanks for the review

PR #9688, opened 2021-05-31
Added unique fields in for interception in events

halfhope:
Added key layout_module_id in array $data['layout_modules'] because I can't identify unique items while use event Controller/design/layout/getForm/after.


TheCartpenter:
Since the branch patch 5 has been deleted, shouldn't this commit be closed?

TheCartpenter:
Ok, already closed.

Issue #9691, opened 2021-06-01
TypeError: Unsupported operand types: string * int


cgcmart:
**What version of OpenCart are you reporting this for?**

Master branch

**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?

Almost can't go anywhere because of error message

TypeError:  Unsupported operand types: string * int
.../admin/controller/report/online.php
Line: 76

TypeError:  Opencart\Catalog\Model\Tool\Image::resize(): Argument #2 ($width) must be of type int, string given in .../catalog/model/tool/image.php on line 4

Undefined array key "customer_token" in .../catalog/controller/account/address.php on line 138

TypeError:  Unsupported operand types: string * int
.../admin/controller/user/user_permission.php
Line: 96

TypeError:  Unsupported operand types: string * int
.../admin/controller/catalog/product.php
Line: 163

TypeError:  Unsupported operand types: string * int
.../admin/controller/customer/customer.php
Line: 189

**To Reproduce**

**Expected behavior**

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server?
 - Operating system Debian 10
 - PHP version 8.0.5
 - Apache version Apache 2.4.38
 - Browser(s) tested with [e.g. chrome, safari - if applicable]

**Additional context**

danielkerr:
i think all this stuff is already fixed.

PR #9692, opened 2021-06-03
Made various changes - admin/controller/sale/order.php file

TheCartpenter:


cgcmart (commented):
your changes are incomplete:
L16, 24: remove (int)

L252: add (int) for correct TypeError

L381: add (int) for correct TypeError, and code for DivisionByZeroError. See the issue #9691

TheCartpenter:
Which is why, these changes weren't included since an issue was already addressed about it whereas a Pull Request should of been addressed about it.

TheCartpenter:
This file contains the most empty lines on the entire platform. Codes needs to be cleaned.

PR #9703, opened 2021-06-10
OC4: Proposition for addStyles and addScript

stalker780:
1. Make possible adding styles to footer.
2. Make possible adding scripts properties, like `$prop = 'sync defer onload="myFunc();" id="my_script_id"'`

This is a must have for frontend templates and extensions, loading external resources.

This is a concept which changes array structure for both scripts and styles. So it requires simple changes but in lots of controllers preparing twig templates. Basically for addStyles and getStyles.

### DIFF
--- a/upload/system/library/document.php
+++ b/upload/system/library/document.php
@@ -104,8 +104,8 @@ class Document {
 	 * @param	string	$rel
 	 * @param	string	$media
      */
-	public function addStyle(string $href, $rel = 'stylesheet', $media = 'screen'): void {
-		$this->styles[$href] = [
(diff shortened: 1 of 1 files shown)


TheCartpenter:
If addStyles or addScripts has a default parameter with its variable, then getStyles wouldn't need one.

chulcha:
> 
> 
> If addStyles or addScripts has a default parameter with its variable, then getStyles wouldn't need one.

Does this somehow change the idea of the proposal?

TheCartpenter:
The rest of the proposal remains to be seen.

danielkerr:
its fine as it is,

PR #9704, opened 2021-06-11
FIX: When all language are DISABLED the Web Catalog (frontend) show an error

cahedral:
When All language are DISABLED, the website catalog show an error. This patch set an default language and avoid set all language to DISABLE.


TheCartpenter:
The $this->request->post['language_id'] should be sanitized as (int) and, perhaps, put in a variable to avoid sanitizing the same variable everywhere in these modifications. The rest goes to clean codes.

cahedral:
Nerd I do something to merge this PR?

cahedral:
Close this PR due to inactivity

Issue #9591, opened 2021-04-08
[Master Branch OC4] Please re-work extension namespaces and other extension issues


mhcwebdesign:
The extension name spaces are quite weird at the moment. For example, the **extension/opencart/admin/controller/module/account.php** currently uses this namespace:

`namespace Opencart\Admin\Controller\Extension\Opencart\Module;`

The namespaces should really fall in line with the underlying folder structure, as is common with many other packages like Symfony, PHPWord, or with Java packages. I suggest the following namespace convention for OC4 extensions:

OC4 extension namespaces should always start with Opencart\Extension\<extension-name>\<application>\

For example, above module would use this:

`namespace Opencart\Extension\Opencart\Admin\Controller\Module;`

If a 3rd party extension were to introduce a new extension names 'foo' with a module named 'bar' (using file `upload/extension/foo/admin/controller/module/bar.php` in its ocmod.zip archive) then its namespace should be:

`namespace Opencart\Extension\Foo\Admin\Controller\Module`

mhcwebdesign:
Not even the opencart.ocmod.zip which contains various former core OC extensions can be properly installed or uninstalled without causing numerous errors, too many to be listed here. When I have some time, I'll take a closer look at it this weekend.

chongshengdz:
You are very very wrong.
If a 3rd party extension were to introduce a new extension package named 'foo' with a module named 'bar' (using file upload/extension/foo/admin/controller/module/bar.php in its ocmod.zip archive) then its namespace should be:

it should be 

namespace Opencart\Admin\Controller\Extension\Foo\Module;
class Bar extends \Opencart\System\Engine\Controller {


and it should be **foo.ocmod.zip**

chongshengdz:
just tested the master version and found it still not stable, cannot save data in the Admin.
below is a sample extension, you can download and test it.
https://www.ectransistors.com/groupfolder.ocmod.zip

stalker780:
@chongshengdz 
Don't waste your time. Latest stable OC4 was before admin ajax commits.
I am working with a copy from 26/March.
At least it could be called beta then.

Now it looks like raw alpha again, with lots of bugs, typos and unfinished functionality in admin area.

chongshengdz:
> @chongshengdz
> Don't waste your time. Latest stable OC4 was before admin ajax commits.
> I am working with a copy from 26/March.
> At least it could be called beta then.
> 
[...]

PR #9710, opened 2021-06-14
OC4: Add extension code to install.xml

stalker780:
Take extension code from install.xml

This will give more freedom for developers in naming their extension packs. For example include extension version to name.

Example: 
Many extensions come with names like
oc2x-my-extension-v1234.ocmod.zip
oc30-my-extension-v3456.ocmod.zip
oc30-my-extension-v3456-seo.ocmod.zip
oc30-my-extension-v3456-journal.ocmod.zip
etc.

This is very comfortable for both customers and developers. 

The way it is done in OC4 now, you cannot ever store additional data in filename. 
Otherwise it will break extension folder structure and create folder "extension/oc40-my-extension-v4567". 
While code inside install.xml may contain correct string "my_extension" for any version and modification.

### DIFF
--- a/upload/admin/controller/marketplace/installer.php
+++ b/upload/admin/controller/marketplace/installer.php
@@ -190,6 +190,14 @@ class Installer extends \Opencart\System\Engine\Controller {
 							$name = '';
 						}
 
+						$code = $dom->getElementsByTagName('code')->item(0);
+
+						if ($code) {
+							$code = $code->nodeValue;
+						} else {
+							$code = basename($filename, '.ocmod.zip');
+						}
+
 						$version = $dom->getElementsByTagName('version')->item(0);
 
 						if ($version) {
@@ -230,7 +238,7 @@ class Installer extends \Opencart\System\Engine\Controller {
 							'extension_id'          => 0,
 							'extension_download_id' => 0,
 							'name'                  => $name,
-							'code'              	=> basename($filename, '.ocmod.zip'),
+							'code'              	=> $code,
 							'version'               => $version,
 							'author'                => $author,
 							'link'                  => $link,


mhcwebdesign:
> The way it is done in OC4 now, you cannot ever store additional data in filename.

What exactly are you talking about here? What data, in what filename?

Besides, the whole OC4 extension structure still has issues anyway, see e.g. https://github.com/opencart/opencart/issues/9591

stalker780:
@mhcwebdesign 
I am talking about storing at least extension version in filename. You cannot add it. I have gave examples.
Because installer uses it as extension code and creates folder structure from it. This breaks all paths to templates and urls in extension then, if you rename extension zip (change version number).

As to extension structure issues I may agree. But at least it works and I doubt Daniel will change it before release. 
So I propose minor changes which will make life much easier for developers and customers.

osworx:
Did not try the latest v.4.x code, but I usually use always filenames like following for my extensions:
**My-Custom-Extension-v.1.2.3-OC-4.0.x.ocmod.zip**

Therefore anybody knows for which OpenCart version this package is and which version this extension is.
And in my opinion, all extensions should follow this naming convention.
The more the differences between some OpenCart subversions can be massive.

stalker780:
@osworx That's what I am talking about.

Just check this installer.php code
https://github.com/opencart/opencart/blob/e22ddfb060752cd8134abbfb104202172ab45c86/upload/admin/controller/marketplace/installer.php#L233

And then this code field is used to create folder structure.
https://github.com/opencart/opencart/blob/e22ddfb060752cd8134abbfb104202172ab45c86/upload/admin/controller/marketplace/installer.php#L388

So all your controller, template and url calls in extension must look like... I don't know...
`$this->load->controller('extension/My_Custom_Extension_v123_OC_40x/module/my_module'); `

And if you change extension version, please edit all controllers to correct paths. facepalm.

osworx:
@stalker780 I am fully with you (currently no time to "play" around with messy code, but the current code will never handle such naming conventions).
And @mhcwebdesign will understand our point of view ..

mhcwebdesign:
Perhaps the install.xml should have these in there:
```
<!-- extension version -->
<version>1.0</version>
<!-- extension ID, used as the extension sub-folder name -->
<id>foo</id>
```
And then the extension installer won't have to rely on ocmod.zip file name.

stalker780:
@mhcwebdesign Correct. That is my proposal.

But I called this field 'code', cause that's how it is called in DB.

stalker780:
There should be more changes.

This line tries to read extension zip using code. 
https://github.com/opencart/opencart/blob/e22ddfb060752cd8134abbfb104202172ab45c86/upload/admin/controller/marketplace/installer.php#L278

So I think, there must be a separate field in DB to store real extension zip filename.

osworx:
All would be vey easy to handle, if there would be a central respository with unique extensions IDs.
IDs which are created by Daniel or better by an official module when a new extension is registered.
Such handling would ahve the advantage that only "official" extensions are allowed - and could be checked (also updated).

TheCartpenter:
Or creating dynamic folder names like the storage cache sub-folders are already handled.

PR #9726, opened 2021-06-19
Fixed customer_list to customer - admin/controller/customer/customer.php file

TheCartpenter:


danielkerr:
already added

TheCartpenter:
It is? https://github.com/opencart/opencart/blob/master/upload/admin/controller/customer/customer.php#L337

PR #9732, opened 2021-06-27
Escaping underscore with date function - controller/tool/log.php file

TheCartpenter:


TheCartpenter:
The same could be done on the maintenance branch.

mhcwebdesign:
Why do you want to escape the underscore character?

TheCartpenter:
For the same reason as this one: https://stackoverflow.com/a/16516254

ADDCreative:
You only need to escape recognised format characters.

TheCartpenter:
As part of what the date function can gather, where does it state the underscore between the year date format and time format is recognized?

ADDCreative:
Nowhere, which is why you don't need to escape it.

TheCartpenter:
I would never thought that escaping were specifically meant for recognized characters rather than the other way around.

ADDCreative:
From https://www.php.net/manual/en/function.date.php.
> You can prevent a recognized character in the format string from being expanded by escaping it with a preceding backslash.

So you only need to escape the ones listed here.
https://www.php.net/manual/en/datetime.format.php

TheCartpenter:
It would seem from this page: https://www.php.net/manual/en/datetime.format.php that escaping characters are only meant for literal characters:

```To add literal characteres youi can escape with backslash (\):```

PR #9733, opened 2021-06-27
Used $order_id instead of $this->request->get - catalog/controller/account/order.php file

TheCartpenter:


ADDCreative:
See https://github.com/opencart/opencart/pull/9734#issuecomment-869152467.

TheCartpenter:
Yes, 2nd commit now also deleted.

PR #9738, opened 2021-06-30
fixed RTL issue for front end store

webkul-dilip:


ADDCreative:
Is it wise to make changes to the core Bootstrap files? Would make upgrading a pain.

webkul-dilip:
> Is it wise to make changes to the core Bootstrap files? Would make upgrading a pain.

Noted, we will manage it separately.

PR #9740, opened 2021-07-01
fixed insecure randomness

wr3nch0x1:
fixed insecure randomness

### DIFF
--- a/upload/admin/model/tool/upload.php
+++ b/upload/admin/model/tool/upload.php
@@ -2,7 +2,7 @@
 namespace Opencart\Admin\Model\Tool;
 class Upload extends \Opencart\System\Engine\Model {
 	public function addUpload(string $name, string $filename): string {
-		$code = sha1(uniqid(mt_rand(), true));
+		$code = sha1(uniqid(random_bytes(64), true));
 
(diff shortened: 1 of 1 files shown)


TheCartpenter:
A bit questionable on why the token() helper function is rather not used for this ...

wr3nch0x1:
You can use that too, if you find it more suitable :)

You can use anything except rand()
<http://us.php.net/manual/en/function.rand.php>, srand()
<http://us.php.net/manual/en/function.srand.php>, mt_rand()
<http://us.php.net/manual/en/function.mt-rand.php>, mt_srand()
<http://us.php.net/manual/en/function.mt-srand.php>

ADDCreative:
Don't forget this one. Probably more important.
https://github.com/opencart/opencart/blob/4cfcccf58b2b83f0f8554fa184580c8a904f3faf/upload/catalog/model/tool/upload.php#L5

wr3nch0x1:
Yes. I have few more instances. Please confirm the pull request so that I
can push remaining thanks!

wr3nch0x1:
If this patch is satisfactory for you, can I push remainings with the same?

TheCartpenter:
The user_token would then need to be reviewed as well since it uses the same algorithm as the uploads.

wr3nch0x1:
Can you please confirm it here?
https://www.huntr.dev/bounties/1625139296034-opencart/opencart/

danielkerr:
added manually

PR #9742, opened 2021-07-02
fixed product page  error

webkul-dilip:


ADDCreative:
@webkul-dilip and @WebkulOpencart better to do this in a more consistent and safer manner.

Look how it's done in another part of that file.
https://github.com/opencart/opencart/blob/2a5b358003d1e9e4af964eda2ffcb79abe8229ce/upload/catalog/model/catalog/product.php#L88-L96

So better to do something like below and remove your str_replace in the controller.
```php
$implode = array();

foreach ($category_ids as $category_id) {
	$implode[] = (int)$category_id;
}

$query = $this->db->query("SELECT * FROM " . DB_PREFIX . "product_to_category WHERE product_id = '" . (int)$product_id . "' AND category_id IN(" . implode(',', $implode) . ")");
```

PR #9764, opened 2021-07-18
Solved Admin API cost issues with UPS - catalog/model/extension/shipping/ups.php file

TheCartpenter:


ADDCreative:
This is wrong. The cost should be it the default currency. It gets converted into the customer currency at a later point. Such as.

https://github.com/opencart/opencart/blob/513c307120808aac6279d5fe9553310d59560d77/upload/catalog/controller/checkout/cart.php#L212

TheCartpenter:
Admin API user.

ADDCreative:
What issues are you having with 'Admin API user'? Maybe you would be better asking on https://forum.opencart.com/. The change you suggested will cause the shipping total to be displayed incorrectly for customers who are not using the default currency.

TheCartpenter:
The admin API user also relies on the session data currency. As for the 'cost' versus 'text' field, you will notice that the 'cost' uses the $this->config->get('config_currency') and that the 'text' field uses the $this->session->data['currency'] variable. Surely, I don't need the forum to be reminded of that.

ADDCreative:
As I pointed out in https://github.com/opencart/opencart/pull/9764#issuecomment-882449999. The 'cost' value gets converted later and needs to stay in the default currency, because in will be saved into the order totals in the database. The 'text' value does not and also includes taxes. This is why the 'cost' and 'text' calculations are different. Your change will cause the shipping total to be calculated wrong at checkout. Please test this for yourself.

I've just quickly tested 3.0.3.6, by adding a order in the admin and the values look correct. If you are  having issues it's probably a setting or an extension you are using. In which case ask in https://forum.opencart.com/ or contact the extension developer.

TheCartpenter:
I am using a fresh version of OC v3.0.3.7 release - no extensions outside the scope. As I pointed out above, I am not specifically addressing the issue about checkout but also about the admin API user. The 'text' field does not include taxes but still uses the session data currency is still used as opposed to the config_currency. Which means, the customer using checkout and the API admin user creating the order could both encounter different amounts.

ADDCreative:
The amounts should be different 'cost' is used for internal calculations (before currency conversion) and 'text' is shown to the customer as the quote (after currency conversion and including taxes). This is why one should use `$this->config->get('config_currency')` and the other should use `$this->session->data['currency']`.

TheCartpenter:
Since the order totals have been enforced by shipping service providers by obtaining the payment service provider's ID, there's no point to attempt to balance the totals. Therefore, this commit will now be closed.

PR #9766, opened 2021-07-18
Added currency convert to costs in Royal Mail - catalog/model/extension/shipping/royal_mail.php file

TheCartpenter:


TheCartpenter:
Or, the $cost variable could be converted in the quote_data directly.

PR #9767, opened 2021-07-18
Added currency convert to cost and text in PF 48 - catalog/model/extension/shipping/parcelforce_48.php file

TheCartpenter:


TheCartpenter:
Closing since the actual issue is about the hardcoded amount of 1 dollar with the currency format as opposed to the defined currency value by store owners.

PR #9769, opened 2021-07-18
Added currency convert to cost in ec SHIP - catalog/model/extension/shipping/ec_ship.php file

TheCartpenter:


TheCartpenter:
Closing since the actual issue is about the hardcoded amount of 1 dollar with the currency format as opposed to the defined currency value by store owners.

PR #9774, opened 2021-07-19
Fix problem in pagination on Downloads page

opencartbot:


ADDCreative:
What about.
https://github.com/opencart/opencart/blob/f760a9a5e9bbb6264cc35f5f29e85e41cce5001f/upload/catalog/controller/account/download.php#L44

3.0.x.x_Maintenance branch also needs fixing.

ADDCreative:
@eka7a This one was wrong.

https://github.com/opencart/opencart/blob/1d65465e3b072f055157b555b89b6ca14c4c11f9/upload/catalog/controller/account/download.php#L44

Is still using a hard coded value of 10.

The correct fix would of been to change.
```php
'limit' => $this->config->get('config_pagination'),
```
To.
```php
'limit' => 10,
```

PR #9777, opened 2021-07-19
Update install.php

SoftwareOnbekend:
Seperated long lines of code to multiple lines to inprove readability.
Less srolling from richt to left.


SoftwareOnbekend:
I use a screen in portraitstand  for reading the code. I think that there are more developers use that. Now we have to move a lot from left to richt and back to see the full code.

Issue #9410, opened 2021-03-05
RTL support in OpenCart


WebkulOpencart:
**What version of OpenCart are you reporting this for?**
Both the 3.x and master version
**Describe the bug**
We can add the RTL support in default OpenCart.

WebkulOpencart:
@danielkerr Please confirm so that we can work on the same

WebkulOpencart:
#fixed frontend RTL CSS [https://github.com/opencart/opencart/pull/9744](https://github.com/opencart/opencart/pull/9744)

danielkerr:
im closing this. i have no idea what oc looks like with rtl language

PR #9778, opened 2021-07-20
add admin end RTL CSS

WebkulOpencart:
add admin end RTL CSS


WebkulOpencart:
3.x.x.x RTL CSS added [https://github.com/opencart/opencart/issues/9410](https://github.com/opencart/opencart/issues/9410)

PR #9789, opened 2021-07-25
Made various changes - catalog/controller/checkout/confirm.php file

TheCartpenter:


TheCartpenter:
It doesn't really make sense to put $data with $json in the same method in this case since AJAX response returns a JSON headers,

TheCartpenter:
Closing. File has changed, since.

PR #9792, opened 2021-07-27
Made various changes - catalog/controller/checkout/register.php file

TheCartpenter:


TheCartpenter:
@eka7a: https://github.com/opencart/opencart/pull/9792/files#diff-4cf126608aeaef3471a08ba1c74131fb1da267625cb88c0ed3a3a405a059f588R395 . zone_id also needs to be changed to: shipping_zone_id with the filter_var . Just noticed it.

ADDCreative:
@eka7a There are a couple things wrong with this. First `$this->request->post['payment_zone_id']` can be zero. The way `!filter_var($this->request->post['payment_zone_id'], FILTER_VALIDATE_INT)` is used a zero value will become true, so will error. You need `filter_var($this->request->post['payment_zone_id'], FILTER_VALIDATE_INT) === false`.

You can test this be selecting a country with no zones and select None as the zone.

Second `utf8_strlen(trim($this->request->post['payment_city'])) > 32)` is wrong and can lead to data loss. See my similar comments https://github.com/opencart/opencart/pull/7733#issuecomment-562137753 and https://github.com/opencart/opencart/pull/7733#issuecomment-562189305.

In fact none of the trims are even necessary as POST data is timed in the request.
https://github.com/opencart/opencart/blob/feb5f084a1deebfb835af3e63bae4d90ce631110/upload/system/library/request.php#L47

TheCartpenter:
While the trim function is also located elsewhere with string type validations, it would be correct about the filter_var ... I forgot to test the zeros and must include the === false lookup.

PR #9797, opened 2021-07-27
Sanitized array and int type - catalog/controller/checkout/checkout.php file

TheCartpenter:


TheCartpenter:
Closing. File has been modified, since.

PR #9844, opened 2021-07-28
Added int type - admin/controller/setting/store.php file

TheCartpenter:


TheCartpenter:
The else statement of the config_ prefixes needs to be properly conditioned in this controller.

TheCartpenter:
Closing. Not affecting a variable.

PR #9847, opened 2021-07-28
Added int type - admin/controller/user/user_permission.php file

TheCartpenter:


TheCartpenter:
@eka7a: Do not merge these yet. They need to be reviewed first. I'll take a look at the comments this week.

TheCartpenter:
Need to work on the catalog-end side as well.

eka7a:
@TheCartpenter These changes are not needed. Int is already defined in model files.

TheCartpenter:
The logs keeps increasing from the controllers since the variables being posted on the browser are not sanitized as integers when needed regardless of the models. It is the same concept as the page variable and, below, my modifications where you'll notice that the same variables I sanitized are also already sanitized in the $data array for these reasons. The same should, then, be applied via the model parameters from the controllers as well as the selected arrays which also originates from the browser and submitted by HTML forms.

eka7a:
@TheCartpenter I still don't think this is necessary. Because, values sent to the model file are defined (int, float, string etc) in model. It is unnecessary to define it twice. The page, sort, order etc values is used on frontend and needs to be defined on controller.

If @danielkerr approves of these changes, i can merge similar changes.

And it will be easier if you send the same changes with a single commit. Sending different requests for each file makes this job very difficult. It requires much more time to check each request.

TheCartpenter:
If we take a look at this e.g: https://github.com/opencart/opencart/blob/master/upload/catalog/controller/common/pagination.php . The page variable is inside an array but still is sanitized inside the controller. Then, if we look at almost any controllers, the page variable is sanitized prior to be passed on any $this->load->controller('common/pagination', array()) calls. If you still think this is unnecessary, then these page variables would be sanitized for the wrong reasons.

TheCartpenter:
@eka7a: Another and more relative e.g already used in the core: https://github.com/opencart/opencart/blob/master/upload/catalog/controller/api/order.php#L195

TheCartpenter:
Closing. Not affecting a variable.

PR #9853, opened 2021-07-29
Removed '/' after br - extension/opencart/admin/view/template/fraud/ip.twig file

TheCartpenter:
Also requires spaced DIVs but not sure where to set the distance between the containers.


TheCartpenter:
It's actually the opposite. Closing.

PR #9861, opened 2021-07-29
Aligned date_added variable - admin/view/template/mail/returns.twig file

TheCartpenter:


TheCartpenter:
@eka7a: Today's changes are smaller. 5-6 commits that needs to be in.

PR #9865, opened 2021-07-30
Added true case - catalog/controller/account/address.php file

TheCartpenter:


TheCartpenter:
There used to be a battled discussion about the ways to sanitize the address ID in the past. Now that the CSRF token lookup has been added into the catalog startup, this should reduce concerns due to the customer_token that is also integrated into the account address controller.

ADDCreative:
You have `=== true` for this one as well.

TheCartpenter:
Just fixed it at the same time. I am working on the trim function in the mean time.

TheCartpenter:
Closing due to conflicts.

PR #9872, opened 2021-07-30
Removed empty line on break - catalog/controller/checkout/payment_method.php file

TheCartpenter:
Not much difference it would bring but it's how it looks in the codes in general


TheCartpenter:

PR #9880, opened 2021-07-30
Removed trim function - admin/controller/catalog/category.php file

TheCartpenter:


TheCartpenter:
It's actually removed, not added.

TheCartpenter:
Changed my mind since it comes from a looped array.

PR #9895, opened 2021-07-31
Removed trim function - admin/controller/user/user.php file

TheCartpenter:


TheCartpenter:
The admin part has been completed for the trim function removal.

PR #9897, opened 2021-07-31
Added new method in product model - catalog/model/catalog/product.php file

TheCartpenter:


TheCartpenter:
Closing. Other factors could be involved by pulling the total downloads from the catalog.

PR #9913, opened 2021-07-31
Replaced string with filter_sanitized_string - admin/controller/common/column_left.php file

TheCartpenter:


ADDCreative:
This is wrong. To give a extreme and purely hypothetical example. Now you can `$this->request->get['user_token']` be something like `1ab1b2b1b<script src="javascript.js">` and `$this->session->data['user_token'])` be `1ab1b2b1b` and they will be equated as equal. You do not need to sanitise, only check.

Doing the following would be enough. The cast to string would only cause an "Array to string conversion" warning, if it did happen to be an array.
```php
&& ($this->request->get['user_token'] == $this->session->data['user_token'])
```

TheCartpenter:
While the use of string cast is correct, on the opposite side, when $this->request->get['user_token'] gets compromised with the $this->session->data , it still won't make any difference on that side. We'd still need something in the middle to prevent that to happen globally.

TheCartpenter:
Changed it to hash_equals helper function. However, I don't see this helper function anywhere else on the platform anymore. Hard to say if it's because it needs to be removed or will it be used for something else.

TheCartpenter:
Closing. This commit may not be needed anymore.

PR #9915, opened 2021-08-01
Other localisation validation changes - admin/controller/customer/customer.php file

TheCartpenter:


TheCartpenter:
Closing since the commission amount is allowed to be set from the API and from the catalog/model/checkout/order.php file.

PR #9918, opened 2021-08-01
Added validation for posted amount - admin/controller/customer/customer.php file

TheCartpenter:
Now that the master branch uses the API from the admin-end side, the order ID may not be passed with the addTransaction but the posted amount can still be tricky if that API is used as opposed to before when the admin did not had an API built-in.


TheCartpenter:
Closing since the affiliate commission from the API and catalog/model/checkout/order.php file does allow an amount of 0.

PR #9919, opened 2021-08-01
Added error_amount definition text - admin/language/en-gb/customer/customer.php file

TheCartpenter:


TheCartpenter:
Closing since the API and catalog/model/checkout/order.php file does allow the commission amount to be set to 0.

PR #9920, opened 2021-08-01
Added flter_validate_float validation - admin/controller/sale/voucher.php file

TheCartpenter:


ADDCreative:
This is wrong. You shouldn't allow negative or zero vouchers.

TheCartpenter:
Ok, fixed.

ADDCreative:
Still wrong. You shouldn't allow negative.

TheCartpenter:
Fixed it with this solution: https://www.php.net/manual/en/filter.filters.validate.php#116924 .

TheCartpenter:
Closing. May not need this one.

Issue #9909, opened 2021-07-31
Config session expire - default duration to be added in oc_setting.


anytizer:
config_session_expire - a new variable seems to be present/introduced.

Please add it to the installer sql: install/opencart.sql

Hints:

    INSERT INTO `oc_setting` (`code`, `key`, `value`) VALUES ('config', 'config_session_expire', '1440');

Affected:
admin/controller/setting/setting.php
admin/controller/startup/session.php

Or otherwise, an error is generated:
TypeError:  Unsupported operand types: int + string in ___\admin\controller\startup\session.php on line 23

PR #9921, opened 2021-08-01
Default session time

agtt:
Default session time
First installation fix


agtt:
config_session_expire value not have default value for that when i login admin area i get error.
We have set when we install opencart.

admin/controller/startup/session.php
time() + $this->config->get('config_session_expire') 

or 

$this->config->get('config_session_expire')  ? $this->config->get('config_session_expire')  : 3600

anytizer:
#9909

PR #9922, opened 2021-08-01
Added true cases - catalog/controller/account/wishlist.php file

TheCartpenter:


ADDCreative:
You don't need to set true for link as it's not a redirect.

TheCartpenter:
Might still be pullable as a link, however, since this Ajax request is not used in an HTML data type in this case.

ADDCreative:
It is use in a HTML context.

https://github.com/opencart/opencart/blob/09378cdca7e797400793891573c187f9a72ea656/upload/catalog/view/javascript/common.js#L322-L326

TheCartpenter:
https://github.com/opencart/opencart/blob/master/upload/catalog/view/javascript/common.js#L313

TheCartpenter:
A better e.g would be in the common.js file from the install folder where the getURLVar function is used. In this e.g, it is also using an HTML context but the URL is added in manually from the part array.

ADDCreative:
dataType just defines how the data is returned, not how it's used. As `success` is used in the `.prepend()`, so best to leave it encoded.

PR #9923, opened 2021-08-01
Made various changes - catalog/controller/checkout/cart.php file

TheCartpenter:


ADDCreative:
`$json['success']` is use in an HTML context, so no need to set true for link.

https://github.com/opencart/opencart/blob/09378cdca7e797400793891573c187f9a72ea656/upload/catalog/view/javascript/common.js#L209-L213

TheCartpenter:
Ok, fixed.

ADDCreative:
You have unfixed it with one of your changes.

TheCartpenter:
Changes were not made fast enough since this commit was created. checkout/cart.php has been modified today with the *.php but the modules array should still read as: [] , Changes not done.

PR #9924, opened 2021-08-01
Made various changes - catalog/controller/account/register.php file

TheCartpenter:


TheCartpenter:
Closed without closure notes. Int casts with posted variables on the browser will cause error activity message outputs without sanitizing. Array types are also validated elsewhere on the platform with $this->config->get . Changes not done.

PR #9926, opened 2021-08-01
Added int cast - upload/catalog/controller/api/cart.php file

TheCartpenter:


TheCartpenter:
Closed without closure notes. Browser requires sanitizing on posted integers. Otherwise, it ends up in error activity logs. Changes not done.

PR #9927, opened 2021-08-01
Added int casts - catalog/controller/api/address.php file

TheCartpenter:


ADDCreative:
I think you mean `=== false` not `=== true`.

TheCartpenter:
Ok, fixed.

PR #9929, opened 2021-08-01
Made various changes - catalog/controller/checkout/payment_address.php file

TheCartpenter:


ADDCreative:
filter_var will never return true.

TheCartpenter:
Ok, fixed.

PR #9931, opened 2021-08-01
Update returns.php - catalog/controller/account/returns.php file

TheCartpenter:


TheCartpenter:
I think that settles all the int casts sanitizing and the trim functions.

PR #9933, opened 2021-08-01
Added validation to $response_info - extension/opencart/admin/controller/currency/fixer.php file

TheCartpenter:


ADDCreative:
This is wrong. You can't use in_array like that. in_array looks for values not keys.

TheCartpenter:
Closing. This one may not be needed.

PR #9935, opened 2021-08-01
Added int cast and 1 - catalog/controller/account/account.php file

TheCartpenter:


ADDCreative:
Totally unnecessary. It's not a integer, it's a boolean stored in a integer. Anything, non-zero should be considered true.

TheCartpenter:
It's as necessary as it needs to be along with other integers by $this->config->get .

ADDCreative:
The 1 effectively means not 0 or false. It doesn't need to be the actual value of 1. Have a read or https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting.

TheCartpenter:
Of course. Which is why all of the following files also uses their int cast from each controllers on the master branch already:

- ((int)

admin/controller/startup/application.php
admin/model/catalog/product.php
catalog/controller/account/address.php
catalog/controller/account/affiliate.php
catalog/controller/account/edit.php
catalog/controller/account/register.php
catalog/controller/account/voucher.php
catalog/controller/api/address.php
catalog/controller/api/order.php
catalog/controller/checkout/checkout.php
catalog/controller/checkout/payment_address.php
catalog/controller/checkout/register.php
catalog/controller/checkout/shipping_address.php
catalog/controller/information/contact.php
catalog/controller/mail/review.php
catalog/controller/product/product.php
catalog/controller/startup/application.php

- (int)$this->config->get

admin/controller/sale/api.php
admin/controller/sale/order.php
admin/controller/startup/application.php
catalog/controller/account/address.php
catalog/controller/account/affiliate.php
catalog/controller/account/edit.php
catalog/controller/account/register.php
catalog/controller/api/address.php
catalog/controller/api/order.php
catalog/controller/checkout/payment_address.php
catalog/controller/checkout/shipping_address.php
catalog/controller/mail/order.php
catalog/controller/product/manufacturer.php
catalog/controller/startup/application.php

ADDCreative:
I can't see any `== 1` in those files. As I said anything not zero should be considered true, not just 1.

TheCartpenter:
Anything is a bit overrated in this case since the option values uses the integers of 0 and 1. Everything else loaded by extensions don't use the int cast nor == 1 .

TheCartpenter:
So to add, the enforcement of int cast from those that are already enforced from the core already don't use the true / false bools either but rather a number.

ADDCreative:
The status is just a boolean represented as an integer. To test for equal to 1 it totally unnecessary. All that is important is whether it's zero or not.

PR #9944, opened 2021-08-01
Added array casts - catalog/model/account/download.php file

TheCartpenter:


TheCartpenter:
Closed without closure notes. Array is used elsewhere on the platform with order statuses involving conversion from JSON into arrays during startup.

PR #9948, opened 2021-08-02
Main opencart extensions should't delete

agtt:
If you delete the plugin, the website will not work for that we should add control


BigManatee:
Minor issue but the comment `// Should'nt be able to uninstall a main extension` should be `// Shouldn't be able to uninstall a main extension`

agtt:
I edited the comment. This issue can cause big problems.

danielkerr:
already fixed

PR #9949, opened 2021-08-02
Made various changes - extension/opencart/catalog/model/fraud/ip.php file

TheCartpenter:


eka7a:
This change cancel the fraud ip check.

PR #9965, opened 2021-08-03
Removed break line from first mb-3 required - catalog/view/template/checkout/payment_address.twig file

TheCartpenter:


TheCartpenter:
```<div class="row row-cols-2">``` might be missing as opposed to the shipping_address.twig file.

TheCartpenter:
Closing. This commit may not be needed.

PR #9979, opened 2021-08-04
Aligned label with checkbox type - catalog/view/template/product/search.twig file

TheCartpenter:


TheCartpenter:
Concludes the checkbox labels alignment.

PR #10085, opened 2021-08-05
TLS Version is changed to use TLS1.2

Asif-Mahmood:
changed the crypto type passed to stream_socket_enable_crypto in case of TLS host. As the current option of "STREAM_CRYPTO_METHOD_TLS_CLIENT" defaults to TLS 1.0 after PHP 5.6 which in turns fails on many mail hosts. As TLS 1.0 is not recommended by many mail hosts.

### DIFF
--- a/upload/system/library/mail/smtp.php
+++ b/upload/system/library/mail/smtp.php
@@ -142,7 +142,7 @@ class Smtp {
 
 				$this->handleReply($handle, 220, 'Error: STARTTLS not accepted from server!');
 
-				stream_socket_enable_crypto($handle, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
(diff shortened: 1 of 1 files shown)


ADDCreative:
This change would prevent TLS 1.3 from being used. Which doesn't seem like a good idea. The PHP manual is a bit out of date. Best to look at the source code.

As of PHP 7.2 "STREAM_CRYPTO_METHOD_TLS_CLIENT" means TLS 1.0, 1.1 and 1.2.
https://github.com/php/php-src/blob/php-7.2.34/main/streams/php_stream_transport.h#L172-L176

As of PHP 7.4 "STREAM_CRYPTO_METHOD_TLS_CLIENT" means TLS 1.0, 1.1, 1.2 and 1.3.
https://github.com/php/php-src/blob/php-7.4.22/main/streams/php_stream_transport.h#L171-L176

If your getting TLS 1.0 when using PHP 7.2 or greater, there is probably an issue with your server settings or cipher suites.

PR #10138, opened 2021-08-11
Greater pagination size for admin

anytizer:


TheCartpenter:
I don't agree with this. 100 results per page is too large. The default was normally 30 in previous releases than OC 4. It may have changed since.

PR #10166, opened 2021-08-18
No ellipsis for short product descriptions on category page

dabbmedia:
By default, catalog/controller/product/category.php appends '..' to all product descriptions, after shortening to the product_description_length from theme config. This update does not append the '..' to descriptions shorter than the theme product_description_length.


osworx:
And what for should this change be good?
It would be much better to cut words NOT somewhere in the word, instead cut after or before the whole word.

dabbmedia:
My goal in this commit is to remove the ellipsis when the product description isn't shortened, because it looks like an error to customers or site visitors when '..' is added to a complete sentence. It is also implies there is more to be read on the product page when there is no additional text.

I agree trimming at the end of the last word before the product_description_length would be better, but people understand the content is cut short because of the '..'. Also, in American English, it is more common to use 3 dots, not 2; so maybe that string should be moved to the language file or settings?

Thanks for the feedback. Sorry I somehow missed the notification on your comment until now.

dabbmedia:
I updated the code to work for both of our concerns.

I also added a text_ellipsis field to the language file and referenced it in the controller.

dabbmedia:
@osworx any thoughts on the code updates I made to address both of our concerns? I've been working around this minor bug since OC version 2, would love to see it disappear.

PR #10169, opened 2021-08-19
should add status ='1'

chongshengdz:
should add status ='1' if you disable some countries, then the disable countries will not show up in the frontend.


TheCartpenter:
`status` with backward escaped quotes = '1'.

Issue #10084, opened 2021-08-04
Bypass getLoginAttempt indefinitely by logging in from a different IP address


photographer-mememe:
**What version of OpenCart are you reporting this for?**
3.x

**Describe the bug**
It's possible to indefinitely bypass the getLoginAttempt check.

**To Reproduce**
1. Go to the account/login or checkout/login
2. Attempt failed logins with email = test@test.zzz from a first IP address = 111.111.111.111, let's say total = 2 times, which is below the config_login_attempts = 5.
3. Then switch IP address = 222.222.222.222 and attempt again with the same email = test@test.zzz, always with a different IP address from the first IP address, for total = indefinite number of times.
4.  $json['error']['warning'] is never triggered, so you can continue to brute force the login.

**Expected behavior**
In catalog/model/customer, addLoginAttempt() will add multiple rows for the same email if failed login attempts are from multiple IP addresses. Because of this:
`SELECT ..... WHERE username = ......  AND ip =`

But getLoginAttempts() only returns the first row of the query.

So with the example above with indefinite failed login attempts, total for the second query row will rise indefinitely, but the validation always gets total = 2 from the first query row.

On a successful login attempt, deleteLoginAttempts() will delete all rows with the same email.

**How to Fix**
Option 1 - log only a single row
1. Delete all rows from customer_login to make sure there are no duplicate entries for the same email address.
2. In addLoginAttempt(), Remove "AND ip =" from the query, so that logging in from a different IP address will overwrite the single instance row of the same email.
3. Then only the IP of the last failed attempt will be recorded, and total will be cumulative, no matter how many IP addresses the failed attempts are made from.

Option 2 - log multiple rows
1. In getLoginAttempts(), change the return to `$query->rows`, to return all multiple rows of the same email.
2. In catalog/controller/account/login and checkout/login, change to a foreach loop to check all rows.

```
//adds totals of multiple rows
$attempt_total = 0;
foreach ($login_info as $avalue) {
	$attempt_total += $avalue['total'];
}
unset($avalue);


//optionally send an alert email to admin when total attempts equals config_login_attempt, so that it only sends one time. Needs to occur before the next addLoginAttempt, otherwise this number may be skipped.
if ($attempt_total == $this->config->get('config_login_attempts')) {
	$subject = sprintf('%s - Too many login attempts for a customer', $this->config->get('config_name'));
			
	$message  = sprintf('There were too many login attempts for customer %s', $this->request->post['email']) . "\n\n";
[...]

PR #10172, opened 2021-08-24
Fixed multiple IP issue of login attempts

prabhatkumaroc082:
Fix for https://github.com/opencart/opencart/issues/10084


TheCartpenter:
If this option would need to be considered, then line 113 still looks for the IP even though the lookup was suggestingly removed.

prabhatkumaroc082:
@TheCartpenter 
please refer https://github.com/opencart/opencart/issues/10084#issuecomment-904479316

PR #10187, opened 2021-08-29
Removed duplicated createInvoiceNo() method - admin/controller/sale/api.php file

TheCartpenter:


TheCartpenter:
Updated previous PR.

PR #10191, opened 2021-08-31
Avoid division by zero error

anytizer:
I got this error when trying to run OC in headless manner, without browser.


TheCartpenter:
I am not sure this would be ideal since it would only enforce the limit to be pushed at 10 otherwise as it would harden the troubleshooting due to an expected integer that could either be 0 or not an integer.

PR #10202, opened 2021-09-07
More language compatibility

osworx:
Changed utf8_general_ci to utf8mb4_general_ci


TheCartpenter:
In PHPMyAdmin, I noticed this is the default setting for charset when creating a new table nowadays.

eka7a:
@osworx Also the character set needs to be utf8mb4.

osworx:
> 
> 
> @osworx Also the character set needs to be utf8mb4.

Correct, forgot it - sorry.

PR #10223, opened 2021-09-22
Pushed replace in accordance of the other models - admin/model/catalog/category.php file

TheCartpenter:


TheCartpenter:
Closing. REPLACE should not contain spaced parentheses.

PR #10246, opened 2021-10-19
add MyFatoorah Payment Extention

rsaeed-mf:
Add MyFatoorah Payment Extention
You can find more details on the online market store: 
https://www.opencart.com/index.php?route=marketplace/extension/info&extension_id=40652&filter_search=myfatoorah

We provide easy-to-integrate payment gateway services for the OpenCart websites. We target GCC and Egypt.
myfatoorah.com/


TheCartpenter:
These classes declaration won't work for the master branch which is where you have suggested this extension to be added into. To use these classes, better to suggest these files in the maintenance branch.

danielkerr:
use the contact us for requesting added to the core. we don't do it for free.

PR #10252, opened 2021-10-23
Added sorting for Scripts and Styles.

trydalcoholic:
In my experience, I have encountered the problem of including scripts/styles files. Previously connected files distorted the operation of files included later. Simple sorting after calling the getStyles()/getScripts() methods resolved such cases.


condor2:
When the function getScripts is replaced you get errors on product form. **ckeditor is not a function**

trydalcoholic:
Thanks for reporting this!
As you correctly pointed out, the "logical" order of inclusion of scripts was violated if the value of `$sort` is the same (and by default it is the same for all):
```
$ this->document->addScript ('view/javascript/ckeditor/ckeditor.js'); // Should enable on earlier
$ this->document->addScript ('view/javascript/ckeditor/adapters / jquery.js'); // Should enable on later
```
Unfortunately, I did not understand how to solve this with `array_multisort`, which is often used in Opencart code. I have now changed the sort to `uasort`. This solution doesn't seem to break the current work of Opencart.

PR #10257, opened 2021-10-31
3.0.x.x maintenance

Vilgar:
Update session lib file session.php


Hkr32:
@TheCartpenter, твой выход)

TheCartpenter:
@Hr32: What does that even mean? Nowhere in here I am related to this commit ...

PR #10286, opened 2021-12-04
The Countable interface compatibility in php-8.1.

glodov:
Fix for the warning in php-8.1:
Unknown: Return type of Twig\Node\Node::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in ../opencart/upload/system/storage/vendor/twig/twig/src/Node/Node.php on line 148


eka7a:
a15bc1157e4b981019851de089f6534c7508f154

PR #10293, opened 2021-12-07
Fixed separated IFs with elseIFs statements - catalog/controller/api/sale/payment_method.php file

TheCartpenter:
Only one dimensional $json['error'] array is created here. They'd be something odd to set separated IF statements in the lookup since only one error at a time would be returned in this case.


TheCartpenter:
Closing. Empty lines already resolved, since.

PR #10301, opened 2021-12-12
Solve return type error while using PDO

anytizer:
Solve return type error while using PDO: Due to return type defined in the function and the return is in between the body of the function. Fix the IDE experiences an error.


danielkerr:
already fixed

PR #10314, opened 2021-12-19
Added text_default with config_name - admin/controller/sale/order.php file

TheCartpenter:


TheCartpenter:
Closing. Wrong stores array location.

PR #10329, opened 2021-12-31
small fixes .. php8.1

egproo:


condor2:
Is not necessary to make changes in twig files,
Just TWIG should be updated to latest version because now is 3.1.1

https://github.com/twigphp/Twig/blob/3.x/CHANGELOG

eka7a:
a15bc1157e4b981019851de089f6534c7508f154

Issue #10348, opened 2022-01-03
ParseError: Unclosed '{' on line 723 does not match ')' in ../admin/controller/setting/setting.php on line 726


cgcmart:
https://github.com/opencart/opencart/blob/bffc4254f395bbae7740e1189b6f9ebec2c7dd54/upload/admin/controller/setting/setting.php#L726

Fix:
`if ($this->request->post['config_file_max_size'] > $upload_max_filesize) {`

slicktdog08:
I just encountered this also. Thanks for fix. Please merge into project source soon

slicktdog08:
Got another error after making your change

`ParseError: syntax error, unexpected token ";" in /[redacted]/html/admin/controller/setting/setting.php on line 581`

slicktdog08:
Sorry I made change in wrong place. Your change did solve the problem for me :)

PR #10349, opened 2022-01-04
Removed extra ')' setting controller

condor2:


condor2:
https://github.com/opencart/opencart/issues/10348

PR #10352, opened 2022-01-07
Couple of fixes testing on PHP8.1

howdu:


upload/catalog/controller/checkout/payment_method.php
```diff
@@ -57,7 +57,7 @@ public function index(): string {
 			if (isset($this->session->data['payment_methods'])) {
 				$data['payment_methods'] = $this->session->data['payment_methods'];
 			} else {
-				if (!isset($this->session->data['payment_address'])) {
+				if (!empty($this->session->data['payment_address'])) {
```
TheCartpenter, line 60: It's already changed to isset for the shipping_address from !isset in the commits. !empty would not be necessary in this case.

upload/catalog/view/template/product/category.twig
```diff
@@ -33,15 +33,21 @@
...
+            <div class="col-sm-3">
+              <ul>
+                {% for category in categories) %}
                   {% for child in category %}
                     <li><a href="{{ child.href }}">{{ child.name }}</a></li>
                   {% endfor %}
-                </ul>
-              </div>
-            {% endfor %}
+                  {% if loop.index is divisible by(4) and not loop.last %}
```
TheCartpenter, line 42: For some reason, this change is still based on an older way to put it but does make more sense due to it's simplicity to validate this. It would be, then, to know from the PHP TWIG Engine which PHP function would be more optimized to accomplish this.

danielkerr:
it works fine on php8.1 now

PR #10368, opened 2022-01-16
Escaped quotes correction - upload/catalog/model/localisation/country.php file

TheCartpenter:


TheCartpenter:
Closing. Query has been modified, since.

PR #10375, opened 2022-01-18
Added backward escaped quotes - catalog/model/checkout/subscription.php file

TheCartpenter:


TheCartpenter:
Closing. Query has already been modified.

Issue #7633, opened 2019-09-30
If smtp email fails, the order is missing


dimkoss:
Before customer is redirected to success page, opencart is sending order's confirmation emails to customer and administrators. But if for any reason the smtp connection fails, the program breaks  and the customer doesn't know that he made a successfully online payment.

In system/library/mail/smtp.php the throw new \Exception code just raises the exception and breaks the program. This should be handled in a different way to avoid double orders or payments.

danielkerr:
make sure your smtp or php mail settings are setup correctly.

danielkerr:
also yuou can find missing orders from the admin order page.

simkalp82:
yes. my smtp configs correctly. 
I sell 500-1000 products daily. 50 missing order per day. 

Your solution check 50 order per day?
> "also yuou can find missing orders from the admin order page."

waste time! 

Opencart is a simple software. if you are a big company use another software..

simkalp82:
> also yuou can find missing orders from the admin order page.

yes. my smtp configs correctly.
I sell 500-1000 products daily. 50 missing order per day.

Your solution check 50 order per day?

"also yuou can find missing orders from the admin order page."

waste time!

Opencart is a simple software. if you are a big company use another software..

osworx:
> Opencart is a simple software. if you are a big company use another software..

I would ask myself: "**why do I have so many missing orders?**"

Maybe your checkout is a crap, maybe your products, maybe your prices, maybe something other ..
Not "recommending" other software, that is a bad advice!

simkalp82:
> > Opencart is a simple software. if you are a big company use another software..
> 
> I would ask myself: "**why do I have so many missing orders?**"
> 
> Maybe your checkout is a crap, maybe your products, maybe your prices, maybe something other ..
> Not "recommending" other software, that is a bad advice!

Because If smtp email fails, the order is missing.. 
Opencart order system = smtp mail..

danielkerr:
IT COULD BE THE site IS NOT RECEIVING A CALLBACK FROM THE  PAYMENT GATEWAY.

or could be you deleted a order status that is still used by the order. u should ask in the forums or pay some one for support.

ADDCreative:
This can be an issue, I've also seen it happen. As the exception will stop execution of the rest of the code and the trigger for the mail event is before (catalog/model/checkout/order/addOrderHistory/before). It means the order status doesn't get updated, so the order goes as missing.

[...]

PR #10395, opened 2022-01-30
Update smtp.php - invalid mailbox

leonidukg:
If the user does not specify the mail correctly, this error causes fake problems.
SMTP replies: 550 Message was not accepted -- invalid mailbox.
But why does the store stop all the work of the site, it is not a critical error. For example, it is not possible to change the order status of a product. All these errors should just write errors into the log, not stop the store in any way.

### DIFF
--- a/upload/system/library/mail/smtp.php
+++ b/upload/system/library/mail/smtp.php
@@ -222,7 +222,7 @@ class Smtp {
 
 			fputs($handle, '.' . "\r\n");
 
-			$this->handleReply($handle, 250, 'Error: DATA not accepted from server!');
+			//$this->handleReply($handle, 250, 'Error: DATA not accepted from server!');
 
 			fputs($handle, 'QUIT' . "\r\n");
 


TheCartpenter:
Server specifics ... not sure why this commit has been posted here ...

leonidukg:
mail.ru is the largest provider of e-mail services in Russia. 
I do not think that postal problems at all should be fatal and cause problems with mail processing. 
I'm not saying that my commit is the solution to all problems, but the SMTP module should be redesigned and just logged for errors, but not to stop the store because of these little things.

TheCartpenter:
It's also to know about the probabilities that store owners will notice their error logs prior to create their new topics on the OC forum or new issues on this repository. Over the years, the results shows pretty low on that.

leonidukg:
This is a fine-tuning of the store and if the owner notices that the mail does not work, then he should contact a specialist, if he himself does not understand how and what to do.
Store owners mostly can't do anything at all, but use what they've been made to do.
An additional problem is that now many people do not use mail at all, but only indicate the phone number. And the order form requires mail, so customers specify the wrong mail or just random letters to place an order. Yes, this is a problem for the store owner. But I personally experienced that the store could not process more than 10 orders because of this SMTP error.

Translated with www.DeepL.com/Translator (free version)

TheCartpenter:
If that tuning would be as fine as described, then this free platform would only be about the specialist, since constantly asking people to resolve technical issues doesn't necessarily bring better service to it, especially by the time store owners do find the locations of these logs.

Error triggers can be disabled on websites and only kept to the logs server-based without the need to remove the error display on the screen from Opencart while a store runs on a production server.

leonidukg:
Let the error be stored in the logs, I don't mind. Right now a fatal ajax error is caused when trying to change order status.

ADDCreative:
The error handling of the mail classes definitely need changing. SMTP servers are never 100% reliable. An error shouldn't cause everything to fail. Best just to log the error. This has been mentioned a few times. https://github.com/opencart/opencart/issues/7633

TheCartpenter:
One way this could work is by adding a mail method into the loader registry and call the mail library the same way the languages are handled. In the loader registry, the log library could be used by using the registry object.

danielkerr:
i think sorted by changing the from address

PR #10407, opened 2022-02-08
Fix default report settings

ADDCreative:


danielkerr:
already fixed and removed

PR #10417, opened 2022-02-10
Initial dockerization

sneils:
Goals:
- dockerize the application, a.k.a. starting a container and having the application up out of the box
- no application code changes!

Notes:
- install/cli_install.php doesn't seem to be working correctly and fails with some missing constants (e.g. APPLICATION)
- install/cli_cloud.php doesn't create the config files and therefor is also not usable standalone
- this caveats mean this was created by running the manual install process and then copying the created config files, while replacing the hardcoded values to source from environment variables


sneils:
Sorry, I first wanted to create this PR on my fork and test some more.

PR #10421, opened 2022-02-12
Showing both version of text redirection - admin/language/en-gb/tool/upgrade.php file

TheCartpenter:


TheCartpenter:
Showed two different versions of what the text redirection should be. The way it previously was, using 'redirecting', was fine. Otherwise, I have added another sentence of what it could also be if the '...' is not added at the end of text line.

danielkerr:
i think code has been moved

PR #10423, opened 2022-02-13
DDOS & other tries to connect to DB

osworx:
While the current solution is better as it was before, I would suggest the above changes.
With them, in no case an output is displayed, instead it is logged (is optional).

Optional the timeout could be also defined (in cases of DDOS, etc.) to cut the server load.

### DIFF
--- a/upload/system/library/db/mysqli.php
+++ b/upload/system/library/db/mysqli.php
@@ -10,8 +10,10 @@ class MySQLi {
 
 		try {
 			$mysqli = @new \MySQLi($hostname, $username, $password, $database, $port);
-		} catch (\mysqli_sql_exception $e) {
-			throw new \Exception('Error: Could not make a database link using ' . $username . '@' . $hostname . '!');
+			$mysqli->options(MYSQLI_OPT_CONNECT_TIMEOUT, 5); // optional, see comment
+		} catch (\Exception $e) {
+			error_log("\n" . date('Y.m.d H:i:s') . ' Error: no connection to database [' . $database . '] with user [' . $username . '] pw [' . $password . '] on host [' . $hostname . '] and port [' . $port . '] msg [' . $e->getMessage() . ']', 3, DIR_LOGS . 'error.log' );
+			die();
 		}
 
 		if (!$mysqli->connect_errno) {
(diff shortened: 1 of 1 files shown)


ADDCreative:
Setting MYSQLI_OPT_CONNECT_TIMEOUT after connection is pointless. You would need to set it before and use real_connect to connect.

This is only of any use for badly configured servers, which could be a lot judging by the forums. I would still throw a generic error message with no credentials, as will as the log. Otherwise there will be a load of forum posts about white screens.

osworx:
@ADDCreative yes, you are right about the timeout.
Regarding the display of message, a generic and harmless text may better for some, I use in such cases only 2 dots: ..
Then I know there is something .. wrong ..
Or instead showing a message, redirecting to a generic .html "error page" ?

On the other side, if those "website owners" do not know what a "white screen" means, they have a problem in general!

Khnaz35:
I have tried the @osworx solution but i am getting the following error:

`Notice: Undefined variable: mysqli in /home/xyz/public_html/system/library/db/mysqli.php on line 17Notice: Trying to get property 'connect_errno' of non-object in /home/xyz/public_html/system/library/db/mysqli.php on line 17Notice: Undefined variable: mysqli in /home/xyz/public_html/system/library/db/mysqli.php on line 18Warning: Creating default object from empty value in /home/xyz/public_html/system/library/db/mysqli.php on line 19

Fatal error: Uncaught Error: Call to undefined method stdClass::set_charset() in /home/xyz/public_html/system/library/db/mysqli.php:20 Stack trace: #0 /home/xyz/public_html/system/library/db.php(31): DB\MySQLi->__construct('localhost', 'oxyzzzwed_bassJo...', '12eeeee-%E', 'xyzeyss', '3306') #1 /home/xyz/public_html/system/framework.php(80): DB->__construct('mysqli', 'localhost', 'oxyzzzwed_bassJo...', '12eeeee-%E', 'xyzeyss', '3306') #2 /home/xyz/public_html/system/startup.php(104): require_once('/home/xyz/...') #3 /home/xyz/public_html/index.php(21): start('catalog') #4 {main} thrown in /home/xyz/public_html/system/library/db/mysqli.php on line 20`

ADDCreative:
> @ADDCreative yes, you are right about the timeout. Regarding the display of message, a generic and harmless text may better for some, I use in such cases only 2 dots: .. Then I know there is something .. wrong .. Or instead showing a message, redirecting to a generic .html "error page" ?
I personally would bother with a redirect. Just simple text or even as you suggest 2 dots.

osworx:
Because Daniel had made a change in the file: https://github.com/opencart/opencart/commit/da5dd4e92a453bdab665faed8ce2219bd9f0b299
I am closing this PR.

PR #10425, opened 2022-02-14
Installer upgrade

NikitaGuminsky:
Allows to install several extensions at once


TheCartpenter:
While I do agree with the getXMLData to be added as a method, the $calc variable may still be inaccurate by calculating 1024 * 1024 on server OSes.

TheCartpenter:
This foreach portion may help since it's already in the core: https://github.com/opencart/opencart/blob/master/upload/admin/controller/tool/backup.php#L70 .

PR #10427, opened 2022-02-14
Modified the max file size into a helper function - admin/controller/marketplace/installer.php file

TheCartpenter:


TheCartpenter:
Closing. JQuery code is being used instead.

PR #10430, opened 2022-02-16
Added isset lookup - install/controller/startup/database.php file

TheCartpenter:


TheCartpenter:
Closing. DB_PORT constant is being used instead.

PR #10434, opened 2022-02-20
Added isset lookup - install/controller/upgrade/upgrade_2.php file

TheCartpenter:


TheCartpenter:
Closing, since it is supposed to match by default.

PR #10435, opened 2022-02-20
captcha check after form validation

sadn3r:
checking captcha only after form passed validation -> more useful for end users because google recaptcha invalidates captcha every failed form request. !do not use this logic when checking login or any other security forms!


TheCartpenter:
I second this. ```!$json``` should be included in those lookups.

danielkerr:
bot sure what this  changes with the current code

PR #10437, opened 2022-02-22
Updated mysqli.php to check function parameter is not null then call real_escape_string

chittaranjan7:


upload/system/library/db/mysqli.php
```diff
@@ -50,7 +50,10 @@ public function query(string $sql): bool|object {
 	}
 
 	public function escape(string $value): string {
-		return $this->connection->real_escape_string($value);
+		if ( $value != '' ) {
+			return $this->connection->real_escape_string($value);
+		}
+		return ''
```
TheCartpenter, line 56: Missing semi-colon.

chittaranjan7: Thanks, @TheCartpenter.  Corrected it.

PR #10457, opened 2022-03-04
Don't use strftime as deprecated in PHP 8.1

ADDCreative:
vendor\braintree would also need updating in someway.


mhcwebdesign:
We have an updated system/storage/vendor/braintree in our forked repository at 
https://github.com/mhcwebdesign/opencart-php8/tree/3.0.x.x_Maintenance
I haven't created a pull-request yet for above.

Issue #10444, opened 2022-02-27
SEO not working - TypeError in model seo_url.php


cgcmart:
https://github.com/opencart/opencart/blob/24273376c71ad8498e0aded785ed386c5dafafd2/upload/catalog/model/design/seo_url.php#L10

TypeError:  Opencart\Catalog\Model\Design\SeoUrl::getSeoUrlByKeyValue(): Argument #2 ($value) must be of type string, null given in

TheCartpenter:
https://github.com/opencart/opencart/commit/cff71bb0e8f8aa18bfc39746088016135accff68#diff-63ec18d41a843f6c9ed99b8043c976087464dd6c9bd4637a96e6086d74f9519bR38

has been changed to: LIKE but does not use: % .

cgcmart:
I am waiting for authorized seo_url update complete, have not tested the update. Do you mean L126 and L132 `=` need to change to `LIKE`?

TheCartpenter:
R38 in the URL means L38.

cgcmart:
That is already updated. My question is whether it's applicable to L126 andL132

TheCartpenter:
Applicable on L38.

TheCartpenter:
Such as this change: https://github.com/opencart/opencart/commit/dc1e971539ad323269e18420ba4db3ca1cf2d395#diff-6bf9946bb8aefb6bbb29a3ed833261396d8814a3bd38463b43c17500e3e75761R5 . Not sure if the keyword and value both needs to contain a /% with LIKE or only one of them.

PR #10458, opened 2022-03-04
fix seo url

chongshengdz:
fix seo url


TheCartpenter:
Might also need to make this change for the models but not a 100% sure, yet: https://github.com/opencart/opencart/issues/10444#issuecomment-1057285354 .

PR #10468, opened 2022-03-09
Added int cast on order_product_id - catalog/model/checkout/subscription.php file

TheCartpenter:


danielkerr:
refrence field is required

PR #10473, opened 2022-03-09
PCI Compliance for CC number - extension/opencart/catalog/controller/payment/credit_card.php file

TheCartpenter:


danielkerr:
how is this pci complaince changing **** **** **** to str_repeat? there are spaces between the numbers because credit cards are normally spaced out between 4 digitsa

TheCartpenter:
Yes, that would be correct here. In fact, the last four digits in the core extensions of the OC v3.x releases do have fields for that specifically. It just deluded me to close this commit since.

PR #10483, opened 2022-03-14
Update marketplace.php

boxaltcoin:
Fix list for admin marketplace


TheCartpenter:
Correct. The list() function does not exist.

PR #10484, opened 2022-03-14
Fixed remaining value in the array - system/library/cart/cart.php file

TheCartpenter:


danielkerr:
no it was required to be duration in remaining because it set how many are left to do.

PR #10493, opened 2022-03-20
[3.0.x.x] Fixed affiliate custom_field not show in customer account when you edit from admin

condor2:
…it from admin


condor2:
@mhcwebdesign merge this

PR #10501, opened 2022-03-28
Don't remove leading slash from PHP_SELF

ADDCreative:
REQUEST_URI has leading slash, so don't remove it when using PHP_SELF.


TheCartpenter:
https://github.com/opencart/opencart/pull/10502#discussion_r836402250

PR #10504, opened 2022-03-30
Added getTotalRecurringProducts() method - admin/model/catalog/product.php file

TheCartpenter:


TheCartpenter:
By profile ID suffice.

PR #10506, opened 2022-03-30
Added recurring in marketing/contact - admin/controller/marketing/contact.php file

TheCartpenter:


TheCartpenter:
Would need the recurring ID to keep the array in one dimension.

PR #10508, opened 2022-03-30
Added filter_product_id in getRecurrings() method - admin/model/catalog/recurring.php file

TheCartpenter:


TheCartpenter:
Reversed lookup. Will use it in admin products page instead.

PR #10509, opened 2022-03-30
Added autocomplete for Recurrings in products list - admin/controller/catalog/recurring.php file

TheCartpenter:


mhcwebdesign:
You create tens of little pull requests each day! Could you please combine them into one big per day or per week, that would make things easier for all. Thanks.

TheCartpenter:
It was a long run but I think that settles it for the missing features with the product and order recurring on the Maintenance Branch.

danielkerr:
code does not exist any more.

PR #10520, opened 2022-03-30
[Bug Fix] - Added space on WHERE clause - catalog/model/account/return.php file

TheCartpenter:


TheCartpenter:
Closing this commit, since the catalog/model/checkout/recurring.php file looks for a reference string that is empty. Furthermore, the catalog/controller/event/activity.php file, addReturn() method, looks for an order status ID that is not yet defined which means that the probability for store owners to have a reference string at this stage is pretty low.

upload/catalog/model/account/return.php
```diff
@@ -27,7 +27,7 @@ public function getReturns($start = 0, $limit = 20) {
 	}
 
 	public function getTotalReturns() {
-		$query = $this->db->query("SELECT COUNT(*) AS total FROM `" . DB_PREFIX . "return`WHERE customer_id = '" . $this->customer->getId() . "'");
+		$query = $this->db->query("SELECT COUNT(*) AS total FROM `" . DB_PREFIX . "return` WHERE customer_id = '" . $this->customer->getId() . "'");
```
TheCartpenter, line 30: This portion of the commit would need to be kept, since this is a bug.

PR #10521, opened 2022-03-30
Added recurring lookup - catalog/controller/account/return.php file

TheCartpenter:


TheCartpenter:
Closing this commit, since the catalog/model/checkout/recurring.php file looks for a reference string that is empty. Furthermore, the catalog/controller/event/activity.php file, addReturn() method, looks for an order status ID that is not yet defined which means that the probability for store owners to have a reference string at this stage is pretty low.

PR #10524, opened 2022-03-31
Added recurring lookup - admin/model/sale/return.php file

TheCartpenter:


condor2:
You have some typo's

this

LIKE '%" . $this->db->escape($data['filter_recurring']) . "%'

and not this
LIKE '" . $this->db->escape($data['filter_recurring']) . "%'

And change **or** to **`or`**

condor2:
You have some typo's

this

LIKE '%" . $this->db->escape($data['filter_recurring']) . "%'

and not this
LIKE '" . $this->db->escape($data['filter_recurring']) . "%'

And change **or** to **`or`**

TheCartpenter:
Even with the $implode array, it shows AND, not OR. As for the '%', it also shows at the end of the string, not at the beginning in this old branch. As for the escaped quotes, couldn't agree more on that one but. unfortunately, dinosaurs have  been brought back to life and will only slow us down to make these corrections.

It's funny, in a sense. Yesterday, I was able to make these changes within 10 hours with the recurring filters and returns. To be able to pull a request about the same request; to add backward escaped quotes, it would take an eternity.

TheCartpenter:
> If you don't add backward escaped quotes for or when you try to filter by recurring will crash. Without them or is taken as OR command ad not as table name or (order recurring).

Yup, I know. I might need to know how to create pull requests all combined within the same pull request without the need to enter each filenames in the command lines. That could help to speed things up if somebody could provide their two cent on that.

mhcwebdesign:
For the time being, I just delete all the email copies from your pull requests.

You could a better github client software, such as SmartGit.

TheCartpenter:
Deleting all email copies is up to your total discretion as it is your choice to make. However, I will take a look at SmartGit.

TheCartpenter:
It's asking for a password in SmartGit to create a new pull request.

TheCartpenter:
The backward escaped quotes have now been added on the recurring model modifications.

PR #10536, opened 2022-04-05
Fix double slash in session path

ADDCreative:
https://github.com/opencart/opencart/pull/6093 for 3.0.x.x_Maintenance.


danielkerr:
i hope this is not in the released version

mhcwebdesign:
@ADDCreative : You are right, there shouldn't be double slashes in the path.

mhcwebdesign:
Note: The same fix was already applied to the master branch, too.

PR #10538, opened 2022-04-07
Replaced hash with sha1_file function - system/library/template/template.php file

TheCartpenter:
https://www.php.net/manual/en/function.sha1-file.php

### DIFF
--- a/upload/system/library/template/template.php
+++ b/upload/system/library/template/template.php
@@ -68,7 +68,7 @@ class Template {
 	}
 
 	protected function compile(string $filename, string $code): string {
-		$file = DIR_CACHE . 'template/' . hash_file('sha1', $filename . $code) . '.php';
+		$file = DIR_CACHE . 'template/' . hash('sha1', $filename . $code) . '.php';
 
 		if (!is_file($file)) {
 			file_put_contents($file, $code, LOCK_EX);
--- a/upload/system/library/template/template.php
+++ b/upload/system/library/template/template.php
@@ -68,7 +68,7 @@ class Template {
 	}
 
 	protected function compile(string $filename, string $code): string {
-		$file = DIR_CACHE . 'template/' . hash('md5', $filename . $code) . '.php';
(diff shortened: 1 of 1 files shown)


ADDCreative:
You are trying to calculate the sha1 hash of a file that doesn't exist! ($filename . $code) is only part off a filename.

cgcmart:
`hash_file('sha1', $filename . $code)`

TheCartpenter:
> `hash_file('sha1', $filename . $code)`

Yes, I had this one in mind as well. Will make the change.

ADDCreative:
`$filename . $code` is not a file that will exist. Trying to hash it is pointless weather useing hash_file or sha1_file. The original code is making a unique filename from the name and the template contents.

TheCartpenter:
md5 is more likely where it should be improved for that.

ADDCreative:
> md5 is more likely where it should be improved for that.

What does that even mean?

TheCartpenter:
It means that md5 is no longer in use.

ADDCreative:
The hash algorithm is irrelevant. You can't hash something that doesn't exist.

TheCartpenter:
There are other ways to do this. MD5 algorithm is simply history these days. It's good to say that it works but it's not as efficient as it used to be nowadays.

TheCartpenter:
More information about this would be to use the sha256 with buffering as opposed to MD5: https://stackoverflow.com/a/52183613 .

ADDCreative:
You are completely missing the point. The original code was using a hash to make a filename for the cache file. You change is trying to hash the contents of a file that doesn't exist.

TheCartpenter:
I didn't missed the point. I was also addressing one in the meantime. However, I will replace the hash function but that MD5 parameter should no longer be there.

danielkerr:
its fine as is. and no reason to use sha instead of md5. its not a security issue. its just hashing for unique name,

PR #10540, opened 2022-04-07
Fix mail for PHP 8

ADDCreative:
PHP 8 mail() requires all headers to end CRLF.
Split base64 encoded lines.

### DIFF
--- a/upload/system/library/mail/mail.php
+++ b/upload/system/library/mail/mail.php
@@ -26,45 +26,51 @@ class Mail {
 			$to = $this->to;
 		}
 
-		$boundary = '----=_NextPart_' . md5(time());
-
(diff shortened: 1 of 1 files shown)


ADDCreative:
@danielkerr smtp.php doesn't need changing only mail.php, as it's only an issue with the mail() function. SMTP does its own changing in the line endings anyway.

https://github.com/opencart/opencart/blob/11c4bb6c7c8375a76b7a4919d5f19d0dce940202/upload/system/library/mail/smtp.php#L210-L211

https://github.com/opencart/opencart/blob/11c4bb6c7c8375a76b7a4919d5f19d0dce940202/upload/system/library/mail/smtp.php#L221-L225

PR #10547, opened 2022-04-17
[Master] Added suggestion - exclude core libraries - system/engine/event.php file

TheCartpenter:


TheCartpenter:
Closing. Vendors will take the lead over OCMod and Events in the future on anyhow.

PR #10554, opened 2022-04-21
Adding in a "Made To Order" stock feature.

Motorhead1991:
Handy for those that don't stock one-off objects.


cgcmart (commented):
The premise is to config display stock, show out of stock, and stock checkout.

TheCartpenter:
These kinds of request could create a lot of debate by store owners in the future. It's been encountered a few times in the past as well. I would suggest to provide such ideas by using Event Triggers by re-using the $data['stock'] and re-condition it in there in case store owners may want to use modified versions of what they need as oppose to suggest changes to core codes that may interfere with the config display stock.

Motorhead1991:
It does appear to break the "stock" feature. I'll have to revert it on my end and tinker with it more.

PR #10567, opened 2022-04-22
[Master] - Added language definition with filesize - extension/opencart/admin/controller/dashboard/order.php file

TheCartpenter:


danielkerr:
not required to use text_ for single characters

PR #10609, opened 2022-04-23
[Master] - Removed reference row - catalog/view/template/account/subscription_info.twig file

TheCartpenter:


danielkerr:
reference is required for some subscription services like braintree

PR #10680, opened 2022-04-30
[3.0.x.x] Added SCSSPHP 8.1 warning fix

condor2:
https://github.com/opencart/opencart/blob/master/upload/admin/controller/startup/sass.php#L18

### DIFF
--- a/upload/admin/controller/startup/sass.php
+++ b/upload/admin/controller/startup/sass.php
@@ -14,7 +14,7 @@ class ControllerStartupSass extends Controller {
 					$scss = new \ScssPhp\ScssPhp\Compiler();
 					$scss->setImportPaths(DIR_APPLICATION . 'view/stylesheet/');
 
-					$output = $scss->compile('@import "' . $filename . '.scss"');
+					$output = $scss->compileString('@import "' . $filename . '.scss"')->getCss();
 
 					$handle = fopen($stylesheet, 'w');
 


mhcwebdesign:
Thank you for pointing it out.

The SCSSPHP is a 3rd party vendor script integrated into OpenCart via composer.

So rather then directly patching it, could you look into a newer version of scssphp (e.g. at gitgub.com/scssphp) which is compliant with PHP8.1?  Then we can just update the composer.json and do a refresh of the vendor plugins.

condor2:
With updated SCSSPHP you will get deprecated function because the function was changed

The "ScssPhp\ScssPhp\Compiler::compile" method is deprecated. Use "compileString" instead

https://github.com/scssphp/scssphp/blob/master/src/Compiler.php#L345

mhcwebdesign:
Thanks for your explanations. Will the updated SCSSPHP still work with bootstrap 3.3.5?

condor2:
yes

condor2:
But if you needed I can give you this version with jquery 3.6.0, summernote 0.8.20, bootstrap 3.4.1

mhcwebdesign:
I'll try it all out on my local test server first later today or tomorrow.

mhcwebdesign:
SCSSPHP now updated, along with your suggested change, see https://github.com/opencart/opencart/pull/10684

PR #10681, opened 2022-04-30
[Maintenance] - Replaced addReference with editReference - catalog/controller/extension/payment/pp_express.php file

TheCartpenter:


mhcwebdesign:
Was this a bug? Can someone test it please?

TheCartpenter:
Just check the method inside the route. addReference() doesn't exist. It's editReference that was meant to be there. The wrong method name's been used because somebody got distracted at that time.

mhcwebdesign:
OK, now merged.

PR #10683, opened 2022-05-01
[3.0.x.x] Deprecated function SCSSPHP

condor2:


mhcwebdesign:
Now added along with an update of the SCSSPHP, see https://github.com/opencart/opencart/pull/10684

PR #10754, opened 2022-05-01
[Maintenance] - Added backslashes on new instances - catalog/model/catalog/review.php file

TheCartpenter:
Should use JSON to submit review and use an event to submit the email in the mail folder along with the others.


TheCartpenter:
Closing. Resolved on opencart-3 repository.

PR #10760, opened 2022-05-01
[Maintenance] - Made various changes - catalog/model/extension/payment/bluepay_redirect.php file

TheCartpenter:


TheCartpenter:
Using port 443 in cURL request. Not sure if this should remain there since the use enforcement of SSL / TLS.

PR #10777, opened 2022-05-01
Added backslashes on new instances - catalog/model/extension/payment/realex_remote.php file

TheCartpenter:
Multiple cURL init within same models should be removed. addOrderHistory methods aside from catalog/model/checkout/order.php file should not be integrated into the core as an e,g. Event Triggers should be used instead rather than creating order histories manually on separated constructors from the core.


TheCartpenter:
Closing. Resolved on opencart-3 repository.

PR #10835, opened 2022-05-09
Fix error handler for PHP 8

ADDCreative:
error_reporting() no longer returns 0 when suppressed with @.


mhcwebdesign:
Thanks, well spotted.

PR #10837, opened 2022-05-10
[Maintenance] Define version header

haxcop:
Without this addition the upgrade will throw a lot of warning lines, Feel free to suggest a better dynamic method

### DIFF
--- a/upload/install/controller/common/header.php
+++ b/upload/install/controller/common/header.php
@@ -2,7 +2,7 @@
 class ControllerCommonHeader extends Controller {
 	public function index() {
 		$this->load->language('common/header');
-		
+		define('VERSION','3.0.3.8');
 		$data['title'] = $this->document->getTitle();
 		$data['description'] = $this->document->getDescription();
 		$data['links'] = $this->document->getLinks();


TheCartpenter:
Not sure why we would need that in there ...

haxcop:
> Not sure why we would need that in there ...

If you propose a better way I'm happy with it too :)

mhcwebdesign:
@haxcop: You really need to provide some more details about what exactly you, from where you upgraded from, and how. otherwise nobody will be able to reproduce it here, let alone be able to help you.

haxcop:
> @haxcop: You really need to provide some more details about what exactly you, from where you upgraded from, and how. otherwise nobody will be able to reproduce it here, let alone be able to help you.

My apologies:

From version 3.0.3.6 to the Latest 3.0.3.8
PHP V 7.3 and 7.4

Warning: Use of undefined constant VERSION - assumed 'VERSION' (this will throw an Error in a future version of PHP) in /home/.../system/storage/modification/system/library/template/twig.php on line 37 

With the proposed commit this warning dissapears

condor2:
@mhcwebdesign if you notice it is /system**/storage/modification/**system/library/template/twig.php

condor2:
No, but who know what ocmod extension changed twig file :)

danielkerr:
its added in index.php

PR #10843, opened 2022-05-15
[Master] - Made various changes - install/view/template/install/step_4.twig file

TheCartpenter:
Also added a note for product quality sold online regarding dangerous goods in order to inform store owners prior to add any products to their stores.

### DIFF
--- a/upload/install/view/template/install/step_4.twig
+++ b/upload/install/view/template/install/step_4.twig
@@ -7,10 +7,11 @@
 	</div>
 	<div class="container">
 		<div class="alert alert-danger alert-dismissible"><i class="fa fa-exclamation-circle"></i> {{ error_warning }} <button type="button" class="close" data-bs-dismiss="alert"></button></div>
+    <div class="alert alert-info"><i class="fa fa-exclamation-circle"></i> {{ text_sale_info }} <button type="button" class="close" data-bs-dismiss="alert"></button></div>
 		<div class="card">
 			<div class="card-header"><i class="fas fa-pencil-alt"></i> {{ text_step_4 }}</div>
 			<div class="card-body p-4">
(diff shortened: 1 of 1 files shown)


TheCartpenter:
This commit is about the beginning of this section in the discussion whereas store owners should be informed regarding the quality of products that needs to be sold online as to complete these types of transactions with their payment and shipping service providers prior to add any products online: https://forum.opencart.com/viewtopic.php?f=202&t=228371#p843074 .

ADDCreative:
> line 10 is not required. mot sure why u added that

@danielkerr You would probably need the other pull request anyway which has the message in. https://github.com/opencart/opencart/pull/10844 I can't see the point of a dangerous goods message myself.

@TheCartpenter This is exactly the sort of thing @mhcwebdesign and myself have been warning about. Spreading and mixing changes across multiple pull requests because you can't or won't use a proper Git client. Now, because it's too difficult to follow which ones are related only half the change has been added.

TheCartpenter:
@ADDCreative: Regarding the pull requests, I have been assigned a repository, and I am using it with combined commits since yesterday with Smartgit. Before that time, I could not create combined commits due to the same reason I have explained 'several times' now; a password was required and I could not go beyond that point other than creating individual commits.

As per my e.g on the paragraph, please take a look at the new repository and you'll notice that 'I have indeed' posted combined commits after sometime since yesterday.

TheCartpenter:
> line 10 is not required. mot sure why u added that

I referred the discussion to why line 10 has been added since there are store owners believes it's an Opencart issue while it has nothing to do with it. If the message in the installation is not the right location to add such message, I would suggest another location to add it. As a suggestion, perhaps in the readme file of your repository.

PR #10845, opened 2022-05-15
[Master] - Added missing utf8_substr - admin/controller/common/filemanager.php file

TheCartpenter:


TheCartpenter:
If that being the case, then what you're also implying is that utf8_substr that's been put into other similar IF statements are also wrong prior for me to add the last modification.

ADDCreative:
Yes it looks like the other if statements are wrong in that file.

danielkerr:
ok please do a pull request with it correct?

ADDCreative:
> ok please do a pull request with it correct?

@danielkerr https://github.com/opencart/opencart/pull/11224

PR #10864, opened 2022-05-20
Update order.php for vivification error in PHP 8.1

picasolutions:
$session->data is not yet an array at line 391.  So PHP 8.1 warns that auto vivification is deprecated.
Check for the condition, and ensure that it is an array.


ADDCreative:
This has already been fix the correct way. https://github.com/opencart/opencart/pull/10414

picasolutions:
Thank you @ADDCreative - does this mean that PHP 8.1 fixes are scattered in different branches?

Or am I missing something?

PR #10867, opened 2022-05-21
Removed extra language variable from analytics.php

condor2:


condor2:
And same for all language files from same location

PR #10888, opened 2022-05-25
Fixed CKEditor language interface bug

condor2:


eka7a:
Also, return theme file load ($this->load->view) names should also be changed or  theme file name changes should be undone.

cgcmart:
rename file name returns.php to return.php will consist with other file names.

TheCartpenter:
Yes, I agree with this. It needs to be renamed since it cannot be escaped like database fields and tables. It was confusing in the past to have partial prefixes for module_, payment_, and shipping to know what the nature of their $this->config->get was . It is the same with reserved terms of 'return'. All needs to be renamed to: 'returns'; file names, routes and method names regardless if it's about a controller, model or view.

TheCartpenter:
Aside from the ckeditor modification the category_form.twig file, the return term in the opencart-3 repository has now been renamed into returns: https://github.com/opencart/opencart-3/commit/0522a0133ed45f4f5eaec6c3f9d112271fd9d4ab .

TheCartpenter:
Database change for returns: https://github.com/opencart/opencart-3/commit/4abcc96373102193ee82b76d6f1e43f5fb5948c4 .

TheCartpenter:
Head duplication have been solved  and can use returns instead of return as, of this day, in the opencart-3 repository.

PR #10901, opened 2022-05-26
Update contact.php

PhamToan:
less spam box and fix bug when using SMPT like sendgrid


ADDCreative:
I did a pull request https://github.com/opencart/opencart/pull/8932 for this back in November 2020.

PR #10918, opened 2022-05-27
[Master] - Replaced recursive_move method with controller - install/controller/upgrade/upgrade_1.php file

TheCartpenter:
To avoid method replications


TheCartpenter:
Closing this. Since the master branch already uses Ajax via JSON to handle the files, the installation for, either, step 1 and step 4 should do the same to address the same URL that would take the parameters with ->set. Then, to apply the recursiveMove(param1, param2) class since the master branch has been converted into JSON already.

PR #10929, opened 2022-05-29
return to returns

BuslikDrev:


stalker780:
Strange. Why `returns`?

Why not same as `category`, `product`, `attribute`, `order`, `language`, etc?

It should be called `return`.

chulcha:
[upload/catalog/controller/account/**returns**.php]

BuslikDrev:
> Strange. Why `returns`?
> 
> Why not same as `category`, `product`, `attribute`, `order`, `language`, etc?
> 
> It should be called `return`.

The creators could not cope with something https://github.com/opencart/opencart/pull/10888#issuecomment-1137856250

stalker780:
> The creators could not cope with something [#10888 (comment)](https://github.com/opencart/opencart/pull/10888#issuecomment-1137856250)

Perhaps it's better to rename it to some synonym. `revoke` for example, or `cancellation`. I am not that strong in english, but sure there should be a synonym which will suit and not be a reserved php word.

TheCartpenter:
> > The creators could not cope with something [#10888 (comment)](https://github.com/opencart/opencart/pull/10888#issuecomment-1137856250)
> 
> Perhaps it's better to rename it to some synonym. `revoke` for example, or `cancellation`. I am not that strong in english, but sure there should be a synonym which will suit and not be a reserved php word.

```return``` is a reserved php word, not ```returns```. In other e-commerce platforms, ```returns``` is also used. It could not be named ```revoke``` or ```cancellation``` since these terms are already designated for order statuses and could be brought to confusion to the users whereas the Event Triggers would be used with the ```cancellation``` from the core vs. the payment extension modules that also uses ```revoke``` depending on the payment service providers required to store owners.

TheCartpenter:
Furthermore, the ```cancelled``` order status does not have the same significance as the ```return``` order status.

PR #10944, opened 2022-05-30
OC4: fix broken utf8_substr with default 0 length

stalker780:
Now it always returns an empty string


stalker780:
However it is not quite correct too.
If there is a need to use 0 length, it will never work. It will always return full string.
Behavior will be different from original php substr function.

ADDCreative:
https://github.com/opencart/opencart/pull/11073 should be correct.

PR #10956, opened 2022-05-31
TypeError: Cart\User::isLogged():

BuslikDrev:
TypeError: Opencart\System\Library\Cart\User::isLogged(): Return value must be of type int, null returned in

https://github.com/opencart/opencart/issues/10971

### DIFF
--- a/upload/system/library/cart/user.php
+++ b/upload/system/library/cart/user.php
@@ -1,9 +1,9 @@
 <?php
 namespace Opencart\System\Library\Cart;
 class User {
-	private $user_id;
-	private $username;
-	private $user_group_id;
+	private int $user_id = 0;
+	private string $username = '';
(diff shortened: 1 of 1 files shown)


upload/system/library/cart/user.php
```diff
@@ -1,9 +1,9 @@
 <?php
 namespace Opencart\System\Library\Cart;
 class User {
-	private $user_id;
-	private $username;
-	private $user_group_id;
+	private int $user_id = 0;
+	private string $username = '';
+	private int $user_group_id = 0;
 	private $permission = [];
```
TheCartpenter, line 7: ```private array $permission = [];```

TheCartpenter:
That might be one of the cause to why users from the OC forum reported issues to stay logged in in their OC admin ...

PR #10963, opened 2022-05-31
Removed old Jquery files

condor2:


cgcmart:
@condor2 

It seems jquery-ui never used, is it removed, too?

PR #10974, opened 2022-05-31
Cache Problem : here I recommond these lines to solve cache problem i…

smohamadabedy:
…n opencart pages. the main problem is whole page will be cached and cause bad ux but I figure out these two lines help to disable caching for entire page, the problem is observed even in product, catalog , .. forms when all javascript actions will not work properly on submiting or reloading the page.

```
$response->addHeader('Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0');
$response->addHeader('Pragma: no-cache');
```


ADDCreative:
The Cache-Control header is now set twice on the catalog side. If setting in the framework the following should be removed, else set individually in each session start up.
https://github.com/opencart/opencart/blob/9e1c9fcc6bcd3b4a0e8212a63f6092d406eeea4a/upload/catalog/controller/startup/session.php#L55

Expires header is also needed if Apache mod_expires is used with an ExpiresDefault set.

Also as it stands a redirect doesn't set the headers. There are cases when the redirect will be cached by the browser. For example going to checkout with an empty cart, then when next going to checkout with something in the cart you get redirected back to the cart page and not taken to checkout.
https://github.com/opencart/opencart/blob/373430288855fe81aad4fa0242ab1c1afa70d2ef/upload/system/library/response.php#L49-L52

PR #10997, opened 2022-06-02
pdo problem

smohamadabedy:
To solve the PDO problem in the issues section for searching and ...

### DIFF
--- a/upload/catalog/model/catalog/product.php
+++ b/upload/catalog/model/catalog/product.php
@@ -69,7 +69,7 @@ class Product extends \Opencart\System\Engine\Model {
 				$words = explode(' ', trim(preg_replace('/\s+/', ' ', $data['filter_name'])));
 
 				foreach ($words as $word) {
-					$implode[] = "pd.`name` LIKE '%" . $this->db->escape($word) . "%'";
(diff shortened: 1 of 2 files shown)


TheCartpenter:
That may not solve it for other DB libraries than PDO. Perhaps the escape function or the preg_replace function should be re-validated prior to use the query.

smohamadabedy:
> That may not solve it for other DB libraries than PDO. Perhaps the escape method or the preg_replace function should be re-validated prior to use the query.

smohamadabedy:
errors occure only in these two file, I checked others models and all of them pass % inside the skape.

cgcmart:
`$implode[] = "pd.`name` LIKE '%" . $this->db->escape_like_str($word) . "%'";`

TheCartpenter:
For MySQLi, this could differ based on the locations where the percentage is placed since the % would be also escaped.

TheCartpenter:
This solution may help to use otherwise with LIKE: https://stackoverflow.com/a/51338452 .

danielkerr:
already fixed

smohamadabedy:
note that the problem is with `%`  and as far as I know real_escape_string function will NOT escape _ (underscore) and % (percent) signs, so there is no conflict to move `%` inside escape function.

PR #10998, opened 2022-06-02
Added GLOB_BRACE with $next - admin/controller/common/filemanager.php file

TheCartpenter:


danielkerr:
why because i did it on one do u then decide i should do it for all

TheCartpenter:
It's not really about a 'decision', it's more about a 'proposal' based on what I noticed in the codes. Besides, while some changes were not required, based on your replies, some were also merged in the meantime which means that, somewhere, it was simply meant to be ...

danielkerr:
U ONLY NEED TO ADD IT IF U ARE COPYING OR DELETING FILES

Issue #9417, opened 2021-03-09
Validate Product Review


chulcha:
**What version of OpenCart are you reporting this for?**
master and early
**Describe the bug**

Add review

**To Reproduce**
Any POST request, for example from spammers 

**Expected behavior**

https://github.com/opencart/opencart/blob/1215717947c07704e163ca538dd11690db91817d/upload/catalog/controller/product/product.php#L619

```
if (isset($this->request->get['product_id'])) {

}
```

It may be necessary to check the existence of the product

TheCartpenter:
Added commit in opencart-3 repository: https://github.com/opencart/opencart-3/commit/4e6bb54e790f409f9bd25157df95864bfa2c3f60 .

PR #11008, opened 2022-06-03
Let Users restrict reviews to customer with purchased item

smohamadabedy:
I think this is good to let user restricts  reviews  and only customers with purchased  item can add reviw. 
admin -> Setting and catalog -> product are updated.


TheCartpenter:
I never really liked these sorts of features because whenever customers wants to purchase their goods, and they never tried them before, they'd have to send these reviews before officially buying the items when the philosophy is rather about: Try before you buy .., not: rate before you try.

upload/catalog/controller/product/product.php
```diff
@@ -416,6 +416,13 @@ public function index(): void {
 				$data['review_guest'] = false;
 			}
 
+			if ($this->config->get('config_review_on_purchase') && $this->customer->isLogged()) {
+				$data['review_on_purchase'] = $this->model_catalog_product->cutomerHasPurchase($this->customer->getId(),$product_id) > 0 ? true : false;
+			}else{
+				$data['review_on_purchase'] = true;
+			}
+			var_dump($this->config->get('config_review_on_purchase') ,$this->model_catalog_product->cutomerHasPurchase($this->customer->getId(),$product_id) > 0,$data['review_guest'] );
```
TheCartpenter, line 424: Is this needed for specific reasons?

smohamadabedy: no actually was for testing I'll delete it.

smohamadabedy:
> I never really liked these sorts of features because whenever customers wants to purchase their goods, and they never tried them before, they'd have to send these reviews before officially buying the items when the philosophy is rather about: Try before you buy .., not: rate before you try.

I'll saw this in feature release To do section, And actually I really like it to be in opencart as I close review section for my opencart customers most of the times.

TheCartpenter:
Yes, once the item has been purchased. Not before would be a great fit to the core.

danielkerr:
https://github.com/opencart/opencart/issues/9417

Issue #10887, opened 2022-05-25
[Bug] Shipping estimate and text success language


condor2:
**What version of OpenCart are you reporting this for?**
Latest from github

**Describe the bug**
If you use other language then english, text_succes is loaded from english file and not from other language.

Also after shipping method is applied, some translation are switched to english
![cart_bugs](https://user-images.githubusercontent.com/14251939/170253786-76ab4732-68f6-48f4-a598-16ea114fa4af.png)

If I remove/delete english from admin, everything is ok.

TheCartpenter:
Yes, this is a persistent issue even from previous OC versions. The part where it's difficult to see is if a language pack is either published accurately or if it contains errors to result this.

condor2:
Language is OK because after I removed english, all variables are read.

TheCartpenter:
But when removing other language packs is when the language text definition is not translated after a refresh and that can also be caused by the extension language pack by its distribution. Have you tried other language packs by other distributions to reproduce the issue?

condor2:
No issue on 3.0.3.8

TheCartpenter:
Might need further testing on this one.

BuslikDrev:
> It is a bug 100%,

https://github.com/opencart/opencart/issues/10922#issuecomment-1140285134

nmpetkov:
I tested in my test OC4 site, and all works fine in my opinion. Everyone can also test there:
https://octest40.cmstory.com/
Just select Bulgarian language and test.

mhcwebdesign:
@nmptekov: It took me only a few seconds to come across a bug in https://octest40.cmstory.com/canon-eos-5d-%D0%BA%D0%B0%D0%BC%D0%B5%D1%80%D0%B0?route=product/product&language=bg , namely the add to cart doesn't work!
Tested in Firefox 101.0

The bottom line is: OpenCart 4.0.0.0 is not yet suitable for live sites!

nmpetkov:
@mhcwebdesign ?? Add to cart works, but this product needs to specify "Select" :-) Just specify the item on the page colored in red -)

mhcwebdesign:
@nmptekov: Sorry, no it doesn't work! And of course I selected the required option.
Tested with Chrome Version 102.0.5005.61 (Official Build) (64-bit) and Firefox 101.0.

OpenCart 4.0.0.0 has so many bugs, when I have some time I'll post them all here on github.

nmpetkov:
[...]

PR #11019, opened 2022-06-05
Add language parameter where missing in catalog side

nmpetkov:
Thi resolves the issue #10887

Description: Since in Opencart 4 the selected language in the front part of the site is transmitted as a GET parameter, for all links and ajax calls that depend on the language, they must contain the selected language as a parameter. This is done in general, in this PR are added in places where it is missing and this leads to unexpected changes in the language for the user.

### DIFF
--- a/upload/catalog/controller/account/address.php
+++ b/upload/catalog/controller/account/address.php
@@ -84,6 +84,8 @@ class Address extends \Opencart\System\Engine\Controller {
 	public function form(): void {
 		$this->load->language('account/address');
 
+		$data['language'] = $this->config->get('config_language');
+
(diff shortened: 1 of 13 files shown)


nmpetkov:
Mentioned PR above resolves the problem in the front of this issue. Can check in the test site: https://octest40.cmstory.com/.

condor2:
On register.twig you write url and index with UPPERCASE :)

condor2:
If you are still here :), add **alert-dismissible** on $('#alert').prepend('<div class="alert alert-success">

shipping.twig

nmpetkov:
> On register.twig you write url and index with UPPERCASE :)

For some reason a technical error from using the shortcut ... corrected!

condor2:
@nmpetkov  don't forget this.

> If you are still here :), add **alert-dismissible** on
> 
> `$('#alert').prepend('<div class="alert alert-success">`
> 
> shipping.twig

Issue #11026, opened 2022-06-05
Is not possible to select *.ocmod.zip files using the extension installer upload file dialog 


dharmawebstudio:
**What version of OpenCart are you reporting this for?**
4.0.0.0

**Describe the bug**

- Browse to extension installer admin page. 
- Click on upload extension button
- File dialog window does not allow to select .ocmod.zip file.  

**Expected behavior**
- Be able to select ocmod..zip files

**Screenshots / Screen recordings**

Consider the following screenshot. Is not possible to select the *.ocmod.zip file.

![Captura de pantalla 2022-06-05 a las 21 11 26](https://user-images.githubusercontent.com/8374245/172067000-a79206e2-ccde-43bc-845e-0cdcd2ef2ce1.png)

**Server / Test environment (please complete the following information):**
 - Google Chrome in Mac OS. Versión 98.0.4758.109

dharmawebstudio:
Not sure if somebody else is facing this issue, but I propose to change on this line: 

https://github.com/opencart/opencart/blob/4.0.0.0/upload/admin/view/template/marketplace/installer.twig#L54

```
accept="ocmod.zip".
```

Needs to be replaced by:

```
accept=".zip".
```

Then check in JS or server side if the file name also contains ".ocmod.zip".

stalker780:
No problem in Windows 10/11, Firefox 101/Chrome 102
Current `accept=".ocmod.zip"` works correctly.

stalker780:
Looks like MacOS file manager does not recognize extensions in input accept property. Only mime types

https://stackoverflow.com/questions/47386981/input-type-file-attribute-accept-not-working-in-safari-on-macbook

https://stackoverflow.com/questions/37357970/safari-doesnt-respect-input-type-accept-extension

danielkerr:
google chrome is fine. seems to be a safari issue which i cannot install on my windows pc. ffs i dont want to start messing around with vm's

padaliyajay:
I have also same issue in my Mac. this issue is confirmed if you are using macOS. no matter what browser you are using.

mhcwebdesign:
This is not an OpenCart bug, but it's an macOS file manager issue, hence it should be reported to https://developer.apple.com/bug-reporting/

Also, be aware that the Safari web browser has quite a few well-known bugs, better to use a proper web browser like Chrome or FireFox.

padaliyajay:
this issue is also in Chrome if you are using macOS.

[...]

PR #11040, opened 2022-06-06
accept="application/zip,.zip" for  Safari

BuslikDrev:
=( https://github.com/opencart/opencart/issues/11026

### DIFF
--- a/upload/admin/view/template/marketplace/installer.twig
+++ b/upload/admin/view/template/marketplace/installer.twig
@@ -51,7 +51,7 @@ $('#button-upload').on('click', function () {
     if (!$('button-upload').prop('disabled')) {
         $('#form-upload').remove();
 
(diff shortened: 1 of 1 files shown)


TheCartpenter:
I would not suggest that change to be made since the suggested change is about removing the ocmod. entirely because of a single browser.

BuslikDrev:
> Я бы не советовал вносить это изменение, поскольку предлагаемое изменение связано с удалением ocmod. полностью из-за одного браузера.



> I would not suggest that change to be made since the suggested change is about removing the ocmod. entirely because of a single browser.

```
	if (/macintosh|iphone|ipod|ipad/.test(window.navigator.userAgent.toLowerCase())) {
		var accept = ' accept="application/zip,.zip"';
	} else {
		var accept = ' accept=".ocmod.zip"';
	}
```

TheCartpenter:
That change would impose that any other browsers would be using the .ocmod.zip extension as part of the installer except Safari.

Issue #10941, opened 2022-05-29
httponly = false 


BuslikDrev:
Why would OpenCart 4 let js scripts get a session cookie?

danielkerr:
httponly = false got nothing to do with js being able to set cookies. sick of playing guessing games.

BuslikDrev:
> httponly = false got nothing to do with js being able to set cookies. sick of playing guessing games.
<button onclick="console.log(document.cookie);" type="button">Get Session Cookie</button>
![image](https://user-images.githubusercontent.com/32706609/172239245-c633485f-72ca-45aa-a301-06cdd6e0d546.png)

PR #11043, opened 2022-06-06
session security

BuslikDrev:
https://github.com/opencart/opencart/issues/10941


danielkerr:
then setting cookies during testing will be impossible. u so smart!

Issue #11011, opened 2022-06-04
CSS need to be updated - 1 product/row- design issue


condor2:
**What version of OpenCart are you reporting this for?**
latest github
![40_row_prod](https://user-images.githubusercontent.com/14251939/171994877-1e3b68af-d594-42e1-82fb-661950a594c3.png)

vs 3.0.x.x
![30_row_prod](https://user-images.githubusercontent.com/14251939/171994908-2ab16d5e-7d92-405f-ae77-04388548da07.png)

Also grid not same height
![grid_height](https://user-images.githubusercontent.com/14251939/171994940-92fd942e-e1ad-42e8-95ef-40d7a15cf8f9.png)

danielkerr:
a pull request please.

condor2:
@danielkerr 

`id="product-list"` was intentionally deleted? from featured, latest, bestseller

danielkerr:
id="product-list" might be duplcate with multiple modules

wakson:
I actually notice this occuring on v4.0.0.0 when the screen is at resolution 2560 x 1440 but also occurs when i am changing screen size. It really looks where when the height is not the same...

danielkerr:
@[wakson](https://github.com/wakson) nobody cares ur problems! 

its been reported. it will be investigated, fixed and closed. we dont or neeed ur input

danielkerr:
[condor2](https://github.com/condor2) suggesting adding id="product-list" is dumb!

condor2:
It was a **question** not a suggestion you dumb

danielkerr:
i ment who ever put it there. because an id= would have appeared multiple times.

PR #11066, opened 2022-06-07
products Height problem, issue #11011

smohamadabedy:
single commit with less changes for issue #11011


danielkerr:
i think will break add to cart

smohamadabedy:
> i think will break add to cart

I tested this one and worked fine. I changed form position but hidden inputs are still inside the form tag and it works properly

PR #11084, opened 2022-06-07
Added array|bool on getOrder() method - catalog/model/account/order.php file

TheCartpenter:
Could also be void ...


danielkerr:
needs to be array

PR #11087, opened 2022-06-07
Added array type on getLinks() method - system/library/document.php file

TheCartpenter:
Could also be void ...


danielkerr:
its array only

PR #11088, opened 2022-06-07
Added array and string type on get and set methods - system/library/language.php file

TheCartpenter:


ADDCreative:
Set doesn't return anything.

PR #11097, opened 2022-06-08
Trimmed <br /> to <br/> - system/library/db/mysqli.php file

TheCartpenter:


condor2:
same in footer.twig
theme from design

PR #11099, opened 2022-06-08
Replaced <br /> with <br/> - system/library/db/pgsql.php file

TheCartpenter:


mhcwebdesign:
This is a prime example of unnecessary pull requests. Also, rather then posting tens of these pull requests within a few minutes, you could have combined them all into a single pull-request, to make life easier for others! 

See also https://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br

PR #11120, opened 2022-06-09
OC4: Improved admin getProducts function

nmpetkov:
Admin product model function getProducts is improved to be used from Opencart and module developers:
- fully backward compatible
- possibility to filter by all product table columns
- possibility to sort by all product table columns
- possibility to specify sql comparison operator
- build in count total (no need to use getTotalProducts)
- possibility to filter by category, manufacturer, store, SEO keyword

If this PR is accepted, I can make in similar way PRs for opders and customers.

### DIFF
--- a/upload/admin/controller/catalog/product.php
+++ b/upload/admin/controller/catalog/product.php
@@ -152,6 +152,7 @@ class Product extends \Opencart\System\Engine\Controller {
 
 		$data['products'] = [];
 
+        $product_total = 0;
 		$filter_data = [
 			'filter_name'     => $filter_name,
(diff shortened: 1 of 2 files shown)


nmpetkov:
@TheCartpenter 

If a module developer decides to filter by of the columns in products other than those in the current getProducts function, he will not be able to. In the new version of the function can. It is same for sorting.

Regarding $product_total: in the offered variant of the function this data is obtained with the same where conditions as in the filtering. In the old version, these were 2 different functions.

danielkerr:
ity won't work.  there will be times it requires LINK or ><

PR #11144, opened 2022-06-10
Removed config_cookie_id - admin/controller/setting/store.php file

TheCartpenter:
Not showing in admin/view/template/setting/store_form.twig file


TheCartpenter:
It is also showing in admin/language/en-gb/setting/store.php file. It might also be a missing block of code in the store_form.twig file ...

TheCartpenter:
Closing. Looks like it needs to be added.

Issue #11137, opened 2022-06-10
SMTP Mail


ProjectMyst:
If you use SMTP for sending mail, you can't make an order in the admin section. When you try to place an order you get an error.

I'm using tls:\\ with port 587

TheCartpenter:
Regarding email settings issues, please address them on the Opencart forum > << Your OC version >> > General Support section.

danielkerr:
hpow u know its a bug?

looks more like the port is wrong

ProjectMyst:
Because I tried with all ports, and it doesn't work with any of them. For example, using tls:// with port 587 works just fine when placing an order through the customers side but it will not work when putting an order through the admin side.

TheCartpenter:
Then, you'd need to contact your host or to request deeper troubleshooting from the OC forum as a custom job. At this point, it's not an OC issue.

ProjectMyst:
Hi

Sorry to be a pain but I'm not getting it, can you please explane why that is, why would it work on example.com/store and not example.com/store/admin

I mean, if it was the port then it wouldn't work full stop. If I customers makes an order over at example.com/store then the order works just fine and the invoice email gets sent out. But if I make an order over at example.com/store/admin for a customer then it will not work and gives me the error! It's clearly nothing to do with my hosting provider.

Would you be able to help me understand so I can try and fix the problem.

Thanks in advance.

ProjectMyst:
So I've done more test and worked it out what the problem is.

If you using SMTP for sending out mail and have extension Anti-Fraud IP enabled, it will not let you make orders through the admin backend but it will work through the customers front end.

danielkerr:
it will change it to what ever status u add there.

ProjectMyst:
There's nothing on the server logs, but on the console:

jquery-3.6.0.min.js:2

GET https://projectmyst.com/store1/admin/index.php?route=sale/order|call&user_token=c308440ca7697b9d0fc5208bc40e1a53&action=sale/order|confirm 500
send @ jquery-3.6.0.min.js:2
ajax @ jquery-3.6.0.min.js:2
(anonymous) @ index.php?route=sale/order|info&user_token=c308440ca7697b9d0fc5208bc40e1a53:3078
dispatch @ jquery-3.6.0.min.js:2
v.handle @ jquery-3.6.0.min.js:2
index.php?route=sale/order|info&user_token=c308440ca7697b9d0fc5208bc40e1a53:3125 
error

Network:

[...]

PR #11151, opened 2022-06-10
Changed int|void to int|bool - extension/opencart/catalog/model/fraud/ip.php file

TheCartpenter:
As per: https://github.com/opencart/opencart/issues/11137#issuecomment-1152733931

### DIFF
--- a/upload/extension/opencart/catalog/model/fraud/ip.php
+++ b/upload/extension/opencart/catalog/model/fraud/ip.php
@@ -1,7 +1,7 @@
 <?php
 namespace Opencart\Catalog\Model\Extension\Opencart\Fraud;
 class Ip extends \Opencart\System\Engine\Model {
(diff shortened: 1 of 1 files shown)


ProjectMyst:
> As per: https://github.com/opencart/opencart/issues/11137#issuecomment-1152733931

Pull request didn't merge.

TheCartpenter:
It won't be, since this suggestion was to let pass of an IP address lookup behind the scenes while managing an order within an API whereas, either, this extension should rather be more informative to the admin API users regarding a restriction imposed or for this extension to be entirely removed.

TheCartpenter:
An admin API group feature would need to be added by design in order to achieve this without affecting the fraud extensions as well as the orders based on an event /before trigger with JSON.

TheCartpenter:
Or, by using a fraud extension lookup by service.

PR #11165, opened 2022-06-11
[3.0.x.x] Fixed Affiliate custom fields not updated from Affiliate ac…

condor2:
…count


danielkerr:
3.x goex here: https://github.com/opencart/opencart-3

PR #11166, opened 2022-06-11
[3.0.x.x] Removed is_null from filter price-Product model

condor2:


danielkerr:
3.x goex here: https://github.com/opencart/opencart-3

PR #11208, opened 2022-06-12
Pushed parentheses on both common.js

TheCartpenter:


TheCartpenter:
I think I need to delete branches once submitted.

PR #11210, opened 2022-06-12
Pushed parentheses on both common.js file

TheCartpenter:


TheCartpenter:
Screw it. Directly from the site is way simpler.

PR #11217, opened 2022-06-12
SVG support and alert issue in filemanger

smohamadabedy:


danielkerr:
svg will never be supported because its not a real image format

PR #11227, opened 2022-06-12
OC4: Admin, Sale, Order, Filter: added date from/to possible

nmpetkov:
**Version**: current in master branch Opencart 4

**Location**: In the admin panel, Orders, filter section.

**Problem**: Currently in the filter section for "Date added" and "Date modified" there is an option to enter a date, but not from / to. This makes the report almost unusable. In most cases, the desire is to see the orders after date, before date, or between dates.

**Solution**: The date controls change, with a date "from" and a date "to" for each of "Date added" and "Date modified" (see picture below).

image


danielkerr:
i already fixed it.

nmpetkov:
@danielkerr It is completly different problem! Please look at it!

PR #11242, opened 2022-06-12
Added JSON array encoding - catalog/model/checkout/order.php file

TheCartpenter:
Also added isset lookup in $this->db->escape method.


TheCartpenter:
Closing. Lookup was meant to be removed.

PR #11259, opened 2022-06-13
Removed unused code from customer controller

condor2:


condor2:
On previous code structure in Customer list was displayed together: customers+affiliates

TheCartpenter:
Yes, correct. Which is why it is a bit hard to tell by design for v4 which side it is since the filters from PHP are still available basically.

condor2:
I think Daniel forgot to delete them :).
We will see

PR #11260, opened 2022-06-13
Removed unused affiliate code from customer model

condor2:


TheCartpenter:
Yes, confirmed. This key doesn't exist anymore and must be removed.

PR #11261, opened 2022-06-13
Added format_size() function - admin/controller/catalog/download.php file

TheCartpenter:


danielkerr:
it also annoying to see millions of pull requests about spaces. if there is a bug do a pull requ8est. im not that bothered about code formatting for now.

PR #11273, opened 2022-06-13
Added format_size() helper function - catalog/controller/checkout/shipping_address.php file

TheCartpenter:


danielkerr:
im not doing format size!

TheCartpenter:
The helper function is already in the core as you seem to have already used it in your admin/controller/marketplace/installer.php file. If you're referring to this commit specifically, however, I will simply close it.

danielkerr:
some one added it noty me. and igf they fucked with my installer code its a problem for me.

danielkerr:
im removing it. i should not have avccepted it

PR #11276, opened 2022-06-13
Added error_format_x language definition - catalog/language/en-gb/en-gb.php file

TheCartpenter:


danielkerr:
uploads should be done in megabyte size.

PR #11284, opened 2022-06-14
Update common.js autocomplete (good)

BuslikDrev:
The latest version of Chrome has to click on the field 2 times for the window to appear. Also, the window quickly disappears on the first move of the mouse.

### DIFF
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -424,7 +424,37 @@ var chain = new Chain();
             var $this = $(this);
             var $dropdown = $('#' + $this.attr('list'));
 
+            this.timer = null;
             this.items = [];
+
+            $.extend(this, option);
+
+            // Focus
+            $this.on('focus', function () {
+                $this[0].blur();
+                $this.unbind();
+                this.request();
+            });
+
+            // Keydown
+            $this.on('input', function (e) {
+                this.request();
+
+                var value = $this.val();
+
+                if (value && this.items[value]) {
+                    this.select(this.items[value]);
+                }
+            });
+
+            // Request
+            this.request = function () {
+                clearTimeout(this.timer);
+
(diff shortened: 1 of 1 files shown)


danielkerr:
im using latest and its not doing it to me

BuslikDrev:
> im using latest and its not doing it to me

[Edit category](https://demo.opencart.com/admin/index.php?route=catalog/category|form): click on the input "Parent" field. Do you immediately see a window with a list of options? Try to enable "save forms" in chrome - attribute autocomplete="on".

danielkerr:
Start typing something should come up

BuslikDrev:
> Start typing something should come up

When I type, the queries work. I don't get a listbox on the first click on the field (ajax works).

danielkerr:
im not here to fix ur personall issue. if u have noticed no one else reporting this

BuslikDrev:
> im not here to fix ur personall issue. if u have noticed no one else reporting this

You won't be able to catch the problem because of my bad English.
You ask others if they have an issue with the popup when autocomplete is enabled in google chrome. Have you ever tried adding non-existent data to any field on the demo site and clicking save so that the browser saves the options and starts offering them to you next time? Have you tried releasing the mouse quickly after the popup appears?

BuslikDrev:
Due to new data appearing in the "tag datalist" on the focus event, the popup window disappears. The window appears if chrome offers options from autosave. It also happens when you click again and quickly move the mouse out of the field.

stalker780:
I can also add a problem with autocomplete groups

Win10, Firefox 101

image
image

BuslikDrev:
> I can also add a problem with autocomplete groups
> 
> Win10, Firefox 101

This problem is not in the js script, but in the sql query.

stalker780:
No. JSON is fine.
But autocomplete selectbox does not group category options correctly
This is options structure

image

PR #11285, opened 2022-06-14
Added backquotes - admin/model/catalog/category.php file

TheCartpenter:


mhcwebdesign:
What is it these days with these backquote obsession? It's neither needed, nor is it a bug.

TheCartpenter:
It ain't about these days. It's about server specs where even partial characters of reserved terms would prevent SQL queries to load. It's been like this a long time with PHP. The rest just goes to the people to catch up on it.

mhcwebdesign:
Well, category_id isn't one of the reserved words: https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-I

TheCartpenter:
What that site doesn't reveal is how PHP has its unique case compared to the rest of the programming industry since it's the only language where backquotes needs to be used aside from the others even on the reserved terms whether it's partial or full.

mhcwebdesign:
No, not inside PHP strings, e.g. **"I can have any keyword inside my string, such as category namespace class etc"**.

danielkerr:
it is weird!

PR #11297, opened 2022-06-14
Added entry_address_format language definition - admin/language/en-gb/setting/setting.php file

TheCartpenter:


TheCartpenter:

PR #11298, opened 2022-06-14
Added address_format row - admin/view/template/setting/setting.twig file

TheCartpenter:


TheCartpenter:

PR #11300, opened 2022-06-14
Added backquotes - admin/model/customer/customer.php file

TheCartpenter:


danielkerr:
your just wasting time posting stuff like this. 

stop posting anyd look for real bugs by tersting pages notycode

TheCartpenter:
If you think I am wasting your time by addressing these issues, one thing straight you're gonna read here, regardless if that close button is a virtue for you to use, I didn't created PHP, the PHP industry did. If you have any feedbacks to provide to them, then go babycry to them about it, not to me. Besides, only minimal of them remains and it's where you're starting to complain about it. I don't know what you've been through in your life dude but I see since yesterday that you're becoming constantly defensive in your response while there's no reason for that to happen especially from someone who doesn't document process and expecting people to know on first sight. I don't have to take that from you.

PR #11313, opened 2022-06-15
Added config_affiliate_expire - install/opencart.sql file

TheCartpenter:
Own doing.


TheCartpenter:
- admin/controller/setting/setting.php
- admin/view/template/setting/setting.twig
- catalog/controller/startup/marketing.php

PR #11333, opened 2022-06-19
Added json['redirect'] - catalog/view/template/account/tracking.twig file

TheCartpenter:


danielkerr:
no reason

TheCartpenter:
If that json['redirect'] was posted for no reason, as you state, then this code is for Santa Claus:

https://github.com/opencart/opencart/blob/master/upload/catalog/controller/account/tracking.php#L79-L83

!

PR #11353, opened 2022-06-24
Added getSubscriptionByOrderProductId method - admin/model/sale/subscription.php file

TheCartpenter:


TheCartpenter:
Found another way.

PR #11368, opened 2022-06-26
Update installer.php

hbkpanchal:
Fixed file path when unlinking.


TheCartpenter:
Since ```$this->request->files['file']['tmp_name']``` is a super global variable, I think it was intended it was unlinked that way.

hbkpanchal:
Oh yeah true. that was nerd from me.

PR #11391, opened 2022-06-26
Added backquote on total - admin/model/sale/order.php file

TheCartpenter:


mhcwebdesign:
Not needed. There are more important issues to be addressed with OpenCart 4.

TheCartpenter:
If it wouldn't be important, other fields wouldn't contain it. The rest is simply based on opinions.

PR #11392, opened 2022-06-26
[OC 4.0 MASTER] Installer, uninstall: recursive delete directories

nmpetkov:
Installer, uninstall: recursive delete directories to prevent errors and partial uninstall.

Problem: When function uninstall in admin/controller/marketplace/installer.php tried to delete directories, and they are not empty, the procedure ends, but errors are registered in the log, and the extension directory remains undeleted.
This is a problem because if you try to install the same extension again, the procedure will not start.

With this PR the problem is solved.

### DIFF
--- a/upload/admin/controller/marketplace/installer.php
+++ b/upload/admin/controller/marketplace/installer.php
@@ -413,7 +413,7 @@ class Installer extends \Opencart\System\Engine\Controller {
 					if (is_file($path)) {
 						unlink($path);
 					} elseif (is_dir($path)) {
-						rmdir($path);
+						$this->recursiveDelDir($path);
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


TheCartpenter:
You could also refer to your admin/controller/common/developer|theme method rather than creating duplicated codes with slight different codes.

nmpetkov:
It is not dublicate code. Here in this PR is used recursive scan and delete, not one level.

As for the code you are pointing: in that places when deleting cache, it is also better to use recursive deletion, just in case directories are not empty.

TheCartpenter:
However, copying with recursive mode can also create side issues with shared or VPS servers as opposed to dedicated servers while moving files from the front-end.

nmpetkov:
@TheCartpenter Here nothing is connected to the kind of the server ot fronend of the site. This is just directory where particular extension is located. The purpose is to delete this directory with success.

PR #11394, opened 2022-06-27
Fixed the setting upgrade lookup - install/controller/upgrade/upgrade_8.php file

TheCartpenter:


TheCartpenter:
Will re-submit a one patch solution.

PR #11395, opened 2022-06-27
Fixed upgrade settings lookup - install/controller/upgrade/upgrade_8.php file

TheCartpenter:


TheCartpenter:
> why are u removing the prefix?

Took it from the opencart.sql file but when I removed the oc_ prefix, I forgot some spots where DB_PREFIX needed to be added in the upgrade yesterday.

TheCartpenter:
Should be all done now.

PR #11402, opened 2022-06-29
Fixed label FOR and select ID - admin/view/template/setting/setting.twig file

TheCartpenter:


TheCartpenter:
Dashes goes there, not underscores.

PR #11403, opened 2022-06-29
Added quotes on sort_order - admin/model/catalog/product.php file

TheCartpenter:


mhcwebdesign:
Seriously, have you nothing better to do than this obsession about quotes?  The DB-field sort_order is of type 'int' anyway!

danielkerr:
why does fixcking

PR #11415, opened 2022-07-02
Fixed language_admin - upload/admin/controller/startup/language.php

cbekir:


danielkerr:
no it was changed to just admin

cbekir:
It makes the admin language show when you select a language other than English. If you do not use this patch, the admin language is only English. You can add a second language and test it.

PR #11430, opened 2022-07-04
PHP 8.x support for catalog/model/product.php

mhcwebdesign:


ADDCreative:
@mhcwebdesign getPopularProducts also need updating, not that it is used by the core. I would do a pull request, but they just get closed for the 3.0.x.x_Maintenance branch.

https://github.com/opencart/opencart/blob/9e2d736d2e406d0c4b9dc705822a912949cb35c9/upload/catalog/model/catalog/product.php#L277-L283

mhcwebdesign:
> @mhcwebdesign getPopularProducts also need updating, not that it is used by the core. I would do a pull request, but they just get closed for the 3.0.x.x_Maintenance branch.
> 
> https://github.com/opencart/opencart/blob/9e2d736d2e406d0c4b9dc705822a912949cb35c9/upload/catalog/model/catalog/product.php#L277-L283

Thanks for noticing it, it's now updated.

PR #11434, opened 2022-07-05
[OC 4 master] Fix post install security checks

LapinChat:
### Post install security steps:

- **Remove installation folder:**
    - wrong trim in glob -> fixed
- **Move storage folder:**
    - Wrong trim in glob -> fixed
    - Source storage folder was not deleted -> fixed
- **Rename admin folder:**
    - Wrong trim in glob (`function admin()`, `function delete()`) -> fixed
    - Error while building redirection url -> fixed

Screen Shot 2022-07-04 at 9 23 25 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@
 
 # IDE Project files
 /upload/.idea/*
+/.idea/*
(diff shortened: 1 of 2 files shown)


danielkerr:
rtrim is not required

PR #11452, opened 2022-07-09
Use instant log writes

mhcwebdesign:
This is better for developers.

Currently, log entries often show up in quite a delayed manner in the DIR_LOGS.'error.log', which makes it harder for testing and development.

### DIFF
--- a/upload/system/library/log.php
+++ b/upload/system/library/log.php
@@ -12,8 +12,7 @@
 */
 namespace Opencart\System\Library;
 class Log {
-	private string $file;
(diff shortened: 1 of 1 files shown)


danielkerr:
doies not matter

mhcwebdesign:
Actually, it does matter! Right now, only when the destructor is called will the log file be generated. I have scenarios where even after multiple requests to the server no error.log file was written even though I had $this->log->write calls in several of my controllers for testing purposes.

PR #11456, opened 2022-07-11
Add for attribute to labels

MarvinKlein1508:
This is required in order to select the checkbox when the user clicks on the label.


danielkerr:
can not because popup link will not work

PR #11464, opened 2022-07-14
Fix validation for custom_fields in checkout

MarvinKlein1508:
@mhcwebdesign 
Both checkout guest and register validate all custom fields. It is right to be done for customer and address but affiliate should not be validated here. Otherwise the checkout does not work when you have cretaed a required custom field for affiliate.


mhcwebdesign:
Good point, just added it to the 3.0.0.0_Maintenance branch, too.

PR #11473, opened 2022-07-21
Installer upgrade

NikitaGuminsky:
Allows you to install multiple extensions at once. Allows you to automatically upgrade extension versions after upload and downgrade extension versions after upload if desired.


danielkerr:
the extension sytem is fine with what i have done.

PR #11475, opened 2022-07-22
[OC 4.0 Master] Admin Checkboxes Unchecked Values

nmpetkov:
Completion of the promised work in https://github.com/opencart/opencart/pull/11431

Description: Handle form 0/1 checkboxes, namely when the value is toggled between 0 and 1 with a checkbox, and the checkbox is unchecked, in the post is missing an entry with value 0.

### DIFF
--- a/upload/admin/view/template/catalog/category_form.twig
+++ b/upload/admin/view/template/catalog/category_form.twig
@@ -128,6 +128,7 @@
                 <label for="input-top" class="col-sm-2 col-form-label">{{ entry_top }}</label>
                 <div class="col-sm-10">
                   <div class="form-check form-switch form-switch-lg">
+                    <input type="hidden" name="top" value="0"/>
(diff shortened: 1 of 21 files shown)


mhcwebdesign:
So what exactly are you trying to accomplish? By its very nature, checkboxes, when unchecked, don't submit anything, see e.g. https://www.w3schools.com/tags/att_input_type_checkbox.asp

The controller needs to use something like this:

```
if (empty($this->request->post['my-checkbox'])) {
  // checkbox wasn't set
} else {
  // checkbox was set
}
```

nmpetkov:
Using only checkboxes with value 1 and relaying on existent or not in POST is not compatible with previous Opencart versions, and in some cases is not enough, in cases you are adding new variables with default value not 0, but 1.
You are right is many cases there is no problem, but it is better to have consistent way to handle on/off variables in post forms.

danielkerr:
its bootstrap that uses checkboxes as sliders so need to add default value

danielkerr:
cheers

PR #11502, opened 2022-08-03
Installer upgrade

NikitaGuminsky:
Allows you to upgrade extension versions after upload and downgrade extension versions after upload if desired.


danielkerr:
u need to speak to dimitry. the current version is fine. im not accepting ur changes

PR #11543, opened 2022-08-15
Fix DivisionByZeroException when limit in URL is not a number

MarvinKlein1508:
When the user provides a non numeric value per get parameter, the store throws a DivisionByZero exception. This fills up server logs when bots trying out some values.

This fixes it for the category side in front-end.

### DIFF
--- a/upload/catalog/controller/common/pagination.php
+++ b/upload/catalog/controller/common/pagination.php
@@ -16,10 +16,14 @@ class Pagination extends \Opencart\System\Engine\Controller {
 
 		if (isset($setting['limit'])) {
 			$limit = (int)$setting['limit'];
+			if($limit == 0) {
+				$limit = 10;
+			}
 		} else {
 			$limit = 10;
 		}
 
+		
 		if (isset($setting['url'])) {
 			$url = str_replace('%7Bpage%7D', '{page}', (string)$setting['url']);
 		} else {
--- a/upload/catalog/controller/product/category.php
+++ b/upload/catalog/controller/product/category.php
@@ -37,9 +37,13 @@ class Category extends \Opencart\System\Engine\Controller {
 
 		if (isset($this->request->get['limit'])) {
 			$limit = (int)$this->request->get['limit'];
+			if($limit == 0) {
+				$limit = $this->config->get('config_pagination');
+			}
 		} else {
 			$limit = $this->config->get('config_pagination');
 		}
+
 
 		$data['breadcrumbs'] = [];
 
--- a/upload/catalog/controller/product/search.php
+++ b/upload/catalog/controller/product/search.php
@@ -61,6 +61,9 @@ class Search extends \Opencart\System\Engine\Controller {
 
 		if (isset($this->request->get['limit'])) {
 			$limit = (int)$this->request->get['limit'];
+			if($limit == 0) {
+				$limit = $this->config->get('config_pagination');	
+			}
 		} else {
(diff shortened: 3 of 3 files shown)


MarvinKlein1508:
Might be necessary for other paginations as well, such as in product/search.php

danielkerr:
can notr replicate and not going to put ur garbage code in there

MarvinKlein1508:
@danielkerr to replicate visit any category and and add &limit=abc to url

MarvinKlein1508:
@danielkerr here you go:
https://demo.opencart.com/index.php?route=product/category&language=en-gb&path=57&limit=asd

@mhcwebdesign maybe you can tell him. He seems a bit ignorant to me.

mhcwebdesign:
I think it may make more sense to put your check, e.g.

```
			if (isset($this->request->get('limit') && (int)$this->request->get('limit')==0) {
				$this->request->get('limit') = 10;
			}

```
somewhere in the startup sequence, or in a pre-action, rather than modifying the code in so many different PHP files.

danielkerr:
Ur both dumb. It does not do this on any of the other pages.

danielkerr:
So this issue is the category page yet u clowns want me to change a value with code that that affects all pages. It's dumb. I can fix the category issue very quickly so it works like every other page

danielkerr:
Another point is u were to stupid to post the page so playing guessing games. If ur to stupid to post an issue correctly u should not coming up with fixes.and dictating to a professional how to so his job.

danielkerr:
In fact MarvinKlein1508 post another issue u Get banned. Ur don't know what ur doing

MarvinKlein1508:
As you may habe seen this also effects the search page and every page within admin that includes a pagination. As a professional you should have seen that I have edited the code gor category.php and search.php instead of instantly closing the pull request. 

You could have given some advice why you closed this instead of calling everyone dumb.

Putting the code into general doesn‘t hurt anyone and it will fix every page at once since the attribute is only changed when it‘s set via URL. Otherwise I can tell you whenever you add another page in the future which holds a pagination, you will miss this again.

The same thing applies to my other pull request. Why is it garbage code? You are literally doing the same just with Jquery. Except your approach isn‘t working. Also why do you wanna use a third party library to trigger a click event? Instead of just writing „“more garbage code“ you could have specified why you rather have leave the not working code instead of applying a simple fix which works a 100% and does not rely on a third party library.

BTW. It is very unprofessional to hold back an update for Opencart for three months so nobody could have deleted their products in their store. 

Von meinem iPhone gesendet

danielkerr:
its garbage because u are adding 3 lines when it can be done on one

		$limit = (int)$setting['limit'];
			if($limit == 0) {
				$limit = 10;
			}

just one:

&& (int)$this->request->get['limit']) {

amateur time.

MarvinKlein1508:
@danielkerr ah look a explanation which helps everyone. Everyone has profited from it. I became a better coder and you have this issue fixed. Wasn't that hard, right?

You also should add validation to admin/setting/setting for Default Items Per Page (Admin). When the user enters a non numeric value, every pagination which relies on this will crash. Could happens for example when there is an whitespace entered by mistake.

danielkerr:
admin does not crash

MarvinKlein1508:
@danielkerr it does when you set this value to a non numeric value
grafik

Then open any page that relies on this value. eg.g admin/index.php?route=catalog/category

Why aren't you using input type="number" for the numeric fields at all?

PR #11547, opened 2022-08-16
Update install.php

PhamToan:
hot fix bug Index column size too large. The maximum column size is 767 bytes (maybe need another better way) for table *_description for index 'name' column   => it can't create table (tested on localhost with wamp64, 10.5.13-MariaDB)
https://stackoverflow.com/questions/54541448/how-to-fix-1709-index-column-size-too-large-the-maximum-column-size-is-767-by


danielkerr:
no idea if this is good idea or not

PR #11550, opened 2022-08-16
Update install.php

Gabrym:
During the upgrade process, it doesn't work when the storage is moved to another directory.


TheCartpenter:
By doing that, the internal system folder during upgrades won't be able to grasp a next upcoming changes to the storage folder.

Gabrym:
When you change location your storage by first authorization to admin you can't use upgrade process in new version because it not works properly.
You need to patch install.php and upgrade scripts in 1 step manually.

danielkerr:
yes it does accont for the storage folder changes

danielkerr:
aj9thyer fucking dumb pull request for imagined issue

PR #11552, opened 2022-08-16
Log message never flush into file

oprachun:


mhcwebdesign:
@ oprachun : The current system/library/log.php does not always work, using on the **__destruct** function to flush it to the log file is not reliable. But you can easily replace the log.php with the one from Opencart 3 which actually works.

Issue #11545, opened 2022-08-15
Error using CLI install script. 


dharmawebstudio:
**What version of OpenCart are you reporting this for?**

- Latest version: 4.0.1.0
- Master branch.

**Describe the bug**

When running command similar to:

```
php cli_install.php install --username    admin
                            --email       email@example.com
                            --password    password
                            --http_server http://localhost/opencart/
                            --db_driver   mysqli
                            --db_hostname localhost
                            --db_username root
                            --db_password pass
                            --db_database opencart
                            --db_port     3306
                             --db_prefix   oc_ 
```

Get the following errors:

```
Undefined constant "APPLICATION" in /var/www/html/system/framework.php:4.
```

I think in order to fix that issue, something like the following next to be added:

```
define('APPLICATION', 'Install');
```

If we continue this way, then you should get the following error: 

```
Fatal error: Uncaught Error: Undefined constant "HTTP_SERVER" in /var/www/html/system/config/install.php:3
```

If that constant is defined in the file, something like: 

```
define('HTTP_SERVER', 'http:://whatever.com');
```

However I don't think this is the best way to define this constant, but I am just trying to sort out the error. It seems define that class require a refactor of the cli_install.php. 

Even after fix the error of Undefined constant "HTTP_SERVER"; then there is still another error, which is the following

```
Error: Call to undefined function Opencart\Install\Model\Install\db_schema() 
```

I think the reason behind this last error is because system/helper class are not loading via autoloader class; and to fix it is possible require the file in system/startup.php; or even better; include it only in cli_install.php because seems to be this file the only one using it. 

Finally another error will appear, which is coming from this line. 

```
$db->query("UPDATE `" . $db_prefix . "product` SET `viewed` = '0'");
```

The column viewed is not in that table since 4.0.1.0. 

**How to reproduce it**

Using [this docker environment](https://github.com/dharmawebstudio/docker-opencart), should be easy to test it. 

```
git clone https://github.com/dharmawebstudio/docker-opencart.git ./
bin/setup
bin/cli php /var/www/html/install/cli_install.php install --username ${OPENCART_USERNAME} \
                                                      --email ${OPENCART_EMAIL} \
                                                      --password ${OPENCART_PASSWORD} \
                                                      --http_server "https://${OPENCART_DOMAIN}/" \
                                                      --db_driver mysqli \
                                                      --db_hostname db \
                                                      --db_username ${MYSQL_USER} \
[...]

PR #11553, opened 2022-08-17
Remove outdated query from CliInstall

dharmawebstudio:
This is the first PR from a serie, which I have locally, to propose a solution to fix the cli_install.php, partially described in #11545. I think for code review purposes is easy to deal which each error in a separated PR. 

In this case, is simple, is removing an outdated query, because "DB_PREFIX . product" table does not contain a column "viewed" since 4.0.1.0. 

That column is now in a "DB_PREFIX . product_viewed"


danielkerr:
already removed

PR #11557, opened 2022-08-18
Fixed subscription data - catalog/model/checkout/order.php file

TheCartpenter:


waxedperfection:
I think an extra ]; was added. line 40.

cgcmart:
> I think an extra ]; was added. line 40.

Not add  but remove line 41.

PR #11579, opened 2022-08-23
Replaced commented code with TWIG compliance - catalog/view/template/common/header.twig file

TheCartpenter:


danielkerr:
codes already been removed.

PR #11580, opened 2022-08-23
Added quote on points value - admin/model/customer/customer.php file

TheCartpenter:


expert4pro:
It is no idea for good this. 
customer_reward.points is INT type.
You added the quote on points value that the database has to do additional work to convert the string to a number...

expert4pro:
and using it in a WHERE clause will prevent the optimizer from using indices (explain plan should show that)

TheCartpenter:
If that'd be the case, the rest of the locations using the quotes, regardless of subjects put with their integers, would result the same thing as opposed to those two last remaining locations that wasn't quoted.

PR #11588, opened 2022-08-24
Update opencart.sql

nexadomain:
There is no `config_password` in the `setting` table and in its form.
On deep inspection, I found `config_security` is missing from 'opencart.sql', and the order of keys is also showing - it is `config_security`


nexadomain:
This is for setting,
"Two-factor security for admin users"

TheCartpenter:
Confirmed. We'll also need this: https://github.com/opencart/opencart/pull/11583/commits/edce396497c47a3f5b81361393cf6643fb9ee2b5

danielkerr:
u dont need the config_security setting added. no errors show up without the setting.

PR #11590, opened 2022-08-25
Update order.php

PhamToan:
fixed bug when filter order status = missing order => order status id = 0


danielkerr:
that code is not there whos code are u working on?

PhamToan:
if (!empty($data['filter_order_status'])) {  => this variable for multi select  (the filter form doesn't has it), you can try by set url on browser. you will see it's wrong ...  i have customed for my client... so i knew that
But while choosing they can choose single value and has data = 0 only

PR #11614, opened 2022-08-30
Use seperate success message for guest checkout

iamsyh:
Maybe we should use a different success message when a user has selected "Guest Checkout" while checking out, to avoid confusion. Because the message right now says "Success: Your account has been created!".


danielkerr:
should be guest_success

PR #11625, opened 2022-09-02
Update free.php

osworx:
Inside the "Free" shopping method, this wording is confusing:
https://github.com/opencart/opencart/blob/8c7bb184fd94f9fa6045ca2b2b2c3a7283bb3663/upload/extension/opencart/admin/language/en-gb/shipping/free.php#L11

After it says:
https://github.com/opencart/opencart/blob/8c7bb184fd94f9fa6045ca2b2b2c3a7283bb3663/upload/extension/opencart/admin/language/en-gb/shipping/free.php#L17

While both should be **Sub-Total**


TheCartpenter:
By doing that, it would also disregard the products where taxes may still be charged depending on the countries the product originates from.

PR #11655, opened 2022-09-14
Update order_info.twig

sabeti05:
Edit order - add product - file option is not filling the correct input after successful upload.


danielkerr:
cheers

Issue #11662, opened 2022-09-17
php Warning:  Undefined array key "status" 


cgcmart:
https://github.com/opencart/opencart/blob/ef2e45a654b9f61932669da121cc07475f1641dc/upload/admin/controller/extension/module.php#L64

go to Admin Navigation menu >> Extensions
choose the extension type: Modules

![php_error_module_status](https://user-images.githubusercontent.com/14033183/190876213-b26731dc-7f72-43f8-ada0-abbd4e27b6e2.jpg)

TheCartpenter:
Confirmed. Done: https://github.com/opencart/opencart/pull/11663

cgcmart:
replace [opencart/upload/admin/controller/extension/module.php](https://github.com/opencart/opencart/blob/ef2e45a654b9f61932669da121cc07475f1641dc/upload/admin/controller/extension/module.php#L64) Line 64
`'status' => $setting_info['status'] ? $this->language->get('text_enabled') : $this->language->get('text_disabled'),`

with
`'status' => $this->config->get('module_' . $code . '_status') ? $this->language->get('text_enabled') : $this->language->get('text_disabled'),`

The php Warning is gone!

TheCartpenter:
By doing that, it will also skip the $setting_info from the rest of the setting key where the extensions will also be installed by remaining with one config key.

danielkerr:
they all need status other wise they will not load

PR #11664, opened 2022-09-17
Added if lookup - admin/view/template/extension/module.twig file

TheCartpenter:
As per: https://github.com/opencart/opencart/issues/11662


TheCartpenter:
That won't do it. Closing.

PR #11675, opened 2022-09-22
Update customer.php

nexadomain:
no `error_regex` exist in 'admin/language/en-gb/customer/customer.php'
`error_custom_field` exists


danielkerr:
sl ur idea is to remove the regex

PR #11681, opened 2022-09-24
Improved can not with cannot - admin/controller/common/security.php file

TheCartpenter:


ADDCreative:
You have // 1. twice and the second commas are not needed.

PR #11696, opened 2022-09-24
Sentence correction. Methods unfinished - admin/controller/mail/subscription.php file

TheCartpenter:
These methods might still be under development. I am not sure.


ADDCreative:
Should be subscription status and not order status. Also it's not a function that returns anything, it's a variable that is or isn't.

PR #11698, opened 2022-09-24
Aligned equal alignments - admin/language/en-gb/catalog/product.php file

TheCartpenter:


ADDCreative:
Now the alignment is a mix of tabs and spaces. You should stick to one. I believe OpenCart has always used tabs for indentation and spaces for alignment.

PR #11699, opened 2022-09-24
Improved sentence - admin/controller/user/user.php file

TheCartpenter:


ADDCreative:
The original was better. The new wording doesn't sound correct. You can't enforce the user, only rules, etc.

PR #11704, opened 2022-09-24
Improved sentence - catalog/controller/event/activity.php file

TheCartpenter:


ADDCreative:
The order status does not 'returns 0'. It should read "If the last order status id was 0 and new order status is not 0".

PR #11705, opened 2022-09-24
Improved sentence - catalog/controller/mail/order.php file

TheCartpenter:


ADDCreative:
The original made more sense. The order status doesn't 'return', it 'is' or 'is not'.

PR #11706, opened 2022-09-24
Improved sentence - catalog/controller/product/compare.php file

TheCartpenter:


ADDCreative:
This makes less sense than the original. Also it you are going to refer to the numeric value returned by count, then the compair is greater than 3. It original comment was fine as it was, as it referred to the new number of products.

PR #11759, opened 2022-09-28
Bootstrap 5.2.1 fix tooltips

stalker780:
Tooltips are shown in the bottom of a page when hovered second time. 
Added a fix destroying an old tooltip instance before showing.

First hover:
image

Second hover:
image

### DIFF
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -79,7 +79,14 @@ $(document).ready(function () {
 $(document).ready(function () {
     // Tooltip
     var oc_tooltip = function () {
-        // Apply to all on current page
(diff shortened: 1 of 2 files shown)


danielkerr:
this fucked the tooltips up and shopwing a error in the console

stalker780:
@danielkerr 
> this fucked the tooltips up and shopwing a error in the console

Strange. I have checked in Firefox and Chrome. There was no any error.
I will check again.

stalker780:
Yes. I see now. There is an error in console if you move quickly between buttons.
It does not affect interface, but I agree there should be no JS errors.
I will prepare another pull request cause tooltips still should be fixed.

PR #11767, opened 2022-10-03
Update common.js

padaliyajay:
Variable scope should be define and local otherwise error will show when strict mode is enabled.


punitkumar:
Use of let than var is more recommended

for (let key in json['error']) {
instead of
for (var key in json['error']) {

PR #11772, opened 2022-10-04
Removed root composer.json

qaharmdz:
Removed root composer.json contains wrong information.
As a note, the composer.json was previously removed on commit cd7486c.


danielkerr:
no idez who put it there.,

PR #11781, opened 2022-10-08
Added IF statement on customer payment ID - catalog/model/checkout/subscription.php file

TheCartpenter:


TheCartpenter:
If an extension is required to add a new payment method, then an IF condition might be needed for the customer payment ID with the subscriptions.

PR #11789, opened 2022-10-10
Added the rest of mandatory subscription fields - admin/model/sale/subscription.php file

TheCartpenter:
These fields needs to be added in the array in order to complete the transaction with the addTransaction() method as to fill the mandatory fields on the database. The subscription_id is not included because we are already mentoring the user to lookup the subscription prior to add it to the subscription transaction table.


TheCartpenter:
Closing this one. The getSubscription() can already pull the required fields.

PR #11796, opened 2022-10-13
Renamed array name: result to payment_method - catalog/model/checkout/payment_method.php file

TheCartpenter:


TheCartpenter:
Not an issue.

PR #11816, opened 2022-10-17
Event support for library methods

NikitaGuminsky:
Adding event support for library methods


TheCartpenter:
You're suggesting to use the entire Registry object into the library/template and TWIG files. I am not sure this would be ideal as the same to remove the language code from the new Language class ...

PR #11822, opened 2022-10-18
Update module.php

PhamToan:
sub-module uses setting status instead of main config status


danielkerr:
i think pointless change

PhamToan:
@danielkerr  try to install module banner then add some sub-module you can see on module list ... the status is alway disable

PR #11823, opened 2022-10-18
Event support for library methods

NikitaGuminsky:
Adding event support for library methods.


danielkerr:
wasting everyones time

PR #11824, opened 2022-10-18
Update contact.php

PhamToan:
we can break line for comment


TheCartpenter:
To consider this, an HTML email would need to be sent. An example can be seen from admin/controller/mail/returns.php file in regard to nl2br function with the comment field.

PR #11826, opened 2022-10-18
Bootstrap fix tooltips onclick

stalker780:
Found another bug with tooltips.

After a button is clicked, tooltip does not work anymore because of `$('.tooltip').remove();`

We should use the same approach as on `mouseenter` to hide a tooltip correctly `tooltip.hide();`

### DIFF
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -90,8 +90,11 @@ $(document).ready(function () {
 
     $(document).on('mouseenter', '[data-bs-toggle=\'tooltip\']', oc_tooltip);
 
-    $(document).on('click', 'button', function () {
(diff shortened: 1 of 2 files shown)


stalker780:
Here is a video of this bug. Same for frontend cart/compare/wishlist buttons.

https://user-images.githubusercontent.com/6297070/196480204-de8ac5db-bd3d-4f91-a051-dbfe0278a575.mp4

danielkerr:
stop messing with this u dojnt know what u are doing

stalker780:
Just fixed a small annoying bug.
Instead of wasting your words pointing everyone how worthless they are, you could just give me a hint where I am wrong :)

Anyway, BS tooltips must be operating with tooltip object and not jQuery selectors.

PR #11837, opened 2022-10-19
Removing empty strings - system/library/mail.php file

TheCartpenter:


danielkerr:
u should noot remove the

TheCartpenter:
Ok, the specific instanced variables as per these commits are now restored.

PR #11871, opened 2022-10-22
Fixes the maintenance page from startup - catalog/controller/startup/maintenance.php file

TheCartpenter:
Obviously, the system/library/cart/user.php file is enforcing its validation to be either set to true or false. Therefore, this fix resolves the issue.

### DIFF
--- a/upload/catalog/controller/startup/maintenance.php
+++ b/upload/catalog/controller/startup/maintenance.php
@@ -18,7 +18,7 @@ class Maintenance extends \Opencart\System\Engine\Controller {
 			// Show site if logged in as admin
 			$user = new \Opencart\System\Library\Cart\User($this->registry);
 
-			if (substr($route, 0, 3) != 'api' && !in_array($route, $ignore) && !$user->isLogged()) {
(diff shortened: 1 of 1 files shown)


danielkerr:
it does return true or false.

TheCartpenter:
Yes, it does 'return' true or false but it does not 'validate' a method from a constructor as a returned true or false value while using '!' in front. What you are implying in your validation by adding a: '!' in front is: != true and it is not what you're using in your thermal operator in the isLogged() method. You are inquiring a: !== true . When trying both results, using the: !== true validation does return the expected result. As an addition, I even added parentheses as a priority lookup in case there'd be concerns about it. However, there are no 'or' statement in that IF line condition, it shouldn't matter with the added parentheses in this case since 'and' conditions follows the same priorities altogether when no parentheses are used in these conditions.

danielkerr:
i ficxed it another way.

NickR2021:
Hi, I added this fix and it works but now anyone can view the site while under maintenance mode . I'm currently in PHP 8.1 and  OC  Ver 4.0.1.1.  using JournaI theme . I also check the default theme and the site is viewable too. any help would be helpful. Thanks

danielkerr:
just tested and it works fine

PR #11872, opened 2022-10-23
Fixes for deprecated dynamic properties for php 8.2

LinuxPractice:
Fix for deprecated dynamic properties.  

MYSQLI_REPORT_ERROR is already a property of mysqli_driver see https://www.php.net/manual/en/mysqli-driver.report-mode.php

request seems to have been forgotten as a public property of the request class

Added extends stdClass to proxy and all classes in library/cart

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@
 /upload/.idea/*
 /.idea/*
 /.DS_Store
+*.bak
--- a/upload/system/engine/proxy.php
+++ b/upload/system/engine/proxy.php
@@ -11,7 +11,8 @@
  * Proxy class
  */
 namespace Opencart\System\Engine;
-class Proxy {
+
+class Proxy extends \stdClass {
 	/**
 	 * __get
 	 *
--- a/upload/system/library/cart/cart.php
+++ b/upload/system/library/cart/cart.php
@@ -1,6 +1,6 @@
 <?php
 namespace Opencart\System\Library\Cart;
-class Cart {
+class Cart extends \stdClass {
 	private array $data = [];
 
 	/**
--- a/upload/system/library/cart/currency.php
+++ b/upload/system/library/cart/currency.php
@@ -1,6 +1,6 @@
 <?php
 namespace Opencart\System\Library\Cart;
-class Currency {
+class Currency extends \stdClass  {
 	private array $currencies = [];
 
 	/**
--- a/upload/system/library/cart/customer.php
+++ b/upload/system/library/cart/customer.php
@@ -1,7 +1,7 @@
 <?php
 namespace Opencart\System\Library\Cart;
 use \Opencart\System\Helper as Helper;
-class Customer {
+class Customer extends \stdClass  {
 	private int $customer_id = 0;
 	private string $firstname = '';
 	private string $lastname = '';
--- a/upload/system/library/cart/length.php
+++ b/upload/system/library/cart/length.php
@@ -1,6 +1,6 @@
 <?php
 namespace Opencart\System\Library\Cart;
-class Length {
+class Length extends \stdClass  {
 	private array $lengths = [];
 
 	/**
--- a/upload/system/library/cart/tax.php
+++ b/upload/system/library/cart/tax.php
@@ -1,6 +1,6 @@
 <?php
 namespace Opencart\System\Library\Cart;
-class Tax {
+class Tax extends \stdClass  {
 	private array $tax_rates = [];
 
 	/**
--- a/upload/system/library/cart/user.php
+++ b/upload/system/library/cart/user.php
@@ -1,6 +1,6 @@
 <?php
 namespace Opencart\System\Library\Cart;
-class User {
+class User extends \stdClass {
 	private int $user_id = 0;
 	private string $username = '';
 	private int $user_group_id = 0;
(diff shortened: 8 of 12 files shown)


danielkerr:
im stuck on this issue right now. god dam php for messing with mysql fuctions

mhcwebdesign:
> im stuck on this issue right now. god dam php for messing with mysql fuctions

Perhaps my recent pull request helps: https://github.com/opencart/opencart/pull/11877

mhcwebdesign:
@LinuxPractice : Instead of extending various library classes from the stdClass, it makes more sense to actually add the missing properties needed in those classes, see my pull request https://github.com/opencart/opencart/pull/11877

LinuxPractice:
My thinking was that by extending stdClass, you can keep the class properties dynamic and possibly not break any extensions or modifications.

LinuxPractice:
> im stuck on this issue right now. god dam php for messing with mysql functions

Did my mysql_driver fix work for you?

mhcwebdesign:
> Did my mysql_driver fix work for you?

I was using **mysqli_report** which is just an alias of [mysqli_driver->report_mode](https://www.php.net/manual/en/mysqli-driver.report-mode.php). Either way should work.

mhcwebdesign:
> My thinking was that by extending stdClass, you can keep the class properties dynamic and possibly not break any extensions or modifications.

True, though using dynamic properties always has a higher security risk, hence I think it is better practice to use regular class pre-defined properties where possible, as is common with object-oriented programming.

LinuxPractice:
> > My thinking was that by extending stdClass, you can keep the class properties dynamic and possibly not break any extensions or modifications.
> 
> True, though using dynamic properties always has a higher security risk, hence I think it is better practice to use regular pre-defined class-properties where possible, as is common with object-oriented programming.

I agree in a closed environment, my concern is is areas like the shopping cart end users may need to add properties.  For example, I have a heavily modified system that requires a quote id and a design id to be added to the cart product after a customer configures and designs a product.  Easy enough with ocMod, but how will I do that with events?

mhcwebdesign:
> I agree in a closed environment, my concern is is areas like the shopping cart end users may need to add properties. For example, I have a heavily modified system that requires a quote id and a design id to be added to the cart product after a customer configures and designs a product. Easy enough with ocMod, but how will I do that with events?

Using an event handler for this would require a trigger after system/library/cart/getProducts which of course doesn't yet exist in OpenCart 4. But I was thinking about using an extension with a decorator class which would wrap up the Cart class, and then define an event handler for e.g. 'controller/*/after' which would replace the Cart instance in the registry with the new decorator instance, e.g. something like 

`$this->registry->set('cart', new \Opencart\System\Extension\MyExtension\Engine\Decorator( $this->registry->get('cart'));`

I'll have to do some experiments during the next few days on this, the decorator design pattern is fairly straight forward, and different extensions can wrap up the cart instance in decorators, as above, resulting in nested decorator instances around the cart instance.

LinuxPractice:
> > I agree in a closed environment, my concern is is areas like the shopping cart end users may need to add properties. For example, I have a heavily modified system that requires a quote id and a design id to be added to the cart product after a customer configures and designs a product. Easy enough with ocMod, but how will I do that with events?
> 
> Using an event handler for this would require a trigger after system/library/cart/getProducts which of course doesn't yet exist in OpenCart 4. But I was thinking about using an extension with a decorator class which would wrap up the Cart class, and then define an event handler for e.g. 'controller/*/after' which would replace the Cart instance in the registry with the new decorator instance, e.g. something like
> 
> `$this->registry->set('cart', new \Opencart\System\Extension\MyExtension\Engine\Decorator( $this->registry->get('cart'));`
> 
> I'll have to do some experiments during the next few days on this, the decorator design pattern is fairly straight forward, and different extensions can wrap up the cart instance in decorators, as above, resulting in nested decorator instances around the cart instance.



> > I agree in a closed environment, my concern is is areas like the shopping cart end users may need to add properties. For example, I have a heavily modified system that requires a quote id and a design id to be added to the cart product after a customer configures and designs a product. Easy enough with ocMod, but how will I do that with events?
> 
> Using an event handler for this would require a trigger after system/library/cart/getProducts which of course doesn't yet exist in OpenCart 4. But I was thinking about using an extension with a decorator class which would wrap up the Cart class, and then define an event handler for e.g. 'controller/*/after' which would replace the Cart instance in the registry with the new decorator instance, e.g. something like
> 
> `$this->registry->set('cart', new \Opencart\System\Extension\MyExtension\Engine\Decorator( $this->registry->get('cart'));`
> 
> I'll have to do some experiments during the next few days on this, the decorator design pattern is fairly straight forward, and different extensions can wrap up the cart instance in decorators, as above, resulting in nested decorator instances around the cart instance.

In my use case, the quote id is used to retrieve a set of options and calculate the price based on an external formula and lookup tables so the new property would need to be available at library/cart run time.

I can see other use cases such as real-time production estimates and real-time delivery estimates in the shopping cart as well as line item shipping.  

Either way, you are still enabling dynamic content to be injected into the class

mhcwebdesign:
@anytizer : The very nature of the Proxy class in OpenCart is to allow for dynamic properties, hence it won't make sense to add a limited number of properties to this particular class, unlike the other ones, see my pull request https://github.com/opencart/opencart/pull/11877

anytizer:
@mhcwebdesign You are right. I realize, modifying proxy class isn't effective as fixing individual classes.

PR #11873, opened 2022-10-23
Added date_next field - system/library/cart/cart.php file

TheCartpenter:
model/checkout/order requires the date_next field from the ```$product['subscription']['date_next']```


TheCartpenter:
Since the ```date_next``` field is for extensions, I will close this.

PR #11877, opened 2022-10-24
Support for PHP 8.2

mhcwebdesign:
Added missing properties to various classes so PHP 8.2 won't complain about assignments to undeclared properties.


TheCartpenter:
This solution is equivalent to the ```$this->data``` we used to have in OC v1.5x releases. It propagates instanced variables throughout the platform from the same namespaces inside the constructors while the Registry Object is already taking care of that. We cannot and must not fall back into those days. It is as if no constructors would be built at all, since propagating variables like these simply over crosses the Engine by also adding the ```\Redis``` manually instead of using the namespace from the system\library\cache while this file is already handled by the framework.

As for the mysqli_report suggestion, however, it could probably happen but it's only speculations at this time since PHP 8.2 won't be out officially as a final release until next month (considering there are no postpones to the date by the industry).

mhcwebdesign:
> This solution is equivalent to the `$this->data` we used to have in OC v1.5x releases. It propagates instanced variables throughout the platform from the same namespaces inside the constructors while the Registry Object is already taking care of that. We cannot and must not fall back into those days. It is as if no constructors would be built at all, since propagating variables like these simply over crosses the Engine by also adding the `\Redis` manually instead of using the namespace from the system\library\cache while this file is already handled by the framework.

Can you translate in easy to understand English please? Using defined properties in classes in normal practice in object oriented programming, unlike dynamic properties.

TheCartpenter:
You're using a fallback solution, it's as most English as this can get.

mhcwebdesign:
> You're using a fallback solution, it's as most English as this can get.

Using class member variables (https://www.php.net/manual/en/language.oop5.properties.php) is a standard OOP practice, as it should have been the case with OpenCart in the first place.

TheCartpenter:
Exactly, then what your first post suggests is to bring this back.

danielkerr:
im closing this. i will do the updates my self as soon s as php8.2 is released

PR #11891, opened 2022-10-30
Permission needs to be checked before ID - admin/controller/sale/subscription.php file

TheCartpenter:


TheCartpenter:
By improving the IF condition, it will identity the fact that people might report a different order ID originating from extensions whereas the addTransaction() method will valid its match. Besides, the sale/subscription TWIG file does use the serialize() function to send all the form values per form IDs. Therefore, better yet not to take any chances.

PR #11909, opened 2022-11-03
Fixed typo of canceled to cancelled term - admin/view/template/setting/setting.twig file

TheCartpenter:


mhcwebdesign:
Both spellings are correct; Americans favor canceled (one l), while cancelled (two ls) is preferred in British English.

I think there are more important issues to be addressed and/or fixed for OpenCart 4.

TheCartpenter:
It is, after all, en-gb by default for the language code out-of-the-box. Which means, the platform is not considered for Americans until a language has been installed as an extension to show otherwise. However, it is correct to state that either two Ls or one L are both considered. My changes focuses on the language code.

TheCartpenter:
Since this issue looks more like a political one, I'll close these.

PR #11910, opened 2022-11-03
Added upgrade fix for canceled term - install/controller/upgrade/upgrade_8.php file

TheCartpenter:


TheCartpenter:
Closing. Not the replacement I was expecting. Will create a new pull.

PR #11922, opened 2022-11-05
Added subscription plan price validation for cron - admin/controller/catalog/subscription_plan.php file

TheCartpenter:


TheCartpenter:
An extension could validate this from the charge() method in the cron by adding an extra log output. I don't know why this suggestion was proposed by others in the past. Closing this.

PR #11945, opened 2022-11-09
Suggesting this change for TWIG file - catalog/view/template/mail/subscription.twig file

TheCartpenter:
A bit of work might still be needed. For now, could be as a suggestion in case something else might be needed.

### DIFF
--- a/upload/catalog/view/template/mail/subscription.twig
+++ b/upload/catalog/view/template/mail/subscription.twig
@@ -55,6 +55,18 @@
   <table style="border-collapse: collapse; width: 100%; border-top: 1px solid #DDDDDD; border-left: 1px solid #DDDDDD; margin-bottom: 20px;">
     <thead>
       <tr>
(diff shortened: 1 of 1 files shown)


nexadomain:
Doctype - many email clients do not support HTML5.
Therefore, the earlier doctype was better.

TheCartpenter:
@danielkerr: This situation seem to match equally between the results of doctypes of HTML 4.01 vs. other TWIG files that don't have them where 5 TWIG files do, 5 others don't as followed:

- admin/view/template/common/header.twig
- admin/view/template/mail/voucher.twig
- admin/view/template/sale/order_invoice.twig
- admin/view/template/sale/order_shipping.twig
- catalog/view/template/common/header.twig
- catalog/view/template/mail/order_add.twig
- catalog/view/template/mail/order_invoice.twig
- catalog/view/template/mail/voucher.twig
- catalog/view/template/mail/subscription.twig
- install/view/template/common/header.twig

PR #11960, opened 2022-11-12
Seperate help disription for trial language file

pemaier:
because the trial duration start with 1 and the normal duration start with 0


TheCartpenter:
It could somewhat help a bit with the ```duration``` since there's a validation involved but you are also repeating the ```cycle``` and ```frequency``` text.

PR #11967, opened 2022-11-13
Fixed conflict of error_default language key - catalog/controller/account/address.php file

TheCartpenter:
save() and delete() methods leads to the same language key ```error_default``` but one is meant to check if it's saved but the other is meant to check if it can be deleted as the default address.


TheCartpenter:
Changed my mind. Closing this.

PR #11977, opened 2022-11-15
Added seo_url table lookup for previous versions - install/controller/upgrade/upgrade_9.php file

TheCartpenter:
Since new database fields are involved, prior versions than OC 4 will not succeed to manage their SEOs in their admin panel. Therefore, since the database fields have been recreated by design between these two versions, we need to handle these fields and new values. No else statement is necessary to replace: '|' to: '.' since the server can already take care of it whether these criteria are found or not.

### DIFF
--- a/upload/install/controller/upgrade/upgrade_9.php
+++ b/upload/install/controller/upgrade/upgrade_9.php
@@ -746,9 +746,9 @@ class Upgrade9 extends \Opencart\System\Engine\Controller {
 
                 foreach ($seo_urls as $seo_url) {
                     $value = preg_replace('/[^a-zA-Z0-9_|\/\.]/', '', str_replace('|', '.', $seo_url['value']));
-                    
+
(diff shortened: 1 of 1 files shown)


danielkerr:
to many things commited. can not see what the changes are supposed to do. 

but this code witll mess a lot up

                if (!filter_var($value, FILTER_VALIDATE_INT)) {
                        $value = $this->db->escape($value);
                    } else {
                        $value = (int)$value;
                    }

TheCartpenter:
Without it, people upgrading from previous OC versions than OC 4 will not be able to use the ```oc_seo_url``` table since you changed database fields by design. However, if it's just a matter of not putting int with the value, we could just leave it with ```$this->db->escape``` if need be. The reason why being on adding an int cast is because some of your array keys contains numeric values rather than string numbers as opposed to others that do contains string numbers in your arrays on the MB as well.

PR #11978, opened 2022-11-15
Added subscription status in filters - admin/controller/sale/order.php file

TheCartpenter:


TheCartpenter:
Since new orders are required to create new subscriptions, then only the most recent orders can show an active status state from the subscriptions. Otherwise, the subscription cannot be pulled for previous orders.

TheCartpenter:
Closing this. Since the integration of the promotional features in mail/subscription, the previous period must complete its cycle before concluding that the subscription for the next subscription period can be used. Therefore, the last two orders must be verified with the active subscription status rather than focusing specifically on the last created order.

PR #11994, opened 2022-11-18
Finalized mail/subscription - catalog/controller/mail/subscription.php file

TheCartpenter:
Should be good this time.


TheCartpenter:
That file was a real pain in the ass!

TheCartpenter:
Finished converting the models' conventional names in OC-3. This issue should no longer occur with the exception of config_currency and language prefixes that may need to be compared a bit before and after posting nowadays.

PR #12005, opened 2022-11-20
Removed trial_remaining field - catalog/model/account/subscription.php file

TheCartpenter:
Does not exist on the database. The remaining event must be used by transactions in order for an extension to make the next move as per this detailed explanation: https://github.com/opencart/opencart/pull/12003#issue-1456937274 .


TheCartpenter:
Closing. These fields are in the subscription table already.

PR #12014, opened 2022-11-22
Added user-defined date_added in addTransaction() - catalog/model/account/subscription.php file

TheCartpenter:


TheCartpenter:
Also removed the ```transaction_id``` field since it does not exist in the ```subscription_transaction``` table.

TheCartpenter:
The reason why being to use a user-defined date is because a charge can only be initiated on the billing cycle date whereas a transaction, based on customers' requests, may require adjustments with his membership plans where the changes may only be reflected on the next billing cycle, since a ```date_next``` field is obviously required with the subscriptions. 

Therefore, while the ```date_next``` field is validated from the cron/subscription and mail/subscription controllers, we need to ensure that the determined time period is always in the future whenever there are promotional features involved. This can also be used on current orders but I am sure extensions can already take care of those.

TheCartpenter:
Closing. No longer necessary.

PR #12020, opened 2022-11-23
Added LOWER SQL function on email field - admin/model/customer/customer.php file

TheCartpenter:


mhcwebdesign (commented):
Why this change?

LCASE and LOWER are synonymous, if you check the official docs it says as much. The reason they're both there is for familiarity, SQL Server and DB2 use lcase, Oracle and PostgreSQL use lower.

There are more important things to be done the OpenCart 4 master branch.

TheCartpenter:
These are leftover results compared to the rest on the MB that uses LOWER functions. Statistically, there are currently more results showing LOWER than LCASE. To avoid confusion, while you are correct to state they are both alias, the mount of results of LCASE is only minimal of use compared to LOWER. As for other issues to solve, I would still have to agree on that.

However, dealing with the basics while we're at it is simply another step forward.

danielkerr:
use LCASE

TheCartpenter:
Ok, will rename all to LCASE.

TheCartpenter:
Can be closed now.

PR #12027, opened 2022-11-23
Added store_id and country fields in customer_ip - install/controller/upgrade/upgrade_8.php file

TheCartpenter:
Customers won't be able to login to their account from the catalog-end without adding these fields when upgrading from previous OC versions,


danielkerr:
uses_customer never used varchar

PR #12028, opened 2022-11-24
Added addTransaction handled by type of value 0 - catalog/controller/mail/subscription.php file

TheCartpenter:
After several reviews of this file, the best conclusion is to set the ```type``` parameter as 0 so when the sale/subscription/addTransaction() will be on the verge of a subscription cancelation state, the ```type``` field will be required on anyhow.


TheCartpenter:
Closing due to conflict.

PR #12041, opened 2022-11-26
Added mail_subscription_cancel event - install/opencart.sql file

TheCartpenter:


eka7a:
There is no cancel function in mail/subscription file. Transaction function exists but it is trigger by admin/controller/sale/subscription/addTransaction/after

Issue #12043, opened 2022-11-26
[4.0.1.1] Captcha doesn't work on Returns page


MagDevel:
**What version of OpenCart are you reporting this for?**
4.0.1.1 and the current master

**Describe the bug**
Captcha doesn't work on the Product Returns page

**To Reproduce**
1. Go to 'Extensions -> Extensions -> Captchas -> Basic captcha -> click 'Edit'.
2. Set Status to enabled, click 'Save'
3. Go to 'System -> Settings -> Default store -> click 'Edit'
4. Go to 'Option' tab, scroll down to 'Captcha' section
5. Select Basic Captcha in Captcha field
6. Pick all options in Captcha Pages field, click 'Save'
7. Go Returns page in front-end

See that Captcha does not appear, and captcha verification is not performed

**Expected behavior**
Captcha should appear on the Returns page, captcha validation should work

**Additional context**
The problem is different values used for config_captcha_page setting:

'returns' value is used in admin/controller/setting/setting.php
'return' value is used in catalog/controller/account/returns.php

TheCartpenter:
I replied to your pull request: https://github.com/opencart/opencart/pull/12045

danielkerr:
@TheCartpenter stop fucking around with stuff!

TheCartpenter:
@danielkerr: WTF are you talking about dude?? This fix was meant to be!

danielkerr:
@TheCartpenter yes sorry. i just checked

MagDevel:
@danielkerr Thank you. 
Please fix the same on line 408

PR #12045, opened 2022-11-26
Renamed return to returns in catalog/controller/account/returns.php

MagDevel:
Issue #12043

### DIFF
--- a/upload/admin/controller/setting/setting.php
+++ b/upload/admin/controller/setting/setting.php
@@ -389,7 +389,7 @@ class Setting extends \Opencart\System\Engine\Controller {
 
 		$data['captcha_pages'][] = [
 			'text'  => $this->language->get('text_return'),
-			'value' => 'returns'
+			'value' => 'return'
 		];
 
 		$data['captcha_pages'][] = [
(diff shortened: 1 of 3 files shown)


TheCartpenter:
If we revert this, it will create issues with the term: ```return```. The reason this has been renamed was to resolved issues that were encountered in the past about this reserved terminology.

MagDevel:
Ok. In that case, we have to rename "return" to "returns" in catalog/controller/account/returns.php as well.

TheCartpenter:
> Ok. In that case, we have to rename "return" to "returns" in catalog/controller/account/returns.php as well.

Yes, correct, as this one's already done: https://github.com/opencart/opencart-3/blob/main/upload/catalog/controller/account/returns.php .

danielkerr:
returns is used as return is a reserved word!

danielkerr:
stop changing things u are wasting my time and will cause more issues.

TheCartpenter:
This case is in regard to rename return into returns that was missing: https://github.com/opencart/opencart/pull/12045/commits/e67170778e9fea1cc23302d13d68d553aa5e3fa6 .

PR #12068, opened 2022-12-02
Update catalog.php

PhamToan:
need to load Language before loading SEO

### DIFF
--- a/upload/system/config/catalog.php
+++ b/upload/system/config/catalog.php
@@ -18,9 +18,9 @@ $_['session_engine']     = 'db'; // db or file
 // Actions
 $_['action_pre_action']  = [
 	'startup/setting',
+	'startup/language',
 	'startup/seo_url',
 	'startup/session',
-	'startup/language',
 	'startup/customer',
 	'startup/currency',
 	'startup/tax',


chulcha:
SEOURL  makes it possible to get language_id

TheCartpenter:
The startup/language controller was recently moved down on the MB. I am not sure why it would be suggested to move it back up.

PhamToan:
> The startup/language controller was recently moved down on the MB. I am not sure why it would be suggested to move it back up.

SEO URL builds with language_id, we need to load language first, I have submitted another merger in the past. but now it is back

TheCartpenter:
I don't see the language ID in the SEO URL startup file: https://github.com/opencart/opencart/blob/master/upload/catalog/controller/startup/seo_url.php .

TheCartpenter:
@danielkerr: Found it. It is stored in the design/seo_url model. It queries the store_id and the language_id .

danielkerr:
SEO URL does not require language Id. It works fine in master.

PhamToan:
> SEO URL does not require language Id. It works fine in master.

try with multi languages

PR #12071, opened 2022-12-03
Added PHP 8.2 support

mhcwebdesign:
Update of several classes so as to solve errors about deprecated creation of dynamic properties.

### DIFF
--- a/upload/admin/controller/tool/store.php
+++ b/upload/admin/controller/tool/store.php
@@ -42,11 +42,12 @@ class Store extends \Opencart\System\Engine\Controller {
 		$registry->set('load', $loader);
 
 		// Create a dummy request class so we can feed the data to the order editor
-		$request = new \stdClass();
+		$request = new \Opencart\System\Library\Request();
 		$request->get = [];
 		$request->post = [];
 		$request->server = $this->request->server;
 		$request->cookie = [];
+		$request->files = [];
 
 		// Request
 		$registry->set('request', $request);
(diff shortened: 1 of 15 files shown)


TheCartpenter:
By using namespaces to declare instanced variables is not good practice. Besides, if one states that modifying each constructor is better than modifying the proxy class (which I still don't agree on), then adding namespaces on instanced variables wouldn't provide better experience especially for extension developers.

mhcwebdesign:
> i will do the fixes myself

No problem. But at least take a look at it, we need support for PHP 8.2 as soon as possible.

danielkerr:
> By using namespaces to declare instanced variables is not good practice. Besides, if one states that modifying each constructor is better than modifying the proxy class (which I still don't agree on), then adding namespaces on instanced variables wouldn't provide better experience especially for extension developers.

agreed

mhcwebdesign:
I think using **object** as a type is very generic, though it works, too.

Anyway, please add all the needed fixes for PHP 8.2 support, you still missed out on a few.

It still complains about these:

Opencart\System\Library\Mail::$parameter
Opencart\System\Library\Mail::$smtp_hostname
Opencart\System\Library\Mail::$smtp_username
Opencart\System\Library\Mail::$smtp_password
Opencart\System\Library\Mail::$smtp_port
Opencart\System\Library\Mail::$smtp_timeout

with **Creation of dynamic property deprecated** errors, caused by the lines 135 to 140 at

catalog/controller/information/contact.php

Also, the call to the **str_split** function needs to be fixed in system/library/mail/smtp.php

For a complete list, see my pull request.

danielkerr:
U block the ability to interchange the objects when u force namespaces

PR #12083, opened 2022-12-06
Updated returns pagination code like in downlod controller

condor2:


danielkerr:
who ever puty it there was wrong to do it.

TheCartpenter:
Yes, I thought so. I also ended up dealing with a mix of product_limit and the value of 10 in account/download and account/returns in the opencart-3 repository for an unknown reason which was also found previously.

PR #12090, opened 2022-12-10
Renamed order_id to filter_order_id - admin/controller/sale/order.php file

TheCartpenter:
Also corrected some commented text and array alignment


TheCartpenter:
Will create another.

PR #12092, opened 2022-12-11
add mobile items  on moudle banner

egproo:
if user device is mobile will show item for mobile


upload/extension/opencart/admin/controller/module/banner.php
```diff
@@ -72,6 +72,12 @@ public function index(): void {
 		} else {
 			$data['items'] = 4;
 		}
+		
+		if (isset($module_info['mobileitems'])) {
+			$data['mobileitems'] = $module_info['mobileitems'];
+		} else {
+			$data['mobileitems'] = 1;
```
gerasart, line 79: can u add 1 to constant like

private const  DEFAULT_MOBILE_ITEMS = 1

mhcwebdesign:
OpenCart 4 already uses a responsive web theme. I am not sure what this pull request is all about. How exactly do you define a mobile device? laptop? tablet? what screen size? touch screen? screen size?

It will probably be better if you were to create an OpenCart extension for whatever you are trying to accomplish here.

egproo:
> OpenCart 4 already uses a responsive web theme. I am not sure what this pull request is all about. How exactly do you define a mobile device? laptop? tablet? what screen size? touch screen? screen size?
> 
> It will probably be better if you were to create an OpenCart extension for whatever you are trying to accomplish here.

Try adding a banner or a slideshow divided into four images, and see how it looks on your mobile phone, for example
and must same banner moudle whith same images sizes on all devices

Check the quality and tell me why we do not seek to work to include the differentiation between different screens (we address screen sizes and type of device (with a specific model) and we cannot address all devices with the same image and it appears in a good appearance

try to check site on https://pagespeed.web.dev/
to get advice about images

mhcwebdesign:
Tried https://pagespeed.web.dev/ for my sites, not much of a difference between desktop and mobile (not even sure how it's defined there, you could have a small laptop with touchscreen, which is sometimes seen as mobile, other times as desktop).

It's all down to to implementing good responsive web themes. Unfortunately, the majority of 3rd party OpenCart web themes are of poorest quality, the worst one being Journal3, but others are equally to be avoided, especially from themeforest.net

upload/extension/opencart/admin/controller/module/banner.php
egproo: I did not understand the need for that, as it is a starting point for the number 1 to appear in the settings, and it is kept as the minimum with one banner, or it is modified according to desire and preservation

egproo:
Most of this talk is outside the scope of development for default features

upload/extension/opencart/admin/controller/module/banner.php
gerasart: if opencart 4 start use php 81 we need try do more readable code

PR #12099, opened 2022-12-11
Made some text correction - admin/language/en-gb/marketplace/marketplace.php file

TheCartpenter:


TheCartpenter:
Will create another.

PR #12115, opened 2022-12-19
Removed print_r - system/library/mail/mail.php file

TheCartpenter:


TheCartpenter:
As for this post, the print_r function was added on your behalf and from you only which I least I had the decency to address the issue about it. If you aren't done with it, it is still up to you to mention it in your commit description which is why Github allows the use of two boxes rather than a single. If you want to stick to a single box, that's totally up to you but, while you can say anything else on that, the reality is that change was indeed valid.

TheCartpenter:
Nor was the topic requesting / suggesting to. It's ok, though; the missing descriptions or reported issues will still be on-going even though you're gone later on if people haven't decided to move on by then regardless of what one might believe in their response.

PR #12149, opened 2023-01-20
[master] HTTP_CATALOG not defined error

nmpetkov:
HTTP_CATALOG not defined error

- Plane new master install on real server
- Go to Settings, select the store, go to Mail tab, select mail function
- Go to frontend of the site and try to create order
- Go to error log and see the error

One side comment: Why after plane install no mail method is selected by default?


nmpetkov:
I see it is now corrected in the code.

PR #12150, opened 2023-01-20
[master] json_decode $order_info['custom_field'] twice

nmpetkov:
Resolve an error.

To reproduce:
- Plane new installation with current files from server master branch
- Create new order
- Go to admin panel and try to view the order

Error obtained:
TypeError:  json_decode(): Argument #1 ($json) must be of type string, array given
.../admin/controller/sale/order.php
Line: 616

Reason:
It is already decoded in model getOrder


nmpetkov:
I see now it is changed in:
https://github.com/opencart/opencart/commit/79549983a203f6ef8a930f0c88c291cc539fac8d

PR #12151, opened 2023-01-20
[master] Fix sending mail is not working

nmpetkov:
This PR es a fix sending mail is not working in current state of master branch.

To reproduce the problem:
- clean install master master branch
- try to create order or add to history of an order

Mail is not sent, and in error log you can see the errors generated.

Other way to resolve the problem is to create mail or smtp clasess just before sending. Here proposed way is similar to Opencart 3: transfer options property before sending. To do this we should make these properties public.


danielkerr:
i think already fixed

PR #12157, opened 2023-01-23
Small corrections

opencartbot:
Using namespace \Opencart\System\Helper as Helper мust be removed from all extensions


danielkerr:
because of out of everything u posted stuf was already fixed or changed there was only 2 i had not noticed.

eugengarkusha:
@danielkerr i'm on opencart 4.2 and language switching func-ty in admin-pannel is still broken

PR #12175, opened 2023-02-09
Update order.php

PhamToan:
make sure it sort DESC
the bug when 2 rows of 1 order has same date  (use custom code for auto change order status)


danielkerr:
no it should be date added.

PR #12190, opened 2023-02-26
white display fix

BuslikDrev:
If the output_buffering ini directive is enabled, then headers_sent() does not work, as a result, compression works and makes a white screen if something is output before compression (error, var_dump).

### DIFF
--- a/upload/system/library/response.php
+++ b/upload/system/library/response.php
@@ -107,6 +107,10 @@ class Response {
 			return $data;
 		}
 
+		if (ob_get_contents()) {
+			return $data;
+		}
+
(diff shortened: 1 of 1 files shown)


danielkerr:
thats does not solve a compression issue its just turning it off

BuslikDrev:
> thats does not solve a compression issue its just turning it off

This is the point - turn off compression so that there is no white screen when the developer wants to debug the code. Otherwise, he needs to disable compression through the admin panel.

danielkerr:
its just dumping out put out if its finding cached output.

Issue #12217, opened 2023-03-21
Wrong validation on custom_fields when creating new address - payment_address.php


RicardoCostaTW:
Version: 4.0.1.1 

**Describe the bug**
Wrong validation on custom_fields when creating new address.

**To Reproduce**
1 - create custome field with 'address' location
2 - make required
3 - save
4- start a new order
5- create a new payment address and save
 
 The result is `{"error":{"payment_custom_field_31":"Phone required!","payment_custom_field_32":"Vat required!"}}`

**Screenshots / Screen recordings**
![image](https://user-images.githubusercontent.com/1208532/226687289-cfc177c8-460d-457f-a4b8-2991899c1baa.png)

The code is:
`$custom_fields = $this->model_account_custom_field->getCustomFields($this->customer->getGroupId());

			foreach ($custom_fields as $custom_field) {
				if ($custom_field['location'] == 'address') {
					if ($custom_field['required'] && empty($this->request->post['custom_field'][$custom_field['custom_field_id']])) {
						$json['error']['payment_custom_field_' . $custom_field['custom_field_id']] = sprintf($this->language->get('error_custom_field'), $custom_field['name']);
					} elseif (($custom_field['type'] == 'text') && !empty($custom_field['validation']) && !preg_match(html_entity_decode($custom_field['validation'], ENT_QUOTES, 'UTF-8'), $this->request->post['custom_field'][$custom_field['custom_field_id']])) {
						$json['error']['payment_custom_field_' . $custom_field['custom_field_id']] = sprintf($this->language->get('error_regex'), $custom_field['name']);
					}
				}
			}`

danielkerr:
not sure who put that fucking code there. it was working at some point.

PR #12218, opened 2023-03-22
Validation fixed for the custom fields in register.php and shipping_a…

shnki:
fixing the bug where checking if the post request doesn't contain a custom_field always returns true in register.php and shipping_address.php https://github.com/opencart/opencart/issues/12217

in the register.php

`if ($custom_field['required'] && empty($this->request->post['shipping_custom_field'][$custom_field['location']][$custom_field['custom_field_id']])) {
							var_dump($this->request->post);
							$json['error']['shipping_custom_field_' . $custom_field['custom_field_id']] = sprintf($this->language->get('error_custom_field'), $custom_field['name']);
						}`
the if statement will always be true because it depends on a  `post['shipping_custom_field']` which is not sent by the client 

image
here the field is not empty yet the server returns that it is required 

image

using the var_dumb() method to see what the post request looks like, I found out that there's no such as ['shipping_custom_field'] in the body of the post request which explains why the server thinks the field is always empty so replaced the request->post['shipping_custom_field'] with request->post['custom_field'] 

In shipping_address.php 
the same problem in register.php, the server sent a message saying the custom field is required even though it has input. 

The problem occurs when trying to add a new shipping address while having a custom_field in the form.
This happens because the validation if statement
  `if ($custom_field['location'] == 'address') {
					if ($custom_field['required'] && empty($this->request->post['custom_field'][$custom_field['custom_field_id']])) {
						$json['error']['shipping_custom_field_' . $custom_field['custom_field_id']] = sprintf($this->language->get('error_custom_field'), $custom_field['name']);
					}`
checks if the post['custom_field'][$custom_field['custom_field_id']] is empty which will always be empty because the post request look like this 
image

replacing 
`					if ($custom_field['required'] && empty($this->request->post['custom_field'][$custom_field['custom_field_id']])) {
`
with 
`					if ($custom_field['required'] && empty($this->request->post['custom_field'][$custom_field['location']][$custom_field['custom_field_id']])) {
`
fixed the issue

### DIFF
--- a/upload/catalog/controller/checkout/register.php
+++ b/upload/catalog/controller/checkout/register.php
@@ -344,8 +344,8 @@ class Register extends \Opencart\System\Engine\Controller {
 				// Custom field validation
 				foreach ($custom_fields as $custom_field) {
 					if ($custom_field['location'] == 'address') {
-						if ($custom_field['required'] && empty($this->request->post['shipping_custom_field'][$custom_field['location']][$custom_field['custom_field_id']])) {
-							$json['error']['shipping_custom_field_' . $custom_field['custom_field_id']] = sprintf($this->language->get('error_custom_field'), $custom_field['name']);
+						if ($custom_field['required'] && empty($this->request->post['custom_field'][$custom_field['location']][$custom_field['custom_field_id']])) {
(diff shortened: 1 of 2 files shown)


danielkerr:
this would break more things!

PR #12223, opened 2023-03-27
need leading space to prevent appending to link

daretogo:
Without a leading space before "the" the word "The" ends up getting appended to the reset link resulting in users clicking on it and getting an "invalid reset link" message.

### DIFF
--- a/upload/catalog/language/en-gb/mail/forgotten.php
+++ b/upload/catalog/language/en-gb/mail/forgotten.php
@@ -3,4 +3,4 @@
 $_['text_subject']  = '%s - Password reset request';
 $_['text_greeting'] = 'A new password was requested for %s customer account.';
 $_['text_change']   = 'To reset your password click on the link below:';
-$_['text_ip']       = 'The IP used to make this request was:';
+$_['text_ip']       = ' The IP used to make this request was:';


ADDCreative:
I don't believe this isn't a fix for anything. There is no need for a space there There is clearly whitespace in the template already.

https://github.com/opencart/opencart/blob/fc4a0861209bc27d0449eab26339df7706a88f93/upload/catalog/view/theme/default/template/mail/forgotten.twig#L5-L9

I suspect @daretogo is using a version of PHP that is not compatible with the version of Twig they are using. https://forum.opencart.com/viewtopic.php?t=210652#p803802

mhcwebdesign:
@ADDCreative: Some older versions of twig, as used in e.g. OpenCart 3.0.2.0, couldn't handle whitespaces correctly, as documented at https://twig.symfony.com/doc/3.x/templates.html#whitespace-control

However, this shouldn't be an issue with the 3.0.x.x_Maintenance branch because that one uses the latest version 3.6.1 of Symfony Twig. Personally I use the alternative OpenTwig engine (www.opentwig.org) on my client's websites.

ADDCreative:
@mhcwebdesign That was the exact point I was making. No need for this change in the 3.0.x.x_Maintenance branch.

PR #12224, opened 2023-03-27
Redirect to cart page after reorder

padaliyajay:
we have a case, when we wanted to use "reorder" button outside the order info page, in that case, problem is, customer was redirecting to order info page after reorder. so to make reorder work from anywhere, we should redirect customer to cart page.


eka7a:
The customer may want to reorder more products. It will need to go to the order page again.

PR #12237, opened 2023-04-11
Update db_schema for subscription_plan

pemaier:
When you are doing a new install there is an error of the missing price field.
			[
				'name' => 'price',
				'type' => 'decimal(15,4)',
				'not_null' => true,
				'default' => '0.0000'
			],


pemaier:
not relevant anymore

PR #12265, opened 2023-04-18
Display Compatibility info for Extension in Marketplace

condor2:
https://github.com/opencart/opencart/pull/12264


condor2:
This way users know extension compatibility before purchase if they don't access through opencart website

PR #12288, opened 2023-04-23
Create apcu

BuslikDrev:
https://www.php.net/manual/en/ref.apcu.php


danielkerr:
we already have apc

danielkerr:
Yes so don't use it.

ADDCreative:
@danielkerr Which to use then, even the default file cache won't support: https://github.com/opencart/opencart/blob/4ccac08d739fa88fdbf1636786b6d2dc5e859504/upload/admin/controller/tool/backup.php#L307

https://github.com/opencart/opencart/blob/4db99331e95e82374c27d2060bdf91de6a229341/upload/system/library/cache/file.php#L57-L60

PR #12292, opened 2023-04-26
Fixed sub category URL update match all value start with old path

qaharmdz:
Issue: Update sub-category URL using unescape wildcard underscore `WHERE `value` LIKE 'x_%'`.
Example: change to category_id `2` also changes all match path values `20`, `234_567`, `2xxx`, etc.


danielkerr:
@eka7a can u prove theres an issue here or are u just accepting what people say

PR #12296, opened 2023-04-27
bugfix backup / restore

mhcwebdesign:
For admin/model/tool/backup.php, method backup: Some fields can contain NULL value, which isn't compatible with string nor array!


ADDCreative:
Beat me to it.

Master has same issue.
https://github.com/opencart/opencart/blob/764ed15e7fe503ce90dfe8287e659fba8cfff1cf/upload/admin/controller/tool/backup.php#L164-L174

PR #12299, opened 2023-04-28
Fix backup with NULL values

ADDCreative:
https://github.com/opencart/opencart/pull/12296 for master.


danielkerr:
seem to be fixing things\ that have no proof there is an issue

ADDCreative:
@danielkerr See https://forum.opencart.com/viewtopic.php?t=231227#p856492. While OpenCart itself may not use null values in the database. Many extensions do, including the official OpenCart Google Shopping extension https://www.opencart.com/index.php?route=marketplace/extension/info&extension_id=35669.

PR #12300, opened 2023-04-29
Suggestion: New `download` method in Response library.

trydalcoholic:
Hi all!
I want to suggest adding a new method in [Response library](https://github.com/opencart/opencart/blob/master/upload/system/library/response.php).

### [Download](https://github.com/trydalcoholic/opencart/blob/refactor/response/upload/system/library/response.php#L148)
The new `download` method will help reduce code duplication and allow you to declaratively start downloading a file.

**Before:**
```php
if (!headers_sent()) {
	header('Content-Type: application/octet-stream');
	header('Content-Description: File Transfer');
	header('Content-Disposition: attachment; filename="' . $filename . '"');
	header('Content-Transfer-Encoding: binary');
	header('Expires: 0');
	header('Cache-Control: must-revalidate, post-check=0, pre-check=0');
	header('Pragma: public');
	header('Content-Length: ' . filesize($file));

	readfile($file, 'rb');
	exit;
} else {
	exit($this->language->get('error_headers_sent'));
}
```

**After:**
```php
$this->response->download($file, $filename);
```


TheCartpenter:
While the JSON headers and output might be ideal for the use of the APIs across different platforms, it might be less ideal for the downloads since you are also removing the ```$this->language->get``` error message whereas customers are being notified of a failed download. That would, then, become a lack of service from the store without the cause being reported from the browser.

A similarity to the system/helper language parameters with the use of key and value from languages, it wouldn't be the right location to fit such ideas into core libraries.

TheCartpenter:
Furthermore, header_sent lookup from the download and upload controllers are still unique. Therefore, there are also no need to move a unique code into a unique library.

trydalcoholic:
@TheCartpenter thank you for your comment. You are right, I agree with you.
It still seems to me that downloading files can and should be added to the response method. I'll think about it some more.
I moved the proposal for the `json` method [to another pr](https://github.com/opencart/opencart/pull/12301).

PR #12301, opened 2023-04-29
Suggestion: New `json` method in Response library.

trydalcoholic:
Hi all!
I want to suggest adding a new method to the [Response library](https://github.com/opencart/opencart/blob/master/upload/system/library/response.php).

### [Json](https://github.com/trydalcoholic/opencart/blob/refactor/response-json/upload/system/library/response.php#L148)
The new `json` method will help reduce code duplication and allow you to return json declaratively.

**Before:**
```php
$this->response->addHeader('Content-Type: application/json');
$this->response->setOutput(json_encode($json));
```

**After:**
```php
$this->response->json($json);
```

You can still set additional flags to generate json. For example:
```php
$this->response->json($json, JSON_NUMERIC_CHECK | JSON_UNESCAPED_UNICODE);
```

### Motivation

- Controller prepares data, Response library prepares response
- Controller should not prepare response headers
- If we expect to return json reponse, then we can extend the Response library with a declarative method
- Small reduction in duplicated code


TheCartpenter:
You're twisting the words while a new commit was not necessary. What's been said on this commit: https://github.com/opencart/opencart/pull/12300#issuecomment-1528791339 is that JSON switch cases could be useful for the APIs, nothing has been mentioned about the core.

trydalcoholic:
@TheCartpenter I apologize if I misunderstood you.
Opencart returns json response not only in api. According to pr statistics, about 200 files were changed. Correct me if I'm wrong, but most of the files are not apis, but controllers with json responses.

TheCartpenter:
Core files do not need to be modified for this. It can already be done by extension by using this example at your own risk: https://github.com/opencart/opencart/discussions/10251#discussioncomment-5715215 .

TheCartpenter:
> ```
> <?php
> namespace Opencart\System\Library\Extension\Mazaengine;
> 
> class Wrapper {
> 	protected $object;
> 
> 	public function __construct(object $object) {
> 		$this->object = $object;
> 	}
> 
> 	public function __get($key) {
> 		return $this->object->{$key};
> 	}
> 
> 	public function __set($key, $value) {
> 		$this->object->{$key} = $value;
> 	}
> 
> 	public function __call($method, $args) {
> 		return $this->object->{$method}(...$args);
> 	}
> 
> 	public function restore() {
> 		return $this->object;
> 	}
> }
> ```
> 
> Your own cart file
> 
> ```
> <?php
> namespace Opencart\System\Library\Extension\Mazaengine;
> 
> class MyCart extends \Opencart\System\Library\Extension\Mazaengine\Wrapper {
> 	public function getProducts() {
> 		$products = $this->object->getProducts();
>                  // Do your changes in $products
>                  return $products;
> 	}
> }
> ```

trydalcoholic:
@TheCartpenter thanks for your example, it really looks interesting.
But my pr is not motivated by the desire to make a modification. I propose to expand the possibilities in the core of Opencart for the reasons that I added in the pr description as "motivation".

TheCartpenter:
JSON headers' already working in the core.

trydalcoholic:
@TheCartpenter what do you mean? As I see, the json response headers set the methods in the controllers, although it seems to me that this is the responsibility of the reponse library.

julioccguimaraes:
This is the type of mod that breaks compatibility with several modules already developed for OpenCart 4

PR #12323, opened 2023-05-13
Update sub_total.twig

fubaman:
The Save button event can not be found under form-form-* , it shout be form-*


fubaman:
Fix for the save button.

PR #12353, opened 2023-05-30
Fix broken language for modules in front-end

MarvinKlein1508:
Currently added language folders for all default extensions are not beeing recognized by Opencart.

For example:
grafik

As you can see, the totals for example are still in English.

This fix checks for the language code within the framework and applies it to the language so that all language files from modules will be applied. After this fix:
grafik

When a non valid language is specified it will fall back to the default of the store.


hayden-t:
savage

Issue #9236, opened 2021-02-06
MySQL 8 : Incorrect DATE value: '0000-00-00'


SilverFoxA:
**What version of OpenCart are you reporting this for?**
OC 3.0.3.7

**Describe the bug**

The homepage and all product listing pages.

**To Reproduce**
1. Download OC 3.0.3.7 from GitHub Opencart repo releases section
2. Setup on Ubuntu running mysql 8
3. Go to the website

`"PHP message: PHP Fatal error:  Uncaught Exception: Error: Incorrect DATE value: '0000-00-00' ....`

**Expected behavior**
List the products

**Screenshots / Screen recordings**
![Screenshot_3](https://user-images.githubusercontent.com/6983070/107120957-c21d4700-68b5-11eb-9126-fb42ec1251b1.png)

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? - Development server
 - Operating system - Ubuntu
 - PHP version - (php-fpm 7.4)
 - Apache version - (Nginx 1.18)
 - Browser(s) tested with [e.g. chrome, safari - if applicable] - Chrome

TheCartpenter:
See this solution: https://forum.opencart.com/viewtopic.php?f=206&t=222399#p813284 .

danielkerr:
any specific reason im using MySQL 8 and have no issues!

TheCartpenter:
> any specific reason im using MySQL 8 and have no issues!

Simple. All followed aspects shows that:

1 - You are using a database server library version that is not yet mandatory by the Industry
2 - You are enforcing / expecting users to use mySQL v8+ whereas other platforms do not yet insist on doing so from a simple incremented sub-version rather than a complete OC version from v3.x to v4.x releases.

3 - Server packaging may not provide such capabilities to, yet, upgrade to that version to the users as opposed to what your server may already provided specifically.

4 - There are no documented processes where users should understand that this MySQL server library shall be mandatory by Opencart as opposed to other platforms since its distribution date on opencart.com website.

Result: Multiple topics on the forum do report failed upgrades due to MySQLi DB library in Opencart. By reverting to the previous library DB version of its distribution, users are then able to upgrade as shown from this post: https://forum.opencart.com/viewtopic.php?f=206&t=222399#p813328 as the solution delivered from this post as well: https://forum.opencart.com/viewtopic.php?f=206&t=222399#p813349 .

WTG!

mhcwebdesign:
[...]

PR #12365, opened 2023-06-07
Update pdo.php

whmemon:
This is fix issue mentioned in
https://github.com/opencart/opencart/issues/9236
as fixed in mysqli.php

### DIFF
--- a/upload/system/library/db/pdo.php
+++ b/upload/system/library/db/pdo.php
@@ -26,8 +26,8 @@ class PDO {
 		}
 
 		if ($pdo) {
-			$this->connection = $pdo;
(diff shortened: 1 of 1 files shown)


whmemon:
I am using OC Version 4.0.2.1 with Laragon 6 having
```
PHP Version 8.1.10
mysqlnd 8.1.10
Apache/2.4.54
```
It was creating error
` Incorrect DATE value: '0000-00-00'`
After searching found following issue
https://github.com/opencart/opencart/issues/9236
After going through source code of OC found code was update in 
`system/library/db/mysqli.php`
but not in
`system/library/db/pdo.php`

so i updated it. I think i am not only facing this issue.

Thanks

TheCartpenter:
Already fixed in the MB.

Issue #12263, opened 2023-04-18
[4.0.2.0] Language bug


condor2:
**What version of OpenCart are you reporting this for?**
4.0.2.0

**Describe the bug**
Language bugs are not fixed....

On catalog side Opencart 4.0.2.0 should read extensions translation from **extension/your_language/extension** and not from **extension/opencart**.

Admin switch language bug is still there...
https://github.com/opencart/opencart/issues/12081#issuecomment-1512899466

PR #12378, opened 2023-06-21
Fix admin language

Hkr32:
https://github.com/opencart/opencart/issues/12263

I'm not 100% sure, but what I was able to test, it helped me and now the translation of the admin works.


chulcha:
also

https://github.com/opencart/opencart/blob/30b9c0d3963aa3600a0d7630c02d67a8cebc80ed/upload/admin/controller/startup/language.php#L42

Use language->load so it's not triggering infinite loops
```
		if (oc_substr($route, 0, 10) != 'extension/' && self::$extension) {
			$this->language->load('extension/' . self::$extension . '/' . $route, $prefix, $language);
		} else {
			$this->language->load($route, $prefix, $language);
		}
```

PR #12400, opened 2023-06-30
oc_payment_example module mktime() missing day argument, causes month to be missed

hayden-t:
Arguments not supplied to mktime are filled in by current system values:
"Any optional arguments omitted or null will be set to the current value according to the local date and time. "

If this is run when the date is eg 30th, it will skip February as there is no such date.

https://www.php.net/manual/en/function.mktime.php

https://stackoverflow.com/questions/66851868/php-mktime-shows-2-marchs-when-i-dump-foreach-month


ADDCreative:
Looks like quite a few other places where mktime() is used without the month. Surprised this has never been spotted before.

Sales

hayden-t:
doesnt surprise me considering the other bugs ive found and resistance Daniel has to collaboration...

PR #12419, opened 2023-07-03
Optimized Zone Delete Query

Developermayank:
Optimized the Zone Delete query


TheCartpenter:
Use the catalog/model/account/download's implode with AND instead of OR statement. IN statement may not guarantee that other extensions might also want to delete other zone entries one at a time which is why the deleteZone() method is used in a foreach statement since the beginning even though it uses more resources.

Developermayank:
Closing for improving

PR #12421, opened 2023-07-04
optimized zone delete query

Developermayank:
Optimized zone delete query
Now the delete zone may accept either int or array as $zone_id. 
On deletion of multiple zones at a time will result in database efficiency and will also be compatible for other extensions.


Developermayank:
Optimized zone delete method.

TheCartpenter:
Pointless.

- This suggestion sanitizes with PHP, not with SQL
- You're still using IN rather than OR which will only interfere with API requests.

PR #12446, opened 2023-07-09
Completed orders are not counted on the dashboard world map

opencartbot:
If the billing address is disabled, orders are not counted on the dashboard world map

### DIFF
--- a/upload/extension/opencart/admin/model/dashboard/map.php
+++ b/upload/extension/opencart/admin/model/dashboard/map.php
@@ -11,7 +11,11 @@ class Map extends \Opencart\System\Engine\Model {
 		}
 
 		if ($implode) {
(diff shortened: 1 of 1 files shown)


TheCartpenter:
For AND clause, you could use a ```$data['filter_x']``` with an ```$sql``` or ```$implode``` by using the same query instead of splitting the same query in an IF statement.

TheCartpenter:
I still don't see the benefit of it. It's been requested before by people who wants to prioritize shipping over payment since you're using an ELSE statement. Also, you're mixing $sql with $implode instead of using $implode altogether since, in this case, $implode is already used.  I would still use a filter instead of using: ```$this->config->get``` to compare each IFs since it's the OC way almost all over the platform in the models.

PR #12449, opened 2023-07-09
Extension Installer. The last installed extension should be the first…

opencartbot:
Extension Installer. The last installed extension should be the first in the list.

If we have multiple pages of installed extensions, then each new installed extension should be searched on the pagination. We need to set default sort by date_added, and not by the name of the extension.


danielkerr:
pointless

PR #12456, opened 2023-07-10
Added missing line  subscription cron language

condor2:


condor2:
Also these lines are wrong

$_['error_shipping_address'] = 'Warning: Payment method extension could not be found!';
$_['error_payment_address'] = 'Warning: Payment method extension could not be found!';

PR #12467, opened 2023-07-10
Renamed hasSubscription to hasSubscriptions in cart/cart library

TheCartpenter:


danielkerr:
why?

TheCartpenter:
Simple. You're using a ```count``` function which means it may be one or more subscriptions that users should be expecting. If you're going to use a singular term with subscription, then;

1 - Why use a getSubscriptions() method if you know that multiple subscriptions should be expected?
2 - Why use a count function if you only expect to have one subscription in return while you know you should be expecting multiple possible subscriptions

Instead, based on what you used, it's misleading as you chose to close these pull requests.

PR #12475, opened 2023-07-11
Removed duplicated banner_image_description table in upgrade/upgrade_…

TheCartpenter:
…8.php


TheCartpenter:
Reported before, closing.

Issue #9477, opened 2021-03-23
Add order without incorrect mail settings cause ERROR


faca5:
OpenCart v3.0.3.2

Problem:
When OpenCart try to send message with wrong mail settings it show POP UP error.
Error stop processing request example: adding new order

Example:
Mail server is offline for few minutes and customer have that kind of problem.

Solution:
Please add try {} except to avoid any blocking of there any problem with mail server (offline, wrong settings etc...).

TheCartpenter:
Server-specifics.

ADDCreative:
If you are getting a pop up then you display error settings are wrong.

I believe this has been reported before. https://github.com/opencart/opencart/issues/7633

WebkulOpencart:
Please report on the OpenCart forum.

faca5:
How to reproduce:
1. System > Settings > Edit > tab Mail
Mail protocol: SMTP
SMTP host: 127.0.0.2 (testing purpose)
2. Try to send any message from store (history, contact, order etc.)
3. You will get POP UP with error which block 

You will get:
![1](https://user-images.githubusercontent.com/11090866/112264355-0fd8ed80-8c71-11eb-872f-693c10504687.png)



Settings are CORRECT!!!


You can get these error(s):
- server settings example: Error: RCPT TO not accepted from server!
- mail server offline
- wrong mail settings
- etc.

You need bypass this error without breaking store.

faca5:
Problem "Error: RCPT TO not accepted from server!" was related with PHP 8.

But still similar error show when REMOTE MAIL SERVER is offline/unavailable and block customer to add new order.
It this ... just skip error if DEBUG mode OFF.

ADDCreative:
@danielkerr @WebkulOpencart This can be an issue, No mail server is ever going to be 100% reliable. As the exceptions thrown by the SMTP mail class will stop execution of the rest of the code and the trigger for the mail event is before (catalog/model/checkout/order/addOrderHistory/before). It means the order status doesn't get updated, so the order goes as missing.

@faca5 Are you able to use the Mail as the Mail Engine? That one does not throw exceptions.

faca5:
Yes! But we use SMTP to avoid message marked as SPAM. Last year many messages "order_add" has been marked as SPAM on @gmail.com. Now we are using SMTP to avoid this problem. Not all user has been marked as spam only few.
[...]

PR #12486, opened 2023-07-11
Update cod.php - Enable to send mail

padaliyajay:
As opencart developer, i often receive complaint about mail is not sending from checkout after order confirm. and most cases because they use cod as a payment method to test checkout.  so this is my suggestion to enable mail for cod

### DIFF
--- a/upload/extension/opencart/catalog/controller/payment/cod.php
+++ b/upload/extension/opencart/catalog/controller/payment/cod.php
@@ -25,7 +25,7 @@ class Cod extends \Opencart\System\Engine\Controller {
 		if (!$json) {
 			$this->load->model('checkout/order');
 
-			$this->model_checkout_order->addHistory($this->session->data['order_id'], $this->config->get('payment_cod_order_status_id'));
+			$this->model_checkout_order->addHistory($this->session->data['order_id'], $this->config->get('payment_cod_order_status_id'), '', true);
 
 			$json['redirect'] = $this->url->link('checkout/success', 'language=' . $this->config->get('config_language'), true);
 		}


danielkerr:
U enable it in the settings

padaliyajay:
I don't think there is any setting for that. if it is send let me know where

padaliyajay:
Notify is false by default
image

TheCartpenter:
@padaliyajay: The r$notify boolean variable needs to be used from an extension. Please take a look in the extension/payment folder and see the ```addOrderHistory``` how it's used. You can also create events with your extensions which allows you to use this method. It is not an issue from the core, in this case.

padaliyajay:
Yes this is not a issue. but cod by default does not sending mail, so i just want to if you want to make part of core. just check my patch, so you better understand it. but it is upto you if you want to make it part of core or not

TheCartpenter:
Adding a notification from the core for COD transactions may not be necessary, since the customer would not receive a conformed copy of his invoice via email by a service provider. The transaction is done locally.

ADDCreative:
@padaliyajay An order confirmation email is sent for all payment methods. The notify is only for order status updates. If you look in catalog/controller/mail/order.php, the notify flag is ignored when adding a new order.

danielkerr:
already sends an email out as soon as the order history status hits processing status which could be pending. activating notify just sends 2 emails.

padaliyajay:
Thanks for clarification, i got it now

padaliyajay:
One more request, i would recommend to add a mail sending in try catch statement, because if mail server down then sending mail will throw error and order will not confirmed even after payment received. this is a case already happen with my client where SMTP has limit of 100 mail per hour and throw error after that, so fail to send mail should not prevent order being confirm

```
try {
  $mail->send();
} catch (\Exception $e) {
  $this->log->write($e->getMessage());
}
```

ADDCreative:
@padaliyajay The issue with the order failing due to an error with sending the email have also been reported in https://github.com/opencart/opencart/issues/7633, https://github.com/opencart/opencart/issues/9477 and https://github.com/opencart/opencart/pull/10395.

PR #12500, opened 2023-07-16
Fix SQL space in product model

stalker780:


TheCartpenter:
Found another one here: https://github.com/opencart/opencart/pull/12501 .

PR #12520, opened 2023-07-23
Configurable admin image thumbnail size

stalker780:
Allow users to configure admin image thumbnail size.

Sometimes this may even save disk space avoiding 100x100 images which are almost never used in the frontend.

image


danielkerr:
there should be a default width and height already

danielkerr:
i prefer to use image_default_width, image_default_height

stalker780:
> i prefer to use image_default_width, image_default_height

@danielkerr Sorry, I have missed your answer. Renamed variables now.

danielkerr:
its not required

PR #12617, opened 2023-08-11
Update catalog download.php - Added . to filename check

pemaier:
Added a point to the filename check so that it is possible to upload files with an extension (e.g. name.txt)


danielkerr:
u dont need a fike extension for a masked file!

PR #12651, opened 2023-08-28
Update seo_url.php

hayden-t:
If keyword is not set:
(which is described in the admin as possible:  "Leave blank if you just want to remove the perimeter." and works) Except it still adds an  extra slash which is unnecessary
(also it should be spelling "parameter")

**update , it seems this comment is removed from master language, even though it still should work.


danielkerr:
does noot have the text "Leave blank" anymore

hayden-t:
but its not a mandatory field

Issue #12655, opened 2023-08-29
[master] Frontend order - Option text and text area limited text saved in db


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect? 
- Order options text and text area loosing data.
During a customer order of products with option text and/or text area there is saved in the database only the 1st 20 characters. So the order information is not complete. (If putted more info direct into the database it is correct displayed in the back end.)

**To Reproduce**
- Make a customer order and order a product with text and/or text area.
- Fill i the option field a text with more than 20 chars 
- Then finish the order
- and see the order in the back end with option information limited to 20 chars.
[order-option-short-text.webm](https://github.com/opencart/opencart/assets/10166095/f5585297-e228-4ad5-a3c6-6a08868d665c)

![image](https://github.com/opencart/opencart/assets/10166095/32a1bcf3-9b66-49fd-8285-b29a8a9b5c0f)

**Expected behavior**
- I expect the full information saved into the database.

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? linux
 - Operating system php
 - PHP version 8.2
 - Apache version 2.4
 - Browser(s) tested with [e.g. chrome, safari - if applicable] FF
 - Version [e.g. 22] 116.0

**Additional context**
- Back end is working well - problem is only at front end

I think here is the problem:
https://github.com/opencart/opencart/blob/0fb89a349ced64861d0d808c4c71f93c448667af/upload/catalog/model/checkout/cart.php#L49C6-L49C103

BuslikDrev:
@danielkerr Why not limit the length in the template? Why extra arrays, especially in the basket for each page?

stalker780:
@BuslikDrev I agree. When using another template you may need to use option names longer than 20 chars. 
But it is not possible without VQMod or modifying core files.

PR #12658, opened 2023-08-29
Update checkout/cart.php - Save $value longer than 20 char in db

pemaier:
Solution for #12655


danielkerr:
already removed it myself

Issue #12483, opened 2023-07-11
Error handlers needs updating for PHP 8


ADDCreative:
The error handlers need updating for PHP 8 to handle errors suppressed with @ correctly. error_reporting no longer returns 0.

https://php.watch/versions/8.0/fatal-error-suppression#suppress-error-handlers
https://www.php.net/manual/en/function.set-error-handler.php#refsect1-function.set-error-handler-examples

https://github.com/opencart/opencart/blob/04c3ddcecff2029682b3ae4b1fb0489de68d8fff/upload/catalog/controller/startup/error.php#L13
The above and in all other error handlers would need changing to something like the following.

```php
if (!(error_reporting() & $code)) {
```

danielkerr:
fine lets see if works or i get complaints

ADDCreative:
@danielkerr Please read https://php.watch/versions/8.0/fatal-error-suppression#suppress-error-handlers.

error_reporting() will no longer return 0 in normal conditions in PHP 8. It will most likly return ```E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE``` which is 4437, so checking ```=== 0``` is pointless.

danielkerr:
then why didnt u post the solution ! wasting my fucking time

ADDCreative:
I've done a pull request https://github.com/opencart/opencart/pull/12660 to show the sort of changes needed to update the error handlers for PHP 8. Not had time to install the latest master and test.

hayden-t:
seems this was abandoned ?

PR #12660, opened 2023-08-30
Update error handlers for PHP 8

ADDCreative:
For https://github.com/opencart/opencart/issues/12483.


danielkerr:
wasting time there is no more error suppression

hayden-t:
what about here ? https://github.com/opencart/opencart/blob/f0d42f933a3b816e8b862e6ac8892f93d87c7843/upload/system/library/cache/file.php#L79

Issue #93, opened 2012-08-23
Order comment missing from HTML mail confirmation


rickard2:
In the non-html version of the order confirmation e-mail the order comment is present, but not in the HTML version. 

This code adds it to the text-based version: 

(catalog/model/order.php:476)

``` PHP
            if ($order_info['comment']) {
                $text .= $language->get('text_new_comment') . "\n\n";
                $text .= $order_info['comment'] . "\n\n";
            }
```

The value of `$order_info['comment']` is never added to the HTML template and there doesn't seem to be any markup for it either. 

I could maybe look into a pull request if no one else is up for it.

danielmoore123:
> Agree with ADDCreative too. strange logic
1) user input named "comment"
2) mysql field named "comment"
3) echo $template->data['comment'] is not user comment!

So my question is, why does this solution as a whole not work? When a comment is left, I can see it in the database and in the OpenCart administration itself, but it refuses to output a user comment in the order confirmation email!! I tried to echo $ip in the email template and it works great, despite the fact it is in the same table as comment!!! Can I rename the comment section if it conflicts? Is there something more to make this work? I have Googled this and searched it all over Stackoverflow, GitHub, OpenCart Fourms, the answer is nowhere... This thread is closed but, why? It doesn't work!!

ADDCreative:
@danielmoore123 https://github.com/opencart/opencart/pull/94 This is the original pull request I submitted to add the comments to the HTML confirmation email. It is for an old version 1.5.4, by may be helpful to you.

PR #12671, opened 2023-09-04
Add customer comments to the Order Confirmation email

BryanT4212:
OC 4.0.2.2 master at e391de5fc8 (Just before blogs were added)

<s>Make sure to include the customer's comments/instructions when confirming the order.  

If no comment was specified when adding the order, check if an order comment exists and use it when mailing the order confirmation.</s>

See https://github.com/opencart/opencart/pull/12671#issuecomment-1716931811

### DIFF
--- a/upload/catalog/controller/mail/order.php
+++ b/upload/catalog/controller/mail/order.php
@@ -54,15 +54,20 @@ class Order extends \Opencart\System\Engine\Controller {
 	}
 
 	/**
+	 * Event handler to send the Order Confirmation email to the customer.
+	 *
 	 * @param array  $order_info
 	 * @param int    $order_status_id
-	 * @param string $comment
+	 * @param string $payment_instructions Optional text message provided by the
+	 *                                     payment handler detailing any specific
+	 *                                     info regarding the payment of the
+	 *                                     invoice
 	 * @param bool   $notify
 	 *
 	 * @return void
 	 * @throws \Exception
 	 */
-	public function add(array $order_info, int $order_status_id, string $comment, bool $notify): void {
+	public function add(array $order_info, int $order_status_id, string $payment_instructions, bool $notify): void {
 		// Check for any downloadable products
 		$download_status = false;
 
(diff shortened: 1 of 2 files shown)


ADDCreative:
@BryanT4212 The customer's comment would need to be separate. It doesn't make sense to only include the customer's comment if there is no comment from the payment extension.

What happens if you use the cheque payment extension which sets the comment? You would never see the customer's comment in the email,

BryanT4212:
@ADDCreative  OK, then I guess you must never see the customer's comment now?

So... add $data['instructions'] = $comment logic so the template has access to both instructions and comments?  Maybe even change the name of the argument to $instructions instead of $comment?

ADDCreative:
@BryanT4212 No, currently there are no customer's comments in the email. This is by design apparently. See https://github.com/opencart/opencart/issues/93.

BryanT4212:
@ADDCreative By design or by omission?

The issues and pull request seem to have been closed without any justification given.  In fact, they all seem to have been closed prior to any discussion taking place.  So where does that leave us?

ADDCreative:
@BryanT4212 See the comment from the developer here https://github.com/opencart/opencart/issues/93#issuecomment-8200140.

You could try another pull request that correctly adds the customer's comments separately, but don't be surprised if it's rejected.

BryanT4212:
@ADDCreative Well... it's difficult to see https://github.com/opencart/opencart/issues/93#issuecomment-8200140 as much of a resolution of the issue or justification for closure as it is just a question and occurred 8 days after the issue & PR were closed in any case.  I think I will try repurposing this request and try again by updating your PR #94 for v4.

BryanT4212:
@ADDCreative What do you think?  Ready for review this time around?

BuslikDrev:
You are doing nonsense by creating unnecessary load by adding arrays.

BryanT4212:
OC 4.0.2.2

Proposed enhancement to include confirmation of the customer's order comments/instructions in the Order Confirmation email and clarify in the code that what is being passed in to the event handler are the payment details provided by the payment system and not the customer's order comments.  

Some housekeeping modifications:
- Place the payment details immediately after the total amount due for the order/invoice
- Move the formatting of newlines as HTML <BR> elements in the order comments and payment details from the controller into the view with the rest of the HTML generation
- Whitespace cleanup

danielkerr:
for adding the order it should be fine.

ADDCreative:
@danielkerr With the commit https://github.com/opencart/opencart/commit/b35daf7d47e3c3d36e6387f23b432b0df5db6ac7. Now the payment instruction don't appear in the email to the customer. If using cheque or bank transfer, the customer will need the payment instruction to pay.

Also now the customer comments appear under the heading "Instructions", where it should be comments. The order email needs to be like the route=account/order.info page, where is shows both the customer comments and the payment instructions.

BryanT4212:
@danielkerr @ADDCreative Yeah, I'm afraid this is not a functionally compatible/accurate merge of the pull request and won't work as intended.  It turned out the original temp;late did not support customer comments at all so they had to be added separate and distinct from the payment instructions generated by the payment system which requires template as well as controller changes.

danielkerr:
ok

PR #12683, opened 2023-09-13
fix notification

BuslikDrev:
Error: Call to undefined function Opencart\Admin\Controller\Tool\oc_bbcode_decode() in admin\controller\tool\notification.php on line 173
and template rename {{ message }} => {{ text }}


BuslikDrev:
I suggest https://fontawesome.com/icons/bell?f=classic&s=regular&an=shake

Issue #12692, opened 2023-09-16
[master] Textarea over the whole shop backend


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?
The textarea fields have a basic display problem.
Pls. have a view to the screenshot.

**To Reproduce**
1. Go to 'Catalog/Categories'
2. See Description

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app
![image](https://github.com/opencart/opencart/assets/10166095/562d9c56-755f-4abf-b066-52a949748f96)

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? server
 - Operating system linux
 - PHP version 8.2
 - Apache version 2.4
 - Browser(s) tested with [e.g. chrome, safari - if applicable]
 - Version [e.g. 22] 115.2.0esr

mhcwebdesign:
There is nothing wrong with the built-in CKEditor. If you think there is a bug then provide more details.

pemaier:
In the writing area of the textarea is a seperate frame with a P visible which not belongs there.

danielkerr:
I'll check when I get home. Seems a silly bug that got left in there. Possibly missed a tag

stalker780:
@pemaier it's not a bug. It is CKEditor feature.
You can see what type of block is used. p,h2,h3,h4,h5,div
It is convenient when you use not only a <p> tag in your html content.

It can be disabled in admin\view\javascript\ckeditor\config.js setting 
config.startupOutlineBlocks = false;
 
![image](https://github.com/opencart/opencart/assets/6297070/acc8f7a3-a897-44bb-828e-6cbe34cca359)

![image](https://github.com/opencart/opencart/assets/6297070/0c6436cf-2fd0-4db4-a9a5-5bdd686e4a2a)

danielkerr:
seems issue aftert i accepted ckeditor commit

danielkerr:
its not a bug its deigned that way

pemaier:
I can live with that - it's a new function - and it's good:) ... and it may also help all this not so experienced users to understand what they do with copy and paste product tests from websites with this many now visible tags. )

danielkerr:
i will looik into removing the showing html tags

[...]

PR #12693, opened 2023-09-16
CKEditor ShowBlocks alternate config

stalker780:
Here is an alternate ShowBlocks config.
It disables ShowBlocks by default and adds ShowBlocks toolbar switch. 
https://github.com/opencart/opencart/issues/12692

In my opinion Blocks should be always visible. It does nothing to the source code but allows you to control the HTML structure better. It's just a matter of habit.

image


pemaier:
Seems to be a good solution
- Traditional users see no difference at 1st view 
- Experienced users get an additional help
- No extra settings in the config

PR #12712, opened 2023-09-23
[Master] Added config_image_topic in setting controller

condor2:


danielkerr:
pretty sure i already fixed

PR #12719, opened 2023-09-23
Update sitemap.twig

opencartbot:


danielkerr:
i think fixed already

PR #12725, opened 2023-09-25
Removed text/javascript and <!-- wrapping

SecretSergey:


danielkerr:
why when its required to stop the js being indexed

PR #12726, opened 2023-09-25
Extracted a wishlist to an external controller

SecretSergey:
Needed for ajax updates.


danielkerr:
pointless. not setting up its own controller just for a link.

PR #12727, opened 2023-09-25
Fix ajax update product/wishlist

SecretSergey:
Buttons in the form update only 1 block, the fix adds the ability to specify an update block to each button.
Fix btns:
- cart btn on account/wishlist;
- wishlist btn on product/product;
- wishlist btn on product/thumb.


SecretSergey:
#12726 required for this pr

PR #12728, opened 2023-09-26
Update setting.php - Added config_image_topic_height/width

pemaier:
Added settings for topic image that they are present in a new installation


danielkerr:
already done

Issue #12718, opened 2023-09-23
Problem with subcategory name, description and breadcrumb


batumibiz:
**OC version: master branch on 24.09.2023**

Problem with subcategory name, description and breadcrumb

**To Reproduce**
1. Add some category with any description
2. Add another category with any description and assign the above created category as its parent.
3. On frontend, go to above created category->subcategory.

Look at subcategory name, description and breadcrumb.  
**The information is incorrect** and is displayed **from the parent category instead of the current subcategory**.

The problem is in the file [/catalog/controller/product/category.php](https://github.com/opencart/opencart/blob/master/upload/catalog/controller/product/category.php)
On 05.09.2023 there was some strange commit that disrupted the normal operation of this file.  
If you roll back the changes made by this commit, the problem disappears and the subcategories are displayed normally.

maximus15102006:
Thank you, it helped a lot!

PR #12729, opened 2023-09-26
Fix: #12718

shashankrp:
Fix  issue : #12718

The application was not able to render the sub category descriptions, the issue was arising because of an error in the code where `upload/catalog/controller/product/category.php`  where in the code `$parts = explode('_', $path);` was actually splitting the path but by using `array_pop($parts)` method we were only able to fetch the first part of the path and the second part of the path was being removed.

In the current commit i just added another variable `$parts1` which just does the same operation as `$parts` and i am passing it to the for loop for validation and now it seemed to be working, i have tested it as for my knowledge. Except `category.php` file i haven't altered any other files. 

This is my first contribution till day, i hope i did it right. 
Thanks.

### DIFF
--- a/upload/catalog/controller/product/category.php
+++ b/upload/catalog/controller/product/category.php
@@ -49,6 +49,7 @@ class Category extends \Opencart\System\Engine\Controller {
 		}
 
 		$parts = explode('_', $path);
+		$parts1 = explode('_', $path);
 
 		$category_id = (int)array_pop($parts);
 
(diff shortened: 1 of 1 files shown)


batumibiz:
The solution is not complete...
The subcategory name and description are displayed correctly, but **Breadcrumb shows the subcategory name twice**.
If this is corrected, the problem will be solved.

batumibiz:
#12730

danielkerr:
### already fixed

maximus15102006:
Thank you, it helped!

PR #12739, opened 2023-09-28
Update blog_list.twig

chongshengdz:
fix search url don't load correctly


danielkerr:
already soorted

chongshengdz:
> already soorted

why close it?
blog search don't work at all.

danielkerr:
Blog system is not even finished

PR #12740, opened 2023-09-28
Update blog.php

chongshengdz:
fix the language doesn't load to the search url


danielkerr:
we put language at the end of the file

PR #12749, opened 2023-10-02
Table tax_rate added Time added/modified and predefine article Author with user name 

pemaier:
The missing fields trowing an install error.


danielkerr:
so why add them back! they are unneded

Issue #12757, opened 2023-10-05
MySQL SSL Connection  Support


fmarslan:
Hi, I use MySQL with SSL, but the opencart db connection configuration does not support to SSL connection. is there a way to SSL connection without changing the mysqli.php file?

danielkerr:
pretty sure u just have too change port number

PR #12758, opened 2023-10-05
#12757 mysql ssl connection support

fmarslan:
#12757 feature

An Advanced panel was added to step-3 UI while installing. The advanced panel is visible while using the MySQLi driver.

image


danielkerr:
dumb hoow u did the installer. ill fix it

Issue #12765, opened 2023-10-07
[master] cli_install not longer work as described in usage


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?
A install via cli_install.php is not working anymore.

**To Reproduce**
1. upload OpenCart files
2. start cli_install
// Usage:
//
//   php cli_install.php install --username    admin
//                               --email       email@example.com
//                               --password    password
//                               --http_server http://localhost/opencart/
//                               --db_driver   mysqli
//                               --db_hostname localhost
//                               --db_username root
//                               --db_password pass
//                               --db_database opencart
//				 --db_port     3306
//                               --db_prefix   oc_

3. a list of errors appear

**Expected behavior**

I expect a working install process as described.

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? server
 - Operating system linux
 - PHP version 8.2
 - Apache version 2.4
 - Browser(s) tested with [e.g. chrome, safari - if applicable] not relevant

**Additional context**

Problem started after including pull request https://github.com/opencart/opencart/pull/12758

fmarslan:
Hi @pemaier,

the problem fixed  with #12767

danielkerr:
i think people shbould just fucking wait to see me finsih the db changes instead of all jumpingin

PR #12767, opened 2023-10-07
Fix MySQL SSL issue (#12765)

fmarslan:
This commit addresses the following issues:
- Fixed a bug affecting CLI installation
- Fixed a bug affecting Cloud installation
- Fixed a bug affecting Upgrades
- Fixed a bug about saved cert
- Corrected a wrong variable name
- Fixed formatting problems in the code

If you encounter any issues with the DB_SSL_* keys, please consider adding the following lines to your development configuration files (config.php and admin/config.php):

```php
define('DB_SSL_KEY', '');
define('DB_SSL_CERT', '');
define('DB_SSL_CA', '');
```
These additions will help ensure proper handling of SSL keys for your database connections.

### DIFF
--- a/upload/install/cli_cloud.php
+++ b/upload/install/cli_cloud.php
@@ -181,9 +181,9 @@ class CliCloud extends \Opencart\System\Engine\Controller {
 		$db_hostname = getenv('DB_HOSTNAME', true);
 		$db_username = getenv('DB_USERNAME', true);
 		$db_password = getenv('DB_PASSWORD', true);
-		$db_ssl_key  = getenv('DB_SSL_KEY', true);
-		$db_ssl_cert = getenv('DB_SSL_CERT', true);
-		$db_ssl_ca   = getenv('DB_SSL_CA', true);
+		$db_ssl_key  = getenv('DB_SSL_KEY', true) ?: '';
+		$db_ssl_cert = getenv('DB_SSL_CERT', true) ?: '';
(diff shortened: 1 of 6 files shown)


fmarslan:
Additionally, I am currently writing the certificate files to the temporary folder. However, this solution is not ideal as each request creates a new file in the temporary directory. Opencart has caching capabilities, so I plan to use caching in the mysqli.php file. I need to learn more about the caching structure within the project. After addressing this issue, I will create a pull request.

ADDCreative:
I would of thought you would be best just to store the path to the files rather than the actual certificate files. It will make thing a lot easier if you have to change them in the future. It will also save having to write them to a temporary files,

fmarslan:
Yes, the file path is correct. However, when using it within the container, we should either mount external storage or create a new container by extending the original one. The latter option is more complex than just changing environment variables.

I have two concerns regarding the file path method when mounting storage:

There is a potential security issue if a third party gains access to the external storage. However, this issue is not critical for us; it's a concern for the system administrator.

There's also an infrastructure problem. For instance, I'm using the Azure container app, which only supports the SMB protocol for external storage. Unfortunately, the SMB protocol's performance is suboptimal, especially during high-traffic periods.

Here's my idea: We can write the certificates to the temp folder during the first request. The certificate paths are then saved to the cache. Subsequent requests can check the path and retrieve the file from the cache. If the path is correct, it will be used; otherwise, a new one will be created.

Maybe I'm overthinking it, maybe your advice is the best. :) Is there a third person to help with the selection?

Issue #12766, opened 2023-10-07
[4.0.2.3] Error after deleting a language


Hkr32:
**What version of OpenCart are you reporting this for?**
4.0.2.3

**Describe the bug**
Order statuses disappeared after removing an additional language
What section does it affect?
setting/setting

**To Reproduce**
1. Install an additional language
2. Set the language status to enabled
3. In the settings, select the newly installed language
4. Remove the language in extensions
5. Order statuses are missing in the settings

**Expected behavior**
The statuses must remain or when deleting a language, it cannot be deleted if it is active and selected by default.

**Screenshots / Screen recordings**
![order_statuses](https://github.com/opencart/opencart/assets/12296392/48ad9a45-1a06-42d0-9da2-e337a8061abb)

**Server / Test environment (please complete the following information):**
![image](https://github.com/opencart/opencart/assets/12296392/846797fc-fb24-498d-9509-7a61d6e43b47)

I understand that you shouldn’t do this, but I think that there may be problems if someone suddenly does this.

danielkerr:
"[4.0.2.3] Error after deleting a language"

and yet theissue is completely different from what u are describing

danielkerr:
i gave no idea i think it only happens to u!

danielkerr:
maybe change the language back to default

Hkr32:
> maybe change the language back to default

It doesn't allow you to change it because all the fields are not filled in.



> i gave no idea i think it only happens to u!

Great approach to development.

stalker780:
@Hkr32 give a link to this language extension.
Are you sure it's not an extension issue?

Hkr32:
> @Hkr32 give a link to this language extension if it is free.

https://www.opencart.com/index.php?route=marketplace/extension/info&extension_id=45362

> Are you sure it's not an extension issue?

I haven't tested other languages, but I just understand the reason for all this. At least it seems so to me)

stalker780:
@Hkr32 

This happens because your language code is stored in your browsers cookies.
After deleting the language extension your browser still have language cookie set `ru-ru`

![image](https://github.com/opencart/opencart/assets/6297070/aa750cd1-b3e1-42ea-a7f5-a8486580a3cb)

OC tries to select rows with absent language when selecting order_status list.

First you need to set any other language. Then delete your default language extension.

Or you need to clear browser cookies to reset your language code.

Extension developer can add this functionality to extension `uninstall` function to avoid such behavior.
So it is not an OC bug really.

[...]

PR #12768, opened 2023-10-08
Check active language

stalker780:
Prohibit uninstalling active language extension

https://github.com/opencart/opencart/issues/12766


stalker780:
Also added a startup fix. It forces to use default language if language from their cookies was deleted in DB.

It is required if there are several admin users. For example one admin changes a language in his browser and deletes certain language. Other admins who used this language in their browsers will still get 0 language_id issue because of their cookies.

Same for frontend. If admin deletes a language, customers will continue using language code from their cookies. This may break lots of scripts.

This issue is not obvious but may break a store functionality when manipulating with languages.

PR #12769, opened 2023-10-08
Check active language

stalker780:
1. Prohibit uninstalling active language extension
2. Admin startup fix. It forces to use default language if language set in cookies was deleted in DB.

Startup fix is required if there are several admin users. For example one admin changes a language in his browser and deletes certain language. Other admins who used this language in their browsers will still get 0 language_id issue because of their cookies.
And there is no way to fix this except cleaning browser cookies.

This issue is not obvious but may break admin functionality when manipulating with languages.

https://github.com/opencart/opencart/issues/12766


danielkerr:
should be done from the extension being uninstalled

stalker780:
@danielkerr It is not possible to check cookies from extension on page load.

The sense is when there are several admins, one of them can delete a language, another admin will face this language issue, because there is no way to change cookies, only cleaning them in browser manually.

danielkerr:
sure

danielkerr:
if users still using when removing a language it should just default back to the original.

stalker780:
You can easily reproduce this behavior.

1. Set any random not existing language code in browser cookies. Imitate deleted language by one of admins.
2. Open System->Localisation->Order statuses
3. See an empty list

image

stalker780:
I have removed all changes except admin/controller/startup/language.php

AJenbo:
This need a rebase or merge with upstream master

danielkerr:
i have a better solution

stalker780:
Works for language_id in SQL queries, language is en-gb by default. 
But now you cannot change admin language if you have deleted language set in cookies.
There is no language dropdown in header.

image

PR #12775, opened 2023-10-14
Update step_2.twig

TheCartpenter:
Changed 8.0+ to 7.4+

### DIFF
--- a/upload/install/view/template/install/step_2.twig
+++ b/upload/install/view/template/install/step_2.twig
@@ -29,7 +29,7 @@
 								<tr>
 									<td>{{ text_version }}</td>
 									<td>{{ php_version }}</td>
-									<td>8.0+</td>
+									<td>7.4+</td>
 									<td class="text-center">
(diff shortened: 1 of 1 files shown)


TheCartpenter:
Since so many defined constants have been simplified in the install folder for OC 4, why not simply the PHP versions as well instead of keeping them separated into each controllers and TWIG files? !

It requires more changes to be done manually to set PHP versions separately whether it's located in separated controllers or TWIG files.

mhcwebdesign:
> Changed 8.0+ to 7.4+

Why, when OpenCart 4 requires PHP 8.0 or later?

TheCartpenter:
> > Changed 8.0+ to 7.4+
> 
> Why, when OpenCart 4 requires PHP 8.0 or later?

That's because you're a little late in the news: https://github.com/opencart/opencart/commit/893e03c15519bba2eaa43f8505facb39e9747a74

PR #12791, opened 2023-10-16
Fix comment validation: min length 25 chars

opencartbot:


danielkerr:
what about Chinese characters?

PR #12804, opened 2023-10-21
Removed var from html declaration

TheCartpenter:


TheCartpenter:
Already covered. Closing.

PR #12807, opened 2023-10-22
Sticky admin filters

stalker780:
Make right filter sticky when scrolling a page.
Very useful if you have set more then 10 items per page in admin area.
Also lists look more aesthetically pleasing now without empty space in the right column.

image


stalker780:
@eka7a I thought about BS classes. But this does not solve top and bottom positioning margin and z-index. Anyway you need to edit CSS file and a separate class for filter container.
BTW this PR also contains a fix for .sticky-top. Because `#container {overflow:hidden}` was breaking any sticky block inside it.

PR #12809, opened 2023-10-22
Catalog - escape aliases

stalker780:


stalker780:
@danielkerr Daniel, could you please check this PR and commit it if it is ok.

Cause it already contains a lot of fixes provided by TheCartpenter. And this produces branch conflicts.

PR #12810, opened 2023-10-23
currency rates refresh bugfix

mhcwebdesign:
Now using the ECB


mhcwebdesign:
The old Yahoo Finance API hasn't been working for quite a while. This bugfix uses the European Central Bank Currency Converter, see admin backend at _**Extensions > Extensions > Currency Rates > European Central Bank Currency Converter**_.

PR #12811, opened 2023-10-23
Catalog Footer - Show Blog entry only if there are topics

pemaier:
Update of footer.php - integration of query if there are topics
Update of cms/topics - added function getTotalTopics
Update of footer.twig - integration of if there are topics


pemaier:
@danielKerr Please remove the 2 commits  52318d5 and 51583f7 they do not belong to the changes. Sorry for that.
52318d5 - not longer required
51583f7 - you already have included this

PR #12813, opened 2023-10-24
Update startup.php

sabeti05:


danielkerr:
better to use __isset

PR #12814, opened 2023-10-26
Update user.php

rahulabs:
Duplicate Code


rahulabs:
Also user_history does not exist.

PR #12819, opened 2023-10-27
More code alignments from the admin languages

TheCartpenter:


TheCartpenter:
Also includes the pagination limits.

PR #12847, opened 2023-10-29
Update bestseller.php

TheCartpenter:


batumibiz:
Why was type casting removed?
The method returns array

TheCartpenter:
> Contributor

As per other models, we're not fetching the ```$results``` into a ```foreach``` . Therefore, the array does not need to be cast.

batumibiz:
> As per other models, we're not fetching the `$results` into a `foreach` . Therefore, the array does not need to be cast.

What do other models have to do with it?
The current method returns an array.
1. We are trying to get data (which is an array) from the cache.
2. If there is no cached data, we make a request to the database and get the result in the form of an array.
3. As a result, we must return an array.
4. Since the cache->get() method returns mixed, it is very useful that we specify the type of data we receive.

As a result, type casting is useful here and there is no need to remove it.

TheCartpenter:
So are these models:

- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/country.php#L134
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/return_reason.php#L94
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/return_action.php#L94
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/length_class.php#L96
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/return_status.php#L94
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/geo_zone.php#L114
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/order_status.php#L94
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/weight_class.php#L96
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/stock_status.php#L94
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/zone.php#L125
- https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/subscription_status.php#L94
- https://github.com/opencart/opencart/blob/master/upload/admin/model/sale/voucher_theme.php#L95
- https://github.com/opencart/opencart/blob/master/upload/admin/model/setting/store.php#L97
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/catalog/manufacturer.php#L57
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/catalog/product.php#L196
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/catalog/product.php#L365
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/catalog/product.php#L535
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/checkout/voucher_theme.php#L46
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/localisation/order_status.php#L26
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/localisation/zone.php#L28
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/localisation/return_reason.php#L35
- https://github.com/opencart/opencart/blob/master/upload/catalog/model/localisation/currency.php#L49

Each returns an array, they're not casted.

batumibiz:
> So are these models:
> 
> * https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/country.php#L134
> 
> Each returns an array, they're not casted.

Well, add casting there if you want to mess around.
Here the casting was done more correctly.

TheCartpenter:
> > So are these models:
> > 
> > * https://github.com/opencart/opencart/blob/master/upload/admin/model/localisation/country.php#L134
> > 
> > Each returns an array, they're not casted.
> 
> Well, add casting there if you want to mess around. Here the casting was done more correctly.

By having a mixed parameter in the library, and based on the mount of results where ```$this->cache->get``` have more results without casting than those that are casted. I would rather wait and find out which part of the code's on the right track rather than knowing that only a few of them have been casted already.

batumibiz:
> By having a mixed parameter in the library, and based on the mount of results where `$this->cache->get` have more results without casting than those that are casted. I would rather wait and find out which part of the code's on the right track rather than knowing that only a few of them have been casted already.

The method has a clearly specified return type - an array. And only an array cannot be anything else, it does not even have the right to be null, or false. [bestseller.php#L14](https://github.com/opencart/opencart/blob/master/upload/extension/opencart/catalog/model/module/bestseller.php#L14)

TheCartpenter:
> > By having a mixed parameter in the library, and based on the mount of results where `$this->cache->get` have more results without casting than those that are casted. I would rather wait and find out which part of the code's on the right track rather than knowing that only a few of them have been casted already.
> 
> The method has a clearly specified return type - an array. And only an array cannot be anything else, it does not even have the right to be null, or false. [bestseller.php#L14](https://github.com/opencart/opencart/blob/master/upload/extension/opencart/catalog/model/module/bestseller.php#L14)

Exactly. As you described on the above, since the output is only expected to be returned as an array, then there's no need to cast it.

PR #12864, opened 2023-11-02
Update product_purchased.php

TheCartpenter:


mhcwebdesign:

TheCartpenter:
> Why? There's nothing wrong with this DB query!

PR #12867, opened 2023-11-03
Update customer_transaction.php

TheCartpenter:


mhcwebdesign:

batumibiz:
> Why? There's nothing wrong with this DB query!

By the way, this is a rare case when I will support this committee. I'll explain in more detail...

As I can see from other commits, work has begun on adding quotes to SQL query table names.
You yourself know that this is critical only for MySQL, everything will work without quotes, the main thing is that the table names do not conflict with reserved names. To avoid risks, use quotation marks.

It's useful to have some kind of standard for source code. And since such work began in previous commits, it makes sense to bring it to the end and enclose ALL table names in queries in quotes.

TheCartpenter:

kanenas:
Are we talking about single quotes **'** or backticks **`** ?

batumibiz:
> [#12843 (comment)](https://github.com/opencart/opencart/pull/12843#issuecomment-1784198338)



> [#12843 (comment)](https://github.com/opencart/opencart/pull/12843#issuecomment-1784198338)

Do not abuse the trust of the maintainer.
Yes, 'some' of your commits can be useful.
But most of your commits, especially when you create a separate PR just to add one space, create a big moral nuisance for other programmers who could potentially help the project.

That's why there are many complaints.

Yes, the project is managed by Daniel and if he is satisfied with this state of affairs, other programmers will not be able to do anything and complaints to GitHub will not help.

But constantly ignoring the complaints of other programmers can lead to the fact that only Daniel and you will remain among the contributors, and maybe sometimes random assistants will appear...

batumibiz:
> Backticks. I think it will likely break compatibility with other database engines, e.g. PostgreSQL or PDO

By the way, I also had a question about this for a long time...
Why during installation is it offered to choose a database driver between mysqli vs mpdo?
What is the difference?
What to choose?
Wouldn't it be more correct to leave only PDO?
Today, PDO has been well developed and it provides the necessary level of abstraction and support for various databases.

ADDCreative:
Even if non-standard backticks were needed, this pull request only quotes the c alias, misses the other ct and cgd aliases. Maybe they are they going to be done is separate pull requests.

mhcwebdesign:
@ADDCreative : I have now added an issue (see https://github.com/opencart/opencart/issues/12869) because we need to decide whether the other DB engines are ever to be supported. All the recently added backticks have made this highly unlikely, so we might as well drop support for e.g. PostgrSQL, or rewrite the majority of the model classes.

TheCartpenter:
> > [#12843 (comment)](https://github.com/opencart/opencart/pull/12843#issuecomment-1784198338)
> 
> Do not abuse the trust of the maintainer. Yes, 'some' of your commits can be useful. But most of your commits, especially when you create a separate PR just to add one space, create a big moral nuisance for other programmers who could potentially help the project.
> 
> **That's why there are many complaints**.
> 
> Yes, the project is managed by Daniel and if he is satisfied with this state of affairs, other programmers will not be able to do anything and complaints to GitHub will not help.
> 
> But constantly ignoring the complaints of other programmers can lead to the fact that only Daniel and you will remain among the contributors, and maybe sometimes random assistants will appear...

Nobody's ignoring anything here. The only reason why you might believe being the case are due to the countless replies I had to put in the past where those replies were, then, and still, ignored by others. That's the choices they've made and the route they have chosen to be into - themselves. If I wouldn't believe on the Opencart evolution, I simply wouldn't dedicate my time to it.

danielkerr:
backticks are required as some the field names use restricted workds.

PR #12870, opened 2023-11-04
Show in catalog footer entry Blog only if CMS topics available

pemaier:


danielkerr:
bad coding! 

if results

should be getTotalArticles()

pemaier:
You are right.
I'll insert getTotalTopics in modul cms/topics because it is possible that a store has topics but no articles.

PR #12892, opened 2023-11-06
Update antispam.php

TheCartpenter:


mhcwebdesign:
Why? 'total' is not a reserved keyword, see https://docs.actian.com/psql/psqlv13/index.html#page/sqlref%2Fsqlkword.htm%23 or https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-T

The same is true with many of your other pull requests.

There are some urgent issues to be addressed for OpenCart.

TheCartpenter:
> Why? 'total' is not a reserved keyword, see https://docs.actian.com/psql/psqlv13/index.html#page/sqlref%2Fsqlkword.htm%23 or https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-T
> 
> The same is true with many of your other pull requests.
> 
> There are some more urgent issues to be addressed for OpenCart.

PR #12898, opened 2023-11-06
Fix syntax error in marketplace

AJenbo:
This commit broke the syntax of the install file and also isn't properly giving the name space for two classes: https://github.com/opencart/opencart/commit/50b13de22dfb2781c834f08600ceeacbe3bf8fa8) broke somethings.

A similar issue was also introduced in this commit, which has also been fixed here as it affects the marketplace controllers: https://github.com/opencart/opencart/commit/0da669409629a3aca01aefe40c49765b34eb9491

I was actually just looking at the code to see what level of analyzers you run on it, nooks like there is none. Are you open to introducing some to automatically catch errors like this?

**Update:** Looks like the syntax error has now been fixed in https://github.com/opencart/opencart/commit/3591578ad725c0216c823276041e1c13642519a9, but the name space issues remains, the PR has been updated to only address that now.

### DIFF
--- a/upload/admin/controller/marketplace/installer.php
+++ b/upload/admin/controller/marketplace/installer.php
@@ -496,7 +496,7 @@ class Installer extends \Opencart\System\Engine\Controller {
 
 				if ($xml) {
 					try {
-						$dom = new DOMDocument('1.0', 'UTF-8');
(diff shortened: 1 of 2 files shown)


danielkerr:
am working on the it

PR #12899, opened 2023-11-07
Remove dead code from getTotalProductsByProductId

AJenbo:
As there is no way to feed the $data array to the function this code can never be activated. Also it does not look to have been and omission as the call site has no trace of filter variables that would fit here.

The offending code was introduced in this commit which also does not appear to have intended for it usage: https://github.com/opencart/opencart/commit/e49531f95b7a1d5b66001fb4530b938a43ee420e


danielkerr:
its used in best sellers

PR #12916, opened 2023-11-10
Fix getting product info in subscription data

AJenbo:
This code was commented out in https://github.com/opencart/opencart/commit/66fa3b1693f4b91fc1cd4c3505addc3b8322f27d which causes subsequent conditions to fail as $product_info hasn't been populated. No clear reason is given to I'm assuming this was just a left over from local debugging.

### DIFF
--- a/upload/admin/controller/sale/subscription.php
+++ b/upload/admin/controller/sale/subscription.php
@@ -658,9 +658,8 @@ class Subscription extends \Opencart\System\Engine\Controller {
 
 		// Product data
 		if (!empty($subscription_info)) {
-			//$this->load->model('sale/order');
-			//getSubscriptionByOrderProductId
-		//	$product_info = $this->model_sale_order->getProductByOrderProductId($subscription_info['order_id'], $subscription_info['order_product_id']);
+			$this->load->model('sale/order');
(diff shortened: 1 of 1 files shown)


upload/admin/controller/sale/subscription.php
```diff
@@ -658,9 +658,8 @@ public function info(): void {
 
 		// Product data
 		if (!empty($subscription_info)) {
-			//$this->load->model('sale/order');
-			//getSubscriptionByOrderProductId
-		//	$product_info = $this->model_sale_order->getProductByOrderProductId($subscription_info['order_id'], $subscription_info['order_product_id']);
+			$this->load->model('sale/order');
+			$product_info = $this->model_sale_order->getProductByOrderProductId($subscription_info['order_id'], $subscription_info['order_product_id']);
```
tktamas, line 662: I would like to ask for your help, as this function is not implemented in the model. 
Error message:
Notice: Undefined property: Proxy::getProductByOrderProductId in /.../admin/controller/sale/subscription.php on line 671: in /.../public/system/engine/proxy.php on line 85

danielkerr: Can we open an issue

tktamas: Yes, I will open an issue.
I also find multiple subscription related issue.

PR #12952, opened 2023-11-15
Fix preg_replace 

mhcwebdesign:
The regular expression '[^a-zA-z0-9]' needs to be replaced with '/[^a-zA-Z0-9]/' , that is, if not letter or digit.


danielkerr:
im going to changer it soo its a fixed name but will have number on the end if it finds more storage folders.

PR #12960, opened 2023-11-16
Fix invalid regex in admin function

AJenbo:
The string was missing the pattern delimiters meaning the regex didn't really do anything. This issue was introduced in https://github.com/opencart/opencart/commit/9b1b3d23a9dc4a5de119b42a8d06c0e0ea0f0115

### DIFF
--- a/upload/admin/controller/common/security.php
+++ b/upload/admin/controller/common/security.php
@@ -290,7 +290,7 @@ class Security extends \Opencart\System\Engine\Controller {
 		}
 
 		if (isset($this->request->get['name'])) {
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
If anything, it should be '/[^a-zA-Z0-9]/' to start with, not '/[^a-zA-z0-9]/' !

Anyway, Daniel is going to do some more changes here soon, see https://github.com/opencart/opencart/pull/12952#issuecomment-1813613289

AJenbo:
Good catch, fixed.

AJenbo:
> Anyway, Daniel is going to do some more changes here soon, see [#12952 (comment)](https://github.com/opencart/opencart/pull/12952#issuecomment-1813613289)

Oh, it would be good if issues like these where tracked and all changes went through a PR to make to more clear what work is ongoing.

PR #12970, opened 2023-11-16
Display only enabled custom_fields - Address

condor2:


danielkerr:
what about if u wanted to view the old customer field data

condor2:
And what if you have 50 disabled fields assigned to Affiliate (used before), and only one fieled enabled?
Is not annoyng to scrool down?
aff_custom_fields

Issue #12965, opened 2023-11-16
[master] Storage folder do not accept dots anymore


pemaier:
**What version of OpenCart are you reporting this for?**
master of todat
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?
After an installation on the 1st view 3 things are to do. 1. delete install folder, 2. define and move storage folder (here's the problem)
Since about a week the storage folder did not accept a dot anymore in the folder name (e.g. storage.master).
If I write storage.master as folder name I get storagemaster as result.
This behavior occurs since about a week.

**Expected behavior**
I'm awaiting that I get the name I type into the form field.

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? server
 - Operating system Linux
 - PHP version 8.2
 - Apache version
 - Browser(s) tested with [e.g. chrome, safari - if applicable] FF
 - Version [e.g. 22] 115.4 ESR

pemaier:
Not working again. Sorry.

pemaier:
Solution is in pull 12971

PR #12971, opened 2023-11-17
Update security.php - fix preg_replace calls to allow dot in name and…

pemaier:
… path

This is a common uses way for path and name in Linux environments.
Solution for #12965


mhcwebdesign:
Using a dot for a file's basename is not a good idea!

pemaier:
Why filename not and path yes?
Btw. the filename we can control but the path name not.

PR #12972, opened 2023-11-18
Update db_schema.php

TheCartpenter:


danielkerr:
number is to big

TheCartpenter:
```status``` of 1 is smaller than 4.
Those numbers were gathered from the ```subscription``` table that are already built-in.

PR #12991, opened 2023-11-21
fix: Wishlist Stock Status Bug Fixed

serkanboztepe:
Warning: Undefined array key "stock_status"

image

Wishlist Stock Status Bug Fixed

image

### DIFF
--- a/upload/catalog/controller/account/wishlist.php
+++ b/upload/catalog/controller/account/wishlist.php
@@ -95,7 +95,15 @@ class WishList extends \Opencart\System\Engine\Controller {
 				}
 
 				if ($product_info['quantity'] <= 0) {
-					$stock = $product_info['stock_status'];
+					$this->load->model('localisation/stock_status');
+
+					$stock_status_info = $this->model_localisation_stock_status->getStockStatus($product_info['stock_status_id']);
+
+					if ($stock_status_info) {
+						$stock = $stock_status_info['name'];
+					} else {
+						$stock = '';
+					}
 				} elseif ($this->config->get('config_stock_display')) {
 					$stock = $product_info['quantity'];
 				} else {


TheCartpenter:
The ```$product_info['stock_status_id']``` won't do it either since the model/product no longer has the ```stock_status``` alias nor the ```stock_status_id``` field either. Therefore, validating the stock status as a separate model in this case would differ from the query being pulled from the product.

serkanboztepe:
> $product_info['stock_status_id']

I can get the stock_status_id value with the $product_info['stock_status_id'] alias. When you send this value to the stock model, the answer comes. For example: Pre-Order.

TheCartpenter:
> > $product_info['stock_status_id']
> 
> I can get the stock_status_id value with the $product_info['stock_status_id'] alias. When you send this value to the stock model, the answer comes. For example: Pre-Order.

Questions and answers, right? Except it doesn't provide the same logic as the ```stock_status``` that used to be integrated in the model/product from previous series by doing that. OC v4 also uses ```$this->statement``` as opposed to previous series. Therefore, it would create a separate cache as opposed to the same product cache.

serkanboztepe:
> > > $product_info['stock_status_id']
> > 
> > 
> > I can get the stock_status_id value with the $product_info['stock_status_id'] alias. When you send this value to the stock model, the answer comes. For example: Pre-Order.
> 
> Questions and answers, right? Except it doesn't provide the same logic as the `stock_status` that used to be integrated in the model/product from previous series by doing that. OC v4 also uses `$this->statement` as opposed to previous series. Therefore, it would create a separate cache as opposed to the same product cache.

So what is your solution to this problem?

TheCartpenter:
> > > > $product_info['stock_status_id']
> > > 
> > > 
> > > I can get the stock_status_id value with the $product_info['stock_status_id'] alias. When you send this value to the stock model, the answer comes. For example: Pre-Order.
> > 
> > 
> > Questions and answers, right? Except it doesn't provide the same logic as the `stock_status` that used to be integrated in the model/product from previous series by doing that. OC v4 also uses `$this->statement` as opposed to previous series. Therefore, it would create a separate cache as opposed to the same product cache.
> 
> So what is your solution to this problem?

The reason why this portion of code has been initially removed is because there are no definite solution to this problem at this time other than duplicating the query that the localisation/stock_status can already pull but for different purposes to what is expected from the products to pull about.

PR #12993, opened 2023-11-22
&amp; in url

dreanei23:
without this change, it will be like this:

```
$this->url->link('extension/my/module', ['param1' => '1', 'param2' => '2']);

site.ru/index.php/?route=extension/my/module&amp;param1=1&param2=2
```

with the change so:

`site.ru/index.php/?route=extension/my/module&amp;param1=1&amp;param2=2`

it is necessary either to replace the ampersand everywhere, or nowhere

for example sitemap.xml (and in general, the xml format) does not accept ampersand. `Accepts only &amp;`

### DIFF
--- a/upload/system/library/url.php
+++ b/upload/system/library/url.php
@@ -57,7 +57,7 @@ class Url {
 
 		if ($args) {
 			if (is_array($args)) {
-				$url .= '&' . http_build_query($args);
+				$url .= '&' . http_build_query($args, '', '&amp;');
 			} else {
 				$url .= '&' . trim($args, '&');
 			}


TheCartpenter:
Except it would interfere with SEO URLs depending on the extensions that are also being used.

dreanei23:
> За исключением того, что это будет стимулировать URL-адреса SEO в зависимости от расширения.

now also used &amp;

only not everywhere replaced & by &amp;

which is why the link looks like this: ...&amp;PARAM=1&PARAM=2

chulcha:
)))))))))))
not problem
```
		if (!$js) {
			return str_replace('&', '&amp;', $url);
		} else {
			return $url;
		}
```

dreanei23:
> ))))))))))) not problem
> 
> ```
> 		if (!$js) {
> 			return str_replace('&', '&amp;', $url);
> 		} else {
> 			return $url;
> 		}
> ```

don't understand

why $js?

chulcha:
> > ))))))))))) not problem
> > ```
> > 		if (!$js) {
> > 			return str_replace('&', '&amp;', $url);
> > 		} else {
> > 			return $url;
> > 		}
> > ```
> 
> don't understand
> 
> why $js?

look at the code

dreanei23:
> > > ))))))))))) not problem
> > > ```
> > > 		if (!$js) {
> > > 			return str_replace('&', '&amp;', $url);
> > > 		} else {
> > > 			return $url;
> > > 		}
> > > ```
> > 
> > 
> > don't understand
> > why $js?
> 
> look at the code

`str_replace('&', '&amp;', 'site.ru/index.php/?route=extension/my/module&amp;param1=1&param2=2');
`
result: `site.ru/index.php/?route=extension/my/module&amp;amp;param1=1&amp;param2=2`

`&amp;amp;` )))

chulcha:
> > > > ))))))))))) not problem
> > > > ```
> > > > 		if (!$js) {
> > > > 			return str_replace('&', '&amp;', $url);
> > > > 		} else {
> > > > 			return $url;
> > > > 		}
> > > > ```
> > > 
> > > 
> > > don't understand
> > > why $js?
> > 
> > 
> > look at the code
> 
> `str_replace('&', '&amp;', 'site.ru/index.php/?route=extension/my/module&amp;param1=1&param2=2'); ` result: `site.ru/index.php/?route=extension/my/module&amp;amp;param1=1&amp;param2=2`
> 
> `&amp;amp;` )))

https://github.com/opencart/opencart/blob/8ffe95f7e9ed84ab54acf45a85d42e358751fa2f/upload/system/library/url.php#L70

dreanei23:
had seen, sorry

danielkerr:
public function link(string $route, $args = '', bool $js = false): string {

use the js = true so it does not convert the &

Issue #12978, opened 2023-11-18
[Master] No error message when regex validation is invalid


condor2:
**What version of OpenCart are you reporting this for?**
Master

**Describe the bug**
Add a new Text custom field and write some letters in Validation input (fgfgf) then click save.

No error message it shown.

Console log message error

<b>Unknown</b>: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in <b>admin\controller\customer\custom_field.php</b> on line <b>387

<b>Warning</b>: preg_match(): Delimiter must not be alphanumeric, backslash, or NUL in <b>admin\controller\customer\custom_field.php</b> on line <b>387</b>{"error":{"validation":"error_validation"}}`

condor2:
Closed....and Admin warning message where is?

danielkerr:
@AJenbo im getting a bit worried about some of ur commits now.

AJenbo:
https://github.com/opencart/opencart/pull/12994#issuecomment-1823144109

danielkerr:
should be ok now

PR #12994, opened 2023-11-22
Fix deprication error in custom_field.php

AJenbo:
Fixes https://github.com/opencart/opencart/issues/12978

### DIFF
--- a/upload/admin/controller/customer/custom_field.php
+++ b/upload/admin/controller/customer/custom_field.php
@@ -384,7 +384,7 @@ class CustomField extends \Opencart\System\Engine\Controller {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


condor2:
Warning: preg_match(): Delimiter must not be alphanumeric, backslash, or NUL in admin\controller\customer\custom_field.php</b> on line <b>387</b>{"error":{"validation":"error_validation"}}

Also I don't see an warning message in admin like...This regex is invalid...

No error_validation in language file

condor2:
yes...but there is no warning message to display in admin.....

AJenbo:
Ok, but it's not going to be something I solve in this PR. I would suggest you create another issue about it.

PR #13004, opened 2023-11-25
Update pgsql.php

TheCartpenter:


mhcwebdesign:
As I said a few weeks ago, I don't think any of current OpenCart 4 versions will work with postgreSQL, especially not since all the recent changes introducing  the backticks. Hence, perhaps it's better just to drop support for this DB driver altogether, at least for the time being.

TheCartpenter:
Perhaps. However, by doing so, it would also affect extensions depending on that library.

PR #13030, opened 2023-11-28
[3.0.x.x] Added language variable 'text_no_results' for Manufacturer …

condor2:
…List

When there are no brands the correct mesage is no brands...and not no products...


mhcwebdesign:
Thanks for this correction, now added to the 3.0.x.x branch.

Issue #13028, opened 2023-11-28
Object properties inside models


stalker780:
https://github.com/opencart/opencart/commit/846a5bf016924059a98e73697cdde5a6c8bed7ba#diff-d3c1b486ca583b5faf702ceec3acefd3ee8063edd614cbd450290be77f9984a3L121

After this commit models are reconstructed on every call for example `$this->model_catalog_product->getProduct()`.
You cannot store any variables inside a model class, because they are reset every time a model is called.

Don't know if it was meant to work like that, but this breaks a lot of functionality of PHP objects, also it takes more time to reconstruct every model class. Especially if you want to use own `__construct` method in your extension.

BTW controllers have this issue since OC1.5 or OC2.0 as far as I remember.

IMHO when you code PHP objects and want to use a class properties like `$this->my_var` you must be able to use them and you rely that these properties are saved between calls to an object. Otherwise why are we using PHP objects at all?

I have tried to use previous version of `model` method in loader.php and it works fine. Properties are saved.

danielkerr:
needs to be fixed a different way

danielkerr:
also controllers should not be stored.

PR #13032, opened 2023-11-28
Store model proxy in registry

stalker780:
Issue is described here
https://github.com/opencart/opencart/issues/13028

### DIFF
--- a/upload/system/engine/loader.php
+++ b/upload/system/engine/loader.php
@@ -123,6 +123,9 @@ class Loader {
 						// https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list
 						// https://wiki.php.net/rfc/variadics
 						$proxy->{$method} = function (&...$args) use ($route, $method) {
+							// Create a key to store the proxy object
+							$key = 'proxy_' . str_replace('/', '_', $route);
+
 							$route = $route . '/' . $method;
 
 							$trigger = $route;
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
This change is unneeded. It already uses this logic:

```
if (!$this->registry->has($key)) {
    ....
    $this->registry->set($key, $proxy);
} else {
    throw new \Exception('Error: Could not load model ' . $class . '!');
}

```

stalker780:
@mhcwebdesign I see it. But you can run your own tests
`__construct` method in `model_catalog_product` is executed multiple times on each page. Thus the model is reconstructed on every call from controllers.
Just add echo to `__construct` function and check yourself.

I just propose a fix. Perhaps Daniel can fix this another way. 
As far as I understand he is still implementing ocmod, so there may be more modifications to loaders.

stalker780:
`$this->registry->set($key, $proxy);` stores proxy not a model. Model is still reconstructed on each call from proxy.

mhcwebdesign:
> `$this->registry->set($key, $proxy);` stores proxy not a model. Model is still reconstructed on each call from proxy.

May be the issue could be cause by line 140:

`    $model = new $class($this->registry);
`

?

stalker780:
@danielkerr not fixed.
Here is an example with latest master.

model_catalog_product is still reconstructed on every call of it's methods

`$this->model_catalog_product->getTotalProducts()`
`$this->model_catalog_product->getProducts()`

image

stalker780:
Just add this method to any model or controller and you will see how many times it is reconstructed

```
public function __construct(\Opencart\System\Engine\Registry $registry) {
	echo '<pre>'.__METHOD__.' ['.__LINE__.']: '; print_r('constructed'); echo '</pre>';
	echo '<pre>'.__METHOD__.' ['.__LINE__.']: '; print_r(debug_print_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS)); echo '</pre>';
	$this->registry = $registry;
}
```

stalker780:
https://github.com/opencart/opencart/blob/95810405c98f77cec9530fd0c600bad7e4f42b48/upload/system/engine/loader.php#L150

If I change `static` with `global` it works fine and model is constructed only once:

`global $model;`

danielkerr:
but ur right it needs fixing.

PR #13045, opened 2023-11-28
file upload fix

captainerd:
Fixes file uploads.


danielkerr:
thim not sure what this fixes. seems all the code is old.

captainerd:
You are right its old, i forgot to Synch!


I don't know if you have fixed it now, BUT the "file upload / " is inaccessible for download, once its uploaded it is lost for ever because in many files / instances of the code the following :

	'name'  => $option['name'],
							'value' => $upload_info['name'],
							
Stores the name of the file instead of the "code" thous the $this->model_tool_upload->getUploadByCode($option['value']); 

will fail later.

PR #13055, opened 2023-12-01
Update redis.php

TheCartpenter:


danielkerr:
very dumb way to say empty

PR #13065, opened 2023-12-02
Update attribute_group.php

TheCartpenter:


ADDCreative:
This is another commit that only seems to break the indentation and dosen't comply with https://github.com/opencart/opencart/wiki/Coding-Standards#indentation.

ADDCreative:
> This change has been made under coding standards identically to what other controllers and model files are already describing.

That is just not ture.

TheCartpenter:
> > This change has been made under coding standards identically to what other controllers and model files are already describing.
> 
> That is just not ture.

If you can't find that to be true then, clearly, you're not up-to-date with the change.

ADDCreative:
> If you can't find that to be true then, clearly, you're not up-to-date with the change.

I can't see any change to the coding standards. So replacing the tab indenting with spaces just seems like a waste of time.
https://github.com/opencart/opencart/wiki/Coding-Standards#indentation

If the the coding standards have changed, them please update them.

mhcwebdesign:
All I can see are 2 added blanks, no bugfixes, no changed tab-based indentations.

As I said, these frequent unneeded cosmetic changes will cause a lot of future broken extensions which use OCmod.

PR #13072, opened 2023-12-02
Update article.php

stalker780:
One more attempt to fix this controller. Placeholder var must be set before it is used.

https://github.com/opencart/opencart/blob/cebabb492e46d49c2fbdf7d567199510fe80daee/upload/admin/controller/cms/article.php#L238


stalker780:
image

TheCartpenter:
If the case, then it goes against everything in best practices based on constructor development by relying on each methods since variables inside a method doesn't have to be added to a specific location as long as the variable has been used already from previous lines or if being placed into an instance variable...

danielkerr:
i already fixed

PR #13087, opened 2023-12-03
Update opencart.sql - add permissions for Subscriptions Report

pemaier:


TheCartpenter:
@danielkerr: This one needs to be added as second after this one: https://github.com/opencart/opencart/pull/13089 since subscriptions are being added into the ```user_group``` table while #13089 does not have the ```user_group``` for subscriptions in it.

pemaier:
About that question I was also thinking. The point is that when we have a field where we are setting limits and it should be used if it is required. The user account is (I guess) already limited with the session lifetime so the field is normally not required. Especially you are setting the lifetime already to value  => 3600000000 ... so the question is why we need the field in the settings if it has not really a function? (May I don't really see it in the whole.) 
btw. this is the only field in the settings which is empty.

PR #13088, opened 2023-12-03
Update opencart.sql - added Affiliate Expire time

pemaier:


danielkerr:
not sure why the affiliate needs to expire?

TheCartpenter:
> not sure why the affiliate needs to expire?

- admin/controller/setting/setting.php - 278
- admin/view/template/setting/setting.twig - 944, 946, 947
- admin/controller/startup/marketing.php - 47
- install/controller/upgrade/upgrade_4.php - 261

PR #13092, opened 2023-12-05
Update payment_method.php

captainerd:
Changing is_callable() to method_exists()  as is_callable() produced an error to me breaking the page, [ php8.1 without a payment extension installed ]

### DIFF
--- a/upload/catalog/controller/account/payment_method.php
+++ b/upload/catalog/controller/account/payment_method.php
@@ -90,7 +90,7 @@ class PaymentMethod extends \Opencart\System\Engine\Controller {
 			if ($this->config->get('payment_' . $result['code'] . '_status')) {
 				$this->load->model('extension/' . $result['extension'] . '/payment/' . $result['code']);
 
-				if (is_callable([$this->{'model_extension_' . $result['extension'] . '_payment_' . $result['code']}, 'getStored'])) {
+				if (method_exists([$this->{'model_extension_' . $result['extension'] . '_payment_' . $result['code']}, 'getStored'])) {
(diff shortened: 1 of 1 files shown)


TheCartpenter:
> Changing is_callable() to method_exists() as is_callable() produced an error to me breaking the page, [ php8.1 without a payment extension installed ]

What error?

captainerd:
> > Changing is_callable() to method_exists() as is_callable() produced an error to me breaking the page, [ php8.1 without a payment extension installed ]
> 
> What error?

Notice: Undefined property: Proxy::getStored in /var/www/html/catalog/controller/account/payment_method.php on line 94: in /var/www/html/system/engine/proxy.php on line 85

captainerd:
> > Changing is_callable() to method_exists() as is_callable() produced an error to me breaking the page, [ php8.1 without a payment extension installed ]
> 
> What error?

Apparently method_exists() is better suited for the scenario of methods, when there is active the other bug here

https://github.com/opencart/opencart/blob/master/upload/admin/model/setting/store.php  createStoreInstance() not initializing a factory() it may plays a part on why can't reproduce it? i work and i submit with what is uploaded here right at the moment is there a another dev repo ?

captainerd:
,just tested it again with a fresh download, and i can reproduce both.

if (method_exists($this->{'model_extension_' . $result['extension'] . '_payment_' . $result['code']}, 'getStored'))   {


image


And also about the other bug i was talking, seems to be related. when you miss 

$registry->set('factory', new \Opencart\System\Engine\Factory($registry));

at createStoreInstance() 

image

PR #13097, opened 2023-12-05
Checkbox labels

stalker780:
Checkboxes should be linked to labels like other inputs in admin area

### DIFF
--- a/upload/admin/view/template/catalog/category_form.twig
+++ b/upload/admin/view/template/catalog/category_form.twig
@@ -127,7 +127,7 @@
                 </div>
               </div>
               <div class="row mb-3">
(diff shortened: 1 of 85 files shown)


TheCartpenter:
In OC 4.x releases, ```for=``` needs to go before ```class=``` .

danielkerr:
should not as when u click on the label it clicks the slider

stalker780:
> should not as when u click on the label it clicks the slider

That's the idea. This makes UI more convenient.

danielkerr:
> > should not as when u click on the label it clicks the slider
> 
> That's the idea. This makes UI more convenient.

No it doesn't matter thats why I purposely removed the label.

PR #13108, opened 2023-12-08
Update en-gb.php

TheCartpenter:


ADDCreative:
This is pointless. Are you going to change Text to Texts, Entry to Entries and Error to Errors in all the language files as well?

stalker780:
@ADDCreative yep. One by one. Every single file 🤣

TheCartpenter:
> This is pointless. Are you going to change Text to Texts, Entry to Entries and Error to Errors in all the language files as well?

If need be. So far, I haven't seen any of those enumerated from the above in the core

TheCartpenter:
> Can we just focus on genuine testing and bugfixes please? OpenCart 4 still has a number of issues to be addressed, nobody cares about minor cosmetic changes.

What's in the core is already what's being focused on.

PR #13148, opened 2023-12-16
Update review.php

TheCartpenter:


upload/catalog/controller/product/review.php
chulcha, line 1: add limit in settings

config_review_limit

TheCartpenter: > add limit in settings
> 
> config_review_limit

It is in the admin but is not a mandatory field. Therefore, there are no default value defined either.

chulcha: Where?
If it exists, then why is it not used?
And only event on model for change data (((

TheCartpenter: Apologize. I thought you were referring to the blog setting limit. Back on topic:

- catalog/controller/account/download.php
- catalog/controller/account/reward.php
- catalog/controller/account/returns.php
- catalog/controller/account/order.php
- catalog/controller/account/subscription.php
- catalog/controller/account/transaction.php
- catalog/controller/cms/comment.php

These files also have user defined value limits, therefore, no ```config_x_limit``` defined either in the catalog.

PR #13177, opened 2023-12-17
Add missing symbol visibility

AJenbo:
This was done using php-cs-fixer's visibility_required rule. Once all outstanding PR regarding code formatting have been merged future changes can be checked for issues in GitHub action and developers can run the tool locally to fix any formatting issues with there code.

### DIFF
--- a/upload/catalog/model/setting/startup.php
+++ b/upload/catalog/model/setting/startup.php
@@ -9,7 +9,7 @@ class Startup extends \Opencart\System\Engine\Model {
 	/**
 	 * @return mixed
 	 */
-	function getStartups() {
+	public function getStartups() {
 		$query = $this->db->query("SELECT * FROM `" . DB_PREFIX . "startup` WHERE `status` = '1' ORDER BY `sort_order` ASC");
 
 		return $query->rows;
--- a/upload/extension/opencart/catalog/model/shipping/flat.php
+++ b/upload/extension/opencart/catalog/model/shipping/flat.php
@@ -11,7 +11,7 @@ class Flat extends \Opencart\System\Engine\Model {
 	 *
 	 * @return array
 	 */
-	function getQuote(array $address): array {
+	public function getQuote(array $address): array {
 		$this->load->language('extension/opencart/shipping/flat');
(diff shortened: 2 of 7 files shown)


TheCartpenter:
```public const CACHEDUMP_LIMIT = 9999;```

Any reason why these const should be put public?

AJenbo:
For backwards comparability, previously they would be publicly available and might have been used by mods/extensions/plugins, and it would be best not to break that.
Now that all symbols have had visibility applied we can add a rule that tests that they are set correctly in new code so such mistakes are not made in the future.

TheCartpenter:
Yes, I would agree here that a rule should be set for code structure. In order to achieve this, still, each commits would have to use the same templates - which is totally fine. It's just that not all builders / software uses the same criteria as defined rules.

TheCartpenter:
Should it need a const to be put public for that? It would be more doubtful, since there might be other ways to achieve this from the workflows.

AJenbo:
I have a rule set ready for php-cs-fixer that can be added as soon as thease PRs are evaluated an merged. Then anyone can install that and run it to have there code adjusted automatically.

TheCartpenter:
> I have a rule set ready for php-cs-fixer that can be added as soon as thease PRs are evaluated an merged. Then anyone can install that and run it to have there code adjusted automatically.

Sounds interesting. This might be exactly what is needed in order to submit proposals worldwide especially if it uses the repository's indent settings as opposed to tabs, or different indents, at the beginning of each lines.

AJenbo:
The rules I made are based on PSR12 (with a few adjustments to match the style of OC) + PHP 7.4 modernization. That's how I created the PRs I did today.

But there are still many more rules to evaluate and apply. I was thinking that maybe you would be interested in looking at that, then you can make PRs to make all the code formatted the same and then add the new rule to the configuration.

TheCartpenter:
If we can find out soon enough, that'd be great to give it a try by also assuming it would cover the code structure with opened and closed parentheses along with the rest of each line statements from the code once proposals will be submitted.

PR #13186, opened 2023-12-17
Modernize arrays

AJenbo:
This was done using php-cs-fixer's array_syntax rule. Once all outstanding PR regarding code formatting have been merged future changes can be checked for issues in GitHub action and developers can run the tool locally to fix any formatting issues with there code.

### DIFF
--- a/upload/system/library/db/pdo.php
+++ b/upload/system/library/db/pdo.php
@@ -34,7 +34,7 @@ class PDO {
 		}
 
 		try {
-			$pdo = @new \PDO('mysql:host=' . $hostname . ';port=' . $port . ';dbname=' . $database . ';charset=utf8mb4', $username, $password, array(\PDO::ATTR_PERSISTENT => false, \PDO::MYSQL_ATTR_INIT_COMMAND => 'SET NAMES utf8mb4 COLLATE utf8mb4_general_ci'));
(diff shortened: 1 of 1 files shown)


TheCartpenter:
I am not sure if PHP PDO takes the modern array structures in their parameters. It was discussed before as it seem to be using the array() instead of [] to avoid an error triggered. Not sure if this issue has been fixed with more recent PHP versions though.

AJenbo:
Before PHP executes code it is compiled in to byte code before the interpreted executes the code. both `[]` and `array()` compiles to the same byte code so there is no functional change here. PHP has had support for the short hand syntax since 5.4.

TheCartpenter:
I used to submit a PR about this before but it was closed because, prior to PHP 8, it would throw an error message if the old conventional ways wasn't used on that particular line at the time. Perhaps with PHP 8, things may have improved. However, since OC v4.x releases also supports PHP 7.4, it would be wise to test this on a higher level setting or on a production server since PHP 7.4 has been deprecated by PHP Corporation since the releases of PHP 8.x.

AJenbo:
All PRs test that the code can be compiled to byte code for version 7.4, 8.0, 8.1, 8.2 (and now also 8.3) exactly to avoid issues like that.

TheCartpenter:
While we can never be too sure about server differences, as time keeps running, I guess we'll keep finding that one out.

PR #13187, opened 2023-12-17
Use modern random number generator

AJenbo:
See notes on https://www.php.net/manual/en/function.rand.php


AJenbo:
@TheCartpenter this PR is a good example of how to enable and apply a new code style rule

PR #13189, opened 2023-12-17
Use real function names

AJenbo:
The function names used here where aliases and not the real function names. This was done using php-cs-fixer's no_alias_functions rule. Once all outstanding PR regarding code formatting have been merged future changes can be checked for issues in GitHub action and developers can run the tool locally to fix any formatting issues with there code.

### DIFF
--- a/upload/system/library/mail/smtp.php
+++ b/upload/system/library/mail/smtp.php
@@ -143,7 +143,7 @@ class Smtp {
 
 		if ($handle) {
 			if (substr(PHP_OS, 0, 3) != 'WIN') {
-				socket_set_timeout($handle, $this->option['smtp_timeout'], 0);
+				stream_set_timeout($handle, $this->option['smtp_timeout'], 0);
 			}
 
 			while ($line = fgets($handle, 515)) {
@@ -152,7 +152,7 @@ class Smtp {
 				}
 			}
 
-			fputs($handle, 'EHLO ' . getenv('SERVER_NAME') . "\r\n");
+			fwrite($handle, 'EHLO ' . getenv('SERVER_NAME') . "\r\n");
 
 			$reply = '';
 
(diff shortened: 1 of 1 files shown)


TheCartpenter:
Modifying the email functions now?

AJenbo:
> Modifying the email functions now?

Please see https://www.php.net/manual/en/function.socket-set-timeout.php and https://www.php.net/manual/en/function.fputs

TheCartpenter:
> > Modifying the email functions now?
> 
> Please see https://www.php.net/manual/en/function.socket-set-timeout.php and https://www.php.net/manual/en/function.fputs

https://www.stackoverflow.com/questions/49334133/fputs-vs-fwrite-in-terms-of-safety/49334196#49334196

AJenbo:
> > > Modifying the email functions now?
> > 
> > 
> > Please see https://www.php.net/manual/en/function.socket-set-timeout.php and https://www.php.net/manual/en/function.fputs
> 
> https://www.stackoverflow.com/questions/49334133/fputs-vs-fwrite-in-terms-of-safety/49334196#49334196

That post is about C not PHP, what is said there does not apply to PHP at all.

This PR is just about consistency, the is no functional change here.

TheCartpenter:
PHP is still based on C. Before the release of PHP 8, however, there used to be a discussion a long time ago on this repository about arguments to use fwrite instead of fputs. Let's just hope the PHP improvement also involves the aliases.

AJenbo:
If you search the repository code you can see that the real function names are used in all other places, it's only in this file where aliases where used.

This one https://github.com/opencart/opencart/blob/master/upload/install/cli_install.php#L428 has been there for 11 years, which would be when PHP 5.4 came out.

> PHP is still based on C.

It's written in C, but that's different.

TheCartpenter:
That is correct for the ```fwrite``` function. 14 matches in 39 matches in 20 files have been found as opposed to the ```fputs``` function that only seem to be used in the mail/smtp library file.

PR #13190, opened 2023-12-17
Use null coalescing to check for unset properties

AJenbo:


TheCartpenter:
I do believe it's a shorter version by using Ternary conditional operators such ways as it has been suggested before by others; it may also interfere with previous extensions now that OCMod has been re-integrated as it may with the VQMod engine. Should it now be the right time to introduce this change?

AJenbo:
Since the re-integratied OCMod has not been released yet this would be the right time to make the change imo

TheCartpenter:
Would take longer for extension developers to convert their extension packages though.

AJenbo:
Better then if they had to do it again later.

TheCartpenter:
> Better then if they had to do it again later.

Couldn't agree more on this one. Let's see where this one leads.

danielkerr:
this need to be backwards compatible. not sure if these ?? is in php 7.4

AJenbo:
@danielkerr ?? Support was added in php 7.0, it's part of the php-cs-fixer modinizatios for 7.0 and you can see that the 7.4 github action was able to parse the code (it does a syntax validation of all files).

https://www.php.net/manual/en/migration70.new-features.php

Hope that helps you feel confident that this is fully 7.4 compatible 😀

PR #13191, opened 2023-12-17
Add missing type hints

AJenbo:


upload/install/cli_cloud.php
```diff
@@ -60,7 +60,7 @@
  *
  * @return false
```
TheCartpenter, line 61: @return bool

AJenbo: This is more accurate. We can't do this is the native code since it is only supported on PHP 8.2+, but it's good to have it here to give more accurate information to IDE and anayzers. Once support for PHP 8.1- is dropped the native hint and be updated to match the PHPDoc.

TheCartpenter: Yes, correct. There are still inaccurate results about the ways to use the outputs with PHP v8+ releases as per the PHPDoc. However, it is also the reason why no output has been specified on this method as it may interfere with some operations during the install / upgrade process.

danielkerr:
adding protected \Opencart\System\Engine\Registry $registry; sort of stops some of the classes being interchangeable. just put that is needs to be an object

AJenbo:
It might be good to use an interface for that to make sure they are compatible

PR #13196, opened 2023-12-17
Test with PHP 8.3

AJenbo:


TheCartpenter:
It may actually cover: https://github.com/opencart/opencart/discussions/13128

PR #13244, opened 2023-12-19
[3.0.x.x] Fixed PHP 8 warning Laybuy model

condor2:


ADDCreative:
This is wrong. You can't just change the order of the parameters, as it's used in the controller.

https://github.com/opencart/opencart/blob/5206a4bf3200fafe5d3ef5ab1d351d5b48c40dee/upload/catalog/controller/extension/payment/laybuy.php#L142
https://github.com/opencart/opencart/blob/5206a4bf3200fafe5d3ef5ab1d351d5b48c40dee/upload/catalog/controller/extension/payment/laybuy.php#L245

Since $data is not be optional. You just need.

```php
public function addTransaction($data, $status) {
```

mhcwebdesign:
@ADDCreative  : You are right, I have now fixed it.

PR #13248, opened 2023-12-19
Update header.php

chongshengdz:
empty link if not logged in, we should add a link to it for better seo


danielkerr:
ur commit does not make sense. search engines dont search the account.

chongshengdz:
> ur commit does not make sense. search engines dont search the account.

yes, you are right, but an empty link is not good for seo,
you can check google's [PageSpeed Insights](https://pagespeed.web.dev/)

PR #13249, opened 2023-12-19
Update review.php

chongshengdz:
{{ back }} is undefined in twig, we should add it in controller


danielkerr:
i just removed the back button. should not be there.

PR #13250, opened 2023-12-19
Testing change with multiple files by merging commits

TheCartpenter:
I think it works now!


TheCartpenter:
Now, people can bring back their email notifications from Github. The issue has been resolved. :)

TheCartpenter:
This is also an intended code change proposal to align the TDs since line 713 and line 720 of product_form.twig originally differs one and another.

PR #13252, opened 2023-12-19
Completed the TD alignments in admin

TheCartpenter:


mhcwebdesign:
What exactly are your trying to accomplish here? These are no bugfixes from what I can see!

mhcwebdesign:
This was perfectly fine, what exactly are you trying to accomplish? Can't we focus on genuine bugfixes?

mhcwebdesign:
Your daily weird changes are a good reason for extensions developers to avoid OCmod-based changes and to stay away from the master branch for the time being.

TheCartpenter:
> Your daily weird changes are a good reason for extensions developers to avoid OCmod-based changes and to stay away from the master branch for the time being.

OCMod is not out, yet. It is rather the best time to match those codes as much as we can so that extension developers can dig up the codes with ease.

Issue #13155, opened 2023-12-16
[master] Admin / Customer group after sorting without header/footer


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?
After sorting Customer Group entries only the customer group entries are visible. Missing header/footer.

**To Reproduce**
1. Go to 'Admin/Customer/Customer Group'
2. Click on 'Sort by Customer Group Name' or Sort Order

**Expected behavior**
Admin page complete

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

![customer_group_sort_2023-12-16_23-37-45](https://github.com/opencart/opencart/assets/10166095/408f6e9d-4626-4ed8-8f5e-ba4826756093)

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? server
 - Operating system Linux
 - PHP version 8.2
 - Apache version 2.4
 - Browser(s) tested with [e.g. chrome, safari - if applicable] FF
 - Version [e.g. 22] 115.0.5

Samdz19:
Yes, it's need to edit this : 

> File : admin/view/template/customer/customer_group.twig

And add before `{{ footer }} `
These lines
```
<script>
$('#customer-group').on('click', 'thead a, .pagination a', function (e) {
    e.preventDefault();

    $('#customer-group').load(this.href);
});
</script>
```

pemaier:
@Samdz19  Thank you for your support. I tested it and it is working. I also created a pr.

pemaier:
solved

PR #13256, opened 2023-12-19
Update customer_group.twig - add footer/header after sort and update user_list

pemaier:
solution for issue #13155


TheCartpenter:
It simply takes an event from an extension to pull that list. It is not a bug.

pemaier:
@TheCartpenter If only an event is required I'm very interested in your solution :)

pemaier:
Added Firstname/Lastname and new status style to user list.
image

pemaier:
Added User Group to user list
image

cgcmart:
Great work.
Happy Holiday!

TheCartpenter:
> Added User Group to user list image

There are no validation set in the save() method that checks for active groups. Therefore, your integration can affect many customers at once. Especially from those who are using subscription cycles, and that their cycle period have not yet ended, it could become chaotic.

pemaier:
@TheCartpenter - Interesting to listen what you tell. But I can't see any point where my changes are saving something. I don't see where I'm using the save() method ... there is only reading of data. And on the other side there is also no way to deactivate a user group. Also it is something new for me that a user or user_id is related to subscription cycles. May I don't understand the relation now. I don't see the problem but you are welcome to explain where the problem is with this changes and where the chaos can be?

TheCartpenter:
> @TheCartpenter - Interesting to listen what you tell. But I can't see any point where my changes are saving something. I don't see where I'm using the save() method ... there is only reading of data. And on the other side there is also no way to deactivate a user group. Also it is something new for me that a user or user_id is related to subscription cycles. May I don't understand the relation now. I don't see the problem but you are welcome to explain where the problem is with this changes and where the chaos can be?

A bit misleading since your PR subject is about the ```customer_group``` - not the ```user_group```. My mistake., I guess; I thought your screenshot was still referring to the ```customer_group```.

cgcmart:
@TheCartpenter You probably confused back end user with store front customer / subscriber.

TheCartpenter:
> @TheCartpenter You probably confused back end user with store front customer / subscriber.

I was rather confused by the PR title vs. the screenshot that was posted since it's about two separate things.

danielkerr:
lets not mix up the commits.

Issue #12290, opened 2023-04-25
[4.0.2.1] Administration language doesn't change, always loading the English translation


AlexandraGkogka:
**What version of OpenCart are you reporting this for?**
4.0.2.1 Fresh install

**Describe the bug**
After installing and enabling a new language extension, the administration always displays the English language translation.

The administration language select dropdown (nav-language) changes the flag and sets the correct language cookie, on each language change, but the OpenCart translation always stays to English.

This issue occurs only in the administration.

**To Reproduce**
1. Install a new language extension.
2. In settings, select the new language as the default Language and Administration Language and save.
3. Logout, clear all cookies and login again to the  Administration.
4. The language translation is still English.
5. From language drop down menu, change the language and select the installed language.
6. The language menu selection changes to the new language, but the language translation is still English.

**Server / Test environment (please complete the following information):**
 - Cloud server
 - Operating system: Ubuntu 20.04
 - PHP version: 8.1, 8.2
 - Apache version: 2.4.55
 - Browser(s) tested with: Chrome, Firefox

**FIX Proposal**
Until a better solution is found, the language issue can be fixed by modifying this file:
admin/controller/startup/language.php

Replace this line:
$this->language->addPath('extension/' . $language_info['extension'], DIR_EXTENSION . $language_info['extension'] . '/admin/language/');

With this:
$language = new \Opencart\System\Library\Language($language_info['code']);
$this->registry->set('language', $language);
$language->addPath(DIR_EXTENSION . $language_info['extension'] . '/admin/language/');

pemaier:
The upper solution is not working at all. Here a sample:
Go to order something as customer and go to sales and view the order.
![image](https://user-images.githubusercontent.com/10166095/234564840-7228c9d7-9167-44cc-ad9f-2ab37288ba12.png)
Till here all is ok BUT if you change the order language 
![image](https://user-images.githubusercontent.com/10166095/234565131-5069abc5-141e-49c3-8445-bd83e8deed76.png)
then you get this price view ...
![image](https://user-images.githubusercontent.com/10166095/234565350-899774ae-9d7e-4bf4-97f7-7c37947b8cf3.png)

AlexandraGkogka:
The solution has been tested and works fine for setting the displayed language in all administration pages. As I said it is a working solution until something more solid comes out.

[...]

PR #13262, opened 2023-12-21
[Master] Update language.php

osworx:
The problem: when a cookie is already set (e.g. de-de from another installation on the same server) and the new installation is made fresh with English only, **each admin database call for entries (e.g. stock status, order status, returns, etc.) is NOT successful!** 
Because de-de could not be found - it is still not installed (only en-gb)!

Therefore the language_id in config_language_id is always 0. 
Therefore the db result is always an empty array!

This commit solves this problem, or define the path for all cookies as it is done already for the OCSESSID.
And it will solve what is described here (empty lists without clearing cookies after installation): https://github.com/opencart/opencart/issues/12290

### DIFF
--- a/upload/admin/controller/startup/language.php
+++ b/upload/admin/controller/startup/language.php
@@ -25,20 +25,22 @@ class Language extends \Opencart\System\Engine\Controller {
 
 		self::$languages = $this->model_localisation_language->getLanguages();
 
-		if (isset(self::$languages[$code])) {
-			$language_info = self::$languages[$code];
-
-			// Language
-			if ($language_info['extension']) {
-				$this->language->addPath('extension/' . $language_info['extension'], DIR_EXTENSION . $language_info['extension'] . '/admin/language/');
-			}
-
-			// Set the config language_id key
-			$this->config->set('config_language_id', $language_info['language_id']);
-			$this->config->set('config_language_admin', $language_info['code']);
-
-			$this->load->language('default');
+		if (!isset(self::$languages[$code])) {
+			$code = array_key_first(self::$languages);
 		}
+			
+		$language_info = self::$languages[$code];
+
+		// Language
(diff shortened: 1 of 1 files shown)


danielkerr:
should the cookies not be folder specific.

danielkerr:
not sure u are fixing future issues. if cookies can be spread across folders then its open to abuse.

danielkerr:
probably better to store the language_id in the user table and not use a cookie.

osworx:
> probably better to store the language_id in the user table and not use a cookie.

Yes, that what I thought also.

The "problem" only is, when another (older) installation is made before, 4.x Version checks the language cookie (as OC is done it since the beginning).
So, after the installation of the new instance is made, and you open the administration to define (e.g.) the store settings, this store cookie is checked and the language therefore defined .. but the language (from the cookie) is at this stage not installed.
The results are empty dropdown fields (like other users here already mentioned, like https://github.com/opencart/opencart/issues/12290#issuecomment-1523975106).

Maybe one solution could be, to check if there is already a language cookie stored AND delete (or empty) it during the installation.
That would solve that issue.

stalker780:
I have already proposed a fix 3 months ago https://github.com/opencart/opencart/pull/12769

stalker780:
@danielkerr 
> probably better to store the language_id in the user table and not use a cookie.

This will not help. 
The problem is that if language is not installed/removed in OC, but it's code/id is stored in user's cookies/table you will get language_id = 0 in all SQL queries.

You just need to check the existence of language code on startup. If there is no such language - use default.

stalker780:
It is basically a 3-line fix. And it takes 6 months to fix this 🤣

image

danielkerr:
> I have already proposed a fix 3 months ago #12769

u didnt fix this issue! dont hijack threads.

PR #13265, opened 2023-12-21
Update upgrade_1.php

TheCartpenter:
Code structure - removed extra space on sprintf function.


AJenbo:
This was already done by php-cs-fixer

PR #13266, opened 2023-12-21
Update article.php

TheCartpenter:


AJenbo:
This was already corrected by php-cs-fixer

PR #13269, opened 2023-12-21
Update catalog/model/account/customer.php

Samdz19:
Add safe and commenter to register account, to fixe customer register error.

> Cannot assign null to property Opencart\System\Library\Cart\Customer::$safe of type bool: in C:\wamp64\www\oc4024a\system\library\cart\customer.php on line 84


danielkerr:
u dont need to add safe and commenter at the front.

PR #13277, opened 2023-12-23
[Master] article.php + check for status

osworx:
Add check for status - makes no sense to display a link when all articles are not enabled (status = 1)


osworx:
Have to make a new one with more changes.

PR #13278, opened 2023-12-23
[MASTER] comment.php

osworx:
Check if approval is required, get then correct text


AJenbo:
Does this depend on https://github.com/opencart/opencart/pull/13279 ? If so then please combine them in to a single PR

PR #13281, opened 2023-12-23
[Master] article.php

osworx:
Check if article is enabled, check also if the connected topic is enabled


danielkerr:
i dont think this will work if no topic assigned to an article

PR #13283, opened 2023-12-23
Added control_structure_continuation_position on same line

TheCartpenter:


TheCartpenter:
This generated file also seem to fail with PHP 8.3 checks.

TheCartpenter:
The original file also causes an error: expected ```]``` . I corrected it.

TheCartpenter:
I think it's for all the submissions being submitted, not just this file...

PHP 8.1 now says canceled on recent PRs as well.

AJenbo:
Other jobs are canceled when one fails. look at the failed, not the canceled

TheCartpenter:
> Other jobs are canceled when one fails. look at the failed, not the canceled

Correct. The details shows the error that I described about the ```.php-cs-fixer``` file.

TheCartpenter:
And also about the ```startup/seo_url``` void removal from this branch.

PR #13284, opened 2023-12-23
Removed void phpdoc in startup/seo_url

TheCartpenter:


AJenbo:
Maybe add a return null if they are not supposed to not return anything (void), or an empty string

AJenbo:
Looks like valid returns for action_pre_action controller index methods would be `null|\Opencart\System\Engine\Action|\Exception`.

For this on specifically since the return statement is missing they would fallback to null so we can set that is it's only return type, fixed here: https://github.com/opencart/opencart/pull/13289

PR #13285, opened 2023-12-23
Update modification.php

TheCartpenter:
Added phpdoc titles on methods

### DIFF
--- a/upload/admin/controller/marketplace/modification.php
+++ b/upload/admin/controller/marketplace/modification.php
@@ -175,6 +175,12 @@ class Modification extends \Opencart\System\Engine\Controller {
 		return $this->load->view('marketplace/modification_list', $data);
 	}
 
+	/**
+	 * Refresh
+	 * 
+	 * @return void
+	 * @throws \Exception
+	 */
 	public function refresh(): void {
 		$this->load->language('marketplace/modification');
 
@@ -566,10 +572,20 @@ class Modification extends \Opencart\System\Engine\Controller {
 		$this->response->setOutput(json_encode($json));
 	}
 
+	/**
+	 * Log
+	 *
+	 * @return void
+	 */
 	public function log(): void {
 		$this->response->setOutput($this->getLog());
 	}
 
+	/**
+	 * getLog
+	 * 
+	 * @return string
+	 */
 	public function getLog(): string {
 		$file = DIR_LOGS . 'ocmod.log';
 
@@ -580,6 +596,11 @@ class Modification extends \Opencart\System\Engine\Controller {
 		}
 	}
 
+	/**
+	 * Clear
+	 * 
+	 * @return void
+	 */
 	public function clear(): void {
 		$this->load->language('marketplace/modification');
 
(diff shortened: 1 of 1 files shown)


upload/admin/controller/marketplace/modification.php
```diff
@@ -175,6 +175,12 @@ public function getList(): string {
 		return $this->load->view('marketplace/modification_list', $data);
 	}
 
+	/**
+	 * Refresh
+	 *
```
AJenbo, line 180: There are a few more like this:
```suggestion
	 *
```

TheCartpenter: What I am noticing is that 99% of the users, except one, are failing those requirements where I am the only one receiving those replies. Therefore, I would ask you to provide the templates you are using from your editors as I am sure you can understand how this looks in general when users are submitting code changes involving ... red Xs included by the founder of their own repository. Also, to provide the replies to each users encountering red Xs since it seem only fair that I am not the only one failing these requirements.

AJenbo: I just enable the php-cs-fixer plug in, easy as that.

Also almost no one is failing it, it's just that Daniel pushes directly to master and introduce issues that then causes it PRs based on it to also fail.

TheCartpenter: > I just enable the php-cs-fixer plug in, easy as that.
> 
> Also almost no one is failing it, it's just that Daniel pushes directly to master and introduce issues that then causes it PRs based on it to also fail.

In other words, what you're implying is with the mount of red Xs posted by users, none of them are going to get merged into the core until users are matching your expectations to match the indents and we're all sitting on it.

@danielkerr: Have you agreed to this?

AJenbo: No, I don't even know how you came to that conclusion from anything I said. The repository is not configured to block PRs based on failing tests.

I feel like you are constantly looking to fight over nothing.
Why don't you just run the tool instead?

TheCartpenter: I was not referring to a configured setting to block PRs, I was referring to you on taking the initiative to merge PRs since the integration of this tool. As you may have not noticed, PHP-CS-Fixer plugin does not work on all developers tools. It has been reported in the past by other developers on this repository explaining that validating the plugin does return a: 'ok' response but when the time to use it in the developer's tool - it returns a 'no response' error.

Therefore, I am not sure how confident you really are with this tool or how far you expect to go with it but, as long as this rule keeps running, you are limiting programmers with their developers' tools of their choice to follow your criteria on this repository.

AJenbo: I don't belive I have seen others mentioning it. The tool never returns either of the things you mention, it sounds more like your interpretation so I can't accurately reply to it when you aren't being specific.

Considering that I told you to go explorer the configuration and propose changes to it I don't see how you could think so.

How is it limiting anyone in anyway, has it stopped you from making any of your PRs?

I'm again wondering if it would be easier to communicate in your native language as I feel some things are lost in translation.

TheCartpenter: Now that this issue has been solved with the latest modifications of the `.php-cs-fixer.php` file, this commit will now be closed.

PR #13286, opened 2023-12-23
Update cli_install.php

TheCartpenter:
Added phpdoc titles on methods.

### DIFF
--- a/upload/install/cli_install.php
+++ b/upload/install/cli_install.php
@@ -123,6 +123,13 @@ class CliInstall extends \Opencart\System\Engine\Controller {
 		$this->response->setOutput($output);
 	}
 
+	/**
+	 * Install
+	 * 
+	 * @param $argv
+	 *
+	 * @return string
+	 */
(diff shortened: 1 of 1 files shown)


upload/install/cli_install.php
```diff
@@ -123,6 +123,13 @@ public function index(): void {
 		$this->response->setOutput($output);
 	}
 
+	/**
+	 * Install
+	 * 
+	 * @param $argv
```
AJenbo, line 129: This is invalid without a type

TheCartpenter: Ok, closing commit. Will re-open a new one.

AJenbo: You can also update a branch by adding more commits to it to adjust and update the pr, but if you feel it's easier to make a new one that is also OK

TheCartpenter: I created a new PR: https://github.com/opencart/opencart/pull/13293 . All passed PHP versions.

PR #13289, opened 2023-12-24
Make null return explicit in SeoUrl::index()

AJenbo:


TheCartpenter:
@danielkerr: ```return null``` is not needed here. A method automatically validates with this type when it's not defined. It can also be tested with ```null``` before it returns but it does not require to return a ```null``` type if and when the output is not defined.

AJenbo:
If you don't then you are relying on error recovering and may well get depreciation issues in the future.

PR #13291, opened 2023-12-24
Revert "Pagination"

danielkerr:
Reverts opencart/opencart#13287


stalker780:
@danielkerr 
Is there any problem with this PR or you just don't want to change existing pagination?

If you need to modify `pagination` and `results` you need to modify more then 100 twig templates. That's absolutely stupid taking into account this is a standard module on every list.

PR #13292, opened 2023-12-24
Update .php-cs-fixer.php

TheCartpenter:


TheCartpenter:
> No, but you reverted several rules to different configurations, please make sure you use the latest code when making changes

All I really did was to move the ```setFinder``` parameters on the top into a ```variable```. I then changed the ```PhpCsFixer\Finder``` to: ```Symfony\Component\Finder\Finder```. ```ignoreDotFiles``` has also been added since there isn't really any need for it but you can always add it back if you want to include them during lookups.

I also added this line: ```'control_structure_continuation_position'  => ['position' => 'same_line'],``` .

AJenbo:
What was the change needed? What error where you getting and what version are you using?

TheCartpenter:
> What was the change needed? What error where you getting and what version are you using?

Since you're asking a question that was already answered on a previous commit to you, it means you haven't took the time to read it.

As mentioned, the error message was: ```expecting ]``` . Test it in your console, you should notice this error message.

TheCartpenter:
With this commit, it solves the issue entirely.

AJenbo:
There really isn't any need for you to have this negative and accusatory tone.

I don't think you infact fully answered the question (if so please point to where you mention the version).

I did see the message you wrote about `]`, but that was after you changed the file, not the one that is in the repository. If it was actually broken in the repo then the Action would have been failing also.

I can't merge this PR either since you didn't use the latest file from the repo. That's why it's failing and showing you all the files that the config you made does not align with.

AJenbo:
My current assumption is that you have an out of date version and that is what is causing you issues, but without additional info from you I can't help resolve it for you.

TheCartpenter:
> There really isn't any need for you to have this negative and accusatory tone.
> 
> I don't think you infact fully answered the question (if so please point to where you mention the version).
> 
> I did see the message you wrote about `]`, but that was after you changed the file, not the one that is in the repository, so I expected this to be something different, especially since that isn't what you have changed either. If it was actually broken in the repo then the Action would have been failing also.
> 
> I can't merge this PR either since you didn't use the latest file from the repo. That's why it's failing and showing you all the files that the config you made does not align with.

Maybe it just wasn't the last one... it seem to work fine now and when I just tried it. The version is: 3.42 . This commit will now be closed.

TheCartpenter:
Ok, it works on opencart-3 as well now.

PR #13293, opened 2023-12-24
Added more phpdoc titles on methods

TheCartpenter:


TheCartpenter:
This time, it's good; all passed.

upload/install/cli_install.php
```diff
@@ -123,6 +123,12 @@ public function index(): void {
 		$this->response->setOutput($output);
 	}
 
+	/**
+	 * Install
+	 *
+	 * @param mixed   $argv
```
AJenbo, line 129: Looks like this is actually supposed to be array<int, string>, I corrected it in a follow up

PR #13294, opened 2023-12-25
Renamed param to OC convention in install/cli_install

TheCartpenter:
As per:

- admin/model/setting/setting.php
- system/library/mail.php

### DIFF
--- a/upload/install/cli_install.php
+++ b/upload/install/cli_install.php
@@ -126,7 +126,7 @@ class CliInstall extends \Opencart\System\Engine\Controller {
 	/**
 	 * Install
 	 *
-	 * @param array<int, string>   $argv
+	 * @param array|int|string $argv
 	 * @return string
 	 */
 	public function install($argv): string {


AJenbo:
It's an array of strings with keys of integers. What you have Witten is arrays of unknown, or integers, or strings. This is not correct and not equivalent

TheCartpenter:
Both mentioned files on the above are also written this way. I don't see anywhere else on the platform where the ```@param``` is written with ```<>``` . You're not following the OC convention but only your own with these brackets.

AJenbo:
It's not mine, and I'm thinking it's not in most places in OC because it was left out and has not been filled in yet.

Maybe you would find this article on the subject helpful:
https://dev.to/suckup_de/modern-phpdoc-annotations-11d4

TheCartpenter:
I never mentioned it was yours, what I said is that it is originally there with the use of: ```|``` rather than ```<>```. As to the site you're referring, this is a third-party website. Any official documents you could provide on that?

TheCartpenter:
For instance, here's one directly from PHP.net: https://pear.php.net/package/PhpDocumentor/docs/latest/phpDocumentor/tutorial_tags.param.pkg.html . Take a look at line 32. It uses: ```|``` to separate the params.

TheCartpenter:
While PHPDoc mentions single types within: ```<>``` separated with the ```[name]``` in their example: https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/param.html .

AJenbo:
Please read the article, | is the syntax for "or" where <> is the syntax for "of" they mean completely different things and are not simply interchangeable or a preference.

You are not going to find anything on php.net about phpdoc since php does not maintain that standard and does not have native support for generics, just like php 5 did not have native support for scalar type hints, hopeful it will come in version 9, but until then we have to use phpdoc to express these consepts to our editors etc.

AJenbo:
> While PHPDoc mentions single types within: `<>` separated with the `[name]` in their example: https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/param.html .

The [] syntax is the older more primitive dialect that does not provide a way to express the key type so you would be loosing that information by opting for this.

TheCartpenter:
In all hopes that what you're expected from the above is also what's expected by the cli_install with ```$argv``` parameter once users will run their fresh installs especially during production times of this release.

AJenbo:
I don't think you understand what roll phpdoc play, they do not incluence the runtime

TheCartpenter:
No, but they do influence when parameters themselves don't have their types defined so to know what should be expected in each methods.

PR #13295, opened 2023-12-25
[master] Admin / Update user list view with new style

pemaier:
sorry for mixing up


danielkerr:
if u look how i did customer i used CONCAT in the SQL for the name

pemaier:
My intention was to change as less from your code as possible. But nevertheless I'll change it in your proposed way tomorrow. Thank you for the hint.

pemaier:
I was having a deeper look into it. If I only change it to use CONCAT in the SQL query then I'm loosing the possibility to sort for first- and lastname. So I think to implement a filter function. In that way I also get the status out of the table header. Is that ok for you?

PR #13309, opened 2023-12-29
Use constants instead of functions

AJenbo:
Using constants is both safer and faster, since the functions can error, where constants are pre-computed.


upload/system/library/session/redis.php
```diff
@@ -36,7 +36,7 @@ public function __construct(\Opencart\System\Engine\Registry $registry) {
 	public function read(string $session_id): array {
 		$data = $this->redis->get($this->prefix . $session_id);
 
-		if (is_null($data) || empty($data)) {
+		if (null === $data || empty($data)) {
```
TheCartpenter, line 39: ```if ($data === null || empty($data)) {```

Still not sure why both need to be compared here since ```empty``` could still validate this standalone.

PR #12945, opened 2023-11-14
SEO optimization and automatization of OpenCart 4

ppalashturov:
Hello,
We from OpenCart Bulgaria have an idea to do some extra SEO optimization and automatization of OpenCart 4. We want to apply and make a pull request with the following functions:

Adding a PHP library for the SEO tags and adding the Documents library to add the tags (meta, schema, etc.) to each page. We are planning to use: https://github.com/melbahja/seo

Add functionality to generate sitemap.xml.

Add functionality for Image optimization like selecting image formats jpg or WebP from the admin panel when caching images. Add LazyLoad library.

Editable title or H1 tags in the pages of the default template.

Adding functionality for automatic generation of URL addresses.

As well as several other things for the better SEO.

Best regards!

PR #13310, opened 2023-12-29
Use native function calls

AJenbo:
This should be both faster and allow better error detection by code analyzers.


ppalashturov:
@AJenbo thanks for code review. I close PR  #12945.
Best regards.

danielkerr:
interesting

PR #13318, opened 2023-12-30
Update redis.php

TheCartpenter:


TheCartpenter:
Commit closed without closure notes. Original code in instanced variable is still unaligned.

PR #13326, opened 2023-12-30
Update user.php - Extended version

pemaier:
added filter entries for getTotalUsers

### DIFF
--- a/upload/admin/model/user/user.php
+++ b/upload/admin/model/user/user.php
@@ -155,6 +155,46 @@ class User extends \Opencart\System\Engine\Model {
 	public function getTotalUsers(): int {
 		$query = $this->db->query("SELECT COUNT(*) AS `total` FROM `" . DB_PREFIX . "user`");
 
+		$implode = [];
+
+		if (!empty($data['filter_username'])) {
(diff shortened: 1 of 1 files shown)


upload/admin/model/user/user.php
```diff
@@ -155,6 +155,46 @@ public function getUsers(array $data = []): array {
 	public function getTotalUsers(): int {
 		$query = $this->db->query("SELECT COUNT(*) AS `total` FROM `" . DB_PREFIX . "user`");
 
+		$implode = [];
+
+		if (!empty($data['filter_username'])) {
```
AJenbo, line 160: Where is `$data` supposed to come from?

upload/admin/model/user/user.php
```diff
@@ -155,6 +155,46 @@ public function getUsers(array $data = []): array {
 	public function getTotalUsers(): int {
 		$query = $this->db->query("SELECT COUNT(*) AS `total` FROM `" . DB_PREFIX . "user`");
```
AJenbo, line 156: You forgot to change this:
```suggestion
		$sql = "SELECT COUNT(*) AS `total` FROM `" . DB_PREFIX . "user`";
```

AJenbo (changes requested):
This code is broken and won't work without changes.

PR #13334, opened 2023-12-31
Update affiliate.php

Samdz19:
add sort order to balance column


Samdz19:
https://github.com/opencart/opencart/pull/13333#issue-2060991599

PR #13336, opened 2023-12-31
Provide modern analyzable interface for getting models

AJenbo:
This pull request introduces a method for loading models that enhances analyzability using modern tools. This improvement provides developers with better feedback in their IDEs and facilitates the early detection of bugs during code analysis.

The enhancement involves documenting the `Proxy` object as a [generic](https://phpstan.org/blog/generics-in-php-using-phpdocs) and adding a method to the loader that takes the class name as input, allowing for predictions about the concrete implementation to be returned.

An example usage in `attribute_group.php` has been updated to showcase the new method. While the initial implementation may seem verbose:
```php
$model_catalog_attribute_group = $this->load->getModel(\Opencart\Admin\Model\Catalog\AttributeGroup::class);

$model_catalog_attribute_group->deleteAttributeGroup($attribute_group_id);
```

It can be simplified, especially with namespace imports:
```php
$this->load->getModel(AttributeGroup::class)->deleteAttributeGroup($attribute_group_id);
```

After this change, the IDE can suggest relevant functions for $model_catalog_attribute_group, and tests can assert that deleteAttributeGroup() exists and is given the correct arguments, preventing potential bugs.

### DIFF
--- a/upload/admin/controller/catalog/attribute_group.php
+++ b/upload/admin/controller/catalog/attribute_group.php
@@ -114,9 +114,9 @@ class AttributeGroup extends \Opencart\System\Engine\Controller {
 			'limit' => $this->config->get('config_pagination_admin')
 		];
 
-		$this->load->model('catalog/attribute_group');
+		$model_catalog_attribute_group = $this->load->getModel(\Opencart\Admin\Model\Catalog\AttributeGroup::class);
 
-		$results = $this->model_catalog_attribute_group->getAttributeGroups($filter_data);
+		$results = $model_catalog_attribute_group->getAttributeGroups($filter_data);
 
 		foreach ($results as $result) {
 			$data['attribute_groups'][] = [
@@ -148,7 +148,7 @@ class AttributeGroup extends \Opencart\System\Engine\Controller {
 			$url .= '&order=' . $this->request->get['order'];
 		}
 
-		$attribute_group_total = $this->model_catalog_attribute_group->getTotalAttributeGroups();
+		$attribute_group_total = $model_catalog_attribute_group->getTotalAttributeGroups();
 
 		$data['pagination'] = $this->load->controller('common/pagination', [
 			'total' => $attribute_group_total,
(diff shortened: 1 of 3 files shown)


TheCartpenter:
Except it would affect extensions between previous OC releases and OC 4 whereas each comparison would need to be made rather than keeping the same load models.

AJenbo:
> Except it would affect extensions between previous OC releases and OC 4 whereas each comparison would need to be made rather than keeping the same load models.

Your are wrong. The previous method is still available. This is an addition for preventing bugs in the core.

TheCartpenter:
> > Except it would affect extensions between previous OC releases and OC 4 whereas each comparison would need to be made rather than keeping the same load models.
> 
> Your are wrong. The previous method is still available. This is an addition for preventing bugs in the core.

Based on the example you're providing above, you are not adding but rather removing and replacing codes from previous codes into a new concept that's been developed from the Loader Engine.

AJenbo:
Then please read the full diff if you have a hard time understanding just the example.

TheCartpenter:
> Then please read the full diff if you have a hard time understanding just the example.

Since you are suggesting the load of ```$this->load->getModel``` , with a modified Loader Engine, those lines have been fully read already.

mhcwebdesign:
@AJenbo : Can we not please go overboard with this? Having 2 way of getting a model class, with you suggested new generic approach simply for the sake of some editors supporting some more features is a bit overdoing it. As was the case the sheer amounts of cosmetic formatting changes. Not everyone uses PHPStorm, there are better editors out there for Linux!

AJenbo:
@mhcwebdesign I don't use PHPStorm, I don't find it to be a very good editor. I do use Linux though.

This is primly for making it possible to detecting bugs in the code, better editor support is a side benefit. Without a tool assisted way to check for bugs it's simply not feasible to effectively bug hunt in a large code base like OC.
You can see some of the other bugs I was able to find and fix this way here:
https://github.com/opencart/opencart/pull/12901 https://github.com/opencart/opencart/pull/12902 https://github.com/opencart/opencart/pull/12906 https://github.com/opencart/opencart/pull/12914 https://github.com/opencart/opencart/pull/12920 https://github.com/opencart/opencart/pull/12919 https://github.com/opencart/opencart/pull/12918 https://github.com/opencart/opencart/pull/12917 https://github.com/opencart/opencart/pull/12921 https://github.com/opencart/opencart/pull/12955 https://github.com/opencart/opencart/pull/12956
(I left out the ones that only produced warnings when show errors is enabled on the server)

Unfortunately futter analysis of the code is not possible because of how models are currently being loadedd

I'm thinking it's best not to remove the old method (to keep versions compatible), at least until OC 5, but we could mark the old one as deprecated. What do you think?

TheCartpenter:
https://github.com/opencart/opencart/pull/13336#issuecomment-1872985342
https://github.com/opencart/opencart/pull/13336#issuecomment-1872986485
https://github.com/opencart/opencart/pull/13336#issuecomment-1872998174

danielkerr:
it should not effect extensions.

danielkerr:
i dont like adding $model_catalog_attribute_group = $this->load->getModel(\Opencart\Admin\Model\Catalog\AttributeGroup::class); 

it would be better in the load model method

AJenbo:
> it would be better in the load model method

@danielkerr could you share more details on your preferred approach for this? Since this is a roadblock for doing further analysis of the code I'm unable to track down and fix bugs until it is resolved.

PR #13337, opened 2023-12-31
Update user.php

pemaier:
Updated getUsers and getTotalUsers / added filter entries

### DIFF
--- a/upload/admin/model/user/user.php
+++ b/upload/admin/model/user/user.php
@@ -132,12 +132,54 @@ class User extends \Opencart\System\Engine\Model {
 	 * @return array
 	 */
 	public function getUsers(array $data = []): array {
-		$sql = "SELECT * FROM `" . DB_PREFIX . "user`";
+		$sql = "SELECT *, CONCAT(`u`.`firstname`, ' ', `u`.`lastname`) AS `name`, (SELECT `ug`.`name` FROM `" . DB_PREFIX . "user_group` `ug` WHERE `ug`.`user_group_id` = `u`.`user_group_id`) AS user_group FROM `" . DB_PREFIX . "user` `u`";
+
+		$implode = [];
+
+		if (!empty($data['filter_username'])) {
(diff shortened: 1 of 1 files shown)


upload/admin/model/user/user.php
```diff
@@ -175,7 +217,47 @@ public function getUsers(array $data = []): array {
 	 * @return int
 	 */
 	public function getTotalUsers(): int {
-		$query = $this->db->query("SELECT COUNT(*) AS `total` FROM `" . DB_PREFIX . "user`");
+		$sql = "SELECT COUNT(*) AS `total` FROM `" . DB_PREFIX . "user` `u` ";
+
+		$implode = [];
+
+		if (!empty($data['filter_username'])) {
+			$implode[] = "LCASE(`u`.`username`) LIKE '" . $this->db->escape(oc_strtolower($data['filter_username']) . '%') . "'";
```
AJenbo, line 225: This function does not have a $data variable so you need to add it, maybe via a function argument like it is for `getUsers()`.

```php
public function getTotalUsers(array $data = []): int {
```

pemaier:
@AJenbo Thank you for the hint. Pls. be so kind and add it ... as I think you have the rights to do. Thank you for your support.

AJenbo:
@pemaier for the record you as the PR author also have the right to change files in it.

image

I also noticed that you appear to be creating multiple PRs for a single feature with just one file changed in each PR. It would be much better if you could change all the necessary files in the same PR as they can then be reviewed, tested and merged together.

Normally you would do the changes locally on your computer and then push them to a branch on your fork, but it appears that you mainly do changes directly in the browser. It's still possible to edit multiple files in a single PR from the web interface, I think the easiest might be to go to your initial PR and then press `.` this will open the web version of Visual Code.

image

Here you can go to the file manager and navigate and edit the files you need, then go to the Source Control section and to stage, commit and push your changes.

pemaier:
@AJenbo Thank you :+1: :))

PR #13352, opened 2024-01-01
Update affiliate.twig - added filter reset

pemaier:


danielkerr:
what might have been better is add form to the foilter and then add a real reset button instead of doing it all in js

pemaier:
... something learned again. Thank you.

PR #13365, opened 2024-01-02
Add JWT to OpenCart

ocdevwizard:
I suggest using **JWT** (JSON Web Token) instead of the existing obsolete system, which relies on passing GET parameters like 'token' in previous versions and 'user_token' or 'customer_token' in newer versions of OpenCart. In essence, we eliminate the need to transmit the token in the URL, which is actually stored in the session. This, in turn, regulates the authentication session time for users and avoids the separation of the authentication session time between authorized admins and customers.

Implementing **JWT** brings several improvements, including:
- we can specify the session time for users who authenticate through '/admin' or '/login'.
- there is no longer a need to pass the token as a GET parameter in the URL.
- in the future, it will be possible to specify an individual session duration for each administrator after authorization. 
- additionally, a separate page can be created with a list of active tokens and the ability to destroy them.
- JWT tokens are stored in cookies with settings that provide a certain level of protection for cookies: 'secure' => true, 'httponly' => true, 'SameSite' => 'Strict'.
- JWT typically uses a secret key to sign the token. I suggest obtaining it from $this->config->get('config_encryption') or creating a separate key generated during the installation of OpenCart, similar to how it works for config_encryption.


TheCartpenter:
I am not sure why vendors proposals keeps arising while our local browser can already generate and compare the last token. The proposed code change originates from a vendor which means the algorithm used to assign the ```user_token``` and ```customer_token``` would not only be the same using different strings but also shows that other platforms could also adapt to the same algorithm by using the same vendors' package...

TheCartpenter:
In addition, this PR also indicates conflicts.

upload/system/storage/vendor/jwt/composer.json
```diff
@@ -0,0 +1,42 @@
+{
+    "name": "firebase/php-jwt",
```
AJenbo, line 2: please update composer.json and composer.lock in the project root

julioccguimaraes:
You have to think that this breaks compatibility with existing extensions. Is this the way we want to popularize OpenCart4?

upload/system/storage/vendor/jwt/composer.json
ocdevwizard: fixed it.

TheCartpenter:
> You have to think that this breaks compatibility with existing extensions. Is this the way we want to popularize OpenCart4?

I hope not...

composer.lock
```diff
@@ -1132,7 +1132,50 @@
                 }
             ],
             "time": "2023-08-28T11:09:02+00:00"
-        }
+        },
+	    {
+		    "name": "firebase/php-jwt",
+		    "version": "v6.10.0",
```
AJenbo, line 1138: Please also commit the changes to composer.json or the package will be remove next time a full dependency update is made.

ocdevwizard: sorry, forgot for this.
changed composer.json

danielkerr:
breaks compatibility but i also don't think its going to be secure.

danielkerr:
also this is just another 3rd party solution that might be replaced in another year.

mhcwebdesign:
What issue exactly are you actually trying to resolve here? And what about 3rd party extensions using user_token ?

Basically, as long as the cookies use [HttpOnly](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) and [Secure](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) we got sufficient security for the session  over a HTTPS connection. JWT tokens also tend to have issues with regards to revocation, needing more space, and stale.

ocdevwizard:
> @ocdevwizard : What issue exactly are you actually trying to resolve here? And what about 3rd party extensions using user_token ?
> 
> Basically, as long as the cookies use [HttpOnly](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) and [Secure](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) we got sufficient security for the session over a HTTPS connection. JWT tokens also tend to have issues with regards to revocation, needing more space, and being stale.

I trying to add some user friendly things. User_token is a rudiment. For example, let's consider a situation where there are two authorized administrators in the system. One of them shares a link to a product (https://.../admin/index.php?route=catalog/product.form&user_token=12801e0e900e065903015debbdbdc05b&product_id=42) so that another administrator can make changes to this product. For obvious reasons, when the other administrator open this URL, he redirected to the login page. If JWT is used, and assuming the token is valid, such a problem would not occur. 

Regarding compatibility with other modules, the same situation would arise for them during the transition from Opencart 3.x to 4.x.

stalker780:
@ocdevwizard the same always happen when you have several admin tabs open in one browser.
You need to close and reopen all the tabs all the time if the session in one tab have ended. 

Or you accidently opened a new admin tab and logged in. All other admin tabs become invalid.

You cannot use browser favorites for admin links.

ocdevwizard:
> @ocdevwizard the same always happen when you have several admin tabs open in one browser. You need to close and reopen all the tabs all the time if the session in one tab have ended.
> 
> Or you accidently opened a new admin tab and logged in. All other admin tabs become invalid.
> 
> You cannot use browser favorites for admin links.

Yes, exactly. I am, my clients, and may be thousands of users, years by years suffering with this thing.

mhcwebdesign:
Sharing an admin link to do some changes from another user's browser is a bad idea, and the other user should be redirected to a login prompt first, which is exactly what OpenCart currently does.

ocdevwizard:
> Sharing an admin link to do some changes from another user's browser is a bad idea, and the other user should be redirected to a login prompt first, which is exactly what OpenCart currently does.

If we set aside the goal of making store management more "user-friendly," then who exactly should Opencart work for? 
Opencart is intended for store administrators who may be casual users without deep technical skills. The main goal of any CMS is to facilitate the process of store management and provide ease of use for a wide range of users. I rely on my years-long experience, along with feedback from my clients who have expressed dissatisfaction with such behavior in the CMS, as illustrated in the example I provided earlier, and my voice is not lonely in this regard.

In Opencart's case, if store administrators find the system difficult to use, it can cause dissatisfaction and negatively impact productivity. The software should be intuitive and easy to use so that store administrators can effectively manage their products, orders and other aspects of the business.

Taking into account feedback from administrators and users, Opencart can develop in the direction of improving the interface and functionality, making it more intuitive and user-friendly.

danielkerr:
> > @ocdevwizard : What issue exactly are you actually trying to resolve here? And what about 3rd party extensions using user_token ?
> > Basically, as long as the cookies use [HttpOnly](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) and [Secure](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) we got sufficient security for the session over a HTTPS connection. JWT tokens also tend to have issues with regards to revocation, needing more space, and being stale.
> 
> I trying to add some user friendly things. User_token is a rudiment. For example, let's consider a situation where there are two authorized administrators in the system. One of them shares a link to a product (https://.../admin/index.php?route=catalog/product.form&user_token=12801e0e900e065903015debbdbdc05b&product_id=42) so that another administrator can make changes to this product. For obvious reasons, when the other administrator open this URL, he redirected to the login page. If JWT is used, and assuming the token is valid, such a problem would not occur.
> 
> Regarding compatibility with other modules, the same situation would arise for them during the transition from Opencart 3.x to 4.x.

its not user friendly! the users are not the developers.

ocdevwizard:
> > > @ocdevwizard : What issue exactly are you actually trying to resolve here? And what about 3rd party extensions using user_token ?
> > > Basically, as long as the cookies use [HttpOnly](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) and [Secure](https://en.wikipedia.org/wiki/HTTP_cookie#Secure_and_HttpOnly) we got sufficient security for the session over a HTTPS connection. JWT tokens also tend to have issues with regards to revocation, needing more space, and being stale.
> > 
> > 
> > I trying to add some user friendly things. User_token is a rudiment. For example, let's consider a situation where there are two authorized administrators in the system. One of them shares a link to a product (https://.../admin/index.php?route=catalog/product.form&user_token=12801e0e900e065903015debbdbdc05b&product_id=42) so that another administrator can make changes to this product. For obvious reasons, when the other administrator open this URL, he redirected to the login page. If JWT is used, and assuming the token is valid, such a problem would not occur.
> > Regarding compatibility with other modules, the same situation would arise for them during the transition from Opencart 3.x to 4.x.
> 
> its not user friendly! the users are not the developers.

I cannot agree with this statement for two reasons:

- forcing the user to reauthenticate when navigating to the URL https://.../admin/index.php?route=catalog/product.form&user_token=12801e0e900e065903015debbdbdc05b&product_id=42 does not contribute positively to the administrator's user experience. There are alternative solutions that can avoid this.

- other CMS and frameworks successfully operate without using a similar methodology, such as including the "user_token" as a GET parameter in the URL.

ocdevwizard:
In a real-life scenario with an Opencart store where several managers (3-5 or more) are working, handling content, products, and order management, I frequently encounter similar questions. For instance, managers find it inconvenient to share links with each other for quickly updating content or verifying certain data. So, how can we assist them? We can advise them to use built-in filters to locate what they're searching for or provide them with the entire sentence on how to open what they need. Alternatively, they could simply copy the URL and send it to the manager. Which method is faster? ))

mhcwebdesign:
`Which method is faster?` Wrong question! It should be: Which method is more secure? OpenCart already allows for several admin backend users, with different user groups and access/modify rights as needed.

ocdevwizard:
> `Which method is faster?` Wrong question! It should be: Which method is more secure? OpenCart already allows for several admin backend users, with different user groups and access/modify rights as needed.

can you answer why you think that using user_token instead of JWT will be more securely?

"OpenCart already allows for several admin backend users, with different user groups and access/modify rights as needed." - JWT is not about "user groups and access/modify rights". Access/modify rights this is not what I talking about.

stalker780:
@mhcwebdesign security has nothing to do with the location of token.
There are lots of secure systems which do not use user_token inside url.

We are talking that links with token are not handy for users. I get a lot of complaints from every new OC customer about htme for decade. People don't understand why they have to close and reopen all their tabs if they logged in one tab.

Almost all OC users keep opened many admin tabs in their browsers.
My life hack is to set session life time to a month or two. So token changes not so often. And I did it to all my websites since OC1.4. 
Otherwise user need to reopen all their admin tabs several times a day and they are very nervous when they lose their data or need to waste their time to log in and reopen tabs.

ocdevwizard:
> My life hack is to set session life time to a month or two.

I did the same(((

stalker780:
That is why would be more convenient to store token in cookies. It will be global for all open tabs in one browser.
Once one of the browser tabs token is changed, other tabs will use new token. And this will not require any actions from a user.

danielkerr:
It sounds like u are leaving the admin open to cross site scripting vulnerability.

stalker780:
@danielkerr Otherwise customers refuse to use OC and say we will use another system, which does not f*ck us up every hour with a new login to admin panel. They are yelling to developers "Do something! We cannot work!".

You will be surprised how many people suffer of this url token issue for years.

This was discussed here and in oc forum many times.

ocdevwizard:
> It sounds like u are leaving the admin open to cross site scripting vulnerability.

I have a penetration test report from a Swiss company that conducted a cybersecurity check on my client's Opencart store. Although they did not specifically mention JWT, they did make reference to something related to "user_token." According to report they said - "Session identifiers ought to be transferred via the browser's Cookie header with the HttpOnly and Secure option enabled."

stalker780:
Yep. This can be easily done without using any 3rd party libraries. And even backwards compatible. So that even old extensions which add user_token in url will work just ignoring them.

stalker780:
Just several examples
https://forum.opencart.com/viewtopic.php?t=22436 - year 2010 :)
https://forum.opencart.com/viewtopic.php?t=204066
https://forum.opencart.com/viewtopic.php?t=226804
https://forum.opencart.com/viewtopic.php?t=227672

TheCartpenter:
Extending the session lifetime to such extent won't help the case as it would become even less secured for administrators. Let's not forget that, whether it's a customer or user logging in the account, the same session libraries handles those events - the same for the APIs in OC v3.x releases with backwards compatibility.

In a real life scenario, as stated from the above, in a workplace, on web platforms, especially through the clouds, the session lifetime expiry value is normally set from: '10-15 minutes' before, whether it's a new user, junior or senior user, needs to log back in. I have seen on rare occasions where 30 minutes delay has been set on older platforms but those are uncommon.

OpenCart offers 1 hour interval from the system/library/cart/cart before products gets automatically removed from the cart. That's right - an hour by default which means an administrator would be suggested to leave the session idle time even longer than the customer. 

Did it helped the case where the ```setcookie``` parameters needed some re-work over the recent years? No, it didn't as it was necessary to be compliant with the users and customers as opposed to previous OC versions. Yet, people are still able to login just fine. Did that interfered with the ```user_token``` as per the forum topics above? Did the scenario really changed? No, it hasn't.

Today, we're now being introduced with a proposal to add a third-party vendor to initiate our tokens while OpenCart already has a built-in algorithm for that and also being maintained. Should it require more work on it, also in regard to permissions, it can still be handled from the core, especially since the ```system/config``` profiles were built-in over the years. since permissions are handled from the startup folder while the ```user_token``` originates from the common folder. Which means, it does not prevent access permissions to take place as it is loaded before the common folder does.

Therefore, is it really beneficial for OpenCart to integrate a third-party vendor to maintain our access and security accounts based on multiple tabs being opened, feedback / complaints received by customers in regard to session length or session protection? I doubt it.

TheCartpenter:
As for storing a session cookie, the ```api/account/login``` already takes care of that. Storing another cookie internally to login wouldn't be quite ... encouraged to use.

stalker780:
@TheCartpenter 
We are all talking about different things.

I don't care much about tokens in reviews/comments/uploads. These urls are only for ajax calls under the hood, and customers do not interfere with them.

I am talking only about removing admin user_token from url. In any safe and convenient for users way. 
Because it causes multitab browser session issues, which means in all modern browsers. 
Same for copy/paste issue for admin urls in emails and messengers.
Same for browser favorite links.
Logging in as a user in one tab destroys all the sessions in other tabs in same browser. That's the problem.

Just remove user_token from url and store it globally in browser cookies.
This does not change the logic of authorization at all. 
An only change is that user_token will be taken not from GET parameter, but from COOKIE.
Both cookies and get parameters are public for everyone. So this does not change anything except stupid logouts of multiple tabs.
You still will have to login if your cookie user_token is outdated. But this will login also all other tabs you have open in your browser. As if logging into one tab all other tabs changed their urls to new tokens.

You know a better way?

How WordPress admin panel works without user_token in url?
How hundreds of online systems work safely without user_token in their urls?

No need to integrate 3rd party libraries for this. 

I don't like a solution in this PR either. But it triggered an old issue.

TheCartpenter:
As my previous reply indicates, it does highlight ```user_token``` multiple times.

TheCartpenter:
> @TheCartpenter We are all talking about different things.
> 
> I don't care much about tokens in reviews/comments/uploads. These urls are only for ajax calls under the hood, and customers do not interfere with them.
> 
> I am talking only about removing admin user_token from url. In any safe and convenient for users way. Because it causes multitab browser session issues, which means in all modern browsers. Same for copy/paste issue for admin urls in emails and messengers. Same for browser favorite links. Logging in as a user in one tab destroys all the sessions in other tabs in same browser. That's the problem.
> 
> Just remove user_token from url and store it globally in browser cookies. This does not change the logic of authorization at all. An only change is that user_token will be taken not from GET parameter, but from COOKIE. Both cookies and get parameters are public for everyone. So this does not change anything except stupid logouts of multiple tabs. You still will have to login if your cookie user_token is outdated. But this will login also all other tabs you have open in your browser. As if logging into one tab all other tabs changed their urls to new tokens.
> 
> You know a better way?
> 
> How WordPress admin panel works without user_token in url? How hundreds of online systems work safely without user_token in their urls?
> 
> No need to integrate 3rd party libraries for this.
> 
> I don't like a solution in this PR either. But it triggered an old issue.

As for WordPress, that's an easy answer - it is also the reason why OpenCart 'does' use a ```user_token``` in the URL.

TheCartpenter:
As for logging in with other tabs, that will also lead multiple users and encourage them to leave those tabs opened from multiple locations in the admin-end. I wouldn't see that as a robust way to secure our session lifetime either, since it will only deliver to the users they are allowed to forget rather than remember their tasks.

ADDCreative:
> > It sounds like u are leaving the admin open to cross site scripting vulnerability.
> 
> I have a penetration test report from a Swiss company that conducted a cybersecurity check on my client's Opencart store. Although they did not specifically mention JWT, they did make reference to something related to "user_token." According to report they said - "Session identifiers ought to be transferred via the browser's Cookie header with the HttpOnly and Secure option enabled."

@ocdevwizard The user_token is not a session identifier, it is more of a synchronizer token. It's main use is to protect against Cross-Site Request Forgery (CSRF), Moving the token into a cookie would be vulnerable to CSRF attacks.

AJenbo:
https://security.stackexchange.com/questions/115794/should-i-use-csrf-protection-for-get-requests

Short answer: no

mhcwebdesign:
> As for logging in with other tabs, that will also lead multiple users and encourage them to leave those tabs opened from multiple locations in the admin-end. I wouldn't see that as a robust way to secure our session lifetime either, since it will only deliver to the users they are allowed to forget rather than remember their tasks.

Maybe implement a modification to allow this multiple tabs scenario, and possibly multiple browsers scenario, only for users on the same local network, perhaps configurable in the System > Settings ?

TheCartpenter:
> > As for logging in with other tabs, that will also lead multiple users and encourage them to leave those tabs opened from multiple locations in the admin-end. I wouldn't see that as a robust way to secure our session lifetime either, since it will only deliver to the users they are allowed to forget rather than remember their tasks.
> 
> Maybe implement a modification to allow this multiple tabs scenario, and possibly multiple browsers scenario, only for users on the same local network, perhaps configurable in the System > Settings ?

That would become too complex to deal with. There's already enough with the SEO URLs from the Systems > Settings > Server tab with the ```.htaccess``` file that users need to deal with. We'd be dealing for another decade to run them back in Business to troubleshoot a multiple browsers scenario issue on the same local network.

ocdevwizard:
> The user_token is not a session identifier

It is incorrect to assume that the user_token value is not stored anywhere. The GET parameter user_token is compared with the user token value from your session. 

I meant that using user_token is a vestige and an outdated practice, which does not provide a good user experience when using OpenCart.

ADDCreative:
> You know a better way?
> 
> How WordPress admin panel works without user_token in url? How hundreds of online systems work safely without user_token in their urls?

@stalker780 WordPress puts its token in the POST (_wpnonce I believe). OpenCart could do the same. Any request that perform actions would need to be changed to a POST (many are using GET) and the token added to the POST data instead of the URL. This would also be more secure as a user can't be redirected to a POST and the token is less likely to be leaked via something like an open redirect.

ADDCreative:
> https://security.stackexchange.com/questions/115794/should-i-use-csrf-protection-for-get-requests
> 
> Short answer: no

@AJenbo Long answer: Yes, if the GET request can change a state on the server, which applies to OpenCart. It would be nice if OpenCart followed RFC 7231 and not allow safe methods such as GET to change the state on the server, but that isn't the case. https://www.rfc-editor.org/rfc/rfc7231#section-4.2.1

danielkerr:
> > The user_token is not a session identifier
> 
> It is incorrect to assume that the user_token value is not stored anywhere. The GET parameter user_token is compared with the user token value from your session.
> 
> I meant that using user_token is a vestige and an outdated practice, which does not provide a good user experience when using OpenCart.

thats ur opion but u haveno security against corsss site scripting. u could litterly send an email with a link in that if u are logged into the admin as support ticket operators might, and have js create a new admin user.

to me this sounds like u are really a novice or ur trying to gaslight me.

ocdevwizard:
> > > The user_token is not a session identifier
> > 
> > 
> > It is incorrect to assume that the user_token value is not stored anywhere. The GET parameter user_token is compared with the user token value from your session.
> > I meant that using user_token is a vestige and an outdated practice, which does not provide a good user experience when using OpenCart.
> 
> thats ur opion but u haveno security against corsss site scripting. u could litterly send an email with a link in that if u are logged into the admin as support ticket operators might, and have js create a new admin user.
> 
> to me this sounds like u are really a novice or ur trying to gaslight me.

If the absence of using a user_token as a GET parameter in the URL appears suspicious to you, how would you justify the fact that other CMS or frameworks, such as Django, do not employ such a method? They are try to be trick or what?

ADDCreative:
> > The user_token is not a session identifier
> 
> It is incorrect to assume that the user_token value is not stored anywhere. The GET parameter user_token is compared with the user token value from your session.
> 
> I meant that using user_token is a vestige and an outdated practice, which does not provide a good user experience when using OpenCart.

@ocdevwizard Using a CSRF token is not an outdated practice. It could probably just be implemented in a more user friendly way, such as putting it in the POST data and not the URL and also handling failure instead on logging out.

Your change seems to put the token in in a cookie. If the web browser sends the session cookie then it will also send your token cookie, so will offer no protection against CSRF.

TheCartpenter:
The only advantage by using POST would be with JS integration whenever the submit button from the HTML forms have been triggered so that users won't be able to refresh the page countless times to create duplicate entries on the DB. Other platforms do use these as well along with their CSRF protection.

AJenbo:
> > https://security.stackexchange.com/questions/115794/should-i-use-csrf-protection-for-get-requests
> > Short answer: no
> 
> @AJenbo Long answer: Yes, if the GET request can change a state on the server, which applies to OpenCart. It would be nice if OpenCart followed RFC 7231 and not allow safe methods such as GET to change the state on the server, but that isn't the case. https://www.rfc-editor.org/rfc/rfc7231#section-4.2.1

The more I learn about this system the more shocked I become

danielkerr:
> The more I learn about this system the more shocked I become

which system? opencart?

danielkerr:
> The only advantage by using POST would be with JS integration whenever the submit button from the HTML forms have been triggered so that users won't be able to refresh the page countless times to create duplicate entries on the DB. Other platforms do use these as well along with their CSRF protection.
> 
> However, it may still require the page to be redirected while OC v4.x releases uses JSON to submit form entries.

js will never protect u from refreshing the page 10900's of times as it can be disabled.

PR #13371, opened 2024-01-02
Correct various generic object types to there actual native types

AJenbo:


danielkerr:
does this not restrict interoperability?

AJenbo:
Not as long as they extend the class. It might be better to instead provide interfaces to assure that modifications for fill the expectations without needing to overwrite the parent.

danielkerr:
I don't like interfaces. Must be a better way to describe a class

AJenbo:
That's like saying you don't like legs for walking. I think the only other option is abstract classes. But one does not preclude the other.

danielkerr:
do more testing next time!

AJenbo:
I would but the PHPUnit setup was deleted around the time of OC4 so I will have to set something up from scratch at some point.

PR #13373, opened 2024-01-02
Fix type error when getting unitialized image

AJenbo:
Filename can be false (on error) or null (when uninitialized), simply return null in that case instead of throwing a type error.

This was previously improperly fixed here: https://github.com/opencart/opencart/commit/da308c65667e52583e0a257ba2c9410b6dd8e417

### DIFF
--- a/upload/system/library/image.php
+++ b/upload/system/library/image.php
@@ -88,8 +88,8 @@ class Image {
 	 *
 	 * @return ?\GdImage
 	 */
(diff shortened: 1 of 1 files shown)


AJenbo:
Hm this was in draft since I needed to make an adjustment to fix the issue with PHP 7.4 before it was ready to be merged.

I'll make a follow up to correct it I guess.

PR #13375, opened 2024-01-02
Check for $filename initialization before utilization

AJenbo:
Check that `$filename` exists before we try to use it.


danielkerr:
does not fix anything. not error

danielkerr:
if (empty($this->request->files['file']['name']) || !is_file($this->request->files['file']['tmp_name'])) {
			$json['error'] = $this->language->get('error_upload');
		}

is already there

PR #13378, opened 2024-01-02
Remove unused arguments

AJenbo:
The SSL arguments are never actually used inside the function, removing them doesn't affect comparability as PHP will silently ignore additional augments in invocations.

Intorudced here: https://github.com/opencart/opencart/commit/671d16272e397471e65f1d6a3748197d1953d3bd


danielkerr:
i think its supposed to match the mysqli one

PR #13383, opened 2024-01-03
Update extension.php

Samdz19:
delete module settings on uninstall


stalker780:
I think this should be optional.

On extension update you may want to uninstall old version files but to leave modules settings.

julioccguimaraes:
We need extension update support

PR #13385, opened 2024-01-03
[3.0.x.x] Create .php-cs-fixer.php

condor2:


mhcwebdesign:
What is the purpose of this? Do we really want cosmetic radical reformatting of PHP code for the 3.0.x.x? It would break an awful lot of existing OCmod-based extensions. Perhaps would be more useful for the opencart-3 repository? Or we could start a new 3.1.x.x branch for this?

TheCartpenter:
> What is the purpose of this? Do we really want loads of cosmetic radical reformatting of PHP code for the 3.0.x.x branch? It would break an awful lot of existing OCmod-based extensions. Perhaps it would be more useful for the opencart-3 repository? Or we could start a new 3.1.x.x branch for this?

3.1 is already published on opencart.com and 3.2 also is in production.

condor2:
> What is the purpose of this? 

To check for issues in the code.

Commands who make "cosmetics" ...can be disabled after

AJenbo:
While php-cs-fixer can fix a few issues and update the code to take advantage of modern features, it mostly deals with code style (hens the cs in it's name).

It would probably be best to adjust the rules so that they fit what is already there before merging them to avoid things being applied by mistake or miss understood.

Maybe there are some style changes they are acceptable still, but I think that would be best evaluated on a case by case basis by the regular maintainers of 3.0.

danielkerr:
can we start fixing open issues?

AJenbo:
@condor2 I have changed the rules to a more minimal set, but it appears you based this on an older branch so it needs to be updated before I can enable the rules in the CS so we can see how they fair against he current code base.
Can you please take care of this?

Also the current outstanding issues reported by PHPStan should be addressed first as it won't progress to the code style check when there are more severe issues found (like wise it won't run the PHPStan stage if some of the files are found to contain invalid PHP syntax).

TheCartpenter:
Not so cosmetic after all on the 3.2. I seem to have already taken care of all of those long before the ```.php-cs-fixer``` got integrated on the opencart-3 repository.

condor2:
@AJenbo Not understand what to do. I have updated my repo's

AJenbo:
> @AJenbo Not understand what to do. I have updated my repo's

you need to either rebase the patch-1 branch on the 3.0.x.x branch, or merge the 3.0.x.x branch in to the patch-1 branch. I think you would need to clone the repo to your computer and use a git tool to do so, so if you prefer to just do things from the webinterface then probably the easiest thing to do is to close this one and create a new one so that it's based on an up to date 3.0.x.x branch.

If you don't already have a clone with upstream as the remote you could do the following (assuming you have setup a ssh key with github) using the git cli tool:
```
git clone git@github.com:condor2/opencart.git
git remote add upstream https://github.com/opencart/opencart.git
git fetch --all
git checkout patch-1
git rebase upstream/master
git push -f
```

If you let me know what tools you normally use and how you have things setup then maybe I can better help you with simpler directions that match what you are comfortable with already.

PR #13386, opened 2024-01-03
[3.0.x.x] Create phpstan.neon

condor2:


mhcwebdesign:
@AJenbo: What is this for? IMHO it won't make sense to do tons of cosmetic formatting and/or doc changes to the 3.0.x.x branch, that would clash with loads of 3rd party extensions.

condor2:
@mhcwebdesign To maintain as clean as possible, cosmetic/doc commands can be disabled...

AJenbo:
@mhcwebdesign [It's for finding bugs](https://phpstan.org/). It don't have anything to do with cosmetic changes.

It's how this was found and fixed: https://github.com/opencart/opencart/pull/13384 (copied from my original here https://github.com/condor2/Opencart_23xx/pull/24) and you seemed happy with that one.

I don't expect that there will be a ton of changes from it either, it's only listing 56 potential issue at the moment.

I'm sad that you seam to have gotten a negative view of what I do ever since I took the time to clean up code formatting with a tool based on your suggestion, I only did so because I agreed with you that the way things where being done at the time was creating a lot of unnecessary churn. I hope we can get past this now that it has been finished.

I don't have any intent to apply code formatting to the 3.0 branch if that is what you are worried about.

mhcwebdesign:
@Abenjo: Thank you for your explanation, I was just trying to understand what these tools would do. Your hard work is much appreciated for the OpenCart project. 

Many users, probably the majority, still use OpenCart 3, not OpenCart 4, because the latter is still subject to many bugfixes and development in general, hence the importance to maintain 3.0.x.x carefully, without causing clashes with existing 3rd party extensions.

AJenbo:
@mhcwebdesign ok glad to hear that. Yeah I only see OC3 as being in maintenance mode, it should receive bug fixes, but no improvements.

PR #13388, opened 2024-01-03
[3.0.x.x] .php-cs-fixer.php

condor2:
https://github.com/opencart/opencart/pull/13385/commits/5deb6ed7a9606287b03e4734c44c9f6c2b9bd9b3

### DIFF
--- a/.github/workflows/Lint.yml
+++ b/.github/workflows/Lint.yml
@@ -26,6 +26,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           tools: composer:v2, cs2pr
+          extensions: mbstring
 
       - name: Lint
         run: |
--- a/.github/workflows/Lint.yml
+++ b/.github/workflows/Lint.yml
@@ -25,7 +25,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          tools: composer:v2
+          tools: composer:v2, cs2pr
 
       - name: Lint
         run: |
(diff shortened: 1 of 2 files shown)


condor2:
More fast :))

AJenbo:
This will now pass the correct code, but I also noticed now that it is likely scanning the vendor folder so maybe we didn't have to disable as many rules as i did.

condor2:
> This will now pass the correct code, but I also noticed now that it is likely scanning the vendor folder.

@AJenbo Maybe because there is no exclude folder? At the end of php-cs fixer

mhcwebdesign:
These tools should not scan the vendor folder. We have to be careful with the latter, the 3rd party vendor libraries all have to support the PHP versions 7.4 to 8.3.

AJenbo:
> > This will now pass the correct code, but I also noticed now that it is likely scanning the vendor folder.
> 
> @AJenbo Maybe because there is no exclude folder? At the end of php-cs fixer

Yeah, it just got to late for me to work on it any further.

> These tools should not scan the vendor folder. We have to be careful with the latter, the 3rd party vendor libraries all have to support the PHP versions 7.4 to 8.3.

Right, thought the compat is set to 7.4 so it shouldn't cause any issues for older versions, but we should not be trying to style them to OC uses, it's a waist of time and would probably reduce the number of rules that can be applied.

AJenbo:
Ok, looks like it wasn't causing much of an issue and the initial run with the larger rule set was done properly so the rules I disabled where valid. I fixed the config and this should now work correctly.

AJenbo:
@condor2 I have enabled 117 rules that the code already abides by, that leaves 93 that you can evaluate and see which to leave out and which would be ok to adjust to for the 3.0 branch.

condor2:
@AJenbo I am not the boss for 3.0.x.x :)

AJenbo:
Ok, my guess it that it will just stay like this then :)

Do you intent to solve the 56 issues found by PHPStan?

condor2:
@AJenbo I made a test on a repo with some commands disabled after I have ignored some files.....
And this I have get...maybe for  3.0.x.x maintainers looks cosmetics....

https://github.com/condor2/3.0.x.x-testing-only/commit/8a964a38c5bab419cf94afc6970b0cd6eff35191

AJenbo:
Ok, You will need to create a PR and have the regular 3.0 maintainer evaluate this since I'm not well versed enough in how it works to evaluate if it's breaking or an acceptable level of change.

PR #13389, opened 2024-01-04
Updated phpstan & php-cs-fixer versions in Lint.yml

condor2:


AJenbo:
Thanks, I wanted to make this update as they fixed the alignment of comment indentations. I also updated the wiki and adjusted the new issues that it found.

Issue #9237, opened 2021-02-06
Session Error When Setting Timezone


bustin98:
**What version of OpenCart are you reporting this for?**
OC 3.0.3.7

**Describe the bug**

Unable to log into the dashboard after changing the timezone from the default of none.

**To Reproduce**

System -> Settings -> Store -> Local -> Time Zone

I set the time zone to America/Detroit (-5:00) and applied the changes. The session ended and was unable to log in.

**Expected behavior**
To be returned to the dashboard with changes applied.

**Screenshots / Screen recordings**
![settings-screen](https://user-images.githubusercontent.com/14907248/107122679-1fd67280-6867-11eb-96b0-1139ac4a80de.jpg)

**Server / Test environment (please complete the following information):**

Local development? Deployed to a web server? - Production server
Operating system - Win2012R2
PHP version - 7.4
Web Server version - IIS8
Browser(s) tested with [e.g. chrome, safari - if applicable] - Chrome

**Extra Notes**

I was able to get out of the situation by changing the session engine from db to file in the system/config/default.php and system/config/catalog.php files.

danielkerr:
it should be something to do with your settings for the server time zone

eMintis:
Working on
Debian 10;
Nginx;
PHP 7.4;
MariaDB 10.5.

bustin98:
I checked my settings in MySQL and I do not have a timezone specified in the configuration. Could that play into how the expiration of sessions is set?

lionsgatecreative:
I'm having the same issue with same version of OC.

TheCartpenter:
According to this forum user: https://forum.opencart.com/viewtopic.php?f=201&t=222369&p=815432#p815421 , the issue seem to have been resolved by applying this solution: https://forum.opencart.com/viewtopic.php?f=201&t=222369&p=815432#p813324 .

webkul-dilip:
> According to this forum user: https://forum.opencart.com/viewtopic.php?f=201&t=222369&p=815432#p815421 , the issue seem to have been resolved by applying this solution: https://forum.opencart.com/viewtopic.php?f=201&t=222369&p=815432#p813324 

@lionsgatecreative @bustin98 Did you still face the same issue?

WebkulOpencart:
I just reinstalled the latest 3.0.3.7 version but not found such an issue so closing this issue

ADDCreative:
[...]

Issue #9492, opened 2021-03-31
Warning: You do not have permission to access the API! YES AGAIN NOT A DUPLICATE


hounw:
**What version of OpenCart are you reporting this for?**

3.0.3.7

**Describe the bug**
Unable to edit orders due to api permissions

**To Reproduce**

Installed on xampp local server, tried on macos and windows.

First, changed the timezone -> unable to login...

tried setting timezone in php.ini and my.cnf -> still no login

tried alternative setting `$_['session_engine'] = 'file'; `

SUCCESS able to login... but now it's impossible to edit an order... 

tried a solution on another issue

https://github.com/opencart/opencart/issues/6783

commeting out the three lines suggested here https://github.com/opencart/opencart/issues/6783#issuecomment-412443340

Doesn't work either.

**Expected behavior**

Being able to login to admin and editing orders

**Server / Test environment (please complete the following information):**
 - Local development xampp on macos with PHP Version 7.3.18/Xampp on windows php 7.4.14

**Additional context**

We tried the ocmod here https://www.opencart.com/index.php?route=marketplace/extension/info&extension_id=30437&filter_search=api%20error 

and it works, but it completely removes a check which seems like a potential security issue (but I haven't dug in enough to know):

`	<file path="catalog/controller/api/*.php">	
		<operation error="skip">
			<search><![CDATA[if (!isset($this->session->data['api_id'])) {]]></search>
			<add position="replace"><![CDATA[
			
			if (isset($this->session->data['me_loves_taco'])) {
			
			]]></add>
		</operation>		
	</file>`

ADDCreative:
Did you set ```$_['session_engine'] = 'file';``` in both.
https://github.com/opencart/opencart/blob/2a5b358003d1e9e4af964eda2ffcb79abe8229ce/upload/system/config/default.php#L43
https://github.com/opencart/opencart/blob/2a5b358003d1e9e4af964eda2ffcb79abe8229ce/upload/system/config/catalog.php#L20

Webkul-Opencart:
@hounw Not found this issue.

hounw:
@ADDCreative as I tried to explain, yes, after changing the time zone in settings, and then setting it in php and mysql, I couldn't log in, the only way I could log in was by changing to 'file' on both files you suggest.

After that however I can no longer edit orders or add history to them (or anything else that requires the api I suspect...). When I searched the error the solutions involved this configs too so I think it's somehow related.

[...]

PR #13390, opened 2024-01-04
Fix for timezone change breaking session expiration

ADDCreative:
This is the same as https://github.com/opencart/opencart/pull/9633 that was never applied to the 3.0.x.x branch. Should fix https://github.com/opencart/opencart/issues/9237 that prenvented admin login when the timezone was a negative value. However, it won't fix the issue a negative value causes with the API https://github.com/opencart/opencart/issues/9492. That would require a change in the API session code or a change in the startup order.

### DIFF
--- a/upload/system/library/session/db.php
+++ b/upload/system/library/session/db.php
@@ -14,7 +14,7 @@ final class DB {
 	}
 
 	public function read($session_id) {
-		$query = $this->db->query("SELECT `data` FROM `" . DB_PREFIX . "session` WHERE `session_id` = '" . $this->db->escape($session_id) . "' AND `expire` > '" . $this->db->escape(date('Y-m-d H:i:s', time())) . "'");
+		$query = $this->db->query("SELECT `data` FROM `" . DB_PREFIX . "session` WHERE `session_id` = '" . $this->db->escape($session_id) . "' AND `expire` > '" . $this->db->escape(gmdate('Y-m-d H:i:s', time())) . "'");
 
 		if ($query->num_rows) {
 			return json_decode($query->row['data'], true);
@@ -25,7 +25,7 @@ final class DB {
 
 	public function write($session_id, $data) {
 		if ($session_id) {
-			$this->db->query("REPLACE INTO `" . DB_PREFIX . "session` SET `session_id` = '" . $this->db->escape($session_id) . "', `data` = '" . $this->db->escape(json_encode($data)) . "', `expire` = '" . $this->db->escape(date('Y-m-d H:i:s', time() + (int)$this->maxlifetime)) . "'");
+			$this->db->query("REPLACE INTO `" . DB_PREFIX . "session` SET `session_id` = '" . $this->db->escape($session_id) . "', `data` = '" . $this->db->escape(json_encode($data)) . "', `expire` = '" . $this->db->escape(gmdate('Y-m-d H:i:s', time() + (int)$this->maxlifetime)) . "'");
 		}
 
 		return true;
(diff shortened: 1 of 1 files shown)


danielkerr:
thx!

ADDCreative:
May I ask why this reverted?

mhcwebdesign:
@AddCreative : Just curious: How exactly is the timezone issue to be fixed? I just set up a new OC 3.0.x.x for the 'America/Denver' timezone (UTC-07), and the only way I managed to get it to work was by doing this:

1)
System > Settings > edit Store > tab Locals > Time Zone: America/Denver (-07:00)

2)
In the system/config/default.php replace
`$_['date_timezone']        = 'UTC';`
with
`$_['date_timezone']        = 'America/Denver';`

The latter is only used in the system/framework.php :
`date_default_timezone_set($config->get('date_timezone'));`

Do you have a suggestion how exactly to fix the timezone issue so it won't result in failed admin logins?

ADDCreative:
@mhcwebdesign This issue is caused as the session is started and the expiry time checked in framework.php, before the time zone is set from the value in the database in controller/startup/startup.php, but is written after the time zone is set. So the session expiry time is checked in one time zone and the written in another.

Using gmdate, which isn't affected by time zone changes, instead of date in the session class, will at least prevent the admin from being locked out.

Another way to fix all the issues it causes would be to change the startup order so the session is not started until the final time zone has been set.

This is what was eventually done in 4.0.x as well and using gmdate.
https://github.com/opencart/opencart/blob/456cc98f6e25be73ba99f43c3fcc73ffb7c2fe6c/upload/system/config/admin.php#L25-L27

Whereas 3.0.x does.
https://github.com/opencart/opencart/blob/8a3c27de562613bab5779647cc3d934bea586098/upload/system/config/catalog.php#L32-L34

It would be more work to change the startup order in 3.0.x as the admin session is started in framework.php not a seperate controller/startup/session.php. There may also be other consequences of changing the startup order.

mhcwebdesign:
@AddCreative : So the only way for OpenCart 3.0.x.x, for the admin login, would be to change the **system/framework.php**. E.g. if $application_config='admin' then get the 'config_timezone' from a DB query (table oc_setting.key) before dealing with the session, e.g.

```
....
if ($application_config=='admin') {
	$timezone =  ... get time zone from a DB query ...
	date_default_timezone_set($timezone);
}

// Session
$session = new Session($config->get('session_engine'), $registry);
.....

```

Just thinking aloud....

danielkerr:
please poost again

ADDCreative:
@mhcwebdesign I like the idea, However, it's also need for the catalog as well. So just.

Before.
https://github.com/opencart/opencart/blob/8a3c27de562613bab5779647cc3d934bea586098/upload/system/framework.php#L83-L84

Add something like.
```php
$query = $db->query("SELECT * FROM " . DB_PREFIX . "setting WHERE `key` = 'config_timezone' AND store_id = '0'");

// Set time zone
if ($query->num_rows) {
	date_default_timezone_set($query->row['value']);
}
```

Just tested and it seems to work. Even the API seems to work.

mhcwebdesign:
@ADDCreative : Good point. The basic idea was that the timezone should not be hardcoded in the **system/config/default.php**, rather the timezone should be found from the oc_settings DB table. So do we need the line 15 in the **system/framework.php** at all:

`date_default_timezone_set($config->get('date_timezone'));`

?

mhcwebdesign:
@ADDCreative : So if I understand you correctly all we need is to add

```
$query = $db->query("SELECT * FROM " . DB_PREFIX . "setting WHERE `key` = 'config_timezone' AND store_id = '0'");

// Set time zone
if ($query->num_rows) {
	date_default_timezone_set($query->row['value']);
}

```
before line 83 in the system/framework.php

?

ADDCreative:
@mhcwebdesign Yes, it effectively moves the following blocks of code before the session is started. So the following blocks of code could be removed, as there is no point setting the time zone again.

https://github.com/opencart/opencart/blob/8a3c27de562613bab5779647cc3d934bea586098/upload/admin/controller/startup/startup.php#L15-L21

https://github.com/opencart/opencart/blob/8a3c27de562613bab5779647cc3d934bea586098/upload/catalog/controller/startup/startup.php#L47-L53

PR #13393, opened 2024-01-04
Update address.php

Samdz19:
Add custom field status


danielkerr:
not sure its required. if u disable custom fields u wont be able to see what they contain

Samdz19:
but if don't add status we will got all custom fields displayed with any status when trying to add new address via admin

danielkerr:
Fine

PR #13405, opened 2024-01-05
Correct call to addModule() in extension module

AJenbo:
`addModule()` takes a data array as the second argument, not a title.
https://github.com/opencart/opencart/blob/3f42879ced0f65ac16b0cc80b3402cfc46246e93/upload/admin/model/setting/module.php#L15

This issue was introduced in https://github.com/opencart/opencart/commit/0ce03aab7e9c19457805b7f91483be70d6d09a29

### DIFF
--- a/upload/admin/controller/extension/module.php
+++ b/upload/admin/controller/extension/module.php
@@ -229,7 +229,7 @@ class Module extends \Opencart\System\Engine\Controller {
 
 			$this->load->model('setting/module');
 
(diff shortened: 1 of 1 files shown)


danielkerr:
kicking myself!

im shocked it still works

AJenbo:
I have a few more fun ones I have to look into. This one appears to have been there for a long time 😵
At least oc4 hast some basic type checks but it could be nice of more where structured and used protocols rather than duck typing.

PR #13408, opened 2024-01-05
Fix sending mail on PHP 7.4

AJenbo:
The version check was wrong, see comment below.

### DIFF
--- a/upload/system/library/mail/mail.php
+++ b/upload/system/library/mail/mail.php
@@ -29,7 +29,7 @@ class Mail {
 			$to = $this->option['to'];
 		}
 
-		if (version_compare(PHP_VERSION, '7.4', '>=') || substr(PHP_OS, 0, 3) == 'WIN') {
+		if (version_compare(PHP_VERSION, '8.0', '>=')) {
 			$eol = "\r\n";
 		} else {
 			$eol = PHP_EOL;


ADDCreative:
@AJenbo It's actually a bug. It should be >= 8.0. There was a change to the PHP mail function in PHP 8 in the way it handles line ending in the headers.

See the original fix https://github.com/opencart/opencart/pull/10540.

It seems to have been mistakenly changed to PHP 8.2 here.
https://github.com/opencart/opencart/commit/73962c499c770cc1d5a40ae2d4b6b33a6f974183#diff-fd55e1fbdbbbd13709246e79d1a0c9288a42c1bb90327d36c7ab55ca4c989876

Then to PHP 7.4 here.
https://github.com/opencart/opencart/commit/6440e653694b92da4c3fa7ead794bd915f6995ec#diff-fd55e1fbdbbbd13709246e79d1a0c9288a42c1bb90327d36c7ab55ca4c989876

AJenbo:
Thanks for clarifying that, I tried to find the PHP release not regarding this but couldn't find it, do you happen to remember where it was?

ADDCreative:
It's not well documented. I think this may have been the original bug report. https://bugs.php.net/bug.php?id=47983

The consequence of the fix was that now all headers sent to the PHP mail function need to end ```\r\n``` to match the headers set by PHP.

AJenbo:
Here is the fix, which confirms that the change applies for 8.0.0 så the fix shoulf be correct now 👍

https://github.com/php/php-src/commit/6983ae751cd301886c966b84367fc7aaa1273b2d

PR #13410, opened 2024-01-05
Normalize some unconventional calling conventions

AJenbo:
The somewhat unusual calling convention would only work as long as it's a constructed object with properties containing callback, if an object with actual methods where returned this would have resulted in a fatal error. Instead use the common calling convention.

Introduced in fcf2a8173a983d00c470824a4ae088afdb1931bf, 0a3f518cc3cca9208c69a1f237c5099ec9fe9e62, and 7f16b030abc39fa7763fddb098b876fa8ce5b670

### DIFF
--- a/upload/catalog/controller/api/sale/cart.php
+++ b/upload/catalog/controller/api/sale/cart.php
@@ -25,7 +25,7 @@ class Cart extends \Opencart\System\Engine\Controller {
 
 		$this->load->model('checkout/cart');
 
-		($this->model_checkout_cart->getTotals)($totals, $taxes, $total);
+		$this->model_checkout_cart->getTotals($totals, $taxes, $total);
 
 		$json['products'] = [];
 
--- a/upload/catalog/controller/api/sale/order.php
+++ b/upload/catalog/controller/api/sale/order.php
@@ -440,7 +440,7 @@ class Order extends \Opencart\System\Engine\Controller {
 
 			$this->load->model('checkout/cart');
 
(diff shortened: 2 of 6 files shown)


danielkerr:
this will cause the refences vars to be dropped!

this will canuse a loot of issues/

AJenbo:
@danielkerr do you mean these $totals, $taxes, $total? Could you give me a test case that shows the difference?

danielkerr:
Yes. Do u no about referencing in php? Using &$total. Well models use a magic method in the proxy class. Magic methods drop references so the totals can not accumulate.

AJenbo:
Yeah I know about passing variables by reference. I would suggest maybe using a DTO since objects are always by reference and would avoid such issues. But I can also take a look at changing the Proxy to use a Proxy/Decorator pattern rather then being a constructed objects with callable properties rather then actual methods, that would improve things over all I think, and likely also fix this issues (I'll have to test it a bit).

danielkerr:
I don't care it looks like that. It's a PHP issue. I'm just finding a way to close around it.

AJenbo:
The issue is that it disassociates it from the model which means I can't analyze the code any further at that point so I need it changed to locate further bugs in the system

danielkerr:
they are called anonymous functions

AJenbo:
Sure, or more generically Closure. But that doesn't really address the issue that there signatures is not documented anywhere in OC which means it's not possible to predict if they are used correctly or not, and instead they can only be reasoned about at runtime where they may fail.

PR #13413, opened 2024-01-05
Added new line for each null IDs in models

TheCartpenter:


danielkerr:
dont use null use = '';

danielkerr:
what about we change all the models to code more like:

`	public function addAttribute(array $data): int {
		$this->db->query("INSERT INTO `" . DB_PREFIX . "attribute` SET `attribute_group_id` = '" . (int)$data['attribute_group_id'] . "', `sort_order` = '" . (int)$data['sort_order'] . "'");

		$attribute_id = $this->db->getLastId();

		foreach ($data['attribute_description'] as $language_id => $value) {
			$this->addDescription(language_id , $value);
		}

		return $attribute_id;
	}`

TheCartpenter:
> what about we change all the models to code more like:
> 
> ```
> 	public function addAttribute(array $data): int {
> 		$this->db->query("INSERT INTO `" . DB_PREFIX . "attribute` SET `attribute_group_id` = '" . (int)$data['attribute_group_id'] . "', `sort_order` = '" . (int)$data['sort_order'] . "'");
> 
> 		$attribute_id = $this->db->getLastId();
> 
> 		foreach ($data['attribute_description'] as $language_id => $value) {
> 			$this->addDescription(language_id , $value);
> 		}
> 
> 		return $attribute_id;
> 	}
> ```

```language_id``` is missing ```$``` in ```addDescription``` . However, if we were to do that, the controllers are still calling their methods with full paths as opposed to the previous ways the methods used to be called. It would no longer be the same type of calls between the controllers and the models as their roles would, in fact, be reversed as opposed to previous releases.

TheCartpenter:
As for affecting ```null```, it was already merged previously for another user. I just added the spaces between the declaration and the function use underneath to match with OC's coding standards.

PR #13414, opened 2024-01-06
Fix loading subscription data

AJenbo:
`upload/admin/controller/sale/subscription.php` had an inverted case on line 791 meaning it never loaded the subscription data. This was introduced here: https://github.com/opencart/opencart/commit/e586908b1b17899d62133a13e65b95017e22e053

In `upload/catalog/controller/cron/subscription.php` the editDateNext() is called on a model that does not implement it looks to have been introduced in https://github.com/opencart/opencart/commit/0da669409629a3aca01aefe40c49765b34eb9491

`upload/admin/controller/sale/subscription.php`, `upload/catalog/controller/mail/subscription.php`, `upload/admin/controller/mail/subscription.php` and `upload/admin/controller/customer/customer.php` appears to have been broken since https://github.com/opencart/opencart/commit/23e7072059f9bf2987ec82e204be721a0caba257 and https://github.com/opencart/opencart/commit/8a00dffe8334133698dc863fe944bcf23735f7f4
I implemented some of the missing function, but did not implement support for paging which was previously used in `upload/admin/controller/customer/customer.php`.

### DIFF
--- a/upload/admin/controller/customer/customer.php
+++ b/upload/admin/controller/customer/customer.php
@@ -882,9 +882,9 @@ class Customer extends \Opencart\System\Engine\Controller {
 
 		$data['payment_methods'] = [];
 
-		$this->load->model('customer/customer');
+		$this->load->model('sale/subscription');
 
-		$results = $this->model_customer_customer->getPaymentMethods($customer_id, ($page - 1) * $limit, $limit);
+		$results = $this->model_sale_subscription->getSubscriptions(['filter_customer_id' => $customer_id]);
 
 		foreach ($results as $result) {
 			if (isset($result['image'])) {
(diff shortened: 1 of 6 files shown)


TheCartpenter:
The ```customer_id``` should not be filtered since the first name and last name is already concatinated.

AJenbo:
That does not appear to be a robust way to find the exact ones since two people could have the same names.

TheCartpenter:
> That does not appear to be a robust way to find the exact ones since two people could have the same names.

On the other way around, the customer can also get deleted and another customer could end up with the same customer ID as a different / new customer.

AJenbo:
ID's should not be reused, it's [incrementing](https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html) and keeps track of that. If not then I would say that is a serious bug in OC, in any case this is to align with how things worked before they broke so I don't intent to make such a change to the fix.

TheCartpenter:
> ID's should not be reused, it's [incrementing](https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html) and keeps track of that. If not then I would say that is a serious bug in OC, in any case this is to align with how things worked before they broke so I don't intent to make such a change to the fix.

It's been like that before but when we delete a customer from the database, since recent MySQL library versions, users are allowed to delete a customer ID whereas the ID resumes by incrementing to its last value on the table.

PR #13416, opened 2024-01-06
Assert that model methods are callable

AJenbo:
Validate that the function is callable, and also add support for extensions that uses actual methods rather then callbacks in variables.


danielkerr:
maybe rename from

$model

to 

$model_extension_fraud->

PR #13418, opened 2024-01-06
Correct getRateName() return type

AJenbo:
Introduced in https://github.com/opencart/opencart/commit/977710c5ba995d49261f20273e971482c584970b

### DIFF
--- a/upload/system/library/cart/tax.php
+++ b/upload/system/library/cart/tax.php
@@ -149,9 +149,9 @@ class Tax {
 	 *
 	 * @param int $tax_rate_id
 	 *
-	 * @return string
+	 * @return false|string
 	 */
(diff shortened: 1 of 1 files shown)


TheCartpenter:
The rate could also be returned as an empty string instead of false and rather the use of ```string|false```.

AJenbo:
Or null, but I don't really want to change how things work, just fix the bug as is.

TheCartpenter:
Still indicates that ```false``` is returned in the ```else``` statement. It could be replaced with: ```''``` and set the ```@return``` to: ```string```.

AJenbo:
@TheCartpenter you are welcome to make such a PR yourself if you think it would be an improvement.

PR #13422, opened 2024-01-06
Correct json error format for Cloud::save()

AJenbo:
If a permission error happened it would cause a fatal error if there was also a key or secret error since they treat ['error'] as an array, but it would have been set to a string. This is solved by also treating permission as a key on the error array.

Introduced here: 53496577d702c2ec92506ee7ca31f03d55deb49a


danielkerr:
i have removed this. i was planning on setting up a system to use amazon s3 for uploading.

PR #13425, opened 2024-01-06
[3.0.x.x] Fixed Caught class Template\Exception not found.

condor2:
https://github.com/opencart/opencart/blob/master/upload/system/library/template/twig.php#L120


mhcwebdesign:
Let's use \Exception instead of \Twig\Error\SyntaxError as there could be other causes other than syntax errors, too, for throwing an Exception.

PR #13436, opened 2024-01-06
[3.0.x.x] Undefined variable: $data- upload model

condor2:


TheCartpenter:
Already fixed in opencart-3. The core shows a mandatory array, not an optional parameter for this method.

TheCartpenter:
Looks like it's optional after all. Corrected it, now.

condor2:
Why only $data and not $data = array ?
Is like getTotalProducts from product model

TheCartpenter:
> Why only $data and not $data = array ? Is like getTotalProducts from product model

Replied on the above.

condor2:
Posted at the same time :)

PR #13438, opened 2024-01-06
[3.0.x.x] Duplicated array key -  ec_ship.php

condor2:


TheCartpenter:
Already fixed.

condor2:
> Already fixed.

To bypass the checking the file :)
https://github.com/opencart/opencart-3/blob/main/phpstan.neon#L23

TheCartpenter:
> > Already fixed.
> 
> To bypass checking the file :) https://github.com/opencart/opencart-3/blob/main/phpstan.neon#L23

That's related to something else in regard to the excluded file, not for this error. The suggested fix has been applied already. The reason why we believe it's fixed on the v3.0.x.x branch is because not all the rules as the MB and opencart-3 have been added the CS fixer. The error may, then, remain on the backend until deeper troubleshooting has been done without the specific parameters.

danielkerr:
can u update sagepay with opayo from the extension store. needs updating as they have discontinued the url.

TheCartpenter:
> can u update sagepay with opayo from the extension store. needs updating as they have discontinued the url.

Done: https://forum.opencart.com/viewtopic.php?p=865552#p865552 .

condor2:
I think he want to be added in repo, and removed sagepay.

TheCartpenter:
> I think he want to be added in repo, and removed sagepay.

Which sagepay? Direct, Server or both?

condor2:
I think both. 
Never used this payment.
Also there is module Sagepay

Maybe @danielkerr know better.

condor2:
If Paypal_express was removed..posible this will not work anymore

https://github.com/opencart/opencart/blob/3.0.x.x/upload/catalog/controller/extension/recurring/pp_express.php

TheCartpenter:
> Paypal_express was removed..posible this will not work anymore
> 
> https://github.com/opencart/opencart/blob/3.0.x.x/upload/catalog/controller/extension/recurring/pp_express.php

Yes, correct. It has been removed when the first intro. of PayPal Checkout was added into the 3.0.x.x which is one of the reasons I fixed the PayPal Checkout extension in order to deal with recurring orders / subscriptions directly with PayPal for those who can since it has been provided by Opencart as a free extension as well on the OC Marketplace since quite sometime now.

TheCartpenter:
As for opayo, it has now been added to the opencart repo: https://github.com/opencart/opencart-3/commit/805c476384571cf6367a628a951153faab2fa986 .

PR #13442, opened 2024-01-06
Use DTO instead of scalar types to avoid issues with passing things by reference

AJenbo:
Here is an implementation of using a DTO for tracking the totals rather then doing it using scaler type and having to work around limits of assigning arguments by refere, as talked about here https://github.com/opencart/opencart/pull/13410#issuecomment-1879502757

This could be a breaking change for mods that implement this function so the benefits have to be weight against that.


danielkerr:
not a chance! u are not fixing anything. what im doing is correct.

PR #13448, opened 2024-01-06
Check that rewrite() method exists before calling it

AJenbo:
$rewrite can contain classes that do not implement a rewrite function, check that the function exists to avoid crashing on a bad extension.


danielkerr:
u caused this:

https://github.com/opencart/opencart/issues/13451

AJenbo:
That's not related to this PR, you are just being petty.

Issue #12869, opened 2023-11-03
PostgreSQL support


mhcwebdesign:
**What version of OpenCart are you reporting this for?**
Master branch

**Describe the bug**
It is highly unlikely that the other DB engine PostgrSQL will actually work, especially in view of all the recently added backticks ('`') for table aliases in the various model classes.

In view of this, would it make sense to drop support for PostgrSQL altogether? Or are there plans to fix all the model classes, starting with the many unneeded backticks, and other query fixes?

TheCartpenter:
> **What version of OpenCart are you reporting this for?** Master branch
> 
> **Describe the bug** It is highly unlikely that the other DB engine PostgrSQL will actually work, especially in view of all the recently added backticks ('`') for table aliases in the various model classes.
> 
> In view of this, would it make sense to drop support for PostgrSQL altogether? Or are there plans to fix all the model classes, starting with the many unneeded backticks, and other query fixes?

Sure, go ask the database industry for a 2-minute solution that would also get rid of all the reserved terms they've put into their documentations to inform the public that ```reserved terms```, under their rights I might add, are reserved for the purpose of their libraries development in order to make this work in the mean time while somebody else worries about another and singleton library.

mhcwebdesign:
@TheCartpenter : Do have anything constructive to say here? A decision is needed here as to whether OpenCart should even support other DB engines at all, and if so, many of the model classes need fixes.

TheCartpenter:
> @TheCartpenter : Do have anything constructive to say here? A decision is needed here as to whether OpenCart should even support other DB engines at all, and if so, many of the model classes need fixes.

You damn, I have. The only and definite fixes that really are required is the MySQL Parser from the server libraries which is beyond our control.

Even though there'd be a way to dynamically remove the backticks from the models directly from the SQL string in the DB libraries, that still won't change the fact that those removals would be user-defined as it would create other problems.

danielkerr:
its too much messing around. postgres seems to differ to much from mysql

PR #13455, opened 2024-01-07
[3.0.x.x] Updated pgsql.php

condor2:
https://github.com/opencart/opencart/pull/13449/files

### DIFF
--- a/upload/system/library/db/pgsql.php
+++ b/upload/system/library/db/pgsql.php
@@ -1,59 +1,62 @@
 <?php
 namespace DB;
-final class PgSQL {
-	private $link;
+class PgSQL {
+	private $connection;
 
-	public function __construct($hostname, $username, $password, $database, $port = '5432') {
-		if (!$this->link = pg_connect('hostname=' . $hostname . ' port=' . $port .  ' username=' . $username . ' password='	. $password . ' database=' . $database)) {
+	public function __construct($hostname, $username, $password, $database, $port = '') {
+		if (!$port) {
+			$port = '5432';
+		}
+
+		try {
+			$pg = @pg_connect('host=' . $hostname . ' port=' . $port . ' user=' . $username . ' password=' . $password . ' dbname=' . $database . ' options=\'--client_encoding=UTF8\' ');
+		} catch (\Exception $e) {
 			throw new \Exception('Error: Could not make a database link using ' . $username . '@' . $hostname);
 		}
 
-		if (!mysql_select_db($database, $this->link)) {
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
Is anyone actually using PostgreSQL? Certainly for master branch and the opencart-3 repository it won't work and should be removed, especially after the many unneeded cosmetic changes to the various model classes.

condor2:
@mhcwebdesign Hmm..I don't know. Never used or tested :)
MySQLi is ok for me :)

AJenbo:
@mhcwebdesign by cosmetic changes are you referring the addition of backticks to the sql statements?

https://wiki.postgresql.org/wiki/Things_to_find_out_about_when_moving_from_MySQL_to_PostgreSQL

I do agree that this has probably broken the queries on most databases that are not MySQL variants, and should probably be revered.

mhcwebdesign:
> @mhcwebdesign by cosmetic changes are you referring the addition of backticks to the sql statements?
> 
> https://wiki.postgresql.org/wiki/Things_to_find_out_about_when_moving_from_MySQL_to_PostgreSQL
> 
> I do agree that this has probably broken the queries on most databases that are not MySQL variants, and should probably be revered.
> 
> Or are there other things you think would also cause issues?

Yes, it's mostly about the backticks. As far as I know, no  intensive testing for PostgreSQL was ever done .

AJenbo:
> As far as I know, no intensive testing for PostgreSQL was ever done 

Looks more like zero testing was done, the queries are completely broken for PostgreSQL. Most appears to have been done here: https://github.com/opencart/opencart/commit/3f717942eee0e373145e53b1bbbe851fc05acd5d

This means that since OC 4.0 PostgreSQL has been completely broken. The same is also true for [opencart-3](https://github.com/opencart/opencart-3/commit/885f4897eaca453cc91c8b9c521aca8a958c337f). But so is 3.0.x.x: https://github.com/opencart/opencart/commit/623980369db371f61918bf45871faee7965a1e31. It looks like the only working option for someone on PostgreSQL would be [Opencart_23xx](https://github.com/condor2/Opencart_23xx)

condor2:
@AJenbo If you refer to backward escaped. I have added also on my repo....on admin side

AJenbo:
Ok, then I guess that OC on PostgreSQL is just dead.

mhcwebdesign:
That's why I suggested a while ago to remove the system/library/db/pgsql.php altogether, see https://github.com/opencart/opencart/issues/12869

PR #13456, opened 2024-01-07
[3.0.x.x] Duplicated keys for array - admin config system

condor2:
I think has same effect as before.

### DIFF
--- a/upload/system/config/admin.php
+++ b/upload/system/config/admin.php
@@ -42,6 +42,8 @@ $_['action_event'] = array(
 	'view/*/before' => array(
 		999  => 'event/language',
 		1000 => 'event/theme',
+	),
+	'view/*/after' => array(
 		'event/language'
-	)
+	),
 );
--- a/upload/system/config/admin.php
+++ b/upload/system/config/admin.php
@@ -41,9 +41,7 @@ $_['action_event'] = array(
 	),
 	'view/*/before' => array(
 		999  => 'event/language',
-		1000 => 'event/theme'
-	),
-	'view/*/before' => array(
+		1000 => 'event/theme',
 		'event/language'
 	)
 );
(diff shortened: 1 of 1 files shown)


condor2:
As @TheCartpenter  changed in (https://github.com/opencart/opencart-3/blob/main/upload/system/config/admin.php#L51)

mhcwebdesign:
I think this is wrong. The admin/controller/event/language.php actually uses a view/before event:

```
	public function index(&$route, &$args) {
		foreach ($this->language->all() as $key => $value) {
			if (!isset($args[$key])) {
				$args[$key] = $value;
			}
		}
	}

```

condor2:
Yes....but duplicated keys

	'view/*/before' => array(
		999  => 'event/language',
		1000 => 'event/theme'
	),
	'view/*/before' => array(
		'event/language'
	)

mhcwebdesign:
I'll look into it tomorrow.

mhcwebdesign:
I think it needs to be like this for the system/config/admin.php :

```
....
// Action Events
$_['action_event'] = array(
	'controller/*/before' => array(
		'event/language/before'
	),
	'controller/*/after' => array(
		'event/language/after'
	),
	'view/*/before' => array(
		999  => 'event/language',
		1000 => 'event/theme'
	)
);
```

And the admin/controller/event/language.php needs this :

```
	.....
	public function index(&$route, &$args, &$template_code='') {
		foreach ($this->language->all() as $key => $value) {
			if (!isset($args[$key])) {
				$args[$key] = $value;
			}
		}
	}

```

Need to do some testing tomorrow.

condor2:
@mhcwebdesign You have tested?

mhcwebdesign:
> @mhcwebdesign You have tested?

I have tested it, and added the needed change to the 3.0.x.x .

PR #13457, opened 2024-01-07
Update order_info.twig

Samdz19:
hide disabled currencies

### DIFF
--- a/upload/admin/view/template/sale/order_info.twig
+++ b/upload/admin/view/template/sale/order_info.twig
@@ -167,7 +167,9 @@
                 <div class="form-floating">
                   <select name="currency" id="input-currency" class="form-select">
                     {% for currency in currencies %}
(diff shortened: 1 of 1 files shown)


TheCartpenter:
Not sure if this is ideal, since we'd still need to know whenever they are disabled rather than hiding that information.

Samdz19:
@TheCartpenter , I mean hide the disabled currencies from the order form in admin, so when you try to add an order or edit it via admin side you will see all the currencies with disable and enable status, so the modification will show only the enable currencies in the dropdown field.

TheCartpenter:
> @TheCartpenter , I mean hide the disabled currencies from the order form in admin, so when you try to add an order or edit it via admin side you will see all the currencies with disable and enable status, so the modification will show only the enable currencies in the dropdown field.

In this case, yes, I did read it right on what you meant.

danielkerr:
stop trying to change how the system works and start on the listed issues.

PR #13467, opened 2024-01-08
Update log.php (Replaced string with mixed to accept array)

nexadomain:
With commit [https://github.com/opencart/opencart/commit/6440e653694b92da4c3fa7ead794bd915f6995ec](https://github.com/opencart/opencart/commit/6440e653694b92da4c3fa7ead794bd915f6995ec) made on Oct 14, 2023,  
`$this->log->write()` method no longer support _'array'_
However, it is required to add error log with **arrays** also.

so replaced 'string' type with 'mixed' type for `write()` method, so as it can support string as well as array

### DIFF
--- a/upload/system/library/log.php
+++ b/upload/system/library/log.php
@@ -30,11 +30,11 @@ class Log {
 	/**
 	 * Write
 	 *
-	 * @param string $message
+	 * @param mixed $message
 	 *
 	 * @return void
 	 */
-	public function write(string $message): void {
+	public function write(mixed $message): void {
 		file_put_contents($this->file, date('Y-m-d H:i:s') . ' - ' . print_r($message, true) . "\n", FILE_APPEND);
 	}
 }
--- a/upload/system/library/log.php
+++ b/upload/system/library/log.php
@@ -34,7 +34,7 @@ class Log {
 	 *
 	 * @return void
 	 */
-	public function write(mixed $message): void {
+	public function write($message): void {
 		file_put_contents($this->file, date('Y-m-d H:i:s') . ' - ' . print_r($message, true) . "\n", FILE_APPEND);
 	}
 }


TheCartpenter:
Should be ```?string $message``` not mixed.

nexadomain:
> Should be `?string $message` not mixed.

but errors can be array too.

In Square Extension, errors are returned in array which need to be stored.
But since only string is there, it throwing error which preventing further steps required in payment

TheCartpenter:
> > Should be `?string $message` not mixed.
> 
> but errors can be array too.
> 
> In Square Extension, errors are returned in array which need to be stored. But since only string is there, it throwing error which preventing further steps required in payment

Which is why, we put a question mark in front (?) since, as you claimed, it may not always be a string but also an array.

TheCartpenter:
As indicated here: https://github.com/opencart/opencart/blob/master/upload/extension/opencart/catalog/model/payment/opayo.php#L684 .

nexadomain:
> As indicated here: https://github.com/opencart/opencart/blob/master/upload/extension/opencart/catalog/model/payment/opayo.php#L684 .

`Opencart\System\Library\Log::write(): Argument #1 ($message) must be of type ?string, array given, called in C:\xampp\htdocs\OC\opencart\upload\extension\squareup\system\library\squareup\exception\api.php on line 56: in C:\xampp\htdocs\OC\opencart\upload\system\library\log.php on line 40`

not working

TheCartpenter:
> > As indicated here: https://github.com/opencart/opencart/blob/master/upload/extension/opencart/catalog/model/payment/opayo.php#L684 .
> 
> `Opencart\System\Library\Log::write(): Argument #1 ($message) must be of type ?string, array given, called in C:\xampp\htdocs\OC\opencart\upload\extension\squareup\system\library\squareup\exception\api.php on line 56: in C:\xampp\htdocs\OC\opencart\upload\system\library\log.php on line 40`
> 
> not working

The ```$data``` array needs to have ```?string```, not the message. If there's an array of ```$message```, that should normally be a string passed into the parameter, then set an array as the second parameter as indicated in the ```log``` method I showed above. This is how extensions are normally coded.

TheCartpenter:
Correction: Array is the first parameter, the string is the second.

AJenbo:
Explicit mixed is not supported before PHP 8.0 so instead we should use implicit mixed (no type) to also support PHP 7.4. Insted of a better type would be string|array, and we can still provide this as the type hint in the PHPDoc.

upload/system/library/log.php
```diff
@@ -30,11 +30,11 @@ public function __construct(string $filename) {
 	/**
 	 * Write
 	 *
-	 * @param string $message
+	 * @param mixed $message
 	 *
 	 * @return void
 	 */
-	public function write(string $message): void {
+	public function write(mixed $message): void {
```
AJenbo, line 37: To support PHP 7.4
```suggestion
	public function write($message): void {
```

upload/system/library/log.php
```diff
@@ -30,11 +30,11 @@ public function __construct(string $filename) {
 	/**
 	 * Write
 	 *
-	 * @param string $message
+	 * @param mixed $message
```
AJenbo, line 33: A more accurate type:
```suggestion
	 * @param string|array<string> $message
```

PR #13471, opened 2024-01-08
Update opayo.php

TheCartpenter:


upload/extension/opencart/catalog/model/payment/opayo.php
```diff
@@ -238,40 +238,28 @@ private function getOrderTransactions(int $opayo_order_id): array {
...
 		}
 
-		$item['subscription']['description'] = $subscription_description;
+		$order_info = $this->model_checkout_order->getOrder($this->session->data['order_id']);
 
-		// Create new subscription and set to pending status as no payment has been made yet.
-		$subscription_id = $this->model_checkout_subscription->addSubscription($this->session->data['order_id'], $item['subscription']);
+		$subscription_id = $order_info['subscription_id'];
 
 		$this->model_checkout_subscription->editReference($subscription_id, $vendor_tx_code);
```
AJenbo, line 261: @TheCartpenter there is no such function, can you try and figure out what this should be replaced with?

TheCartpenter: Seem to be a leftover from the original developer. I looked everyone for this method but the reference field no longer exists in the core nor do any extensions on the opencart-3 except ```securetrading_pp``` and the old recurring system. I'll remove that line.

PR #13473, opened 2024-01-08
Correctly create subscription from Opayo

AJenbo:
`order_id` should be an array element, not a separate argument:

https://github.com/opencart/opencart/blob/16951fba3886c7a7d241fecac9ee5a7533c800a4/upload/catalog/model/checkout/subscription.php#L16


TheCartpenter:
No, I removed it from here: https://github.com/opencart/opencart/pull/13471 . The ```addSubscription()``` method is already handled by the checkout/model/order/addHistory method now. Leaving it there will create another subscription of the same order.

AJenbo:
Yep, replaced by https://github.com/opencart/opencart/pull/13472

PR #13474, opened 2024-01-08
Fix Opayo::calculateSchedule()

AJenbo:
The signature of `calculateSchedule()` was was causing `$next_payment` to be cast to a string which was worked around with some string parsing, but the function would still fail when trying to call `->modify()` and so would any function that tried to use the return value.
Also `$cycle` should have been an `int` and not a string.

### DIFF
--- a/upload/extension/opencart/catalog/model/payment/opayo.php
+++ b/upload/extension/opencart/catalog/model/payment/opayo.php
@@ -474,16 +474,14 @@ class Opayo extends \Opencart\System\Engine\Model {
 	 *
 	 * @param string    $frequency
 	 * @param \Datetime $next_payment
-	 * @param string    $cycle
+	 * @param int       $cycle
 	 *
 	 * @return \Datetime
 	 */
-	private function calculateSchedule(string $frequency, string $next_payment, string $cycle) {
+	private function calculateSchedule(string $frequency, \DateTime $next_payment, int $cycle) {
+		$next_payment = clone $next_payment;
 		if ($frequency == 'semi_month') {
-			// https://stackoverflow.com/a/35473574
-			$day = date_create_from_format('j M, Y', $next_payment->date);
-			$day = date_create($day);
-			$day = date_format($day, 'd');
+			$day = $next_payment->format('d');
 			$value = 15 - $day;
 			$is_even = false;
 


TheCartpenter:
There's a specific and limited reason why also the ```\DateTime``` instantiation does not propagate between methods; whether it's the same or from a different constructor in this case. Which is why, using a parameter that enforces DateTime in this scenario might not be suggested.

AJenbo:
I made sure to check all call sites to avoid such issues

TheCartpenter:
The functions I added it do not require compatibility check.

AJenbo:
> The functions I added it do not require compatibility check.

I'm saying I looked at the code to make sure it was the correct fix.

The code you added still fail later in the function since the return value is then incorrect and you can't call `->modify()` on a string. That's why this correction is needed.

TheCartpenter:
> > The functions I added it do not require compatibility check.
> 
> I'm saying I looked at the code to make sure it was the correct fix.
> 
> The code you added still fail later in the function since the return value is then incorrect and you can't call `->modify()` on a string. That's why this correction is needed.

What error message does it provide?

AJenbo:
`PHP Fatal error:  Uncaught Error: Call to a member function modify() on string`

TheCartpenter:
> `PHP Warning: Attempt to read property "date" on string` and `PHP Fatal error: Uncaught Error: Call to a member function modify() on string`

Since OC 3.2 and OC 4 now uses time zones, this may become handy to find alternatives to set the same parameters: https://stackoverflow.com/a/43810382 . The admin settings already has a time zone setting where users can modify. The same value could then be repulled by using ```$this->config->get``` and add it to the time zone function before attempting to modify the date period. I think it's already fair to say that the current time zone settings do propagate from the system/config and from the admin settings reflecting on the entire platform.

The downside would be with the use of other extensions where the system/framework relies by default to initiate the time zone, and goes on with the admin settings and another one for the date conversion. Therefore, if it already can be done with and without an API in regard our sessions, why not do the same with date conversion?

TheCartpenter:
You merged it anyways...

AJenbo:
I applied the suggestion from your link first, the object is now clone so to not affect the original like the suggestion says.

https://github.com/opencart/opencart/pull/13474/files#diff-b001027e37b4c842dfe7e50b3d21fced0bfdaec5cf4191ca75c72b3985dec6ceR482

TheCartpenter:
You still enforced ```\Datetime``` into the parameter...

AJenbo:
Yes, because that is what is given to the function, the code was causing a fatal error like I showed you.

The creation of the `\DateTime` object happens elsewhere in the code and is not affected by this change:
https://github.com/opencart/opencart/blob/12bd1d294d1b7089cbafc3f08e25fe0d360ba3ea/upload/extension/opencart/catalog/model/payment/opayo.php#L264

This is also where time zone is determined so if you want to improve handling of that you should make a change there. It is unrelated to this change.

TheCartpenter:
> Yes, because that is what is given to the function, the code was causing a fatal error like I showed you.
> 
> The creation of the `\DateTime` object happens elsewhere in the code and is not affected by this change:
> 
> https://github.com/opencart/opencart/blob/12bd1d294d1b7089cbafc3f08e25fe0d360ba3ea/upload/extension/opencart/catalog/model/payment/opayo.php#L264
> 
> This is also where time zone is determined so if you want to improve handling of that you should make a change there. It is unrelated to this change.

I'll give you that... the change you made has nothing to do with this change... I'll move some codes around and see if that can solve the issue.

TheCartpenter:
Completed the test. It's an engineering fault. ```Datetime``` requires an array to get out of this problem. Otherwise, the ```new``` instances initiated inside the loops with ```subscriptionPayment``` and ```cronPayment``` will eventually crash after 700x indexes of instantiation. To reduce the problem on happening, the ```cronPayment``` needs to be moved to the cron/opayo folder in order for the tasks to be ran to a minimum of 1 minute that is already in the core instead of all users and bots running the ```cronPayment``` on every microseconds from the browser.

PR #13481, opened 2024-01-09
[3.0.x.x] Updated startup.php

condor2:
I think is the correct solution


mhcwebdesign:
The same is true for catalog/controller/startup/startup.php, which I have now changed, too.

PR #13487, opened 2024-01-09
[3.0.x.x] Update squareup.php

condor2:
https://github.com/opencart/opencart-3/blob/main/upload/admin/controller/extension/payment/squareup.php#L653

### DIFF
--- a/upload/admin/controller/extension/payment/squareup.php
+++ b/upload/admin/controller/extension/payment/squareup.php
@@ -32,6 +32,8 @@ class ControllerExtensionPaymentSquareup extends Controller {
                     $this->config->set('payment_squareup_merchant_id', null);
                 } else {
                     if (!$this->config->get('payment_squareup_locations')) {
+						$first_location_id = null;
+
                         $previous_setting['payment_squareup_locations'] = $this->squareup->fetchLocations($this->config->get('payment_squareup_access_token'), $first_location_id);
                         $previous_setting['payment_squareup_location_id'] = $first_location_id;
                     }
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
So has anyone actually tested the squareup payment method lately?

AJenbo:
@condor2 nice work looks like you are down to just 3 files with issues on level 0

mhcwebdesign:
> @condor2 nice work looks like you are down to just 3 files with issues on level 0

Yes, but the remaining ones will be a bit more tricky. 

For example, the WechatPay extension is quite old and not compatible with the newer current zoujingli vendor library. If I can't find a matching vendor library for the latter with support for all of the PHP 7.4 to 8.3 versions, then we might have to remove the WechatPay altogether.

condor2:
If it was up to me, I would delete almost all of them. 🤣

mhcwebdesign:
> Once level 0 is solved should I try to setup reflection that allows further analysis and increase the level to 1?
> 
> Do we want that for the other forks like opencart_23xx and opencart-3?
> 
> On oc4 I'm slowly progressing on level 4 and level 6 is mostly done as well. The state of level 5 isn't clear to me yet but don't look impossible either.

Let's first deal with the outstanding issues on OC 3.0.x.x before changing the analysis level. What does level 1 do?

AJenbo:
Well that was exactly what I said we should.

Level one does the same type of checks as level 0, but for classes that utilizes magic methods and magic calls. Something that OC makes heavy use of inits proxy and registry so there are lots of areas that would not be checked by level 0.

PR #13491, opened 2024-01-09
PayPal Integration Update

mhcwebdesign:
Updated to latest version 2.2.0


ADDCreative:
@mhcwebdesign This seems to be missing some files. I get ```Could not load template admin/view/template/extension/payment/paypal/order.twig!``` when viewing an order.

There is a admin/view/template/extension/payment/paypal/order.twig in the original source at https://github.com/Dreamvention/paypal/tree/3.x.x. There looks to be other files missing as well.

Also the latest version seems to be 2.2.1.

PR #13500, opened 2024-01-09
Remove inoperable error suppresion

AJenbo:
It's not possible to suppress errors on construct like `new`, and there isn't really an error that it will emit either
https://www.php.net/manual/en/language.operators.errorcontrol.php
https://www.php.net/manual/en/language.oop5.basic.php#language.oop5.basic.new

### DIFF
--- a/upload/system/library/db/pdo.php
+++ b/upload/system/library/db/pdo.php
@@ -34,7 +34,7 @@ class PDO {
 		}
 
 		try {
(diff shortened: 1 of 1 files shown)


TheCartpenter:
Yeah, I don't know why that one's there. It is also obvious that when somebody chooses the PDO library from their config, then it won't be for permanent reasons most of the time but rather to work with multiple databases for specific projects or when using an extension.

ADDCreative:
@AJenbo The @ error suppressor is not supported by OpenCart anyway. Which could be a bit of an issue as any basic PHP notice or warning will cause a redirect to an error page and an exit.

AJenbo:
Personally I prefer not to suppress them and instead have better handling

PR #13503, opened 2024-01-10
Remove unused private methods from Image

AJenbo:
Thease cannot be called from the outside, and is never called within the class


danielkerr:
no! theres lots of reasons to have the other methods. i want the library classes to operate as its own framework

PR #13507, opened 2024-01-10
Fix check for imag type

AJenbo:
The logic here is incorrect, a string will always be different from at least one of two others.

The intent here appears to make sure that it is at least one of the two, for which it should have been `&&` instead of `||`, but to simplify the logic and make it more extensive (and shorter) I have switched it to use an `in_array()` check instead.

This was broken here, and was actually correct before the change: https://github.com/opencart/opencart/commit/5d361b97932b77a6ac9dc2321c4f3354bc16d292

### DIFF
--- a/upload/system/library/image.php
+++ b/upload/system/library/image.php
@@ -185,7 +185,7 @@ class Image {
 			$scale = min($scale_w, $scale_h);
 		}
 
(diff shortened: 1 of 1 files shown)


danielkerr:
stop posting useless pull requests

AJenbo:
They are not useless, you just don't understand the issues they are correcting

PR #13508, opened 2024-01-10
Remove redundant checks in authorize

AJenbo:
`$token_info` has already been checked not to be empty on the previous clause.


danielkerr:
so token might not be found and will through an error

AJenbo:
No because that was already checked on the previous line

PR #13509, opened 2024-01-10
Improve check for wrapper

AJenbo:
The `+ 1` would result in `$start` always being true in the following check. Instead check for a none false value and only increase $start after.


AJenbo:
Looks to have been integrated in to https://github.com/opencart/opencart/commit/01a8470257924a4736a51fe630755b11e2c28bb1

PR #13512, opened 2024-01-10
Fix Tax::calculate()

AJenbo:
The $calculate value can either be a string (like `P` or `F`, but probably also other values), `true`, or `false`. But the signature had been changed to limit it to `bool` types. This would result in the function to never skip unmatched P or F types.

Issue introduced here: https://github.com/opencart/opencart/commit/977710c5ba995d49261f20273e971482c584970b


danielkerr:
i prefer to force type

AJenbo:
Instructions unclear

PR #13514, opened 2024-01-10
Remove redundant check for debug config

AJenbo:
The debug parameter is set top true right above so there is no need to check this here.


danielkerr:
its not redundant

PR #13515, opened 2024-01-11
[3.0.x.x] Typo image library

condor2:


ADDCreative:
This is wrong. Now whatever the mime type, that bit of code will always return if the scale is 1.

condor2:
> This is wrong. Now whatever the mime type, that bit of code will always return if the scale is 1.

https://github.com/opencart/opencart/blob/master/upload/system/library/image.php#L188

ADDCreative:
It's wrong there as well then. In fact there is already a pull request to fix it. https://github.com/opencart/opencart/pull/13507

PR #13526, opened 2024-01-12
Doesn't delete the extension's settings when uninstalling it

julioccguimaraes:
https://github.com/opencart/opencart/discussions/12713

### DIFF
--- a/upload/admin/model/setting/extension.php
+++ b/upload/admin/model/setting/extension.php
@@ -86,7 +86,6 @@ class Extension extends \Opencart\System\Engine\Model {
 	 */
 	public function uninstall(string $type, string $code): void {
 		$this->db->query("DELETE FROM `" . DB_PREFIX . "extension` WHERE `type` = '" . $this->db->escape($type) . "' AND `code` = '" . $this->db->escape($code) . "'");
-		$this->db->query("DELETE FROM `" . DB_PREFIX . "setting` WHERE `code` = '" . $this->db->escape($type . '_' . $code) . "'");
 	}
 
 	/**


pemaier:
Should this not be a decision of the extension developer of the extension? how you like to update an extension if you are loosing all previous information? what is the concept for this?

stalker780:
@pemaier that's what this PR is intended to do. To keep modules/extension settings when uninstalling an extension.

pemaier:
I see the problem from both sides. On the one side we need information for updates or **a concept** for it. On the other side do we get a setting table full of garbage. I think the problem is not that simple ... but in the end the decision should be at developer side.

condor2:
Much better to improve Uninstall function to ask if you want to delete settings or not for a module...

julioccguimaraes:
This change is temporary until the functionality to update an extension is added. I support my modules and every time I have to update them it's a routine of having to reconfigure them again and it costs me time and is tedious.

stalker780:
At least you still can uninstall modules and then uninstall extension. This will clear DB from garbage.

I agree there should be more clear process of extensions management.

For example there should be 2 buttons in extensions list:
1 - delete (which only deletes extension files for future update)
2 - uninstall (which deletes files + clears DB settings)

danielkerr:
what aboutt if ur doing an update and dont want the settings removed

haxcop:
I've been using a sql trigger to accomplish such functionality, would be a good idea to get it implemented by default in opencart.

danielkerr:
No mate

stalker780:
> what aboutt if ur doing an update and dont want the settings removed

As far as I recall you cannot install extension with same code untill you uninstall an old one.
So it is not possible to update an extension without deleting it's settings.
This could have changed. Did not check this functionality for while.

danielkerr:
i need to alter the system so we have install, update and uninstall

mScott224 (approved):

PR #13528, opened 2024-01-12
Fix check for image type

AJenbo:
This code was wrong and would always result in true:

`('image/png' != 'image/png' || 'image/png' != 'image/webp')`

There is no value you can set `$this->mime` to that would ever result in the expression evaluating to false since it will always be unequal the first or second value.


mhcwebdesign (commented):
Agree, the `($this->mime != 'image/png' || $this->mime != 'image/webp')` doesn't make sense.

PR #13529, opened 2024-01-12
Accept type forced to bool

AJenbo:
Since $calculate is being forced to a boolean and both branches here are the same, only in the event that the value is false and $tax_rate['type'] is an empty string would it skip both branches. Since a none strict compare will match true and any none empty non-zero string.

Note: I'm not in favor of this, but it appears the only other solution based on the comment here: https://github.com/opencart/opencart/pull/13512


danielkerr:
im getting as bad as magento

AJenbo:
Why reject my pr if you are just going to do the same change

PR #13538, opened 2024-01-12
Fix time for quickly dipslayed notifications

AJenbo:
If less then 1 sec has passed the values would not be set, so set defaults for that case.


TheCartpenter:
I think this would conflict with this one since the same file on two different PR has been proposed: https://github.com/opencart/opencart/pull/13540/commits/d5918497b2ed497600be4f9b3db46ad11d7a0daa .

AJenbo:
It won't since the changes to the file do not overlap. Even if they do conflict I will just rebase once one of them is merged.

PR #13121, opened 2023-12-11
Update notification.php - time information

pemaier:
correction

PR #13540, opened 2024-01-12
Fix plural form in notifications

AJenbo:
Since `.` takes precedence over `?` the code would execute as:
```php
					$code = ($range . ($value > 1)) ? ($range . 's') : '';
```

Meaning it would always return the plural form.

This was introduced here https://github.com/opencart/opencart/commit/b5884f048e1c38b38fb6c87759da7c3b26e4bd10 and improperly fixed here https://github.com/opencart/opencart/commit/d61ea04a79904a68c56bb4968e531cd5132ca908 despite the issue being pointed out https://github.com/opencart/opencart/pull/13121#pullrequestreview-1785429347

### DIFF
--- a/upload/admin/controller/tool/notification.php
+++ b/upload/admin/controller/tool/notification.php
@@ -95,7 +95,7 @@ class Notification extends \Opencart\System\Engine\Controller {
 			foreach ($ranges as $range => $value) {
 				if ($value) {
 					$date_added = $value;
-					$code = $range . ($value > 1) ? $range . 's' : '';
+					$code = ($value > 1) ? $range . 's' : $range;
 				}
 			}
 


TheCartpenter:
I wonder if either the notifications or the chart controllers' calculations shouldn't replace the ```clone``` with ```\Datetime``` parameter. This could be one way to solve this by converting into days.

AJenbo:
It's working with more then just days, and it's still possible to make the same mistake with DateTime.

TheCartpenter:
> It's working with more then just days, and it's still possible to make the same mistake with DateTime.

The extension/dashboard folder also have more functionalities than just days in the models.

TheCartpenter:
SQL queries could also be used without the need to set any parameters in the meantime since ```$this->db->query``` can be pretty much accessed from everywhere to convert the time periods without the need of use with Datetime().

AJenbo:
This PR does not relate to the time calculation but the text that follows the unit.

danielkerr:
needs to be changed just in case none a-z characters

Issue #13518, opened 2024-01-11
[3.0.x.x] wechat_pay issues


mhcwebdesign:
Is anyone still using the wechat_pay at all? I am thinking about removing wechat_pay from OC 3.0.x.x.

It relies on an abandoned vendor package 'zoujingli/wechat-php-sdk', the composer software always comes up with this warning:

Package zoujingli/wechat-php-sdk is abandoned, you should avoid using it. Use zoujingli/wechat-developer instead.

While OC 3.0.x.x has the recommended 'zoujingli/wechat-developer' vendor package, the wechat_pay doesn't use it yet.

danielkerr:
ill get my staff to contact them

mhcwebdesign:
Any news yet from the wechat_pay developers and/or authors?

mhcwebdesign:
OK, we'll keep the package **zoujingli/wechat-php-sdk** in there till we hear back from the wechat_pay author.

PR #13556, opened 2024-01-13
[3.0.x.x] Include WeChat dependency

AJenbo:


mhcwebdesign:
Please read my posted issue here: https://github.com/opencart/opencart/issues/13518

We should not use the abandoned vendor package 'zoujingli/wechat-php-sdk'! 

Please undo this merge, and ask the wechatpay developer to update the payment extension, to use the recommended 'zoujingli/wechat-developer'. In the meantime we should remove this payment extension altogether.

Also, we need to make sure it supports all of the PHP versions 7.4 up to 8.3!

mhcwebdesign:
Having further tested it, it appears the old vendor package **zoujingli/wechat-php-sdk** doesn't support PHP 8.0 or later, it uses the deprecated PHP **libxml_disable_entity_loader** (see https://www.php.net/manual/en/function.libxml-disable-entity-loader.php).

If we keep the webchat_pay in there, then the least we should do is to restrict it to PHP versions < 8.0, which can be easily implemented in the **catalog/model/extension/payment/wechat_pay.php** and the a**dmin/controller/extension/payment/wechat_pay.php** .

Anyway, Daniel has now also asked his staff to contact the authors of wechat_pay, to see whether they can update it.

AJenbo:
libxml_disable_entity_loader is only deprecated, and will likely continue to work until 9.0.

if oc treats it as a critical error you can supress it eith @

PR #13558, opened 2024-01-14
[3.0.x.x] Define private properties

AJenbo:
Resolves the last of the issues reported by PHPStan on level 0 for 3.0.x.x

### DIFF
--- a/upload/admin/controller/extension/payment/securetrading_pp.php
+++ b/upload/admin/controller/extension/payment/securetrading_pp.php
@@ -1,5 +1,6 @@
 <?php
 class ControllerExtensionPaymentSecureTradingPp extends Controller {
+	private $data = array();
 	private $error = array();
 
 	public function index() {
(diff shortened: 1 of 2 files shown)


AJenbo:
@mhcwebdesign should we increase it to level 1 now?

mhcwebdesign:
We still have to sort out this issue: https://github.com/opencart/opencart/issues/13518
Your solution here (https://github.com/opencart/opencart/pull/13556) is not the right approach.

And I think before increasing the PHPStan level, just a reminder the 3.0.x.x branch is mainly for bugfixes and having made it compatible with PHP 8.x versions, and with a goal of keeping it as much backward compatible for existing 3rd party extensions as possible, including the many OCmod-based extensions.

Once the wechatpay issue has been properly addressed, it might be a good idea then to do a 3.0.4.0 release first.

AJenbo:
> just a reminder the 3.0.x.x branch is mainly for bugfixes and having made it compatible with PHP 8.x versions

Just, that's why PHPStan makes a lot of sens here since it is good at finding code that doesn't behave the same on all versions of PHP as well as general bugs.

AJenbo:
Fixed here https://github.com/opencart/opencart/commit/725bd7a6563cee34d19af288e832ec4e700b4119

PR #13562, opened 2024-01-15
Correct the type of input for Currency::format()

AJenbo:
third parameter expects a float and not a boolean


danielkerr:
not sure why opayo is in the default opencart extensions. 3rd party extensions can go from the marketplace

PR #13564, opened 2024-01-15
Correct types given to mkdir()

AJenbo:
secound argument of mkdir() should be an int and not a string: https://www.php.net/manual/en/function.mkdir.php


danielkerr:
not sure 0644 works? since if its an interger it will just revert to just 644

AJenbo:
@danielkerr numbers starting with 0 are octals: https://www.php.net/manual/en/lnguage.types.integer.php

PR #13565, opened 2024-01-15
Fix missing paramter when calling cart->add() in cron subscription

AJenbo:
4th argument should be subscription_plan_id:

https://github.com/opencart/opencart/blob/406a3fe1133856ff9ec16df153ac8d085fa0ac37/upload/system/library/cart/cart.php#L373


danielkerr:
not sure subscription_plan_id should be there if its already an order from a subscription

danielkerr:
should it not be 0

AJenbo:
That might also work.

PR #13570, opened 2024-01-15
Opayo: Remove unued argument from getCard()

AJenbo:
The first argument was never used, and was being skipped incorrectly


AJenbo:
@no198 what are you up to?

PR #13573, opened 2024-01-15
Use explicit casts when doing number manipulation in string form

AJenbo:


danielkerr:
cn it produce and error? if not then its  wasting time

PR #13574, opened 2024-01-15
Cast strings to int after extracting the ID from a string

AJenbo:


mhcwebdesign:
Unnecessary change, the admin/model/setting/module.php already has the **int** type-hint in the getModule method, and the underlying DB query uses a type-cast as well: `WHERE `module_id` = '" . (int)$module_id . "'"`

PHPStan gone a little crazy here? Can we focus on real bugixes instead?

TheCartpenter:
> While it isn't strictly necessary, making the cast explicit means it's clear that it is intentional and isn't an error.
> 
> Doing so is what allowed me to catch errors like this one: #13565
> 
> > Can we focus on real bugixes instead?
> 
> I understand your concern about bug fixes. Just a reminder, I'm a volunteer, and I don't get paid for this, directly or otherwise. I've tackled my fair share of bugs. Also, overall project health is crucial for being able to trace and solve bugs. If there are specific bugs you are concerned about, why not take a look at them yourself?

I am currently working on this issue with extension defined objects on opencart-3 since I was able to reproduce them with the logs on level 2.

AJenbo:
this is a level 5 issue so you are probably conflating it

PR #13581, opened 2024-01-17
[3.0.x.x]Update .php-cs-fixer.php

condor2:
I think this the best php-cs-fixer configuration for this branch, without changing array type...


condor2:
@mhcwebdesign What you think?

mhcwebdesign:
It results in hundreds of warnings, and would required needed cosmetic code changes in almost every PHP file. This in turn would break loads of 3rd party extensions which use OCmod or VQmod. I think I am going to restore the previous .php-cs-fixer.php file.

PR #13598, opened 2024-01-25
Adjusting the appearance of admin panel forms

batumibiz:
In the forms of the admin panel, since the time of the old Bootstrap, visual artifacts have remained in the form of dividing lines between the elements of forms. In this refinement, I brought the appearance of the forms according to Bootstrap 5, where only Fieldset has a separation line. The forms began to look cleaner and more compact.

form_r1_c1

### DIFF
--- a/upload/admin/view/stylesheet/stylesheet.css
+++ b/upload/admin/view/stylesheet/stylesheet.css
@@ -429,25 +429,6 @@ a {
     }
 }
 
-.row.mb-3 {
-    position: relative;
-}
-
-.row.mb-3 + .row.mb-3 {
-    padding-top: 1rem;
-}
-
-.row.mb-3 + .row.mb-3:before {
-    display: block;
-    content: '';
-    position: absolute;
-    width: calc(100% - calc(var(--bs-gutter-x) / 1));
-    height: 0px;
-    left: calc(var(--bs-gutter-x) / 2);
-    top: 0;
-    border-bottom: 1px solid #ededed;
-}
-
 .form-control:hover {
     border: 1px solid #b9b9b9;
     border-top-color: #a0a0a0;


stalker780:
That's just a matter of habit, with or without lines.

As to artifacts, I have never got any artifacts. This could be your monitor issue. Check your monitor and browser resolution/scale/page zoom.

Furthermore left side is not the latest OC4 screenshot. And not even OC3. This looks like an alpha OC3.1 which was never released.

batumibiz:
My monitor is not to blame.
By the word 'artifact', I meant precisely the dividing lines between the elements of the form.
The screenshot with lines is taken from here: https://demo.opencart.com/admin
The fact that there are separation lines in the Master branch and it seems to me that it is better without them.

stalker780:
Official demo is very outdated. Perhaps two years old

batumibiz:
> Official demo is very outdated. Perhaps two years old

It doesn’t matter the specific version, if I now attach a screenshot from the master branch, nothing will change.
The essence of the question: there are lines and I suggested removing them. Since the new version uses `fieldset` along with `legend`, it makes sense to leave the lines only with them.

stalker780:
@batumibiz To be honest, I like these lines.
It looks much better with them on a big screen.
OC had these lines since OC2.0. And it is about 10 years. Nobody complained.

As I mentioned before it is just a matter of habit.

batumibiz:
> @batumibiz To be honest, I like these lines. It looks much better with them on a big screen. OC had these lines since OC2.0. And it is about 10 years. Nobody complained.

You were quite right in mentioning older versions.
They did not have Fieldset groups and in large forms lines may have helped.

But the new OpenCart 4 uses Bootstrap 5 and Fieldset groups, which are visually separated by lines.
I'm not adding anything, but simply bringing the forms to the classic look that is recommended in Bootstrap 5.

danielkerr:
lines are fine and makes it distinctive

Issue #7008, opened 2018-11-05
Cache control headers not set


ADDCreative:
Since version 3 uses its own session handling, session_start(), quite rightly, is not longer called. However, as session_start() was also responsible for setting the cache control headers, they are now no longer being set by default (apart from a few places such as captcha/basic and downloads).

The session.cache-limiter, which usually defaults to nocache, controls the headers. http://php.net/manual/en/session.configuration.php#ini.session.cache-limiter You can see the headers that used to be set by session_start() here. http://php.net/manual/en/function.session-cache-limiter.php

This is particularly problematic for Internet Explorer 11 (note: not Edge in emulation mode) which aggressively caches responses from ajax GET request. This can cause problems with adding to cart, checkout, payment selection, etc.

Some forum posts confirming the problem.
https://forum.opencart.com/viewtopic.php?f=201&t=207498#p737430
https://forum.opencart.com/viewtopic.php?f=202&t=207703

Could also be the cause of following issues.
https://github.com/opencart/opencart/issues/6771
https://github.com/opencart/opencart/issues/6600

ADDCreative:
Following on from a [discussion on the forums](https://forum.opencart.com/viewtopic.php?f=202&t=207498&p=783947#p783930). Here is a simple example to recreate one of the problems this issue causes.

1. Goto https://demo.opencart.com/ in Google Chrome. Preferably in a clean session with nothing added to the cart.
2. Open the developer console (F12).
3. Goto the Network tab in the developer console and ensure "Disable cache" is not ticked, "Preserve log" in not ticked and only "Doc" is selected.
4. Click on the iPhone product to go to the product page.
5. Click add to cart. You should see the item add to the the cart at the top of the page.
6. Click the web browser's back button. You will find the cart now shows as empty. This can really confuse some customers.

If you now look in the developer tool Network at tab for demo.opencart.com, in the Size column you can see "(disk cache)". The means that Chrome use the last version from it's cache and did not request the page from the server that would of had the updated cart total.

Try this with version 2.3.0.2 or less and you won't see the issue. If you compare the response headers in the development tool, you will see the lack of cache control header in version 3, which are present in 2.3.0.2 and earlier.

This is just a simple example of the issue. There can be a lot [worse problems](https://forum.opencart.com/viewtopic.php?f=202&t=216840) if the web browser starts caching AJAX responses at checkout.

ADDCreative:
[...]

PR #13599, opened 2024-01-25
[3.0.x.x] Add cache control headers

ADDCreative:
Still regular posts on the forum with issues cause by lack of cache control headers, which have never been added to 3.0.x.x.

Use header instead of $response->addHeader to prevent redirects being cached. Which can be a problem on hosting with an ExpiresDefault set. A better way would be to modify the response library to include headers on redirects and only send when a session is started.

Also probably needed to comply with data protection regulations.

### DIFF
--- a/upload/system/framework.php
+++ b/upload/system/framework.php
@@ -72,6 +72,9 @@ $registry->set('request', new Request());
 // Response
 $response = new Response();
 $response->addHeader('Content-Type: text/html; charset=utf-8');
+header('Expires: Thu, 19 Nov 1981 08:52:00 GMT', true);
+header('Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0', true);
+header('Pragma: no-cache', true);
 $response->setCompression($config->get('config_compression'));
 $registry->set('response', $response);
 


mhcwebdesign:
I am not quite sure what issue exactly your are trying to resolve. The OC 4 master branch uses something like this:

```
$response->addHeader('Access-Control-Allow-Origin: *');
$response->addHeader('Access-Control-Allow-Credentials: true');
$response->addHeader('Access-Control-Max-Age: 1000');
$response->addHeader('Access-Control-Allow-Headers: X-Requested-With, Content-Type, Origin, Cache-Control, Pragma, Authorization, Accept, Accept-Encoding');
$response->addHeader('Access-Control-Allow-Methods: PUT, POST, GET, OPTIONS, DELETE');
$response->addHeader('Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0');
$response->addHeader('Pragma: no-cache');
```

Wouldn't that work for the 3.0.x.x, too?

ADDCreative:
Access-Control is something completely different to cache control. For the cache control, that is missing the Expires header which is needed in certain hosting configurations. Also, as I said, headers added with $response->addHeader don't get used on a redirect. So it was only really half fixed in 4.0.x.

A better way would be to modify the response library to include headers on redirects and only set the headers when a session is started.

For more information have a good read of https://github.com/opencart/opencart/issues/7008.

PR #13602, opened 2024-01-26
extensions uninstall fixes: remove permissions, + fixes for phpstan.neon

mhcwebdesign:
When an extension gets uninstalled, the access/modify permissions are now removed from the oc_user_group DB table, too.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -9,7 +9,9 @@ parameters:
         - ./upload/
     excludePaths:
         - ./system/storage/vendor/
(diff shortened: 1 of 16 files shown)


akaunderr:
`upload/admin/model/user/user_group.php` Line 86:
```
function removePermissions( string $route ): void {
```
shoud be
```
public function removePermissions( string $route ): void {
```

mhcwebdesign:
You are right, thanks for spotting it.

PR #13614, opened 2024-01-27
Update seo_url.php

TheCartpenter:


mhcwebdesign:
It should keep the

`return $this->db->getLastId();`

in there for method **addSeoUrl**

TheCartpenter:
Model parameters also don't match with the SQL query.

AJenbo:
getSeoUrls() is also a duplicate function name in this file. Things have been kind of broken for the past couple of weeks.

TheCartpenter:
The filters are still in-use in the controller. I removed the duplicated ```getSeoUrls()``` in the model.

AJenbo:
Ok, I merged https://github.com/opencart/opencart/pull/13618 which fixed the other file that had invalid syntax.
Could you plese merge master in to this PR so that we can see if it gets things to at least pass the syntax check and a status of how things are looking over all?

AJenbo:
Looks like Daniel fixed the duplicate functions in seo_url.php, but broke `admin/model/catalog/product.php`. I fixed it in master, but that means you will have to merge again for the PR to pass. Things are a bit of a mess right now.

TheCartpenter:
> Looks like Daniel fixed the duplicate functions in seo_url.php, but broke `admin/model/catalog/product.php`. I fixed it in master, but that means you will have to merge again for the PR to pass. Things are a bit of a mess right now.

It's been twice on the same day you're doing this. Therefore, you broke it, you fix it.

AJenbo:
> > Looks like Daniel fixed the duplicate functions in seo_url.php, but broke `admin/model/catalog/product.php`. I fixed it in master, but that means you will have to merge again for the PR to pass. Things are a bit of a mess right now.
> 
> It's been twice on the same day you're doing this. Therefore, you broke it, you fix it.

What? I didn't do this, it's Daniel, are you directing this at him?

TheCartpenter:
He's working on new features so I'll close this PR.

PR #13624, opened 2024-01-28
AWS has been removed, dependencies updated

batumibiz:
AWS is not used in OpenCart and it makes sense to completely remove it from the distributive.
More than 2500 files that were hanging like dead weight were freed :)


batumibiz:
So, since there are too many changed files in the commit and GitHub freezes when viewing the changes, I will explain what exactly I did:

- In the `composer.json` file, I corrected the path to the vendor libraries
- In the `composer.json` file, I updated the `scssphp` and `twig` versions
- In the `composer.json` file, I removed AWS
- In the `system/vendor.php` file I removed everything related to AWS (see [vendor.zip](https://github.com/opencart/opencart/files/14076123/vendor.zip))
- Through the console, I executed the `composer update` command, then it updated the dependencies itself and deleted the extra libraries.
- Finally, I checked the work and commit the changes.

No other files were changed.

danielkerr:
no im planning on having an option of s3 for storage

PR #13628, opened 2024-01-28
Update filter.php

TheCartpenter:


mhcwebdesign:
All the add functions in the models should return the new primary key value. For example:

```
$filter_group_id = $this->db->getLastId();
return $filter_group_id;
```
instead of returning void.

PR #13632, opened 2024-01-30
[3.0.x.x lvl 1 phpstan.neon

condor2:


AJenbo:
From what I have seen the structure of OC 2/3 is only good enough to pass level 1 so this should probably be the highest level set for this branch.

mhcwebdesign:
I'll look into it when I have some time. Level 1 will generate a lot of false positives, e.g. useless warnings, hence it will have to be fine tuned to exclude certain messages.

First though we'll have to address some more urgent genuine bugfixes for e.g. Opayo.

PR #13637, opened 2024-01-31
Update customer.php

rahuloberai:
Undefined $address_id variable


AJenbo:
Thanks. But it looks like it was fixed here: https://github.com/opencart/opencart/commit/252d65ab9e6fab14b3b8228b772668f98656015b

PR #13640, opened 2024-02-01
currency_id wrongly escaped in getCurrency method

mehdiRma10:
currency_id should not be escaped but casted instead, because it is supposed to always be an integrer.

### DIFF
--- a/upload/catalog/model/localisation/currency.php
+++ b/upload/catalog/model/localisation/currency.php
@@ -28,7 +28,7 @@ class Currency extends \Opencart\System\Engine\Model {
 	 * @return array<string, mixed>
 	 */
 	public function getCurrency(int $currency_id): array {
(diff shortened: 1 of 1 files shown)


upload/catalog/model/localisation/currency.php
```diff
@@ -28,7 +28,7 @@ public function editValueByCode(string $code, float $value): void {
 	 * @return array<string, mixed>
 	 */
 	public function getCurrency(int $currency_id): array {
-		$query = $this->db->query("SELECT DISTINCT * FROM `" . DB_PREFIX . "currency` WHERE `currency_id` = '" . $this->db->escape($currency_id) . "'");
+		$query = $this->db->query("SELECT DISTINCT * FROM `" . DB_PREFIX . "currency` WHERE `currency_id` = '" . (int)$currency_id . "'");
```
AJenbo, line 31: This was already cast to an in in the function signature so doesn't even need the explicit cast here.

PR #13642, opened 2024-02-03
[3.0.x.x] replaced wget with curl ecb controller

condor2:


mhcwebdesign:
Why? There was no bug!

AJenbo:
Not all serves include wget by default, but curl is part of the standard package set.

condor2:
> Why? There was no bug!

You have tested currency update using hosting Cron?

mhcwebdesign:
> > Why? There was no bug!
> 
> You have tested currency update using hosting Cron?

Yes, works fine on our webhost.

condor2:
With?

`wget -q "website_url" >/dev/null 2>&1`

AJenbo:
@condor2 try saying what error you get when you do that...

condor2:
> @condor2 try saying what error you get when you do that...

Using wget will download a bunch of 0 kb files on server.

`index.php?route=extension%2Fcurrency%2Fecb%2Frefresh
index.php?route=extension%2Fcurrency%2Fecb%2Frefresh.1
index.php?route=extension%2Fcurrency%2Fecb%2Frefresh.2
`

AJenbo:
So that is 2 good reasons to switch to curl in favor of wget.
- wget can leave clutter on the server
- wget is not available on all server configurations

mhcwebdesign:
You are quick coming to the wrong conclusion. The original call was a simple

wget -q "https://www.your-domain.com/index.php?route=extension/currency/ecb/refresh"

for the CRON task, and it works perfectly fine. What webhost doesn't have **wget** ? This one of the most common Linux/Unix commands available on webhosts.

condor2:
> You are quick coming to the wrong conclusion. The original call was a simple
> 
> wget -q "https://www.your-domain.com/index.php?route=extension/currency/ecb/refresh"
> 

I have tested before. On my server download/generate zero-bytes files. so thats why I have tried with **>/dev/null 2>&1**, but same.

So I use curl -s without zero-bytes files.

mhcwebdesign:
According to **wget --help**:

```
GNU Wget 1.20.3, a non-interactive network retriever.
Usage: wget [OPTION]... [URL]...
....
-q,  --quiet                     quiet (no output)
....
```

So ask your webhost to fix your wget then, we have run this command on countless servers/webhosts, never had this issue.

AJenbo:
> You are quick coming to the wrong conclusion. The original call was a simple

You are quick to dismiss a problem with out understanding it. The new call is more compatible.

> wget -q "https://www.your-domain.com/index.php?route=extension/currency/ecb/refresh"
> 
> for the CRON task, and it works perfectly fine, without generated zero-bytes files.

Which vendor is your wget from? It could be a different between for example the BusyBox and GNU implementation.

> What webhost doesn't have **wget** ? This one of the most common Linux/Unix commands available on webhosts.

Lots. curl is part of the ubuntu-server meta package, wget is not, minimalistic setups based on ubuntu-server would be setup like that.
 
> please do some proper testing first

What is it that wasn't tested here?

AJenbo:
> According to **wget --help**:
> 
> ```
> GNU Wget 1.20.3, a non-interactive network retriever.
> Usage: wget [OPTION]... [URL]...
> ....
> -q,  --quiet                     quiet (no output)
> ....
> ```
> 
> So ask your webhost to fix your wget then, we have run this command on countless servers/webhosts, never had this issue.

You are miss understanding that description, "quiet (no output)" means nothing is printed on stdout, not that it will not generate a file. This is used to avoid clutter in the server logs, not to have it not save the downloaded file. I'm starting to think you are the one that didn't test this.

mhcwebdesign:
@Alenjo: OK, you right, the wget should indeed use the **-q --no-check-certificate -O /dev/null** . I stand corrected here!

PR #13643, opened 2024-02-03
Update product.php

TheCartpenter:


AJenbo:
Could you merge the latest master in to this branch so that the change can be properly analyzed. Else it's to much work for me to review.

PR #13644, opened 2024-02-05
if page > 1 delete page=1

BuslikDrev:
image


upload/catalog/controller/common/pagination.php
```diff
@@ -7,7 +7,7 @@
  */
 class Pagination extends \Opencart\System\Engine\Controller {
 	/**
-	 * @param array<string, mixed> $setting
+	 * @param array $setting
```
AJenbo, line 10:
```suggestion
	 * @param array<string, mixed> $setting
```

danielkerr:
not sure of the issue

Issue #13663, opened 2024-02-10
[3.0.x.x] phpstan requires function signatures to be uniqly identifed across the project


mhcwebdesign:
Ever since the phpstan runs on github after each pull-request it keeps showing conflicting results for the 3.0.x.x branch.

For example, this one: https://github.com/mhcwebdesign/opencart/actions/runs/7855632559
as opposed to this one: https://github.com/opencart/opencart/actions/runs/7855612207

Running the same level 1 tests on my local test server reports no errors at all for any of the supported PHP versions.

Can we please remove phpstan from github and replace it with a more reliable alternative? If none exists, I'll just set it back to level 0 for 3.0.x.x and simply ignore it for future maintenance work.

mhcwebdesign:
I also noticed running phpstan on my local test server multiple times, for the same 3.0.x.x, no changed files, and changed PHP version, can result in different errors being reported after each run, you expect it for each run to come up with the same results.

mhcwebdesign:
Besides, your change to the model/extension/payment/opayo.php for the sendCurl is wrong! phpstan needs to be fixed if it can't cope with functions using optional parameters.

AJenbo:
The issue is caused by there being multiple identically identified functions in OC3 that do not match there signatures: 

https://github.com/opencart/opencart/blob/bc6ed93123d1b8896234d11ab9d1a636d83dbc13/upload/catalog/model/extension/payment/opayo.php#L431

https://github.com/opencart/opencart/blob/bc6ed93123d1b8896234d11ab9d1a636d83dbc13/upload/admin/model/extension/payment/opayo.php#L250

Both classes are called `ModelExtensionPaymentOpayo` without a namespace.

You will get different results depending on which is processed first by your system.

Basically there are 3 solution:
A: Add namespaces to classes to avoid collisions, this is what OC4 does, but probably not viable for OC3
B: Make identically identified functions have the same signature (see https://github.com/opencart/opencart/pull/13665)
C: Rename the conflicting functions or class so the different signatures have unique identifiers.

TheCartpenter:
https://github.com/opencart/opencart/pull/13667#issuecomment-1937797181

ondrejmirtes:
My reaction: https://github.com/phpstan/phpstan/issues/10560#issuecomment-1938424247

ondrejmirtes:
[...]

PR #13665, opened 2024-02-10
[3.0.x.x] Make identically identified functions have the same signature

AJenbo:
@mhcwebdesign nice progress on the PHPStan issues. This PR should make the branch pass the check fully.

The reason it's failing is that there are multiple classes with the same name and namespace, coatning method with the same name but different signatures.

This is just a proposed solution, idk if you prefer to solve it by renaming them or some other way.

### DIFF
--- a/upload/admin/model/extension/payment/opayo.php
+++ b/upload/admin/model/extension/payment/opayo.php
@@ -247,7 +247,7 @@ class ModelExtensionPaymentOpayo extends Model {
 		$this->db->query("UPDATE `" . DB_PREFIX . "order_recurring` SET `status` = '" . (int)$status . "' WHERE `order_recurring_id` = '" . (int)$order_recurring_id . "'");
 	}
 
-	public function sendCurl($url, $payment_data) {
+	public function sendCurl($url, $payment_data, $i = null) {
 		$curl = curl_init($url);
 
 		curl_setopt($curl, CURLOPT_PORT, 443);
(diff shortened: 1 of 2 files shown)


mhcwebdesign:
This doesn't explain the phpstan bugs, see https://github.com/opencart/opencart/issues/13663 

For a given set of PHP files, phpstan should always produce the same results, it doesn't! Until then, I suggest we disable phpstan, or come up with a proper fix!

mhcwebdesign:
Besides, your change to the model/extension/payment/opayo.php for the sendCurl is wrong! phpstan needs to be fixed if it can't cope with functions using optional parameters.

AJenbo:
Hm maybe you didn't understand what I wrote in the first message.

The issue at hand is that there are multiple classes in OC3 with the same class name.

> phpstan needs to be fixed if it can't cope with functions using optional parameters.

This has noting to do with optional parameters. It's caused by there being two classes with the same name that both have a method called sendCurl(). One has 3 arguments the other 2. So it depends on which one is processed first and there for will give different results on different computers.
This issue was solved in OC4 by using namespace.

My proposed solution for OC3 is to have identical signatures for identically named signatures.
The alternative would be to make sure they are globally unique.

mhcwebdesign:
Sorry, you are wrong. Different classes in PHP, using the same name for a method, can use different method signatures for this. These are not bugs in OC 3.0.x.x! And developers shouldn't have to worry about how they name class methods, with what signatures, as long the methods are correctly used and called upon.

I reported this bug here: https://github.com/phpstan/phpstan/issues/10560 on the phpstan repository, though I don't have time to write  more sample codes for this on their repo. Busy making OC 3.0.x.x stable again!

AJenbo:
Sure, you are entitled to your opinion of cause.

PR #13667, opened 2024-02-11
[3.0.x.x] Make destinct signatures have unique identifiers

AJenbo:
And alternative solution to https://github.com/opencart/opencart/pull/13665

I went with renaming the functions since that causes the least amount of change.

@mhcwebdesign maybe you will like this one better :shrug:

### DIFF
--- a/upload/admin/model/extension/payment/opayo.php
+++ b/upload/admin/model/extension/payment/opayo.php
@@ -104,7 +104,7 @@ class ModelExtensionPaymentOpayo extends Model {
 			$void_data['SecurityKey'] = $opayo_order['SecurityKey'];
 			$void_data['TxAuthNo'] = $opayo_order['TxAuthNo'];
 
-			$response_data = $this->sendCurl($url, $void_data);
+			$response_data = $this->sendAdminCurl($url, $void_data);
 
 			return $response_data;
 		} else {
@@ -147,7 +147,7 @@ class ModelExtensionPaymentOpayo extends Model {
 			$release_data['TxAuthNo'] = $opayo_order['TxAuthNo'];
 			$release_data['Amount'] = $amount;
 
-			$response_data = $this->sendCurl($url, $release_data);
+			$response_data = $this->sendAdminCurl($url, $release_data);
 
 			return $response_data;
 		} else {
@@ -192,7 +192,7 @@ class ModelExtensionPaymentOpayo extends Model {
 			$refund_data['RelatedSecurityKey'] = $opayo_order['SecurityKey'];
 			$refund_data['RelatedTxAuthNo'] = $opayo_order['TxAuthNo'];
 
-			$response_data = $this->sendCurl($url, $refund_data);
+			$response_data = $this->sendAdminCurl($url, $refund_data);
 
 			return $response_data;
 		} else {
@@ -247,7 +247,7 @@ class ModelExtensionPaymentOpayo extends Model {
 		$this->db->query("UPDATE `" . DB_PREFIX . "order_recurring` SET `status` = '" . (int)$status . "' WHERE `order_recurring_id` = '" . (int)$order_recurring_id . "'");
 	}
 
-	public function sendCurl($url, $payment_data) {
+	public function sendAdminCurl($url, $payment_data) {
 		$curl = curl_init($url);
 
 		curl_setopt($curl, CURLOPT_PORT, 443);
--- a/upload/admin/model/extension/payment/worldpay.php
+++ b/upload/admin/model/extension/payment/worldpay.php
@@ -89,7 +89,7 @@ class ModelExtensionPaymentWorldpay extends Model {
 
 		if ($qry->num_rows) {
 			$order = $qry->row;
-			$order['transactions'] = $this->getTransactions($order['worldpay_order_id'], $qry->row['currency_code']);
+			$order['transactions'] = $this->getAdminTransactions($order['worldpay_order_id'], $qry->row['currency_code']);
 
 			return $order;
 		} else {
@@ -97,7 +97,7 @@ class ModelExtensionPaymentWorldpay extends Model {
 		}
 	}
 
-	private function getTransactions($worldpay_order_id, $currency_code) {
+	private function getAdminTransactions($worldpay_order_id, $currency_code) {
 		$query = $this->db->query("SELECT * FROM `" . DB_PREFIX . "worldpay_order_transaction` WHERE `worldpay_order_id` = '" . (int)$worldpay_order_id . "'");
 
 		$transactions = array();


TheCartpenter:
This means that identical conventional names between the admin and catalog-end must not be used in OC v3.x releases when it comes to the models.

AJenbo:
Or that there signature must be the same. A lot of the methods two matching functions must be the same. In that way it's not really different from how extensions can replace a class with one that has the same functions.

TheCartpenter:
Then, the previous conventional names of ```sendCurl``` as opposed to ```sendAdminCurl``` methods could remain the same since both ends of the platform along with extensions can still make the calls on identical conventional names.

AJenbo:
The signatures of the two functions are different, one takes 2 parameters, the other 2 to 3. So if the one that only takes to wins the race then all calls with 3 arguments given will be issen as incorrect. Please se the linked issue for more details.

TheCartpenter:
> The signatures of the two functions are different, one takes 2 parameters, the other 2 to 3. So if the one that only takes to wins the race then all calls with 3 arguments given will be seen as incorrect. Please se the linked issue for more details.

That simply means that either renaming the conventional model names or using different parameters on the same conventional names are still a no win / win situation, since extensions depending on those methods will still be affected.

AJenbo:
It only affects 2 methods in OC3. `ModelExtensionPaymentWorldpay::getTransactions()` is private, so this should have a pretty limited effect on anything else.

mhcwebdesign:
This is just not acceptable. If different classes decide to have a method named sendCurl, but with different signatures, then so be it. PHP handles it just fine. phpstan should either be fixed, or not be used at all! Renaming perfectly valid methods for the sake of phpstan just isn't a solution.

We need to focus on genuine bugfixes, not on unneeded changes because of phpstan. As I said it is bad enough that phpstan produces different results for the same set of PHP classes!

AJenbo:
> PHP handles it just fine.

No, PHP would throw a fatal error. It's only because you don't actually load both that the issue is avoided, but here we are loading the whole application.

> As I said it is bad enough that phpstan produces different results for the same set of PHP classes!

I have already explained to you why this is, you can run it single threaded if you value deterministisk behavior over speed, but it isn't really solving things.

TheCartpenter:
In regard to conventional names, I would have to agree. PHP corporation already allows unrestricted conventional method names across different object names as the same from different controllers which are still different class names. Therefore, PHPStan should not be restricting any methods especially since we're dealing with the entitled 'personalized' CMS platform.

AJenbo:
>  PHP corporation already allows unrestricted conventional method names across different object names as the same from different controllers which are still different class names. 

PHPStan does not have issues with identical method names on different class names.

TheCartpenter:
> > PHP corporation already allows unrestricted conventional method names across different object names as the same from different controllers which are still different class names.
> 
> PHPStan does not have issues with identical method names on different class names.

But still requires to rename conventional names even though it would be specific.

AJenbo:
Sorry, I don't understand what you are saying since it's not grammatically valid.

TheCartpenter:
> Sorry, I don't understand what you are saying since it's not grammatically valid.

Regardless on how one cannot read between the lines here but the reality is that two people on the topic are already objecting against PHPStan since it requires conventional names to be renamed regardless of the quantity.

mhcwebdesign:
> > PHP handles it just fine.
> 
> No, PHP would throw a fatal error. It's only because you don't actually load both that the issue is avoided, but here we are loading the whole application.
>

I think you misunderstood here something. PHP does not throw errors simply because different classes happen to use the same method name with different signatures! This is perfectly legal in PHP. For example, there are 16 classes in OC 3.0.x.x with a method named sendCurl, with some varying signatures. There is no point in fixing something which isn't broken simply because phpstan can't correctly parse PHP code!

TheCartpenter:
Yes, I agree with this.

AJenbo:
> I think you misunderstood here something. PHP does not throw errors simply because different classes happen to use the same method name with different signatures!

Ok, let me clarify. PHP will throw a fatal error if you load two classes with the same class name. It does not support multiple classes with the same name which prevents this type of issue entirely.

> For example, there are 16 classes in OC 3.0.x.x with a method named sendCurl, with some varying signatures. 

Please, read what I'm writing. This is about the same function existing in classes with the same name. Not the same function names in any class. I have stated this repeatedly.

Maybe this simple example will help illustrate the issue:
```php
<?php

class A {
   function sendStuf() {}
}
class A {
   function sendStuf($something) {}
}
```

mhcwebdesign:
This is not the case here! method sendCurl is defined in differently named classes. The class names are only the same when one is for the admin, the other for the frontend catalog. E.g. 

admin/model/extension/payment/opayo.php

`public function sendCurl($url, $payment_data) {`

catalog/model/extension/payment/opayo.php

`public function sendCurl(string $url, array $payment_data, $i = null) {`

Perfectly valid here, since the frontend model never creates nor calls the admin model here!

mhcwebdesign:
Also, this still doesn't explain the other phpstan bug either: It produces different results, sometimes it reports this particular so-called issue, sometimes it doesn't. Hence we can't rely on this tool when it report **[OK] No errors !**

AJenbo:
> This is not the case here! method sendCurl is defined in differently named classes.

It is the case here, look for you self:

https://github.com/opencart/opencart/blob/bc6ed93123d1b8896234d11ab9d1a636d83dbc13/upload/catalog/model/extension/payment/opayo.php#L2

https://github.com/opencart/opencart/blob/bc6ed93123d1b8896234d11ab9d1a636d83dbc13/upload/admin/model/extension/payment/opayo.php#L2

> Perfectly valid here, since the frontend model never creates nor calls the admin model here!

They can not even be loaded, let alone called. They basically have to be treated as two separate applications.

> Also, this still doesn't explain the other phpstan bug either: It produces different results, sometimes it reports this particular so-called issue, sometimes it doesn't. Hence we can't rely on this tool when it report [OK] No errors !

As I have already explained to you multiple times now, this is because of multi threading so when you have conflicting signatures for the same identifier you end up having race conditions as to which one gets read first and becomes the autoritative one.

mhcwebdesign:
As I said, you don't seem to get it. As I said, simply because there are 2 identically named classes **ModelExtensionPaymentOpayo**, one for the catalog frontend, the other for the admin backend, doesn't mean that we would now not be able to define the **sendCurl** with different signatures.

Also, simply because phpstan is multithreaded, shouldn't cause it to produce different results in a postcode lottery style!

Please fix phpstan, or look for another tool which uses a proper parser technology!

AJenbo:
I do get that, but that isn't relevant since we are not dealing with the runtime here. I wish you understood that, and why that is the case.

I don't want to sound demeaning but from your last comment I think there are some aspect to this you haven't understood and I would suggest you read som of the documentation as it woulg be a lot easier to talk about things then.

ADDCreative:
> They can not even be loaded, let alone called. They basically have to be treated as two separate applications.

The admin and catalog in OpenCart are two separate applications. Any tools need to be configured to treat them as such.

AJenbo:
Agreed: https://github.com/opencart/opencart/pull/13667#issuecomment-1937817762

mhcwebdesign:
@Ajenbo : With due respect, several developers have now pointed it out to you about the issues with phpstan. 

And @AddCreative are right here, the admin and catalog sides of OpenCart are separate, hence using identically named classes, with identically named methods, such as sendCurl, with different signatures between the admin and catalog classes, is perfectly legal, and phpstan came up here with false positives, such as this one:

**Error: Method ModelExtensionPaymentOpayo::sendCurl() invoked with 3 parameters, 2 required.**

Also, whether phpstan uses multithreading or not, how is that of any relevance here? Above listed error is clearly a false positive.

Also, what is even worse, phpstan occasionally reports different results for the exactly the same set of PHP files and same PHP version, see this one: https://github.com/opencart/opencart/issues/13670

So are willing to fix phpstan, or not?

TheCartpenter:
Since the phpstan vendors have closed the topic as their solution to use unique signatures while OC already allows the use of identical signature as part of CMS solutions with a proxy class whereas OC 3 releases shows this error message above on a lower diagnostic level and that OC 4, on a higher diagnostic level does not, then this means that whether one or two decides to deactivate phpstan on OC 3 releases, it won't prevent extension developers to move on to OC 4 as clearly demonstrated on the OC Marketplace already. Therefore, since the OC 4 engine has already been converted to demonstrate these false positive no longer arises, compared to previous releases, then this shows that OC 4 is the way out of this. Otherwise, we'd be showing that deactivating phpstan would result to fallback into many years of troubleshooting on Github rather than troubleshooting in matter of minutes aside from these error messages on OC 3.

ADDCreative:
@AJenbo @mhcwebdesign

AJenbo:
> @AJenbo @mhcwebdesign

https://github.com/opencart/opencart/pull/13667#issuecomment-1937844992

mhcwebdesign:
> @mhcwebdesign There is an option to disable parallel processing if needed. https://phpstan.org/config-reference#parallel-processing

Thanks. Yes, I have seen that, and tried it out on my local test server, the generated reports are now consistent, unlike with the multi-threaded runs. However, it's much slower, and will probably cause timeouts when run on github.

I think your suggestion to use separate configurations for catalog, admin, and install is probably the best workaround for the phpstan issues.

AJenbo:
If you split the application so you avoid collisions, then you won't be loading multiple competing signatures for the same identificer and thus won't get varying results depending on which one get loaded first.

> will probably cause timeouts when run on github.

GitHub provides 2 threads per worker, so the job might increase to about 3 minutes. The default timeout is 360 minutes, so probably not.

PR #13669, opened 2024-02-11
Update pgsql.php

TheCartpenter:


mhcwebdesign:
I thought PostgreSQL would be removed altogether, as most of the SQL statements won't work with that DB engine?

PR #13675, opened 2024-02-12
Update setting.php

TheCartpenter:


TheCartpenter:
All the models are based like this.

PR #13680, opened 2024-02-12
Admin product model fix events

stalker780:
After refactoring many model functions do not trigger events. 
If this fix is fine I will check other models.


cgcmart:
My understand `$this->` refers to the current file and itself here.

PR #13693, opened 2024-02-17
[3.0.x.x] remove old pp_express recurring

mhcwebdesign:
Remove old pp_express recurring, it was incomplete and buggy


mhcwebdesign:
Use the recommended Paypal Checkout Integration extension, which supports recurring payments, too.

PR #13730, opened 2024-02-24
Update article.php

stalker780:


mhcwebdesign:
What is the point of all your model class modifications?

stalker780:
@mhcwebdesign 
model events

mhcwebdesign:
> @mhcwebdesign model events

There are better ways to do that.

stalker780:
@mhcwebdesign propose your solution

mhcwebdesign:
> @mhcwebdesign propose your solution

A simple class extending the model class, using and overridden method, will do just fine.

The only complication arises when multiple 3rd party extensions try to extend the same model class, in which case it should be a chain of dynamically extended classes. 

OpenCart already has (finally) a factory class. This one should be improve to support above described scenario, see e.g. https://stackoverflow.com/questions/1539530/is-it-possible-to-extend-a-class-dynamically on how this could be done in principle. 

Our Override Engine for OpenCart 3 has been doing it for ages. I might port it to OpenCart 4.

stalker780:
I don't see any problem. Make a pr with your override engine.
Meantime OC4 has events engine which does not work with models calling `$this->method()`.

stalker780:
> Events are not always the best approach for modifying the code

I agree with you. But without calling `$this->model_catalog->getProducts()` events do not work at all. So what the sense in events engine if it cannot catch model events?
This approach is used in OC4 from the start. After latest code refactoring many methods are called directly. I am trying to fix this and get events working for new model methods.

No need to convince me in override engine usefulness. I have the same voice value here as you do.
Try to convince Daniel ;)

PR #13733, opened 2024-02-25
Update cli_cloud.php

TheCartpenter:


ADDCreative:
display_errors takses a string, not a bool. While true will convert to '1', false will give an empty string.
https://www.php.net/manual/en/errorfunc.configuration.php#ini.display-errors

TheCartpenter:
Actually, ```1``` can also be used. Therefore, I'll close this.

PR #13755, opened 2024-03-02
Remove unused function from customer.php

AJenbo:
This function is both broken and unused


danielkerr:
no need to remove the function. just needs renaming and fixing

AJenbo:
I wish you would not push so many broken parts, it seems new things get broken every day and it's often not clear when or if they get fixed.

PR #13757, opened 2024-03-03
More phpdocs work

TheCartpenter:


mhcwebdesign:
This adds nothing to a clarification or a better understanding of the classes. See for e.g.


```
	/**
	 * Delete Histories
	 * 
	 * @param int $return_id
	 * 
	 * @return void
	 */
	public function deleteHistories(int $return_id): void {
		....
	}

```

AJenbo:
@mhcwebdesign i agree with you. But this PR is simply following the projects code style, which produces redundant content like this which requires additional work to maintain. It would be better to start a discussion about leaving out redundant information like this. Php-cs-fixer can be configured to strip any such redundancy so the change would be easy to apply if accepted.

danielkerr:
Yes. Install an ai coding thing like copilot and it can generate descriptions for u

PR #13762, opened 2024-03-04
Configure ApiGen using apigen.neon file

AJenbo:
This uses the apigen.neon configuration file rather then using command line arguments.

This update addresses compatibility issues with systems that parse symbols such as `*` (bash). Additionally, it enables the generated documentation to be loaded directly from disk without the necessity of running a web server (although it can still be hosted from one).

Furthermore, the correct project folder is now passed to '--working-dir' to ensure ApiGen can resolve project dependencies, ensuring the correct generation of the last 8 files. The script no longer errors when executed from the command line rather than being called from a web server.

These changes resolve all warnings and errors previously encountered during generation:
```
 Using configuration file opencart/apigen.neon.
 Using Composer autoloader for finding dependencies in opencart/upload/system/storage/vendor.
 Analyzing 869/869 ████████████████████████████ 100% done
 Rendering 1252/1252 ████████████████████████████ 100% done
 [OK] Finished OK                              
 ```


danielkerr:
does not seem to work any more. can u sort it out.

AJenbo:
Looks like you already fixed the conflict in https://github.com/opencart/opencart/commit/48c4fee3172c56bd547f0fd4a9f0ad03adeb359f

PR #13844, opened 2024-04-06
Aligned leftover hidden with text inputs

TheCartpenter:


mhcwebdesign (commented):
The input element is one of the so-called "void elements" or "self-closing tags". These tags do not have a corresponding closing tag because they do not contain any content between an opening and closing tag.


Hence, using HTML5, this change is not needed. We are not using XHTML.

TheCartpenter:
There are still 2196 matches among 327 files that contains these closing tags as we speak - minus this pull request.

mhcwebdesign:
This is a waste of time.

TheCartpenter:
While nobody is looking to waste your valuable time, you do not need to contribute to this. There are already people taking care of it.

mhcwebdesign:
https://developer.mozilla.org/en-US/docs/Glossary/Void_element

Therefore, don't use '/>' for these void elements in HTML5 ('<!DOCTYPE html>')!

TheCartpenter:
It's been discussed numerous times before. On that page, it clearly states:

```Although there is no way to mark up a void element as having any children, child nodes can be added programmatically to the element in the DOM using JavaScript. But that is not a good practice, as the outcome will not be reliable.```

Which is why, void elements are being used.

mhcwebdesign:
Well, it also says:

Self-closing tags (`<tag />`) do not exist in HTML.

See https://developer.mozilla.org/en-US/docs/Glossary/Void_element#self-closing_tags

We'll let Daniel decide whether he wants to go with a proper HTML5, or XHTML.

danielkerr:
XHTML

mhcwebdesign:
In that case, doesn't XHTML require this DOCTYPE:

`<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN"  "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">`

?

PR #13849, opened 2024-04-09
[3.0.x.x] Remove domain from language and currency cookies

### LINKED ISSUE #7872: Problems with settings cookies in case of redirect to subdomain.
Cookies for domains precedes sub-domain cookies. 
In that case, imagine scenario:
- create domain `example.com`
- setup Opencart.
- change Language, Currency as a customer and continue to use site. Everything works well.

Then - You decide to add "www" as main domain to the site.
So, now - all other cookies in the future will be set to `www.example.com` domain.

In this case, - You will not be able to change language, currency, until the old cookie expires.
This is a problem, if in any case, you are using domain `example.com`  as main domain, but do not set up correct redirection to one and only correct domain name. In that case, Cookies fro "domain" will always be with priority, and `startup.php` will get default language from incorrect cookie. 

```bash
language | fr-fr |	.example.com		|	2020-04-09T14:04:54.672Z
language | en-gb |	.www.example.com	|	2020-04-09T14:25:14.619Z
```

Iw user continues to browse www.example.com, then he will not be able to change language to 'fr-fr', because `startup.php`  will always get language from sent cookie and `example.com` will be with higher priority.

 Could it be better to set cookie without domain at all (then it will match if and only specific hostname I suppose). 

It has been chasing me for years. It seems, this is the reason, why it happens sometime in specific situations. Something similar is explained here:

https://www.php.net/manual/en/function.setcookie.php#91140

ADDCreative:
An invalid domain will now cause a fatal error in PHP 8. Don't use HTTP_HOST as it could be set to anything. Also fixes #7872.

### DIFF
--- a/upload/catalog/controller/startup/startup.php
+++ b/upload/catalog/controller/startup/startup.php
@@ -116,7 +116,7 @@ class ControllerStartupStartup extends Controller {
 		}
 				
 		if (!isset($this->request->cookie['language']) || $this->request->cookie['language'] != $code) {
-			setcookie('language', $code, time() + 60 * 60 * 24 * 30, '/', $this->request->server['HTTP_HOST']);
+			setcookie('language', $code, time() + 60 * 60 * 24 * 30, '/');
 		}
 				
 		// Overwrite the default language object
@@ -176,7 +176,7 @@ class ControllerStartupStartup extends Controller {
 		}
 		
 		if (!isset($this->request->cookie['currency']) || $this->request->cookie['currency'] != $code) {
-			setcookie('currency', $code, time() + 60 * 60 * 24 * 30, '/', $this->request->server['HTTP_HOST']);
+			setcookie('currency', $code, time() + 60 * 60 * 24 * 30, '/');
 		}		
 		
 		$this->registry->set('currency', new Cart\Currency($this->registry));


osworx:
What is the purpose to remove the domain?
Without, the cookie is globally valid, therefore if you have several installations on the same server (like we have on our dev servers) there is no difference betwenn them!

ADDCreative:
> What is the purpose to remove the domain? Without, the cookie is globally valid, therefore if you have several installations on the same server (like we have on our dev servers) there is no difference between them!

@osworx That's not correct. If you set the cookie domain it's will actually be valid for that domain and all subdomains. If you leave in empty only and current domain is valid. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#domaindomain-value.

osworx:
Assume I have x installations - same folder level (no subdomain of domain - basically all are subdomains).
Now, when I set a cookie for the installation A, it's also valid for B, C, D, E, etc.

Which cannot be correct, because when I want to have the language cookie set for A, it should NOT be valid also for B, C, D, E etc. ..
That's not correct, because I want to have A with en-gb, B with de-de, C with fr-fr etc.
With your solution we will have the same as we had with OC 1.4.x 1.5., 2.x which I do not want and cannot really work with.

ADDCreative:
If you don't pass a domain to setcookie it will still only be valid for the domain that the request was made to. So if the request was to A, the cookie still will not be valid for B, C, D, E etc. However, without this fix it would've been valid for subdomain1.A, subdomain2.A, etc.

With setcookie NOT providing a domain is actually more restrictive than providing one.

mhcwebdesign:
@AddCreative: Question: If I have a website at https://www.example.com , and a demo site at https://www.example.com/demo (in other words: I am not using a sub-domain), how will your proposed change affect this scenario?

ADDCreative:
@mhcwebdesign It wouldn't have any effect in that scenario. It just stops the cookies being leaked to subdomains (or subdomains of subdomains in your example). Currently the code is setting the cookies to be valid for `www.example.com` and any subdomains of `www.example.com` (whether you are using them or not). Removing the domain from setcookie actually just sets cookies to be valid to `www.example.com` and not any of its subdomains.  The web browser know the domain it sent the request to and sets that as the domain for the cookie.

The domain has already been remove in 4.0.0.0. Just never got ported to 3.0.x.x.
https://github.com/opencart/opencart/blob/c3ee9bd943ae94cc77673b298eee771b20c6fa89/upload/catalog/controller/startup/currency.php#L37-L43

You also get bots trying to manipulate HTTP_HOST, which could now cause a fatal error in PHP 8, so will help stop the PHP error log filling up.

PR #13878, opened 2024-04-21
Update authorize.twig - Automatic send email with code

pemaier:
This change solve the issue that the mail with the code is not send after pressing button Login.


danielkerr:
i already did a fix. it waits to post the code until the document is fully loaded.

PR #13888, opened 2024-04-23
Stock status for stores

pemaier:


danielkerr:
not required to split it to different stores

pemaier:
I was also thinking this but after I was comparing with my customer stores I changed my mind. Some have multistore shops with
- direct order stores where the stock information is (On stock or Available) 
and/or
- pre-order stores where the stock information is (delivery after production)
and/or other combinations. And there it is required in my opinion.

pemaier:
Also I like to draw your attention to the point that with the approved commit the text part was coming in en-gb/setting/store.php  (Line 68 and 106)

TheCartpenter:
> @TheCartpenter To answer your question - Nothing. Because if the store is deleted then the config settings for this store will get also deleted. And the setting config_store_status_id is only used to display the default product store status in this store.

The results would, then, become identical whether it shows: On stock or Available for direct sales since store owners still has the ability to delete the stock statuses from their stores. Therefore, with the exception of the ```product_report``` table, they'd still be loosing track of all the products related to those reflected stock statuses.

pemaier:
@TheCartpenter there is no product_report table where this status is visible. Even not in the product information this status is saved. It is only visible at the catalog product page (product, compare and wishlist). You may correct me if I'm wrong.

TheCartpenter:
> @TheCartpenter there is no product_report table where this status is visible. Even not in the product information this status is saved. It is only visible at the catalog product page (product, compare and wishlist). You may correct me if I'm wrong.

While that analogy is correct, the status statuses is also handled by the store settings as much as the product report feature. Therefore, there shouldn't be any difference between the two since, as you mentioned, the product report is only visible at the catalog product page; so is the stock status implicitly.

pemaier:
@TheCartpenter at the catalog product page is the "Out Of Stock Status" ... not the standard "Stock Status" when the goods are on stock (2 different information). 
The reason why we was doing this is the point to have all (possible) stock statuses at the same place and therefore we removed the text "instock". In that way we removed a redundant stock status information from the shop.
image

TheCartpenter:
That still won't cover different stores being selected for each products located in the admin > catalog > products page. For the same product, how can we differentiate the stock status messages that shows on the catalog product page on the front-end if the same product ID will look up for the current store ID from startup?

pemaier:
OK as I understand you like...

Store 0 (Direct order store) 
Camera Nikon - In Stock
Image software - In Stock 

Store 1 (Pre-Order store) your suggestion is to have the same "In Stock" also in Store 1
Camera Nikon - In Stock
Image software - In Stock 
Event Photo webinar - In Stock 

Store 1 (Pre-Order store) my suggestion is to have the possibility to be more flexible and use e.g. "Available" in Store 1
Camera Nikon - Available
Image software - Available
Event Photo webinar - Available

In the end it is a decision of the shop owner what he like more. But I think it is our job to give him possibilities.

TheCartpenter:
> OK as I understand you like...
> 
> Store 0 (Direct order store) Camera Nikon - In Stock Image software - In Stock
> 
> Store 1 (Pre-Order store) your suggestion is to have the same "In Stock" also in Store 1 Camera Nikon - In Stock Image software - In Stock Event Photo webinar - In Stock
> 
> Store 1 (Pre-Order store) my suggestion is to have the possibility to be more flexible and use e.g. "Available" in Store 1 Camera Nikon - Available Image software - Available Event Photo webinar - Available
> 
> In the end it is a decision of the shop owner what he like more. But I think it is our job to give him possibilities.

Then, as @danielkerr pointed out: https://github.com/opencart/opencart/pull/13888#issuecomment-2075231423 , there's no need to split by stores as what you seem to be rather looking is to split the stock statuses by products.

Issue #13915, opened 2024-05-02
[master] Admin - Product Filter - by Price is not working


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?
Open the admin product page and filter products bey price e.g. 80.00
Result is empty.
Also the result calculation is not working.

**To Reproduce**
see description of the bug

**Expected behavior**
correct results

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

e.g. result for filter price = 0
![image](https://github.com/opencart/opencart/assets/10166095/0855efa7-c12f-45ca-b5fb-f68b00a35c56)

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? server
 - Operating system cloudlinux
 - PHP version 8.2
 - Apache version 2.4
 - Browser(s) tested with [e.g. chrome, safari - if applicable] FF
 - Version [e.g. 22] 115.0

chulcha:
https://github.com/opencart/opencart/blob/0222f208b901116920cd59d12b12640a8f869105/upload/admin/model/catalog/product.php#L880		

~~if (!empty($data['filter_price'])) {~~

if (isset($data['filter_price'])) {

pemaier:
sorry not working

chulcha:
See
https://github.com/opencart/opencart/blob/0222f208b901116920cd59d12b12640a8f869105/upload/admin/model/catalog/product.php#L963

pemaier:
... I tested it and it doesn't work. What you show me I already saw but in line 880 it's not working because of the logic.

chulcha:
Like '0'  - not working
Like '0.0000'  - not working
(float) 0 - 0
(float) 90.0000 = 90


https://github.com/opencart/opencart/blob/0f33cb425f7ce72038ebae0f8db7fb59b3f67e9e/upload/admin/model/catalog/product.php#L964
			$sql .= " AND `p`.`price` LIKE '" . $this->db->escape($data['filter_price']) . "'";

chulcha:
![изображение](https://github.com/opencart/opencart/assets/15030612/870b5979-e964-4218-92dc-b79b5c957c16)

PR #13917, opened 2024-05-02
Update product.php - Solution for #13915

pemaier:


chulcha:
No, I want to filter, for example, %.0001%
Therefore, type conversion is not necessary in this case.
$this->db->escape != (float)

pemaier:
@chulcha The solution is oriented at the history function. If it should get changed it's not on me to decide. It's a question to Daniel.

PR #13929, opened 2024-05-11
More PHPDocs work

TheCartpenter:


AJenbo:
Since the vendor folder is distributed with the project you should make sure to use the minimum supported version of PHP for updating packages, else you may break comparability. With these changes it was now assumes PHP 8.1 since that is the version you where using for running composer.

PR #13932, opened 2024-05-13
Update cart.php

JNJLtd:
Correct stock checking for products with options.


TheCartpenter:
Code and variables are not compliant with OC.

PR #13938, opened 2024-05-17
Update zone.twig - filter reset extended because if a list item is op…

pemaier:
…ened and closed the filter reset is not working anymore.


danielkerr:
reset works fine

PR #13939, opened 2024-05-17
Update country.twig - filter reset extended because if a list item is…

pemaier:
… opened and closed the filter reset is not working anymore.


danielkerr:
reset works fine

PR #13952, opened 2024-05-17
Update product.twig - filter reset extended because if a product is o…

pemaier:
…pened and closed the filter reset is not working anymore.
[Screencast 2024-05-17 11:55:28.webm](https://github.com/opencart/opencart/assets/10166095/99ca1ec2-914d-40f0-897e-78dda5c814b7)

### DIFF
--- a/upload/admin/view/template/catalog/product.twig
+++ b/upload/admin/view/template/catalog/product.twig
@@ -46,7 +46,7 @@
               </div>
               <div class="text-end">
                 <button type="button" id="button-filter" class="btn btn-light"><i class="fa-solid fa-filter"></i> {{ button_filter }}</button>
-                <button type="reset" data-bs-toggle="tooltip" title="{{ button_reset }}" class="btn btn-outline-secondary"><i class="fa-solid fa-filter-circle-xmark"></i></button>
+                <button type="reset" onclick="location.assign('index.php?route=catalog/product&user_token={{ user_token }}');" data-bs-toggle="tooltip" title="{{ button_reset }}" class="btn btn-outline-secondary"><i class="fa-solid fa-filter-circle-xmark"></i></button>
               </div>
             </form>
           </div>


pemaier:
[Screencast 2024-05-17 11:55:28.webm](https://github.com/opencart/opencart/assets/10166095/5bb65069-5512-41e0-b0e9-fddc752afdd4)

TheCartpenter:
I think there are other ways to achieve this with OC 4 than using ```onclick``` ...

pemaier:
Yes - it's possible to make a function and use the button type for example. 
I was choosing this solution because it is the shortest one. 
Other proposals?

TheCartpenter:
> Yes - it's possible to make a function and use the button type for example. I was choosing this solution because it is the shortest one. Other proposals?

Yes, other proposals; Using the BS core attributes that's already assigned in the TWIG files.

osworx:
Sorry @pemaier  .. 
But simply look at the function `$('#button-filter').on('click', function () {` inside the same file.
Then you know how it should work ..

pemaier:
@osworx ... he wants a solution with Symphony/TWIG without onclick/javascript.

osworx:
@pemaier Guess not, BS is BootStrap not Twig or anything else ..
> BS core attributes that's already assigned in the TWIG files.

TheCartpenter:
Furthermore, JS would be implicit with the suggested use since OC has its own oc-x attributes from common.js on anyhow.

pemaier:
<button type="button" id="button-filter-reset" class="btn btn-outline-secondary" data-bs-toggle="tooltip" title="{{ button_reset }}"><i class="fa-solid fa-filter-circle-xmark"></i></button>

$('#button-filter-reset').on('click', function () {
    window.location.assign('index.php?route=catalog/product&user_token={{ user_token }}');
});

TheCartpenter:
That result is the same as mentioned above; ```.assign``` is not a method used in the core. There are already other ways to do this.

TheCartpenter:
A relative example can be found here: https://github.com/opencart/opencart/blob/master/upload/admin/view/template/marketplace/api.twig . Granted, it may not contain a filter in the file but still refers to an ID with the button type and reloads the content with ```window.location.reload()``` .

osworx:
While we have no need for a javascript assign or reload ..

Simply send the request when reset button is clicked and load the content into::
See: https://github.com/opencart/opencart/blob/b979147c753ae824954e88e398c58eb83d37032d/upload/admin/view/template/catalog/product.twig#L58

`$('#button-filter-reset').on('click', function() {
  $('#product').load('index.php?route=catalog/product.list&user_token={{ user_token }}');
});
`

ps: @TheCartpenter `loaction.assign` is a standard, simple javascript (window) function, has nothing todo with any core, BS or/and OC ..

TheCartpenter:
While one might believe that the example above has nothing to do with JS, it is - however - a valid point in regard to ```.load()``` . As for the ```.assign```, it hasn't been said anywhere that it's invalid but rather that it's not a 'method' used in OC.

> While we have no need for a javascript assign or reload ..
> 
> Simply send the request when reset button is clicked and load the content into:: See:
> 
> https://github.com/opencart/opencart/blob/b979147c753ae824954e88e398c58eb83d37032d/upload/admin/view/template/catalog/product.twig#L58
> 
> `$('#button-filter-reset').on('click', function() { $('#product').load('index.php?route=catalog/product.list&user_token={{ user_token }}'); }); `
> 
> ps: @TheCartpenter `loaction.assign` is a standard, simple javascript (window) function, has nothing todo with any core, BS or/and OC ..

danielkerr:
i think u mix up the difference between rest and clear

pemaier:
@danielkerr You are right in function and wording. 
During the work the user awaits a reset to a filter with the initial settings ... so that he get all items in the list. In the current state he get this only if no Item in the list was opened. 
As workaround he has to click the menu entry Products (because breadcrumbs entry is also with filter information) to come back to the initial situation.

osworx:
> i think u mix up the difference between reset and clear

For an(y) enduser: what IS the difference between a "**Reset**" and a "**Clear**" function?
Finally both come to the same result: **a list without any filter**.

The user expects when clicking the button, that he will get the list without any limitations / filter.
So either create 2 different buttons (reset & clear) - which will be more confusing, or add the functionality of that button to get a complete list after clicking and rename that button to a possible **Reset/Clear**

Just my 2 cents here ..

PR #13971, opened 2024-05-28
Upgraded aws/aws-sdk-php vendor to v3.308.4

TheCartpenter:


AJenbo:
@TheCartpenter please stop breaking PHP 7.4 and 8.0 support

TheCartpenter:
> @TheCartpenter please stop breaking PHP 7.4 and 8.0 support

Tell that to the vendor, you're posting this in the wrong location. Besides, from all the people who download these updates, I don't see other complaints on other repositories about that.

mhcwebdesign:
In general, before doing a vendor library update via the composer, always first check for the supported PHP versions. Since OpenCart 4 also is supposed to support the older PHP 7.4, in addition to the PHP 8.x releases, we may not always be able to use latest vendor libraries. What matters is that a vendor library works fine and supports all the PHP version from 7.4 or later.

TheCartpenter:
That can already be ensured by maintaining system/vendor.php file which is why that file is there while the vendor folder is being put up-to-date.

AJenbo:
> > @TheCartpenter please stop breaking PHP 7.4 and 8.0 support
> 
> Tell that to the vendor, you're posting this in the wrong location. Besides, from all the people who download these updates, I don't see other complaints on other repositories about that.

@TheCartpenter 
No, it's you using PHP 8.1 when doing the update. You need to run the minimum required version when updating vendor files or you lock the requirements to the one you run. 

> That can already be ensured by maintaining system/vendor.php file which is why that file is there while the vendor folder is being put up-to-date.

Yes but you are doing it wrong. This is what it now looks like if anyone who uses an older version of PHP then you try to run it:
image

I mentioned it to you last time as well: https://github.com/opencart/opencart/pull/13929#issuecomment-2121594088

TheCartpenter:
According to these readings, vendors are not the ones out-of-date with PHP.

AJenbo:
Here is the fix maybe that can help you understand what the issue is: https://github.com/opencart/opencart/commit/3ab0c21822a40b12d4af9ef8a7de289996bc8a2d

AJenbo:
@danielkerr maybe you should just increase system requirement to PHP 8.1 so @TheCartpenter can do things the way he prefers.

TheCartpenter:
You seem to have reverted the changes for the deprecation-contracts as it looks like this is what you were specifically addressing about as opposed to the aws/aws-sdk-php vendor. Therefore, I submitted another update for aws/aws-sdk-php from this PR which does not rely on dependencies: https://github.com/opencart/opencart/pull/13974

AJenbo:
It's not just the packages () but also the generated composer files requires get set to require the running version as the minimum version.

Your latest PR doesn't appear to have any of the issues.

TheCartpenter:
> It's not just the packages () but also the generated composer files requires get set to require the running version as the minimum version.
> 
> Your latest PR doesn't appear to have any of the issues.

It's settled, then. Since this issue was about dependencies, I'll simply run the queries as stand-alone packages instead.

Issue #13981, opened 2024-05-31
Admin login as customer for store other than 0 not working with cookie strict policy due to redirect chain


hayden-t:
**What version of OpenCart are you reporting this for?**
4.0.1.1 but i think master too

**Describe the bug**
Admin login as customer for store other than 0 not working with cookie strict policy due to redirect chain

**To Reproduce**
Have 2 stores on different domains and try to login as customer from admin panel to store that has domain other than admin

**Additional context**
I initially started this forum thread to explore the issue, which I now understand better as follows:
Im quite happy and confident to make a fix myself for myself, but thought id document the problem here to see what suggestions there are.

- the initial call to domain1.com/administration/index.php?route=customer/customer|login when you click the link to login sends your current cookie (with the critical OCSESSID id to link you to your db session) to setup the process token
- this then redirects to domain2.com/account/login|token to do the front end login which even though no OCSESSID cookie is sent from client with the request, because you have a token it passes
- this then redirects to domain2.com/account/account after success login, also not sending any new/updated cookies, which then fails the load as no cookie and redirects to login page

So it seems the problem is standard redirect behaviour, that it never sends or sets a new cookie for client each step, you only have the one you start with in the redirect chain, which in the case is for a different domain.

I can think up a few workarounds, ill see which works best, roughest is just to break the redirect chain at say the customer/customer|login point, and provide a link that you click on, this will restart the cookie process on the correct domain.

danielkerr:
its not a bug!

danielkerr:
only login from the default store admin

PR #13985, opened 2024-06-04
enable support for different domain login as customer from admin

hayden-t:
enable support for different front end domain, login as customer, from admin

please discuss this with me rather than just closing,

https://github.com/opencart/opencart/issues/13981


hayden-t:
suprise, suprise, too cowardice to reflect

PR #13997, opened 2024-06-11
Admin category list filters

stalker780:
Modern stores have hundreds and even thousands of categories. Category list must have filters by name, parent, status

image

### DIFF
--- a/upload/admin/controller/catalog/category.php
+++ b/upload/admin/controller/catalog/category.php
@@ -16,7 +16,27 @@ class Category extends \Opencart\System\Engine\Controller {
 
 		$this->document->setTitle($this->language->get('heading_title'));
 
+		if (isset($this->request->get['filter_name'])) {
+			$filter_name = $this->request->get['filter_name'];
+		} else {
+			$filter_name = '';
+		}
+
+		if (isset($this->request->get['filter_status'])) {
+			$filter_status = $this->request->get['filter_status'];
+		} else {
+			$filter_status = '';
+		}
+
 		$url = '';
+
+		if (isset($this->request->get['filter_name'])) {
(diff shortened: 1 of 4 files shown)


pemaier:
+1

mhcwebdesign:
@stalker780 : 

> Modern stores have hundreds and even thousands of categories.

Sorry, but the number of products or categories has nothing to do with so-called modern stores, the latter is a matter of personal taste and design. Though I agree having some more filters can be useful. In the meantime why not search the marketplace for a suitable extension?

pemaier:
@mhcwebdesign Thank you for the comment. I also agree that there should not be a endless list of filters. But related to the common philosophy of our project we should give the project a basic filter set. This should be a basic list of filters for the main functionality within OpenCart - not more but also not less.

In Catalog: Categories, Products, Manufacturers, Downloads and Information
Common filters: Name, Store, Status
and in Products also: Model, Category, Manufacturer, Store, Price, Quantity

I think this is the basic list which is required for a comfortable work with small and big shops for our baseline of shop functionality.
For all additional filters (eg. Creation Date, Last Modification, ...) I agree this is an individual set for a marketplace extensions.

stalker780:
@mhcwebdesign are you serious?
Why there are filters for products, reviews, orders?
Everyone could search for an extension on the Marketplace.

We can still use OC1.5 with extensions. No need in OC4.

Sorry for this sarcasm, but what you said can be applied to any OC update or feature.

Still there is a basic functionality which every store must have right out of the box. Otherwise all these product/category/order/customer lists are unusable.

stalker780:
As to marketplace extensions... Extensions->Installer and Extensions->Extensions should have filters by name and status either :)

danielkerr:
fine but seems like ur adding this for no reason

PR #14003, opened 2024-06-21
[3.0.x.x] Fix Pagination limit / 0

Squidysquid1:
Fixed in the main branch: https://github.com/opencart/opencart/commit/61243223184a037fc12aad57fb70d3127dd59a9f

Had errors in my log and saw that it was patched in main but not 3.0.x.x branch.

### DIFF
--- a/upload/catalog/controller/product/category.php
+++ b/upload/catalog/controller/product/category.php
@@ -33,7 +33,7 @@ class ControllerProductCategory extends Controller {
 			$page = 1;
 		}
 
-		if (isset($this->request->get['limit'])) {
+		if (isset($this->request->get['limit']) && (int)$this->request->get['limit']) {
 			$limit = (int)$this->request->get['limit'];
 		} else {
 			$limit = $this->config->get('theme_' . $this->config->get('config_theme') . '_product_limit');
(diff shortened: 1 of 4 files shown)


TheCartpenter:
Except that OC v4.x no longer uses dynamic configurations with the themes in the ```else``` statement whereas OC v3.x releases still uses them.

Squidysquid1:
I apologize I'm not following. If I'm not mistaken it still fixes a divide by 0. In catalog/controller/product/search.php if a user sets limit to be 0 it will cause an error. Granted for searching, for example, a user would have to edit the url to get this error to occur, but still an error with a simple fix.

https://github.com/opencart/opencart/blob/bbaaa74cbcc4c613abd3e943eca47c6b53ac9eea/upload/catalog/controller/product/search.php#L408C1-L408C117 `ceil($product_total / $limit)` This is the line that causes it.

Unless what you are trying to convey that there is also an edge case in the else statement and that we should handle that as well. If this is the case, then I'm not as concerned with that since it would have to be a site admin changing that value. But we could also add a check for that as well.

TheCartpenter:
The MB still implies ``` > 0``` in the lookup. Yours doesn't as an extension could set modify those settings nevertheless.

mhcwebdesign:
Now updated for 3.0.x.x, see https://github.com/opencart/opencart/pull/14013

PR #14007, opened 2024-06-22
FIX problem - SpamAssassin does not like you - 	BASE64_LENGTH_79_INF	…

prometheuscz:
BASE64_LENGTH_79_INF encoded email part uses line length greater than 79 characters


TheCartpenter:
Same here.

prometheuscz:
It is not problem only in SpamAssassin. For example TitanHQ https://docs.titanhq.com/en/25060-built-in-rules.html,... 
In old version of OC, there was standard chunk_split, that use 76 chars.

What is benefit to split it by 950 chars?

PR #14008, opened 2024-06-22
FIX - SpamAssassin does not like you - MPART_ALT_DIFF

prometheuscz:
HTML and text parts are different

Make sure your TEXT version looks similar as the text inserted in your HTML version


TheCartpenter:
That could rather be built as an extension on the Marketplace, since not everybody may use SpamAssasin filtering for their emails.

prometheuscz:
It is not only SpamAssassin - many other spam filters check difference in TEXT and HTML version. 
It is weird and not nice to have "'This is a HTML email and your email client software does not support HTML email!'" in all email in TEXT version. Why dont take there stripper text from HTML version?

PR #14016, opened 2024-06-26
Upgraded aws/aws-sdk-php vendor to v3.315.0

TheCartpenter:


mhcwebdesign:
You have been posting nearly daily updates of the aws-sdk-php vendor library. Don't they have a stable version?

TheCartpenter:
> You have been posting nearly daily updates of the aws-sdk-php vendor library. Don't they have a stable version?

And there's only going to be more of it, since that's a question you're going to have to ask them about. The way I see it, there's a team working in the AWS repository that still maintains their package and their projects. Then, distributes it as opposed to the rest of the vendors on this platform that don't do much. Therefore, I'd say it's worth the daily updates.

PR #14040, opened 2024-07-08
Upgraded aws/aws-sdk-php vendor to v3.316.0

TheCartpenter:


mhcwebdesign:
You should ask the authors of the aws-sdk-php library whether they actually have a stable version of their library.

Besides, as far as I know the aws-sdk-php isn't even used by OpenCart 4. If a future 3rd party extension ever needs it, it should be bundled in there, not in the OpenCart core.

TheCartpenter:
Not that it's in anyone's control except them, it's in the released section of their repository and worldwide. As long as it's in the core, not much that can be done about it.

PR #14042, opened 2024-07-11
Use same path when call functions - order/cehckout model

condor2:


mhcwebdesign:
What is the point of this pull request?

condor2:
To be the same....in some places use same....

https://github.com/opencart/opencart/blob/master/upload/catalog/model/checkout/order.php#L55

stalker780:
Do not do that. You are breaking events for model methods.

On the contrary all model methods should be called using proxy `$this->model_checkout_order->method` to make events work correctly.

condor2:
Then others should be made $this->model_checkout_order->method...

Line 55, line 265...

aaron-muti-420 (approved):
Looks Great

PR #14050, opened 2024-07-15
Docker Desktop

aaron-muti-420:
This is PR talks about running Open Cart Locally using Docker Desk top on windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,9 @@
 
 OpenCart is a free open source e-commerce platform for online merchants. OpenCart provides a professional and reliable foundation from which to build a successful online store.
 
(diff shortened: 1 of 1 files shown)


Squidysquid1:
Windows is misspelled and the wording seems to be quite awkward.

Also in my opinion, if this were to be merged, it would be better in a wiki page under "Setting up a development environment" (or something similar)

aaron-muti-420:
Okay thanks for the feedback i will fix that, where is the Project files for the documentation?

aaron-muti-420 (commented):
Corrected the spelling

PR #14074, opened 2024-08-05
Update header.twig

tmdashwani:
Fav Icon


ADDCreative:
This is wrong. If the icon isn't set, the HTML will be invalid as the href will be empty.

tmdashwani:
@ADDCreative we setup a condition if the icon is not present that not display on HTML

PR #14081, opened 2024-08-13
Upgraded aws/aws-sdk-php vendor to v3.319.4

TheCartpenter:


TheCartpenter:
Due to multiple vendor packages' PHP minimum version requirements, it would be wise to suggest the use of PHP 8.1 rather than PHP 7.4, as suggested by the vendors, nowadays.

PR #14087, opened 2024-08-16
change "account_custom_field" from string to object by json_decode

jian2008:
from the usage in the follow codes, we can know  $data['account_custom_field'] is object
-    admin\view\template\customer\customer_form.twig
-    admin\view\template\sale\order_info.twig
-    catalog\view\template\account\edit.twig

as $customer_info['custom_field'] is read from mysql, so it's string, so we should change the type from string to object  in the follow 3 files by json_decode($customer_info['custom_field'], true):
-   upload/admin/controller/customer/customer.php
-   upload/admin/controller/sale/order.php
-   upload/catalog/controller/account/edit.php

the fix code is reference to version "3.0.3.9". in version  "3.0.3.9"  also use json_decode to convert string to object 
`
$data['account_custom_field']` = json_decode($customer_info['custom_field'], true);
`

### DIFF
--- a/upload/admin/controller/customer/customer.php
+++ b/upload/admin/controller/customer/customer.php
@@ -598,7 +598,7 @@ class Customer extends \Opencart\System\Engine\Controller {
 		}
 
 		if (!empty($customer_info)) {
-			$data['account_custom_field'] = $customer_info['custom_field'];
+			$data['account_custom_field'] = json_decode($customer_info['custom_field'], true);
 		} else {
 			$data['account_custom_field'] = [];
 		}
--- a/upload/admin/controller/sale/order.php
+++ b/upload/admin/controller/sale/order.php
@@ -632,7 +632,7 @@ class Order extends \Opencart\System\Engine\Controller {
 		}
 
 		if (!empty($order_info)) {
-			$data['account_custom_field'] = $order_info['custom_field'];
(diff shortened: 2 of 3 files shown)


danielkerr:
already done in the model 

	public function getCustomer(int $customer_id): array {
		$query = $this->db->query("SELECT DISTINCT * FROM `" . DB_PREFIX . "customer` WHERE `customer_id` = '" . (int)$customer_id . "'");

		if ($query->num_rows) {
			return $query->row + ['custom_field' => json_decode($query->row['custom_field'], true)];
		} else {
			return [];
		}
	}

jian2008:
recently, I check out  newest code, I found 
`$query->row + ['custom_field' => json_decode($query->row['custom_field'], true)];`
is not work,   the 'custom_field'  still  string.


please refer to the article:  https://stitcher.io/blog/array-merge-vs+

so we can change to 
`array_merge($query->row, ['custom_field' => json_decode($query->row['custom_field'], true)])`

or user another way  like in upload/admin/model/catalog/product.php line 844 ~849:
`
		if ($query->num_rows) {
			$product_data = $query->row;

			$product_data['variant'] = json_decode($product_data['variant'], true);
			$product_data['override'] = json_decode($product_data['override'], true);
		}
`

PR #14091, opened 2024-08-18
Upgraded aws/aws-sdk-php vendor to v3.320.2

TheCartpenter:


composer.lock
```diff
@@ -848,7 +848,7 @@
                 "shasum": ""
             },
             "require": {
-                "php": ">=7.1"
+                "php": ">=8.1"
```
AJenbo, line 851: ...

PR #14092, opened 2024-08-19
Fix removing image in edit function of product.php

retrina:
I deleted the code below because the product image is not deleted from the database during edit function and add it into query without "if($data['image'])" :

if ($data['image']) {
			$this->db->query("UPDATE `" . DB_PREFIX . "product` SET `image` = '" . $this->db->escape((string)$data['image']) . "' WHERE `product_id` = '" . (int)$product_id . "'");
		}


retrina:
need to seprate two commits

Issue #11596, opened 2022-08-26
Warning: Zend OPcache API is restricted by "restrict_api" configuration directive in /system/storage/vendor/twig/twig/src/Cache/FilesystemCache.php on line 67


kb-git:
**What version of OpenCart are you reporting this for?**
4.0.1.0

**Describe the bug**
There is Warning message:
Warning: Zend OPcache API is restricted by "restrict_api" configuration directive in /system/storage/vendor/twig/twig/src/Cache/FilesystemCache.php on line 67

During installatin and also after opencart is installed. Warning message is displayed in both frontend and admin. 

**To Reproduce**
1. Go to every page during install/after install opencart.

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server?
 - Operating system: Linux hc1-wd212-6778857fbc-vv8l8 5.4.0-86-generic #97-Ubuntu SMP Fri Sep 17 19:19:40 UTC 2021 x86_64
 - PHP version: 8.0.1
 - Zend OPcache v8.0.1
 - Browser(s) tested with [e.g. chrome, safari - if applicable]: Every browser

**Additional context**
After refresh, warning message will dissapear. But it shows on every new page, which is not cached.

Zend OPcache config in attachment.

![zendopcache](https://user-images.githubusercontent.com/37004034/186867339-17bcc82d-e023-4647-a73b-a1ecd8596a63.jpg)
![Screenshot_1](https://user-images.githubusercontent.com/37004034/186867363-631c1260-4256-48ea-ba7c-4adff4cec830.jpg)

danielkerr:
Zend OPcache is not in the default opencart install

mejzr76:
1. The source of this bug is related to the configuration of the PHP8 API Zend OPcache.
2. The dependencies of this bug are not specified in the provided context.
3. It is not possible to fix this bug by simple modification of the OC code without modifying the PHP8 server configuration?

Thaks a lot,
ONX

ADDCreative:
@mejzr76 The issue is that Twig suppresses the error with @. However OpenCart does not support the @ PHP error suppressor anymore. So any message or warning will cause a issue on some hosting platforms.

https://github.com/opencart/opencart/blob/985c4bf7210baf12a1d7dd6489debe1f56b2e018/upload/system/storage/vendor/twig/twig/src/Cache/FilesystemCache.php#L67

[...]

Issue #6321, opened 2018-01-07
[v3.0.2.0 - Issue] - library cache file.php


straightlight:
It seem the PHP error cache library file.php is still creating error messages. The last solution provided by Daniel did not resolved the issue as a user mentions that the error message still persist after replacing the system/library/cache/file.php : https://forum.opencart.com/viewtopic.php?f=202&t=200896#p709630

ADDCreative:
Maybe something along the lines of this may help to suppress the error if another process has deleted the file, but still detect if the file could not be deleted.

```php
if (is_file($file)) {
    if (!@unlink($file)) {
        clearstatcache();

        if (is_file($file)) {
           // Generate error here.
        }
    }
}
```

PR #14096, opened 2024-08-21
Avoid to fill up log with notices about unlink is not possible

osworx:
Added clearstatcache() to avoid notices in log file

### DIFF
--- a/upload/system/library/cache/file.php
+++ b/upload/system/library/cache/file.php
@@ -15,6 +15,7 @@ class File {
 				if ($time < time()) {
 					if (file_exists($file)) {
 						unlink($file);
+						clearstatcache(false, $file);
 					}
 				}
 			}
@@ -70,6 +71,7 @@ class File {
 			foreach ($files as $file) {
 				if (file_exists($file)) {
 					unlink($file);
+					clearstatcache(false, $file);
 				}
 			}
 		}


ADDCreative:
clearstatcache won't help as it only clears the file status cache for the current process and unlink() clears the file status cache automatically anyway. The unlink messages will likely be cause by a race condition between 2 processes.

Also passing a file is pointless if clear_realpath_cache is set to false.

The following would be enough to suppress the messages.
```php
@unlink($file);
```
If you still wanted to display genuine errors but not those generated by race conditions you you could use.
```php
if (!@unlink($file)) {
	if (is_file($file)) {
		unlink($file);
	}
}
```

osworx:
The problem is, is_file is cached, see: https://www.php.net/manual/en/function.is-file.php
Therefore the function calls a - possible - cached file and is therefore not able to delete (unlink) the file.

Using the @ will suppress only a possible output - should be not the final solution!

Had such issue also with some OC 2.x installations, adding clearstatcache() solved always the problem.
If i run the script without that, the error.log is full is warnings ..

The problem is, that I assume this will happen also at many other users. And the error.log is filling and filling up with stupid entries as long as not new entries are possible (because of the size).
And some clients had then the problem, that the shop began to oprate in "crazy mode" - means not every funcion could be fullfilled and customers had troubles to order!

ADDCreative:
Have a good read of the documentation you posted a link to. https://www.php.net/manual/en/function.clearstatcache.php#:~:text=However%20unlink()%20clears%20the%20cache%20automatically.

It states.
> However [unlink()](https://www.php.net/manual/en/function.unlink.php) clears the cache automatically.

You can also see this in the PHP source code.
https://github.com/php/php-src/blob/450740cbb600fb1b26f2a048d312bbf7fb5a9a24/main/streams/plain_wrapper.c#L1243-L1267

So calling ```clearstatcache(false, $file);``` after ```unlink($file);``` is pointless as the file has already has it's status cache cleared.

Also ```clearstatcache(false, $file);``` is actually an invalid combination of parameters and is the equivalent of ```clearstatcache()```.
https://github.com/php/php-src/blob/e5a30274d8d62f5c392ba7c20e8a21e52ee72a2b/ext/standard/filestat.c#L685-L705

Even if clearstatcache had an effect, you can still get the race condition if another process deletes the file between file_exists and unlink.

mhcwebdesign:
What exact issues are you guys trying to resolve here? What exact error messages?

ADDCreative:
@mhcwebdesign Probably the ```Warning: unlink(/???/storage/cache/cache.???): No such file or directory in /???/www/system/library/cache/file.php``` messages, which are usually changed caused by race conditions (often caused by bots), configuration issues or bugs in extensions. Probably best described at https://forum.opencart.com/viewtopic.php?t=223656#p820470.

As stated in my comment https://github.com/opencart/opencart/pull/14096#issuecomment-2304407187 this change should have no effect (other than a very small change in timings).

mhcwebdesign:
Hmm, in OpenCart 4 it uses something like this:

```
if (!@unlink($file)) {
    clearstatcache(false, $file);
}
```

ADDCreative:
Adding the @ in front of unlink suppresses the warnings. Adding the clearstatcache was originally my idea (See https://github.com/opencart/opencart/issues/6321). However, on testing and looking at the PHP source code it does nothing, so can be remove.

mhcwebdesign:
All right, let's just use this one then:

`@unlink($file);`

See https://github.com/opencart/opencart/commit/23dcc645d783b93a1de03e0d2d0d0f0e349c2dde

hayden-t:
Hi guys. Im on 4.0.1.1 and still getting cache unlink errors.   
[https://github.com/opencart/opencart/blob/8a8f42e5992fc9dfaa25d259f80ebd62837201b7/upload/system/framework.php#L36](https://github.com/opencart/opencart/blob/8a8f42e5992fc9dfaa25d259f80ebd62837201b7/upload/system/framework.php#L36)  
It has this inside the error handler which i guess is meant to stop the errors being reported but they still are for me

```
set_error_handler(function(string $code, string $message, string $file, string $line) use ($log, $config) {
// error suppressed with @
if (@error_reporting() === 0) {
 return false;
}
```

I noticed in master this is not even used at all: so how does it stop them, or doesnt it ?  
https://github.com/opencart/opencart/blob/f0d42f933a3b816e8b862e6ac8892f93d87c7843/upload/system/framework.php#L36

hayden-t:
Turns out when checking error level with error\_reporting() it does not (for me) return 0 when @ is used to suppress, it returns 4437 which can be decoded here [https://maximivanov.github.io/php-error-reporting-calculator/](https://maximivanov.github.io/php-error-reporting-calculator/)  
It should instead be:  
 `if (!(error_reporting() & E_WARNING)) {`  
 `return false;`  
 `}`

hayden-t:
some other related posts:

#12483

#12660

ADDCreative:
@hayden-t OpenCart doesn't support the @ error suppressor any more. Which is a bit of an issue as Twig requires it https://github.com/opencart/opencart/issues/11596#issuecomment-1904250142.

I made a pull request #12660 that adds the following to the error handlers. You would need to make that change to add it back.

``` php
if (!(error_reporting() & $code)) {
	return false;
}
```

hayden-t:
i might just make a pull request to put the cache delete in a try catch as yours got rejected, but pretty sure it will get rejected too,

hayden-t:
#14778

PR #14104, opened 2024-08-26
shift id=option-value={{ option_row }} to table instead tbody

punitkumar:
upload/admin/view/template/catalog/product_form.twig
compare line no 1718 and 803, 815


danielkerr:
its the other way round

PR #14123, opened 2024-09-14
Update seo_url.php

rahuloberai:


danielkerr:
does not make sense because theres no caching using key.

PR #14130, opened 2024-09-17
Update seo_url.php

head-dunce:
Updated seo_url.php to handle product, category and information routes based on the key stored in the database


head-dunce:
Needs more work, working on it.

PR #14131, opened 2024-09-18
Updates to  seo_url.php

head-dunce:
Updated route-setting logic in `index()` where if a route is not set, but a product_id / category_id / information_id has been pulled from the database, set the route accordingly. 

Updated rewrite so that if getSeoURLByKeyValue is ran, the value is stored in the array to prevent duplicate lookups

### DIFF
--- a/upload/catalog/controller/startup/seo_url.php
+++ b/upload/catalog/controller/startup/seo_url.php
@@ -33,6 +33,19 @@ class SeoUrl extends \Opencart\System\Engine\Controller {
 					if ($seo_url_info) {
 						$this->request->get[$seo_url_info['key']] = html_entity_decode($seo_url_info['value'], ENT_QUOTES, 'UTF-8');
 
+						// If product_id is found, set the route to product/product
+                        if ($seo_url_info['key'] == 'product_id') {
+                            $this->request->get['route'] = 'product/product';
+                        }
+                        // If category_id is found, set the route to product/category
+                        if ($seo_url_info['key'] == 'category_id') {
+                            $this->request->get['route'] = 'product/category';
+                        }
+                        // If information_id is found, set the route to information/information
+                        if ($seo_url_info['key'] == 'information_id') {
+                            $this->request->get['route'] = 'information/information';
+                        }
+						
 						unset($parts[$key]);
 					}
 				}
@@ -98,12 +111,12 @@ class SeoUrl extends \Opencart\System\Engine\Controller {
 				$value = '';
 			}
 
-			if (!isset($this->data[$key])) {
-				$this->data[$key] = $this->model_design_seo_url->getSeoUrlByKeyValue((string)$key, (string)$value);
+			if (!isset($this->data[$key][$value])) {
+				$this->data[$key][$value] = $this->model_design_seo_url->getSeoUrlByKeyValue((string)$key, (string)$value);
 			}
 
-			if ($this->data[$key]) {
-				$paths[] = $this->data[$key];
+			if ($this->data[$key][$value]) {
+				$paths[] = $this->data[$key][$value];
 
 				unset($query[$key]);
 			}
--- a/upload/catalog/controller/startup/seo_url.php
+++ b/upload/catalog/controller/startup/seo_url.php
@@ -32,26 +32,27 @@ class SeoUrl extends \Opencart\System\Engine\Controller {
 
 					if ($seo_url_info) {
 						$this->request->get[$seo_url_info['key']] = html_entity_decode($seo_url_info['value'], ENT_QUOTES, 'UTF-8');
-
-						// If product_id is found, set the route to product/product
-                        if ($seo_url_info['key'] == 'product_id') {
-                            $this->request->get['route'] = 'product/product';
-                        }
-                        // If category_id is found, set the route to product/category
-                        if ($seo_url_info['key'] == 'category_id') {
-                            $this->request->get['route'] = 'product/category';
-                        }
-                        // If information_id is found, set the route to information/information
-                        if ($seo_url_info['key'] == 'information_id') {
-                            $this->request->get['route'] = 'information/information';
-                        }
 						
 						unset($parts[$key]);
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
Using

```
switch (true) {
....
}
```

is a strange way to program it. Better use if-else blocks instead of switch.

head-dunce:
Why is a switch strange? It's typically faster and easier to read.

chongshengdz:
how about to add the language routes to to prevent duplicate lookups?

head-dunce:
A lot more needs to be done to this file. Unfortunately this is where I have to leave it for now. Now I see that I'm not moving to OC4 anytime soon so I took a snapshot of the OC4 server I was working on, and I'm working on cleaning up and moving to 3.0.4.0 instead. Hopefully I'll be able to come back and help in the future.

stalker780:
@head-dunce 
Why don't you create your own seo_url extension and execute it instead of the default one? 
Just use events and startup. Override default seo_url controller on startup and that's it.
You can code it the way you want. Modify the url creation/parsing process the way you want.
You just don't know how to cook OC4 ;)

TheCartpenter:
As for this commit: https://github.com/opencart/opencart/pull/14131/commits/aa43fde21ffcc1f23a2f2af2302435763c82a44f , the suggested change may be relative to the PHP version being used as encountered already in the past with OC v3.x releases from the startup files. Still, on this day, it may not affect everyone as also the ```$value``` still needs to be relative with its key without making modifications by extensions.

Your suggestion could still shed some lights to extension developers to code alternative ```$value``` as opposed to the one that's originally delivered from the OC admin.

head-dunce:
Why store the key in the database specifying what type if id it is (product/category/etc) if you still have to specify the route in the url separately? It creates more work for the user. It slows the page load. And it breaks OC3 to OC4 urls. The switch only executes if the route isn't specified.

TheCartpenter:
> Why store the key in the database specifying what type if id it is (product/category/etc) if you still have to specify the route in the url separately? It creates more work for the user. It slows the page load. And it breaks OC3 to OC4 urls. The switch only executes if the route isn't specified.

Simple. Because the SEO URLs from the core have always been used for specific routes, not for all routes. Always has been and if one believes that this method would break the SEOs between OC3 and OC4 releases, then it's more than obvious that the upgrade hasn't been used or that the related database queries handled from the upgrade hasn't been used from PHPMyAdmin in the first place but rather the use of XLS files by extensions to upgrade their stores which, obviously, affects other features in the platform on anyhow. Otherwise, we wouldn't be having this conversation today.

Besides, if the switch only executes if the route isn't specified, then another reason why the suggested code should still not be pointed to specific routes either it's with or without the database.

head-dunce:
Maybe its due to an extension, but none of my OC3 seo urls specify a route.

TheCartpenter:
If none of the SEO URL specifies a route, then less the reason to manually define these routes, without the database, since it will only prevent further management from the OC admin as admins will have less controls via their SEO URLs that are automatically rewritten from the controller/startups especially since OC 4 allows extensions to be written on that specific location as opposed to previous OC releases.

chongshengdz:
> If none of the SEO URL specifies a route, then less the reason to manually define these routes, without the database, since it will only prevent further management from the OC admin as admins will have less controls via their SEO URLs that are automatically rewritten from the controller/startups especially since OC 4 allows extensions to be written on that specific location as opposed to previous OC releases.

this pull request does work, it does reduce almost half of the duplicate queries from database after testing.
need more work for the language=en-gb

chongshengdz:
> this pull request does work, it does reduce almost half of the duplicate queries from database after testing.
> need more work for the language=en-gb and route=product/search

this pull request does work, it does reduce almost half of the duplicate queries from database after testing.
please do more work for the language=en-gb and route=product/search

chongshengdz:
> Maybe its due to an extension, but none of my OC3 seo urls specify a route.

this pull request does work, it does reduce almost half of the duplicate queries from database after testing.
please do more work for the language=en-gb and route=product/search

chongshengdz:
before using this pull request, my database queries is over 10k-12k per second.
now reduce almost a half.
need more work to set the route for the **language=en-gb**

head-dunce:
> before using this pull request, my database queries is over 10k-12k per second. now reduce almost a half. need more work to set the route for the **language=en-gb**

I haven't submitted pull requests on here before, so I ended up having two pulls by mistake when I only wanted one. I'm currently updating my site to 3.0.4.0 with a fresh install, etc. so feel free to run with the idea in your own pull request.

chongshengdz:
> > before using this pull request, my database queries is over 10k-12k per second. now reduce almost a half. need more work to set the route for the **language=en-gb**
> 
> I haven't submitted pull requests on here before, so I ended up having two pulls by mistake when I only wanted one. I'm currently updating my site to 3.0.4.0 with a fresh install, etc. so feel free to run with the idea in your own pull request.

3.0.4.0 is too old boostrap and js.
4.0+ is much better.
you need more work to set the route for the language=en-gb, that would be perfect.

head-dunce:
> > > before using this pull request, my database queries is over 10k-12k per second. now reduce almost a half. need more work to set the route for the **language=en-gb**
> > 
> > 
> > I haven't submitted pull requests on here before, so I ended up having two pulls by mistake when I only wanted one. I'm currently updating my site to 3.0.4.0 with a fresh install, etc. so feel free to run with the idea in your own pull request.
> 
> 3.0.4.0 is too old boostrap and js. 4.0+ is much better. you need more work to set the route for the language=en-gb, that would be perfect.

Yeah, I need to get my site on a new server and I can much more quickly with 3.0.4.0 which did get some boostrap, etc updates. From what I saw, I was able to leave the language out of the url and the pages loaded. I only need one language anyway. I wasn't using a stock version of OC4 though, and I wrote a Perl script to try and shoehorn the OC3 db into it, so my results may not be inline with a stock install. Anyway, too much time wasted on these few lines of code that I'm not using anytime soon. Best of luck - I'm out!

PR #14132, opened 2024-09-20
enable "Print Shipping List" button at  "sale/order.info" page

jian2008:
problems to fix:   at order detail page "admin/index.php?route=sale/order.info", current version  can't click top-right button "Print Shipping List" 

 from  file "upload/install/controller/upgrade/upgrade_9.php" we can see "order" table's  `shipping_method`, `shipping_code`  has merge to one new field "shipping_method", and save as json string

as  "enable" check condition of  top-right "Print Invoice" button is  
`{% if not order_id %} disabled{% endif %}"`

so for  "Print Shipping List" , we  change "shipping_code" to "order_id", that is  the same check condition with "Print Invoice" button
`{% if not order_id %} disabled{% endif %}"`

### DIFF
--- a/upload/admin/view/template/sale/order_info.twig
+++ b/upload/admin/view/template/sale/order_info.twig
@@ -2,7 +2,7 @@
 <div id="content">
   <div class="page-header">
     <div class="container-fluid">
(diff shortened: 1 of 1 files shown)


danielkerr:
thersed an issue but this wont fix it

PR #14138, opened 2024-09-25
TWIG use_yield options

dreanei23:
this option is not in the documentation for version 3 of TWIG, but it is already in use: https://github.com/twigphp/Twig/blob/efc527ea65f0b12c6afda0a1e57e2f220e69afd2/src/Environment.php#L106

In version 4 of TWIG, the option will be true by default, but you can use it now

With this option, memory consumption is much lower and rendering is faster.

### DIFF
--- a/upload/system/library/template/twig.php
+++ b/upload/system/library/template/twig.php
@@ -107,7 +107,8 @@ class Twig {
 				'autoescape'  => false,
 				'debug'       => true,
 				'auto_reload' => true,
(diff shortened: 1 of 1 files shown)


danielkerr:
its not released yet

dreanei23:
@danielkerr and what prevents you from using use_yield=true now? 

It's already working

mhcwebdesign:
I agree with Daniel, we shouldn't use undocumented features. The Symfony Twig template engine is bad enough as it is.

PR #14140, opened 2024-09-27
Update default.php

mavelo-llc:
Remove PHP_SELF check to ensure admin sessions is accessible to catalog

Note: Master branch maintainers might not encounter this problem with OpenCart installed in a subfolder.

https://www.php.net/manual/en/function.setcookie.php If set to '/', the cookie will be available within the entire domain. If set to '/foo/', the cookie will only be available within the /foo/ directory and all sub-directories such as /foo/bar/ of domain.


stalker780:
This PR will break all stores in subfolders like mydomain.com/store1/, mydomain.com/store2/. All cookies will be stored for mydomain.com/

mavelo-llc:
Presume you’re referring to multi-store or multiple installs. Shouldn’t matter for single-store installs.

mavelo-llc:
Alternatively set `session_path` in each app config file...

catalog.php
`!empty($_SERVER['PHP_SELF']) ? rtrim(dirname($_SERVER['PHP_SELF']), '/') . '/' : '/';`

admin.php, install.php
`!empty($_SERVER['PHP_SELF']) ? rtrim(dirname(dirname($_SERVER['PHP_SELF'])), '/') . '/' : '/';`

danielkerr:
this will mess up

PR #14143, opened 2024-09-29
Compatibility with php 7.4 (storage moving, security.php)

opencartbot:
str_starts_with() is php 8 function, it does not allow to move the storage dir, on php 7.4 we get an error: 'Warning: Storage directory name must start with `storage`! e.g. `storage_`.'


stalker780:
https://github.com/opencart/opencart/blob/346ce60580de7b4c79d5f5469ded6189c8a4abb3/upload/system/helper/general.php#L76

stalker780:
Hard to suggest anything here.
Perhaps your new folder name does not start with "storage".
Check it for Cyrillic characters.

PR #14144, opened 2024-09-30
Issue with moving 'storage' on php 7.4

opencartbot:


chongshengdz:
you can upgrade to php 8.3, it loads much faster, php 7.4 is outdated.

PR #14154, opened 2024-10-02
[Master] Various bugfixes

mhcwebdesign:
Needed for the admin order edit


mhcwebdesign:
I won't create more pull requests until we a public beta, because it's a bit too frustrating using github while there are still some many changes by Daniel at the moment.

At any case, the order delete needs to be fixed, as well as the API, and we should also sort out the color styles for the frontend **.btn-primary** and **.btn-danger** buttons.

PR #14181, opened 2024-10-16
Update index.html DOM text reinterpreted as HTML

Shivam7-1:
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.


ADDCreative:
@Shivam7-1 There is no risk of injection as it being set with a string. Also when CodeMirror is updated your change will be lost.

AJenbo:
It's October, he wants free stuff from [DigitalOcean](https://hacktoberfest.com/)

PR #14200, opened 2024-10-29
Update common.js

stalker780:
Reported that timeout 50 does not work with Safari browser. Autocomplete item cannot be selected

But timeout 150 works

Can anyone test with Safari and confirm?


danielkerr:
fine

PR #14210, opened 2024-11-07
[3.0.x.x] Added fix.io currency converter engine

mhcwebdesign:
Useful for currencies not supported by the ECB.


osworx:
Eventually .. you should mention (help text) that calling this service is free only for 100 calls (as I read it per month).
Maybe add something like a check for this limit, otherwise the service will not work unless a paid plan is subscribed.
https://fixer.io/#pricing_plan

PR #14246, opened 2024-12-02
Added check if the information should be displayed in the footer

opencartbot:
Now we have a bug: all informations are added to the footer menu.


danielkerr:
yep

PR #14248, opened 2024-12-03
Make model class properties visible via proxy

billynoah:
Model class properties are not visible via proxy. If they are declared public, this commit makes them visible:
```
extension/opencart/catalog/model/module/example.php
```
```php
namespace Opencart\Catalog\Model\Extension\Opencart\Module;
class Example extends \Opencart\System\Engine\Model {
	public $my_public_prop = 100;
	private $my_private_prop = 200;
}
```
```
extension/opencart/catalog/controller/module/example.php
```
```php
namespace Opencart\Catalog\Controller\Extension\Opencart\Module;
class BestSeller extends \Opencart\System\Engine\Controller {
	public function test_properties() {
		$this->load->model('extension/opencart/module/example');

		// publicly visible
		$data['my_public_prop'] = $this->model_extension_opencart_module_example->my_public_prop;

		// not visible so result will still be null
		$data['my_private_prop'] = $this->model_extension_opencart_module_example->my_private_prop;
	}
}
```

### DIFF
--- a/upload/system/engine/loader.php
+++ b/upload/system/engine/loader.php
@@ -150,14 +150,13 @@ class Loader {
 
 		$proxy = new \Opencart\System\Engine\Proxy();
 
+		// pass the object to proxy in order to provide access to public properties
+		$proxy->setObject($object);
+		
 		foreach (get_class_methods($object) as $method) {
 			if (substr($method, 0, 2) != '__') {
 				$proxy->{$method} = $this->callback($route . '.' . $method);
 			}
-		}
-
-		foreach (get_class_vars($class) as $name => $value) {
-			$proxy->{$name} = $value;
 		}
 
 		// Store proxy object
--- a/upload/system/engine/loader.php
+++ b/upload/system/engine/loader.php
@@ -156,6 +156,10 @@ class Loader {
 			}
 		}
 
+		foreach (get_class_vars($class) as $name => $value) {
+			$proxy->{$name} = $value;
+		}
+
 		// Store proxy object
 		$this->registry->set($key, $proxy);
 	}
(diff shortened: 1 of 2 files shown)


mhcwebdesign:
Not a bug, that's how it should work!

billynoah:
What is the benefit of preventing read access to public class properties?

billynoah:
> Not a bug, that's how it should work!

This commit is an enhancement of the proxy class, not a bugfix.

If you disagree with the intended change can you please explain why? What is the benefit of preventing read access to public class properties?

mhcwebdesign:
> > Not a bug, that's how it should work!
> 
> This commit is an enhancement of the proxy class, not a bugfix.
> 
> If you disagree with the intended change can you please explain why? What is the benefit of preventing read access to public class properties?

Sorry, I may have misunderstood you. Private properties should not be accessible outside the model class. But yes, it's different with public properties, which should indeed have read and write access for their model class instances and/or the proxies.

billynoah:
@mhcwebdesign - actually my original commit only provides read access to public properties. After toying around with things for a bit I discovered that it's really not easy to provide write access in a way which will be readable by the model class methods as well. The last two commits get this working but I have to say it seems a bit overly complex. Please have a look and let me know if you think this is workable or if it can be simplified.

billynoah:
@danielkerr - looking for feedback on this feature. I think it's important to provide access to public class properties but I'm not positive that my approach here is ideal.

billynoah:
> Except that this proposal would also open public classes to the APIs.

How so exactly? Can you provide an example?

mScott224 (approved):

upload/system/engine/loader.php
```diff
@@ -150,6 +150,9 @@ public function model(string $route): void {
 
 		$proxy = new \Opencart\System\Engine\Proxy();
 
+		// pass the object to proxy in order to provide access to public properties
+		$proxy->setObject($object);
```
AJenbo, line 154: Do this via the constructor instead of a setter.

PR #14249, opened 2024-12-03
Specials fixes

stalker780:


danielkerr:
already added

stalker780:
@danielkerr I can see only date_added->type typo fixed.

There are still fixes to be applied

1. admin product_form.twig - lost Subtract option
2. catalog/model/catalog/product.php - lost calculation of percent/subtract/fixed special value

danielkerr:
cheers

PR #14254, opened 2024-12-04
Update product.php

stalker780:
Another attempt to fix specials and discounts
https://github.com/opencart/opencart/pull/14249


TheCartpenter:
```THEN (`p`.`price` - `pd2`.`price`)``` portion does not verify if the amount is higher than 0.00 especially since the admin-end does not verify the entered value either.

TheCartpenter:
Also for: THEN (`p`.`price` - `ps`.`price`) .

stalker780:
@TheCartpenter Don't see any problem. 
What if I want to set a negative value? 
I understand this will be not a discount. But why not? 
It is up to the store admin how he will use this value.

TheCartpenter:
Perhaps adding the + / - like product options already have may resolve this problem since negative values are more based on discounts than specials themselves but now that specials have been migrated with the discounts on the same database table, that may not be as obvious when looking at the equation whether it's for discounts or specials based on the SQL query you're also hitting compared to the cart library.

danielkerr:
i manually added

PR #14256, opened 2024-12-05
Fix URL encoding for product tags in OpenCart 3

head-dunce:
### Issue
Product tags with special characters (e.g., `&`, spaces) were not properly encoded, causing broken phrases

### Fix
Updated the product controller to use rawurlencode() and html_entity_decode() to handle special characters and HTML entities correctly.

### File Modified
- upload/catalog/controller/product/product.php

### Testing
- Tested with tags containing special characters (`&`, spaces, etc.) and verified correct URL encoding.
- Example: `Gas & Oil` now generates `tag=Gas%20%26%20Oil`.
https://www.carguygarage.com/index.php?route=product/search&tag=Gas%20%26%20Oil

### DIFF
--- a/upload/catalog/controller/product/product.php
+++ b/upload/catalog/controller/product/product.php
@@ -450,7 +450,7 @@ class ControllerProductProduct extends Controller {
 				foreach ($tags as $tag) {
 					$data['tags'][] = array(
 						'tag'  => trim($tag),
-						'href' => $this->url->link('product/search', 'tag=' . trim($tag))
+						'href' => $this->url->link('product/search', 'tag=' . rawurlencode(html_entity_decode(trim($tag), ENT_QUOTES, 'UTF-8')))
 					);
 				}
 			}


mhcwebdesign:
Thanks for this bugfix.

ADDCreative:
This doesn't completely fix the issue. You need to also change the encoding on the breadcrumb links, etc.

https://github.com/opencart/opencart/blob/e59e9bc57b7ef3a6402ff8b932d18c9016b3f6cb/upload/catalog/controller/product/product.php#L114

https://github.com/opencart/opencart/blob/e59e9bc57b7ef3a6402ff8b932d18c9016b3f6cb/upload/catalog/controller/product/product.php#L197

https://github.com/opencart/opencart/blob/e59e9bc57b7ef3a6402ff8b932d18c9016b3f6cb/upload/catalog/controller/product/product.php#L490

Maybe also use urlencode to be consistent with the links generated on the search page.
https://github.com/opencart/opencart/blob/e59e9bc57b7ef3a6402ff8b932d18c9016b3f6cb/upload/catalog/controller/product/search.php#L90

head-dunce:
Ah, yeah good catch on the breadcrumbs. I used rawurlencode instead of urlencode since it encodes spaces at %20 instead of + because %20 was how the spaces were in the url before the fix. I don't think it technically matters which way though since it's in the query.

mhcwebdesign:
@ADDCreative : You are right for these other links, e.g. breadcrumbs etc. Just updated and tested it, see https://github.com/opencart/opencart/pull/14257 . Was going for urlencode, as that's what is used in most cases in OpenCart. According to the [W3C](http://www.w3.org/Addressing/URL/uri-spec.html), both '+' or '%20' are valid for URL encoding.

PR #14258, opened 2024-12-08
Update db_schema.php

retrina:
Fix primary column of subscription_option table in install

"order_option_id" => "subscription_option_id"


danielkerr:
fixed already

PR #14310, opened 2025-01-01
French language

stasdrvn:
French language added and sql updated


stalker780:
Languages are installed as extensions in OC.
No one is going to support and update many languages in base core.

Ben-Ajenoui:
We are available for all updates.

Ben-Ajenoui:
> Languages are installed as extensions in OC. No one is going to support and update many languages in base core.
We are available for all updates.

stalker780:
Strange decision.
Many stores do not use even English language.
But it is required as the default language for missing translations. 

Why the base installation should contain hundreds of files for other languages which are never used?

Who will maintain correct translations for new language variables?

Approach with additional languages as extensions looks much more efficient. Each extension has a maintainer who can support his language.

danielkerr:
because i have one  staff that can speak 7 languages.

stalker780:
May I prepare a PR with Ukrainian language then?

Ben-Ajenoui:
> May I prepare a PR with Ukrainian language then?

Where would you like to have this PR in Ukrainian Language?

stalker780:
> Where would you like to have this PR in Ukrainian Language?

Here, in master branch OC 4.1

Ben-Ajenoui:
> > Where would you like to have this PR in Ukrainian Language?
> 
> Here, in master branch OC 4.1

That's not possible, mate. This place is actually for us as developers to communicate and help with the development of Opencart. What we do is for the public. We are aware that all the countries that use French language are in need of ecommerce stores in French. We also have plans to work on other languages as well.

stalker780:
Another strange decision.

OC is very popular in IT, ES, PT, BG, MD, PL, UA, GR, TR, RU... Hundreds of thousands stores. Why only FR and EN then?

Ben-Ajenoui:
> Another strange decision.
> 
> OC is very popular in IT, ES, PT, BG, MD, PL, UA, GR, TR, RU... Hundreds of thousands stores. Why only FR and EN then?

Next, would be Spanish. We want to thank you so much for your comments. We hope that you have a great 2025.

stalker780:
In this situation I may suggest to fill `oc_language` table with popular languages. At least each language would have fixed language_id. And make a check if language folder exists in extension or in basic installation. Make disabled langues disappear in admin forms.

At least this will solve many issues when migrating OC store DB when same language may have different language_ids.

pemaier:
@danielkerr Talented employees in companies is a very good thing! The decision to build up a business case on a worker who can leave the company ... belongs to the owner. 
But what what are the consequences for the community ... 
- if then the translators for this languages are not longer required and go somewhere else?
- if your employee change the company
- if customers like a different style of an language translation (personally, business style, non-gender,....)

I agree with you to serve customers/community with language files. That is a very positive initiative... it is good to get supported. Is there maybe a better way to support them with a language extension file? 
... and a central place for downloads beside the marketplace?
... with responsible a responsible person for each language like in other projects?
[occ_language_german.ocmod.zip](https://github.com/user-attachments/files/18292426/occ_language_german.ocmod.zip)

pemaier:
@Ben-Ajenoui Thank you that you take the care for the languages. German is appended for you as a start point.
German speaking customers will thank you your service. We hope that you have a great 2025.

Ben-Ajenoui:
> @Ben-Ajenoui Thank you that you take the care for the languages. German is appended for you as a start point. German speaking customers will thank you for your service. We hope that you have a great 2025.

I want to thank you so much for your kind comment. We will be looking at the demand for the German Language. At the moment we know for a fact that French and Spanish are our priority. We will look into it. Thank you and I hope you have a great 2025 using Opencart.

eka7a:
If multiple languages ​​are to be added by default, I think it would be better to move the language values ​​in the sql file to a separate sql file. en.sql, fr.sql, tr.sql etc.

pemaier:
@eka7a There is already a base solution integrated in OpenCart for your idea. The language editor need a small change that  it is not longer required to have a language file to be changed. 
1. I'm not sure what it mean for the performance when it is in the database.
2. But in each way it must be implemented in a form that a non expert user can handle the language install on a shop. 
3. I'm sure the customers and integrator's will be happy for a free language service from OpenCart. I only hope that the quality is OK. Right now I see several German translations ... in various quality levels. That's the reason why I'm doing it by myself till now.

May it is more interesting to implement a AI solution which is also translating the customer data also. (In a way like the Google translator solution on the marketplace.)

Various solutions are possible ... but I mean it should not be a quick shot without to consider the different advantages and hints as @stalker780 already said.

eka7a:
@pemaier I'm not talking about the language editor. I'm talking about opencart.sql file in the install.

pemaier:
I was also thinking about to integrate everything in database files. But i was looking around and no serious project integrates the language into the core.
So I suggest to even separate even the texts which are already in the database (status, ...) into text files. 

If as you suggest to put the language files for an installation process into a database ok but for the runtime the texts have to get exported into text files on the host system. That is a performance theme I already had during my SAP times.

osworx:
OpenCart had no additoinal languages beside English all the time since 1st publication.
And that is good as it is.

Shame that some so called "Translators" are selling their language packages, same goes for some who are stealing packages from others and publish that under their own name ( ..hmm @pemaier you know what about I'm talking !?) !

Adding now some languages per default to the core is not the right way.
Blows up the whole package, and needs more maintaince for the core.

OpenCart (Daniel !) should publish additional and qualified (!!) translations from native speakers with no cost finally!
Either by paying or hiring them.

Beside adding new languages to the core, which ones should be added per default?
According to this statistic https://en.wikipedia.org/wiki/List_of_languages_by_number_of_native_speakers shall the top 10 and most used languages included?
What then with variants some langauges have, e.g. as Austrian I really do not need the German (as spoken in Germany) language!
And should be those variants also be added?
What with English for example, currently only en-gb is included, but what about en-au, en-us, and so on .. ?

Finally (for example), when the core has de-de included, who will maintain de-at, or de-ch, de-li and so on?

**Better to leave languages outside the whole package, add only language tranlsations for the installation process and establish a mechanism to install additional languages on demand when needed.**

There is also a problem currently with several languages when added per default: each field offered for a language (see e.g. products with Description, Meta-*, etc.) has to be translated!

_So, better leave additional languages beside English (en-gb) outside of the core package.
And reduce the strings (the most common used in many files !) to a few (or one) file only (would reduce all tranlsations by approx. 1.000 lines!) which can be called when needed._

pemaier:
@osworx ... sorry but if you compare your German translation with my one .... it"s different. 
My customers from Germany and Swiss wanted to have a different style. Sorry but I was doing my translations from start to end by myself!!! 
Does we have some parts translated the same way - yes because it belongs to the situation that German language is not changeable and so we have tom use in some parts the same words. But nevertheless you are welcome to use my translation ([opencart-deutsch.de](https://opencart-deutsch.de)) ... even if you write your name under it. ;)
In exchange you may let me use your Austrian-German variant also?  :))

**This is the best real live sample what it means to have different local translation variants!!!**

In all other things I support you because we have the same view on the situation.

pemaier:
**To all of you a Happy New Year.**
I was thinking about a possible flexible and from many wanted process. You can find it here https://github.com/opencart/opencart/issues/14313

PR #14288, opened 2024-12-25
More PHPDocs on adding model examples

TheCartpenter:

PR #14312, opened 2025-01-01
Applied CS-Fixer and some PHPDocs corrections

TheCartpenter:


danielkerr:
m8 can u not come up with a better example

TheCartpenter:
> m8 can u not come up with a better example

I do have better examples than the one you originally suggested, yes: https://github.com/opencart/opencart/pull/14288#issuecomment-2561921029 . Now that you asked, I'll work on them.

PR #14323, opened 2025-01-03
Some more formate work

pemaier:
Admin complete


pemaier:
Some format work finished Catalog

PR #14324, opened 2025-01-04
More PHPDocs work on admin model examples

TheCartpenter:


TheCartpenter:
This commit: https://github.com/opencart/opencart/pull/14324/commits/01374d3c0e4f9e717559472ba5fe653a2b991a9e is whether to know if, still on this day, we're returning the translation ID since OCMod has been re-integrated or we still don't put it in the PHPDocs.

danielkerr:
ui just uploaded a new example on returns

TheCartpenter:
Since you're repeating the ```$this->load->model```, I presume the ```can be called``` at the top of each models would also have to be removed.

PR #14339, opened 2025-01-09
[3.0.x.x] Fixed php warning on Ionos server - request.php

condor2:


mhcwebdesign:
Already there in line 16 of system/library/request.php :

`	public $request = array();`

PR #14396, opened 2025-01-16
Removed duplicated code from step_2.twig

condor2:


ADDCreative:
magic_quotes_gpc was removed in PHP 5.4 so there is no point checking it at all.

PR #14431, opened 2025-01-23
Fix the max password length in 'en-gb' localisation

opencartbot:


condor2:
https://github.com/opencart/opencart/blob/master/upload/admin/language/en-gb/customer/customer.php#L107

Issue #14351, opened 2025-01-13
[master] Currency and langauge have different design view


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?

The text view and the position is different.

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

![image](https://github.com/user-attachments/assets/e80dd85c-4cfa-4383-b9fb-1728636998ba)

opencartbot:
Fixed

PR #14463, opened 2025-02-01
Fix view of currency and language in the header

opencartbot:
#14351


eka7a:
Please submit a clean pull request.

File oc_ocmod_example.ocmod.zip has nothing to do with this change.

PR #14478, opened 2025-02-03
Fix category edit

stalker780:
Editing category with childs destroys their seo_urls and paths

1. Try saving category `Components` on latest master
2. All child categories get broken paths
3. All child categories get broken urls

image

### DIFF
--- a/upload/admin/model/catalog/category.php
+++ b/upload/admin/model/catalog/category.php
@@ -130,19 +130,22 @@ class Category extends \Opencart\System\Engine\Model {
 			$this->model_catalog_category->addDescription($category_id, $language_id, $category_description);
 		}
 
-		// Old path
+		// Path
 		$path_old = $this->model_catalog_category->getPath($category_id);
+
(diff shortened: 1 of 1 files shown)


stalker780:
@opencartbot 
At least my solution fixes the destructive behavior. 
I understand that categories usually are not edited very often. But in my situation the content manager changed one line in the first level category description and saved it. As a result we got over 200 destroyed category seo_urls. And of course this was noticed much later, when we started getting 404 errors in Google Search console.

mhcwebdesign:
> @opencartbot At least my solution fixes the destructive behavior. I understand that categories usually are not edited very often. But in my situation the content manager changed one line in the first level category description and saved it. As a result we got over 200 destroyed category seo_urls. And of course this was noticed much later, when we started getting 404 errors in Google Search console.

I have created a pull-request for this which just uses the editCategory logic taken from OC 4.0.2.3 where this issue didn't exist, see https://github.com/opencart/opencart/pull/14501

PR #14501, opened 2025-02-12
Update editCategory

mhcwebdesign:
When editing e.g. category 'Components' and then saving it, then all the SEO keywords from its sub-categories, such as 'Components > Mice and Trackballs', 'Components > Monitors', etc., are accidentally deleted.

This bugfix restores the old editCategory function taken from OC 4.0.2.3 where this issue didn't exist.

### DIFF
--- a/upload/admin/model/catalog/category.php
+++ b/upload/admin/model/catalog/category.php
@@ -122,172 +122,124 @@ class Category extends \Opencart\System\Engine\Model {
 	 * $this->model_catalog_category->editCategory($category_id, $category_data);
 	 */
 	public function editCategory(int $category_id, array $data): void {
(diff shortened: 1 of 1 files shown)


stalker780:
OC4.1 have no `column` and `top` fields in `oc_category`

Looks like my solution does the same, but uses new OC4.1 model methods to collect the path and seo_url data instead of direct SQL queries.

mhcwebdesign:
> OC4.1 have no `column` and `top` fields in `oc_category`
> 
> Looks like my solution does the same, but uses new OC4.1 model methods to collect the path and seo_url data instead of direct SQL queries.

OC 4.1.0.0 does not have the 'banner' and 'top' fields anymore in the 'oc_category', but there is still a column field.
I'll try your solution to see whether it does the job.

PR #14518, opened 2025-02-21
[offer] Style of disabled list items

batumibiz:
In this PR, I propose to change the style of highlighting the disabled table list items.
If the PR is accepted, I will prepare a continuation and finalize the rest of the admin panel's tabular lists.
If you don't like it, then just close this PR.

Description of the offer
- To highlight the disabled table rows, the CSS class `.list-disabled` was added
- The code is slightly simplified: it is enough (by the condition) to add a CSS class to the `<TR>` tag
- Lists look more strict and clean
- If desired, the style of all disabled elements can be changed by simply editing the CSS style
- Previously, the line with the product variant was highlighted with a yellow background, but the variant is already highlighted with a yellow button, which is clearly visible.

r1_c1


danielkerr:
good idea!

PR #14535, opened 2025-02-23
Update order.php

TheCartpenter:


mhcwebdesign:
Another useless phpdoc  which adds nothing useful.

@danielkerr : Do we really need this?

TheCartpenter:
> PHP-cs-fixer can be configured to strip all these out via the `no_superfluous_phpdoc_tags` option, it would be nice to do so, but last time daniel asked that the opposite was done (`phpdoc_add_missing_param_annotation`).

Some of the methods are intentionally left out of PHPDocs for specific reasons. By putting automation as suggested, cs-fixer would also include the PHPDocs where they've been intentionally left out and may not be the direction the core is heading.

AJenbo:
> > PHP-cs-fixer can be configured to strip all these out via the `no_superfluous_phpdoc_tags` option, it would be nice to do so, but last time daniel asked that the opposite was done (`phpdoc_add_missing_param_annotation`).
> 
> Some of the methods are intentionally left out of PHPDocs for specific reasons. By putting automation as suggested, cs-fixer would also include the PHPDocs where they've been intentionally left out and may not be the direction the core is heading.

It's already there, we did this over a year ago. Your wrong, the only place where it has been missed is `customer::getAuthorize()`.

TheCartpenter:
> > > PHP-cs-fixer can be configured to strip all these out via the `no_superfluous_phpdoc_tags` option, it would be nice to do so, but last time daniel asked that the opposite was done (`phpdoc_add_missing_param_annotation`).
> > 
> > 
> > Some of the methods are intentionally left out of PHPDocs for specific reasons. By putting automation as suggested, cs-fixer would also include the PHPDocs where they've been intentionally left out and may not be the direction the core is heading.
> 
> It's already there, we did this over a year ago. Your wrong, the only place where it has been missed is `customer::getAuthorize()`.

Of course, which is why each of these locations don't have PHPDocs on top of these specific methods:

- admin/controller/sale/order.php
- system/library/cart/api.php
- system/library/curl.php
- system/library/request.php

TheCartpenter:
> > Of course, which is why each of these locations don't have PHPDocs on top of these specific methods:
> > ```
> > * admin/controller/sale/order.php
> > * system/library/cart/api.php
> > * system/library/curl.php
> > * system/library/request.php
> > ```
> 
> What are you talking about, these follow the php-cs-fixer configuration. The latest change to them is even me running PHP-cs-fixer: image

Not sure what was done 4 months ago but, by looking at this file, I still see empty PHPDocs on top of the methods from the core: https://github.com/opencart/opencart/blob/master/upload/system/library/curl.php .

AJenbo:
> Not sure what was done 4 months ago but, by looking at this file, I still see empty PHPDocs on top of the methods from the core: https://github.com/opencart/opencart/blob/master/upload/system/library/curl.php .

I just told you, I ran PHP-cs-fixer on them 4 months ago, and the automation is what has resulted in what you now saying can't be done with automation.

The reason those methods do not have PHPDoc is because there paramters didn't have any types. Not because of some decision to not have it. If you think other wise point to where that decision is being made instead of just guessing and not even knowing what I have done previously.

TheCartpenter:
> > Not sure what was done 4 months ago but, by looking at this file, I still see empty PHPDocs on top of the methods from the core: https://github.com/opencart/opencart/blob/master/upload/system/library/curl.php .
> 
> I just told you, I ran PHP-cs-fixer on them 4 months ago, and the automation is what has resulted in what you now saying can't be done with automation.
> 
> The reason those methods do not have PHPDoc is because there paramters didn't have any types. Not because of some decision to not have it. If you think other wise point to where that decision is being made instead of just guessing and not even knowing what I have done previously.

It wasn't exactly guessing but it was said in the passed on this repository as well as noticing closed commits about suggestions made over specific methods with PHPDocs. If that decision has been reviewed in order to put those PHPDocs nowadays, I am sure not opposing to that, as quite the opposite; The faster these can be done, the better.

PR #14539, opened 2025-02-24
Fix PHP 8.0 support

AJenbo:
Looks to have been broken here: ee59c8a562199014b3fcfb184d9d6c8c3f9b5a32, because 8.1 was used to update the dependencies.

Also:
- Fix code style
- Update PHPStan
- Run checks on PHP 8.4

If you send me a new invite I might fix some of the issues reported.


TheCartpenter:
```$data['stores']``` and ```$price``` is intentionally repeated rather than using ```+=```. As for system/vendor.php file, the modifications suggests to use new blank lines between IF and foreach statements which doesn't follow the same direction as the core.

AJenbo:
It does follow the same style, the same code style is applied everywhere by php-cs-fixer this isn't manual edits.

TheCartpenter:
> It does follow the same style, the same code style is applied everywhere by php-cs-fixer this isn't manual edits.

I wasn't implying it was manual editing. You may have followed the changes with cs-fxer but what I mentioned above is that ```$data['stores']``` and ```$price``` are intentionally repeated with the affected variable.

AJenbo:
I was replying to the part you now deleted:

> As for system/vendor.php file, the modifications suggests to use new blank lines between IF and foreach statements which doesn't follow the same direction as the core.

TheCartpenter:
> I was replying to the part you now deleted:
> 
> > As for system/vendor.php file, the modifications suggests to use new blank lines between IF and foreach statements which doesn't follow the same direction as the core.

Ok, no problem.

AJenbo:
> `$data['stores']` and `$price` are intentionally repeated rather than using `+=`.

Where was/when this decided, why have it different then every where else?

TheCartpenter:
> > `$data['stores']` and `$price` are intentionally repeated rather than using `+=`.
> 
> Where was/when this decided, why have it different then every where else?

I think this was decided before the re-integration of OCMod but I am not entirely sure if repetitive variables can be used as opposed to cs-fixer suggests, since. It was decided before, however, which is why I tend to keep that line clean from the change, until now.

AJenbo:
It was added here: https://github.com/opencart/opencart/commit/975073a8a913d9ec167827a827dde68a93e65473
It doesn't appear that there is any particular reason, just a human habit. Don't say that it's intentional if you don't actually know.

TheCartpenter:
> It was added here: [975073a](https://github.com/opencart/opencart/commit/975073a8a913d9ec167827a827dde68a93e65473) It doesn't appear that there is any particular reason, just a human habit. Don't say that it's intentional if you don't actually know.

Well, according to these changes, it looks pretty much to me those were intented changes.

Issue #14480, opened 2025-02-03
4.1.0.0 OCMOD Not Working for Specific Case


ocfreeext:
**What version of OpenCart are you reporting this for?**
4.1.0.0

**Describe the bug**

namespace Opencart\Admin\Controller\Catalog;

OCMOD changes WORK inside function index():

	<file path="admin/controller/catalog/product.php">
		<operation>
			<search><![CDATA[public function index(): void {]]></search>
			<add position="after"><![CDATA[
			echo('It is working!');
			]]></add>
		</operation>		
	</file>

But the same OCMOD does NOT work inside function form() :

	<file path="admin/controller/catalog/product.php">
		<operation>
			<search><![CDATA[public function form(): void {]]></search>
			<add position="after"><![CDATA[
			echo('It is working!');
			]]></add>
		</operation>		
	</file>

Although ocmod cache file (extension\ocmod\admin\controller\catalog\product.php) is properly created in both cases.

leandrorppo82:
Hi,
I also noticed that OCmod is not working in the front-end controller and language either.

**Example:**
`<file path="catalog/language/en-gb/common/header.php">
          <operation>
            <search><![CDATA[
      <?php
      ]]></search>
            <add position="after"><![CDATA[ 
//XML  LRPPO              
$_['text_logged']  = 'Hi, test';      
    ]]></add>
       </operation>
</file>`

**And**

`<file path="catalog/controller/common/header.php">   
<operation>
            <search>
                <![CDATA[Opencart\Catalog\Controller\Common;]]>
            </search>
            <add position="replace">
                <![CDATA[Opencart\Catalog\Controller\Extension\Ocmod\Common;]]>
            </add>
        </operation> 
<operation>
           <search><![CDATA[
          $this->load->language('common/header');
      ]]></search>
            <add position="after"><![CDATA[ 
         //XML LRPPO
          $data['test'] = $this->document->getScripts('test');               
          ]]></add>
       </operation>
</file>`

It's creating the cache, but it's not working. Inserting the codes directly into the native files usually works.
Did you get to do these tests?

Thank you for your attention!

leandrorppo82:
Hi,
I also noticed that OCmod is not working in the front-end controller and language either.

**Example:**
```php
<file path="catalog/language/en-gb/common/header.php">
          <operation>
            <search><![CDATA[
      <?php
      ]]></search>
            <add position="after"><![CDATA[ 
//XML  LRPPO         
[...]

Issue #14326, opened 2025-01-04
[master] Modification system: calling functions from a controller or model file does not load the modification


nmpetkov:
What works:
- When the modifications are refreshed, the modified files for controllers, models and twig files appear in the extension/ocmod folder.
- When loading a page without a specific controller function, for example ```admin/index.php?route=catalog/product```, the modified files for controller and twig are loaded correctly.

What doesn't work:
- When loading a specific function in a controller, for example ```admin/index.php?route=catalog/product.list``` (clicking on pagination) – in this case, the modified controller is not loaded, but the original (unmodified) one.
- Calling functions from a model file does not load the modified file, but uses the original one.

danielkerr:
u need to mod the pagination file or do an override.

nmpetkov:
@danielkerr Pagination is used here just as un example. The problem is that modification system is not working as expected!

danielkerr:
Pretty sure it's not expected. Use the events if u need to override stuff

FelipoAntonoff:
@danielkerr In this case, the use of OCMOD should be recommended, after all, it is designed to replace things and generate a cache of them.
While Events are basically for an action or a trigger, such as sending an email after a customer registers.

I would recommend that Opencart 4 have the same resources in OCMOD as Opencart 3 and correct the detail of the editor theme that when editing through the editor stops getting an OCMOD for the edited file.

The purpose of OCMOD is to be able to modify any PHP and Twig file, to avoid having to change the original files.

PR #14622, opened 2025-03-02
Fix for the modification system

vitorsreis:
Hello,

I needed to make simple modifications to the admin/catalog files using OCMOD but was unsuccessful.

```xml
<file path="admin/controller/catalog/category.php">
  <operation>
    <search index="0">
      <![CDATA[public function form(): void {]]>
    </search>
    <add position="after">
      <![CDATA[exit(__FILE__ . ':' . __LINE__);]]>
    </add>
  </operation>
</file>
```

I noticed that the files are correctly created in `<root>/extension/ocmod/...`, but they are not being loaded.

After analyzing `<root>/system/engine/autoloader.php`, I confirmed that they are indeed not being loaded. I am not sure if this was a design decision or an actual issue. However, I found some open issues that are directly related to the problem I am experiencing:
- https://github.com/opencart/opencart/issues/14326
- https://github.com/opencart/opencart/issues/14480
- ... among many others

This is a very important feature, and considering this, I propose a solution.

### DIFF
--- a/upload/system/engine/autoloader.php
+++ b/upload/system/engine/autoloader.php
@@ -76,7 +76,7 @@ class Autoloader {
 		}
 
 		if (isset($file) && is_file($file)) {
-			include_once($file);
+			include_once(modification($file));
 
 			return true;
 		} else {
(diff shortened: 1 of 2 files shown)


danielkerr:
its done differently now. its uses the start-up to over ride directory.

vitorsreis:
> its done differently now. its uses the start-up to over ride directory.

Daniel, sorry for my ignorance, but how exactly is this done? I couldn't find any docs.

How would I modify something in the main files?

If possible, could you provide an example for the XML of this commit

PR #14632, opened 2025-03-05
Added extension language and worked on fr-fr core

TheCartpenter:


TheCartpenter:
Also replaced apostrophized characters with escaped single-quotes. There are still some in the JS core and not sure if they'd have the same effects as PHP files would have.

PR #14633, opened 2025-03-05
Added extension languages and worked on core fr-fr

TheCartpenter:
All set.


TheCartpenter:
There we go. The language keys are now synched between the en-gb and fr-fr. From this point, moving forward, things should get easier.

PR #14634, opened 2025-03-05
Term correction on Expédition (Shipping)

TheCartpenter:


TheCartpenter (commented):
Done.

TheCartpenter:
@danielkerr: Can you please accept this? I need to work in a new patch on the TWIG files for PSR standards. Right now, the MB does not meet PSR compliance as the same as PHP files and there's a lot of work that's been done on this patch right now. I am done with this one.

TheCartpenter:
Changed my mind for TWIG indents. It's still half 2 indents and half user-defined.

PR #14647, opened 2025-03-08
Check OCMOD classes in Factory

alex93x:
Checks whether the modified class exists and loads it instead of the original. However, there is still no way to use the modified classes of system libraries connected in framework.php

### DIFF
--- a/upload/system/config/admin.php
+++ b/upload/system/config/admin.php
@@ -23,11 +23,11 @@ $_['error_display']     = true;
 
 // Actions
 $_['action_pre_action'] = [
+	'startup/extension',
 	'startup/setting',
 	'startup/session',
 	'startup/language',
 	'startup/application',
-	'startup/extension',
 	'startup/startup',
 	'startup/error',
 	'startup/event',
--- a/upload/system/config/catalog.php
+++ b/upload/system/config/catalog.php
@@ -20,6 +20,7 @@ $_['session_engine']     = 'db'; // db or file
 
 // Actions
 $_['action_pre_action']  = [
+	'startup/extension',
 	'startup/setting',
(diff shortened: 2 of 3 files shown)


TheCartpenter:
I am not sure if this suggestion would be ideal, since, according to what the code describes in the change, is that OCMod would prioritize the system/config profiles rather than prioritizing on the core prior to load the extensions. As for the ```class_exists``` in the OC Engine, it also looks like it would check OCMod first rather than loading the core classes.

As a result, it would mean that 80% would be prioritized by extensions and 20% from the core rather than keeping the other way around. It would also alter the ways the controller/startup folders work originally...

alex93x:
Version 3 checked for the presence of a modified file, now it will check for the presence of a class, unless there is another way to override existing functionality. The problem is that the release announced the return of OCMOD, but in fact it is impossible to use it, except to modify the template.

Because of this, developers are in no hurry to adapt their modules, and as long as there are no modules, users will not switch to the new version.

Translated with DeepL.com (free version)

TheCartpenter:
While I understand this suggestion has been posted as a way to reintroduce the extension developers back on track. However, I am pretty sure there are other ways to reintroduce OCMod without the need to integrate such engine into the OC Engine itseld as the controller/startup files already demonstrate. Specific startup files may simply need some adjustments to fit the requirements of the extension developers later on.

danielkerr:
@

> Checks whether the modified class exists and loads it instead of the original. However, there is still no way to use the modified classes of system libraries connected in framework.php

there is a way.

you load the library files using $this->load->library->get('extension/code/library')

danielkerr:
you could replace the object with another object in the registry. this is basic stuff.

PR #14648, opened 2025-03-08
Added sanitized int on some config get IDs

TheCartpenter:


AJenbo:
@TheCartpenter can you just remove PHP 8.0 support if you keep breaking it?

image

Like why do you guys not even bother to look at the failing tests and just keep breaking more stuff.

mhcwebdesign:
According to the docs at https://github.com/opencart/opencart#opencart the master branch is supposed to support PHP version 8.0, too! I just keep ignoring the failing tests from PHPStan.

AJenbo:
It doesn't just run phpstan and this pr made the test fail at installing on php 8.0.

@TheCartpenter keeps updating dependencies with a different version than the minimum supported version and so breaks computability routinely.

7.4 used to also be supported and getting broken and so was eventually just removed. So I guess removing 8.0 would solve the issue of him using 8.1 when updating dependencies.

The phpstan test also used to pass, until broken commits where pushed to master.

mhcwebdesign:
It only keeps mainly complaining about the aws/aws-sdk-php and twig/twig wanting PHP 8.1 or later.

We need to make a decision whether to officially drop PHP 8.0 support, and update the docs accordingly.

mhcwebdesign:
Also, most of the PHP docs are useless, see for example this one:

```
	/**
	 * Success
	 *
	 * @return void
	 */
	public function success(): void {
```

TheCartpenter:
> It doesn't just run phpstan and this pr made the test fail at installing on php 8.0.
> 
> @TheCartpenter keeps updating dependencies with a different version than the minimum supported version and so breaks computability routinely.
> 
> 7.4 used to also be supported and getting broken and so was eventually just removed. So I guess removing 8.0 would solve the issue of him using 8.1 when updating dependencies.
> 
> The phpstan test also used to pass, until broken commits where pushed to master.

It wasn't an intended decision. I just happened to leave it there from the previous vendor updates unintentionally. However, by reading both versions of the story on the topic, whether we should increase those slight PHP minimum requirements to 8.1, as per the vendors, aside from the fact there might still be extensions from the OC Marketplace that are still using deprecated versions, I really think we should increase that requirement to meet the industry standards and requirements since 2025 has arrived.

The same story has been discussed prior with PHP 7.4, it just happened to be on the PHP 8.0's turn this year.

AJenbo:
If you instead update dependencies using the lowest PHP version OC is supposed to support then it will stick to maintenance branches that are compatible with that version, so I really think you should go do that so that it's not broken on 8.0 until you can get Daniel convinced that 8.0 support should be dropped, and at that point update the readme and other parts of OC to require PHP 8.1 instead of breaking support for older php versions again and again and again and again and again and again and again and again

TheCartpenter:
> If you instead update dependencies using the lowest PHP version OC is supposed to support then it will stick to maintenance branches that are compatible with that version, so I really think you should go do that so that it's not broken on 8.0 until you can get Daniel convinced that 8.0 support should be dropped, and at that point update the readme and other parts of OC to require PHP 8.1 instead of breaking support for older php versions again and again and again and again and again and again and again and again

By doing that, it would go the other way around from the previous replies you've mentioned on this PR: https://github.com/opencart/opencart/pull/13971#issuecomment-2138293216 where I've been rather convinced to update vendors packages without the use of dependencies. Otherwise, it would also affect PHP versions from the vendors' requirements.

AJenbo:
I think you just don't understand, but it's very simple:
**Use PHP 8.0 when running composer update**

That's all you have to do to not break things. That's what I have been trying to explain to you each time this happens.

TheCartpenter:
By setting PHP to 8.0, we'd be avoiding this information from AWS: https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-php-runtimes-8-0-x-and-below-in-the-aws-sdk-for-php .

TheCartpenter:
> I think you just don't understand, but it's very simple: **Use PHP 8.0 when running composer update**
> 
> That's all you have to do to not break things. That's what I have been trying to explain to you each time this happens. If you use the minimum supported version when running composer then it will take care of making sure what's installed is compatible with that version.

I've downgraded twig/twig and aws/aws-sdk-php back to their previous versions, now: https://github.com/opencart/opencart/pull/14752/commits/ae32ccd6a2ac750c9afb1db2b31953fc878ad6b9 .

PR #14656, opened 2025-03-11
Add barcode search functionality to product list

fmarslan:
#### Description
This PR introduces a new feature that allows users to search for products using barcode numbers in the product list. The search functionality is case-insensitive and performs a "contains" search on the entered barcode number. The search checks the following fields:
- SKU
- UPC
- EAN
- JAN
- ISBN
- MPN

If the entered barcode number matches any of these fields, the corresponding product will be found.

#### Benefits
This feature is particularly useful for sellers who use a barcode system, making it easier and more practical to find products.

#### Testing
Tested with MariaDB version 11.5.2.

image

### DIFF
--- a/upload/admin/controller/catalog/product.php
+++ b/upload/admin/controller/catalog/product.php
@@ -21,6 +21,12 @@ class Product extends \Opencart\System\Engine\Controller {
 		} else {
 			$filter_name = '';
 		}
+		
+		if (isset($this->request->get['filter_barcode'])) {
+			$filter_barcode = $this->request->get['filter_barcode'];
+		} else {
+			$filter_barcode = '';
+		}
 
 		if (isset($this->request->get['filter_model'])) {
 			$filter_model = $this->request->get['filter_model'];
@@ -90,6 +96,10 @@ class Product extends \Opencart\System\Engine\Controller {
 			$url .= '&filter_name=' . urlencode(html_entity_decode($this->request->get['filter_name'], ENT_QUOTES, 'UTF-8'));
 		}
 
+		if (isset($this->request->get['filter_barcode'])) {
+			$url .= '&filter_barcode=' . urlencode(html_entity_decode($this->request->get['filter_barcode'], ENT_QUOTES, 'UTF-8'));
+		}
+
 		if (isset($this->request->get['filter_model'])) {
 			$url .= '&filter_model=' . urlencode(html_entity_decode($this->request->get['filter_model'], ENT_QUOTES, 'UTF-8'));
 		}
@@ -153,6 +163,7 @@ class Product extends \Opencart\System\Engine\Controller {
 		$data['list'] = $this->load->controller('catalog/product.getList');
 
 		$data['filter_name'] = $filter_name;
+		$data['filter_barcode'] = $filter_barcode;
 		$data['filter_model'] = $filter_model;
 		$data['filter_category_id'] = $filter_category_id;
 		$data['filter_manufacturer_id'] = $filter_manufacturer_id;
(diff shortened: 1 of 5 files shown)


TheCartpenter:
This code change suggests to use multiple product fields from a single filter whereas the product filters from the catalog/model/product already handles. Barcodes should be handled separately where no field combinations should be used for each features that need to be built-in.

fmarslan:
We understand that current filters offer detailed search options, which meet general needs. However, in the order preparation process, using a barcode scanner requires searching different fields each time, slowing things down. In many POS and ERP systems, all barcode types are searched through a single field, making product identification faster and improving efficiency.

Handling barcode management separately from other product fields and providing a unified barcode search feature is common in similar systems. This method streamlines order processing and enhances user experience.

For example erpnext

barcode define screen
image

order create screen
image

TheCartpenter:
What you are suggesting is the implementation of search groups where one product field would not need to affect multiple fields per subjects that may need to be implemented in the catalog model product file. Otherwise, when each users are acquiring a new feature to be built-in, the catalog model product would expand in any inconceivable ways whereas each product fields would set classifications on multiple meanings of different searches that could affect the same product fields and that would create more confusion, core-wise, as it would by extensions.

osworx:
Searching after / by ISBN or EAN/GTIN should be added into the core.
As @fmarslan mentioned, many users search after one of them and when the system is able to filter, it will be a great benefit for all (users and customers).

For example, I've made a "Stock- & Inventory Manager" for OpenCart a few weeks ago.
It use also barcode, generating, printing, etc.
And searchable for .. procedure is similiar this request ..

danielkerr:
i think we need a sperate table for product code types

fmarslan:
Yes, having a separate table for product code types could be useful, especially if users need to add different codes. But it’s something to think through carefully.  

For ERP-like systems, this is a must, but for OpenCart, it’s not really needed right now. A new table would make queries more complex, slow things down due to joins, and add extra work for development and maintenance.  

GTIN already includes EAN, UPC, and JAN, so maybe this part could even be simplified further. In most cases, OpenCart users don’t need multiple GTINs for the same product. On the other hand, SKU and MPN can have different values at the same time. In general, the cleanest approach is to use SKU, MPN, and a single GTIN number for each product. The current PR already does a good job without adding extra cost, especially for searches.  

If multiple GTINs are allowed, the next step will probably be handling different units and pricing per unit, which could make things even more complicated.

danielkerr:
https://en.wikipedia.org/wiki/Universal_Product_Code
Universal Product Code = barcode

danielkerr:
Global Trade Item Numbers

should i just call the new table product_code

TheCartpenter:
> Global Trade Item Numbers
> 
> should i just call the new table product_code

```product_type``` . Zen-Cart already has this feature built-in as well a bit of other platforms.

TheCartpenter:
Except that, creating a barcode table, will only be about one subject whereas product types can refine the search for many more features that are linked with the main products.

fmarslan:
Each variant of the product needs to have a separate GTIN number. OpenCart, with its current structure, already creates a separate product record for each variant, so there won't be any issues here. When searching, users will get the correct results whether they search for the parent product or the variant. No extra processes like joins are needed, and the current filtering system will continue to work as it should.

There might be a discussion about adding more detailed filters, but adding too many details could complicate the system. It would be helpful to start a discussion by gathering specific user needs. Understanding what features people truly need is important for creating a more efficient solution. Otherwise, the current simple and efficient design could be disrupted by unnecessary complexity.

As an example, the filtering feature used in the Frappe/ERPNext system is both useful and minimalist. Instead of overcomplicating things, it would be healthier to open a discussion with user requests in mind.

In conclusion, the current design easily meets the needs at hand. It would be more efficient to improve this basic structure without creating unnecessary complexity, keeping future needs in mind.

fmarslan:
What is your decision about this feature? Will you continue with the child table?

danielkerr:
its hard to think of the table name. maybe product_code_type

fmarslan:
if you use for just codes, may be product_barcode and product_code_id field name may be barcode_name

fmarslan:
@danielkerr, there is no barcode filter on the product list page. If you're not working on this now, I can take care of the filter.

PR #14667, opened 2025-03-14
Added identifiers in fr-fr

TheCartpenter:


TheCartpenter:
admin/language/en-gb/localisation/product_identifier.php is also missing.

PR #14668, opened 2025-03-15
OCMOD fixes for middlewares and models

alex93x:
Fixed the modifiication.model event, added it to the action_event config, create a middleware wrapper where the modification.controller event is triggered


TheCartpenter:
By looking at the code 'itself', it is not bad. It may address 'some' good points in it. However, the analysis of these modifications are as followed:

1. It is another direction going straight to the engine, especially since it is then loaded as a ```pre_action``` in the system/framework .
2. The conventional names are a bit... concerning especially to be put on an e-commerce platform
3. system/config/catalog.php file uses the conventional name like: 'model' for the debugs which represents OCMod in this case

Is there a specific reason for these conventional names and also to modify most of the engine to achieve one goal?

@AJenbo: I would also like to bring you to this discussion, since it was already discussed before that the OC 3 and OC 4 engine did had ... it's differences about the PHPStan diagnostic levels. Therefore, by considering such extent of modifications such as these ones, would that also rectify the situation, would it be unchanged, or more importantly, simply making it worst?

upload/system/framework.php
```diff
@@ -5,6 +5,11 @@
 $autoloader->register('Opencart\Extension', DIR_EXTENSION);
 $autoloader->register('Opencart\System', DIR_SYSTEM);
 
+// Register OCMOD
+$this->autoloader->register('Opencart\Catalog\Controller\Extension\Ocmod', DIR_EXTENSION . 'ocmod/catalog/controller/');
```
vitorsreis, line 9:
```php
$this->autoloader->register('Opencart\\' . APPLICATION . '\Controller\Extension\Ocmod', DIR_EXTENSION . 'ocmod/' . strtolower(APPLICATION) . '/controller/');
```

upload/system/framework.php
```diff
@@ -5,6 +5,11 @@
...
+$this->autoloader->register('Opencart\Catalog\Model\Extension\Ocmod', DIR_EXTENSION . 'ocmod/catalog/model/');
```
vitorsreis, line 10:
```php
$this->autoloader->register('Opencart\\' . APPLICATION . '\Model\Extension\Ocmod', DIR_EXTENSION . 'ocmod/' . strtolower(APPLICATION) . '/model/');
```

vitorsreis (changes requested):
The approach seems to be functional despite adding a bit more complexity; it would achieve the objective.

I only suggest the following modifications to make it truly functional.

upload/catalog/controller/event/modification.php
```diff
@@ -48,14 +48,22 @@ public function controller(string &$route, array &$args): void {
 	 * @return void
 	 */
 	public function model(string &$route, array &$args): void {
+		$pos = strrpos($route, '.');
```
AJenbo, line 51: use str_contains or judging by the next line str_starts_with

AJenbo:
> did has

Sorry I don't really understand what you are trying to say. I'm guessing your are referring to the issue that it's ambiguous if something is operating on the Admin or Catalog namespace.

While this issue might be caused by the same ambiguity I don't think the proposed solution would address it in a way that solves it in general.

TheCartpenter:
> > did has
> 
> Sorry I don't really understand what you are trying to say. I'm guessing your are referring to the issue that it's ambiguous if something is operating on the Admin or Catalog namespace.
> 
> While this issue might be caused by the same ambiguity I don't think the proposed solution would address it in a way that solves it in general.

It then also covers the answer on why we shouldn't need additional engines for each features in the core.

upload/system/framework.php
alex93x: fixed it

upload/catalog/controller/event/modification.php
alex93x: implemented this

vitorsreis (approved):
great!

TheCartpenter:
If we have to conclude that a: ```*$``` at the end of a ```preg_replace``` would fix it as much as setting a constant name in a ```namespace```, I'm pretty sure this kind of code has already been used on this location in the system/engine/factory.php file:

```$class = 'Opencart\\' . $this->registry->get('config')->get('application') . '\Controller\\' . str_replace(['_', '/'], ['', '\\'], ucwords($route, '_/'));``` . Therefore, it doesn't seem like it requires a defined constant...

TheCartpenter:
So to add, on the topic, it is also important to note from system/library/customer.php file:

```private object $request; // Do not add namespace as it stops devs being able to extend classes```

alex93x:
> So to add, on the topic, it is also important to note from system/library/customer.php file:
> 
> `private object $request; // Do not add namespace as it stops devs being able to extend classes`

Not quite sure how this applies to the current situation. The solution to the above remark is usually to use the interfaces

danielkerr:
m not accepting this. ocmod is not a good system to use. everything can be achieved using the event system. im think of having the ocmod system as an extension now. so people can modify it separately.

mhcwebdesign:
>  'Ka Extensions' uses class inheritance and that is the best approach for extending a store functionality. If you remove OCMOD from Opencart, people will juse use VQMOD as it was in Opencart 1.5.x

The main issue with class extensions is the scenario where different extensions happen to extend the same core class, in which case we'd need a dynamic chain of class extensions. Another issue is the way OpenCart 4 uses non-folder based namespaces for extensions. I started working on a new override engine for OpenCart 4 which will address both issues.

For the time being, we use event handlers and occasionally simple decorator design patterns for modifying or extending core classes for our OpenCart 4 extensions.

PR #14675, opened 2025-03-17
Applied CS-Fixer

TheCartpenter:


TheCartpenter:
@daniellkerr: If these two PRs are to be considered:

https://github.com/opencart/opencart/pull/14677
https://github.com/opencart/opencart/pull/14676

Make sure to accept this one before these other twos, since mine does not have the removal requests for Germany in it.

TheCartpenter:
Note: Since I am also not around that area, and that the list of additions to the zones is quite small, it may also be quickly reviewed based on their locations which all includes the ISO 3611-2.

pemaier:
To include the ISO-3166-2 codes in this table is not working well because it is required to be in a separate table for maintenance reasons and for more flexibility. And it is working like this because of the language variants (e.g. de-DE, de-AT, de-CH, ...)
To build a link into to the table is easy we have all what we need to have the reference. e.g. iso-3166-2:DE or iso-3166-2:CH for the German language used in Germany and Switzerland. For the French language  e.g. iso-3166-2:FR or iso-3166-2:CH it's working this way for France and Switzerland.

TheCartpenter:
> To include the ISO-3166-2 codes in this table is not working well because it is required to be in a separate table for maintenance reasons and for more flexibility. And it is working like this because of the language variants (e.g. de-DE, de-AT, de-CH, ...) To build a link into to the table is easy we have all what we need to have the reference. e.g. iso-3166-2:DE or iso-3166-2:CH for the German language used in Germany and Switzerland. For the French language e.g. iso-3166-2:FR or iso-3166-2:CH it's working this way for France and Switzerland.
> 
> Also I suggest to get a working Version asap and then we can bring new developments into the development branch so we have a quick way to get new versions and patch versions.

Yes, I think it is good to say that these changes are only appliable for en-gb and fr-fr as we speak until the new locale folders gets changed as well as for the database to follow.

TheCartpenter:
Closing this, too many conflicts. Will be working on the rest of the PHPDocs for multilanguages.

PR #14677, opened 2025-03-17
Update opencart-en-gb.sql to make German entries is0-3166-1 conform

pemaier:


TheCartpenter:
Are you sure they're meant to be removed rather than renamed to their respective languages on both SQL files?

pemaier:
Yes I'm sure. Because this are no official subdivision related to ISO. 
Because David likes to make a release as I read it's important to be clean and save.

TheCartpenter:
According to this ISO's list: https://github.com/dr5hn/countries-states-cities-database/blob/master/csv/states.csv , at index: 3007, it pretty much looks like it is related as these CSV files are follows up together...

PR #14683, opened 2025-03-18
product code unique check

fmarslan:
This PR introduces the unique check feature. If accepted and merged, the unique check will be implemented.

https://github.com/opencart/opencart/pull/14656

### DIFF
--- a/upload/admin/controller/catalog/product.php
+++ b/upload/admin/controller/catalog/product.php
@@ -1196,6 +1196,14 @@ class Product extends \Opencart\System\Engine\Controller {
 			if ($identifier_info && $identifier_info['validation'] && !oc_validate_regex($product_code['value'], $identifier_info['validation'])) {
 				$json['error']['code_' . $key] = sprintf($this->language->get('error_regex'), $product_code['code']);
 			}
+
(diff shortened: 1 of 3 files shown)


TheCartpenter:
'should' and 'doit' have two different meaning. 'doit' means: 'must'. As per the SEO URL error right above, it also says: must.

fmarslan:
> 'should' and 'doit' have two different meaning. 'doit' means: 'must'. As per the SEO URL error right above, it also says: must. Also, the controller validating the product code is not standard to OC.

I don't know french :( if you share correct sentence, I could change.

TheCartpenter:
French is not the issue, here. The English version still says: 'should' and OC languages normally doesn't use that word if you look at other: ```error_exists``` language keys.

fmarslan:
fixed language.

fmarslan:
> Also, the controller validating the product code is not standard to OC.

Another issue is that I don’t fully understand the meaning. What do you mean by "standard"? What should I pay attention to?
Are you suggesting that this validation should be moved to the validation.php file under the helper folder?

https://github.com/opencart/opencart/blob/729bd8d836d003d41caecf08dd1ef340c1cb4396/upload/admin/controller/catalog/product.php#L1200

danielkerr:
No.

PR #14684, opened 2025-03-18
Version number update to reflect minor change

mhcwebdesign:
The version is broken down into 4 points e.g 1.2.3.4 We use MAJOR.MINOR.FEATURE.PATCH to describe the version numbers.

A MINOR is when there are significant changes that affect core structures. This increment would likely break some 3rd party modules.

See the recent changes, such as the new oc_country_description and oc_zone_description, which will affect some 3rd party extensions.

### DIFF
--- a/upload/admin/index.php
+++ b/upload/admin/index.php
@@ -1,6 +1,6 @@
 <?php
 // Version
-define('VERSION', '4.1.0.1');
+define('VERSION', '4.2.0.0');
 
 // Configuration
 if (is_file('config.php')) {
--- a/upload/index.php
+++ b/upload/index.php
@@ -1,6 +1,6 @@
 <?php
 // Version
-define('VERSION', '4.1.0.1');
+define('VERSION', '4.2.0.0');
 
 // Configuration
 if (is_file('config.php')) {


stalker780:
Actually it is getting too annoying.

In the last 3 years there was no stable OC4 release. All the releases had too much critical bugs and include breaking changes for extensions.

Something is terribly wrong with OC4 release workflow.

Why can't we just stop adding features for 6 months and release a stable OC4?
Make 4.5 or 5.0 as a separate branch for breaking changes?

mhcwebdesign:
> Actually it is getting too annoying.
> 
> In the last 3 years there was no stable OC4 release. All the releases had too much critical bugs and include breaking changes for extensions.
> 
> Something is terribly wrong with OC4 release workflow.
> 
> Why can't we just stop adding features for 6 months and release a stable OC4? Make 4.5 or 5.0 as a separate branch for breaking changes?

I agree, some of the recent changes should have been done in the 'development' branch, not in the 'master' branch. Whatever version number we use for the next release from the master branch, is should labeled as a public beta release, to allow end users to do a full testing first before a final release for live deployments.

batumibiz:
It would be useful to abandon such chaos with versions altogether.
Wouldn't it have been better to use [Semantic Versioning](https://semver.org/)?
Everything is simple, logical and described in detail...

PR #14693, opened 2025-03-19
Updated PHPDocs

TheCartpenter:


danielkerr:
'date_modified' => '2021-01-31'
and date added should not be there.

PR #14700, opened 2025-03-20
More PHPDocs work

TheCartpenter:


danielkerr:
@TheCartpenter I would like to thank you for your continued support of OpenCart! Even after many members have criticized u for doing many minor updates in separate commits! I believe you have save me many 100's of hours of the past few years.

PR #14703, opened 2025-03-21
[fix] The update procedure does not work

batumibiz:
There are critical error in the installer that make it impossible to update.
Steps 7, 8, 9 are not executed due to an error in the model.
Step 10 has an SQL error

@danielkerr possible errors `$json['error']` are not displayed in any way. This is very inconvenient and makes it very difficult to find and fix errors.


Cubuss:
Still doesn't upgrade properly. Tried the patch and it still hangs at step 9

PR #14709, opened 2025-03-21
Update to admin marketplace/installer.php

milsaware:
I created an xml file and changed its extension to ocmod.zip and went to install. Thereafter, every time I attempted to load the extension installer page, I was greeted with a PHP unauthorised zip file error and the page just wouldn't load. This update checks for errors with the zip files, if there's an error it will delete the file and continue through the loop without crashing the page

### DIFF
--- a/upload/admin/controller/marketplace/installer.php
+++ b/upload/admin/controller/marketplace/installer.php
@@ -109,28 +109,6 @@ class Installer extends \Opencart\System\Engine\Controller {
 			if (!$install_info) {
 				// Unzip the files
 				$zip = new \ZipArchive();
-				
-				// Zip error codes
-				$ZIP_ERROR = [
-					\ZipArchive::ER_EXISTS => 'File already exists.',
-					\ZipArchive::ER_INCONS => 'Zip archive inconsistent.',
-					\ZipArchive::ER_INVAL => 'Invalid argument.',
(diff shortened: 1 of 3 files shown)


stalker780:
What for `$ZIP_ERROR` is declared? It is not used anywhere

milsaware:
Sorry, you're right. I was logging the error on the screen, then I wanted to portray the message to the user on the installer page, but I haven't figured out how to display the message yet - I deleted the print code, do you think I should remove $ZIP_ERROR as well?

stalker780:
Well. Anyway your code is not finished and does nothing except silent deleting zip files. I do not think anyone it should be merged. It does not fix anything. I would prefer to see an error in the log and my files preserved rather than deleted without explanation.

milsaware:
Well, it does do something. It got the installer page loading again and removes the offending file that crashes it. I agree the error needs to be logged, it's just setting up the warning in the admin panel after the fact - any suggestions?

milsaware:
Okay, I think I got it. I shifted it to the upload function, so it now checks the temporary file before storing it to the server. If there's an error that wasn't previously picked up, it'll display an error message to the end user during the upload process

danielkerr:
looks like this is requiried. could be loads of reasons the files go corrupt

PR #14713, opened 2025-03-22
removed country_id

farshadfmr:
removed country_id from the function as it was unnecessary.

### DIFF
--- a/upload/admin/model/localisation/country.php
+++ b/upload/admin/model/localisation/country.php
@@ -320,8 +320,8 @@ class Country extends \Opencart\System\Engine\Model {
 	 *
 	 * $this->model_localisation_country->deleteDescriptionsByLanguageId($language_id);
 	 */
-	public function deleteDescriptionsByLanguageId(int $country_id, int $language_id): void {
-		$this->db->query("DELETE FROM `" . DB_PREFIX . "country_description` WHERE `country_id` = '" . (int)$country_id . "' AND `language_id` = '" . (int)$language_id . "'");
+	public function deleteDescriptionsByLanguageId(int $language_id): void {
+		$this->db->query("DELETE FROM `" . DB_PREFIX . "country_description` WHERE `language_id` = '" . (int)$language_id . "'");
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


TheCartpenter:
By doing that, you'll be crashing the delete countries and delete zones from both controllers, since the ```$country_id``` and ```$zone_id``` are both required during the ```foreach``` loop.

farshadfmr:
These two function are only used here:
opencart\upload\admin\model\localisation\language.php
$this->model_localisation_language->deleteLanguage($language_id);
It is just a bug since the deleteLanguage function only has one param.
You must have mistaken it with deleteDescriptions function.

TheCartpenter:
https://github.com/opencart/opencart/blob/master/upload/admin/controller/localisation/country.php#L499

TheCartpenter:
As for the zones, not sure if this has been modified because it was unnecessary, since the ```deleteDescriptionsByLanguageId``` method is still in-use...

mhcwebdesign:
See also this issue: https://github.com/opencart/opencart/issues/14715

TheCartpenter:
Won't work without removing the ```$country_id``` from the controller.

TheCartpenter:
It will also affect all languages from the controller when the language ID will only be parametered and, surely, it's not what we want to happen.

TheCartpenter:
Because en-gb is the default language of the platform and deleting that language would require manual re-insert for people who may want to reuse that language if not ending up doing a fresh install all over again. As for setting up a headsup for other languages, that could be ideal though.

farshadfmr:
Ok I just checked https://github.com/opencart/opencart/blob/master/upload/admin/controller/localisation/country.php#L499. It has more issues at the moment than we anticipated. some functions needs to get renamed eg: `$zone_to_geo_zone_total = $this->model_localisation_geo_zone->getTotalZoneToGeoZoneByCountryId($country_id)`
to 
`$zone_to_geo_zone_total = $this->model_localisation_geo_zone->getTotalZonesByCountryId($country_id);`

Furthermore, I believe that when deleting a country, all its related languages must be also removed. If people plan to re-use the deleted country they must manually re-insert its information. Opencart is not responsible for keeping track of deleted information by user. Also another reason for this, is that when the language of admin section is other than 'en-gb', then the country will not be deleted completely and when the user tries to edit it, no information would be set in the form fields because the `getCountry` function relies on the language of admin (which is as of this example other than (en-gb).

farshadfmr:
I just fixes some of these bugs with the new edits:
https://github.com/opencart/opencart/pull/14718

The skipping of 'en=gb' unchanged.

PR #14718, opened 2025-03-22
fixed bug in admin country & zone controllers

farshadfmr:
Fixed some function names issues from the delete method.
Also some fixes based on the previous pull:
https://github.com/opencart/opencart/pull/14713


danielkerr:
i removed the code ur tryng to add

farshadfmr:
Was there a conflict or something?

PR #14719, opened 2025-03-22
Added deleteCountryDescriptionsByCountryId() method in localisation/country

TheCartpenter:
In compliance to: https://github.com/opencart/opencart/blob/master/upload/admin/model/catalog/option.php#L477-L498 .


TheCartpenter:
Got a better option.

PR #14720, opened 2025-03-22
Modified localisation/country controller and model delete descriptions method

TheCartpenter:
In accordance to: https://github.com/opencart/opencart/blob/master/upload/admin/model/catalog/option.php#L477 with a matched conventional name and only one parameter.


danielkerr:
dont use names like deleteCountryDescriptionsByCountryId

PR #14724, opened 2025-03-23
Modified delete country controller and fixed some PHPDocs description languages

TheCartpenter:


danielkerr:
i dont know what this code is for!

TheCartpenter:
To delete the country descriptions without affecting the default language code. It could be done with an error message as well whenever the default language is selected. I can take care of that.

TheCartpenter:
The default country is already validated, got it.

PR #14741, opened 2025-03-25
More PHPDocs work and added fr-fr definitions for ssr and account/address

TheCartpenter:


TheCartpenter:
@danielkerr: Since you have the ```es-es``` folder, not to forget the install SQL catalog file as well.

PR #14745, opened 2025-03-26
Applied CS-Fixer

TheCartpenter:


cgcmart:
This kind of commit is really worthless. In frankly, so called PHPDocs work is outdated, that doesn't impact any php function.

TheCartpenter:
While you can keep your opinion for yourself, the setting and event tables in the fr-fr SQL catalog also needed to be modified. If you believe that's worthless, then perhaps you should set yourself in the French peoples' position trying to install the French version whereas setting values are missing from a fresh install.

PR #14748, opened 2025-03-27
More PHPDocs work - fr-fr database work - updated aws/aws-sdk-php and twig/twig vendors

TheCartpenter:


TheCartpenter:
That CS-Fixer is crap.

PR #14751, opened 2025-03-28
Allow more depths files of extension to list in permission

padaliyajay:
Currently if controller file is inside nested folder in more than 2 level then it is impossible to access it because of permission.

This PR will fix this issue

### DIFF
--- a/upload/admin/controller/user/user_permission.php
+++ b/upload/admin/controller/user/user_permission.php
@@ -296,12 +296,14 @@ class UserPermission extends \Opencart\System\Engine\Controller {
 		$data['extensions'] = [];
 
 		// Extension permissions
-		$results = glob(DIR_EXTENSION . '*/admin/controller/*/*.php');
+		$results = glob(DIR_EXTENSION . '*/admin/controller/*');
 
 		foreach ($results as $result) {
-			$path = substr($result, strlen(DIR_EXTENSION));
-
-			$data['extensions'][] = 'extension/' . str_replace('admin/controller/', '', substr($path, 0, strrpos($path, '.')));
+			if(pathinfo($result, PATHINFO_EXTENSION) == 'php'){
+				$path = substr($result, strlen(DIR_EXTENSION));
+	
+				$data['extensions'][] = 'extension/' . str_replace('admin/controller/', '', substr($path, 0, strrpos($path, '.')));
+			}
 		}
 
 		if (isset($user_group_info['permission']['access'])) {


TheCartpenter:
This has been suggested before as I would not advise that. This code change also means that other file extensions than PHP files can be gathered in the ```$data['extensions']``` array whereas Opencart does not support as it also uses more resources to result an inconsistent list of data. Therefore, OCMod or Event Triggers would be a better choice rather than digging extensions based on admin permissions, since OCMod can already be pulled from the startup and if, and since, OCMod can be pulled from the startup; so can extensions' list on anyhow.

batumibiz:
> OCMod or Event Triggers would be a better choice

What does OCMod have to do with it?
This PR is not about that.

> This code change also means that other file extensions than PHP files can be gathered in the array whereas Opencart does not support

This PR takes into account only PHP files

batumibiz:
This is useful PR.
[After this comit](https://github.com/opencart/opencart/commit/6723be65bf40bbf30d7233baab24fa3cdf6e10e5), some good extensions stopped working normally.
This PR fixes this problem.

TheCartpenter:
> > OCMod or Event Triggers would be a better choice
> 
> What does OCMod have to do with it? This PR is not about that.
> 
> > This code change also means that other file extensions than PHP files can be gathered in the array whereas Opencart does not support
> 
> This PR takes into account only PHP files

These lines do not take only PHP files into account: https://github.com/opencart/opencart/pull/14751/commits/8417e5683efac0f4b0345f6de9a1b298064b3482#diff-92affeb247c3ede55906480cc02f685c6bdd545880e4ae2e8b298bfcc13de98aR299-R305 . ```*.php``` is removed from the original code and the ```$data['extensions']``` array is above the if statement that looks up for PHP file extensions.

batumibiz:
Well, then you need to [revert the changes from this comit](https://github.com/opencart/opencart/commit/6723be65bf40bbf30d7233baab24fa3cdf6e10e5).
This also solves the problem.

TheCartpenter:
> Well, then you need to [revert the changes from this comit](https://github.com/opencart/opencart/commit/6723be65bf40bbf30d7233baab24fa3cdf6e10e5). This also solves the problem.

The word 'also' in this scenario is based on your observation but providing two separate solutions where one suggests the removal of PHP extensions in the glob and the other to revert is still another angle to what the original poster is suggesting on this PR.

batumibiz:
> > Well, then you need to [revert the changes from this comit](https://github.com/opencart/opencart/commit/6723be65bf40bbf30d7233baab24fa3cdf6e10e5). This also solves the problem.
> 
> The word 'also' in this scenario is based on your observation but providing two separate solutions where one suggests the removal of PHP extensions in the glob and the other to revert is still another angle to what the original poster is suggesting on this PR.

**There is a problem that needs to be solved.**
If you have a solution - send your code, say thank you.
If there is no solution, then please do not interfere.

TheCartpenter:
The most similar way you could try, on your end, would be the ```path()``` method of the ```design/translation``` controller. If you are able to pull the expected path and files from that location, so should you from the user group permissions by using the code this way. However, there aren't much complaints about the user group permissions not pulling up the extensions in general.

padaliyajay:
It is not big change. just same things but allow deep access to nested folders of controllers

tell me what is different between those two code?

Original
```
$results = glob(DIR_EXTENSION . '*/admin/controller/*/*.php');
}
```

PR
```
$results = glob(DIR_EXTENSION . '*/admin/controller/*');
    if(pathinfo($result, PATHINFO_EXTENSION) == 'php'){
    }
}
```

PR #14753, opened 2025-03-29
Retrieve more in depth files from extension directory for permission

padaliyajay:
this is revised PR of previous https://github.com/opencart/opencart/pull/14751

Currently if controllers file is inside nested folder at more than 2 level then it is impossible to access it because of permission.

This PR will fix this issue

### DIFF
--- a/upload/admin/controller/user/user_permission.php
+++ b/upload/admin/controller/user/user_permission.php
@@ -296,7 +296,7 @@ class UserPermission extends \Opencart\System\Engine\Controller {
 		$data['extensions'] = [];
 
 		// Extension permissions
-		$results = glob(DIR_EXTENSION . '*/admin/controller/*/*.php');
+		$results = glob(DIR_EXTENSION . '*/admin/controller/{,*/,*/*/,*/*/*/}*.php', GLOB_BRACE);
 
 		foreach ($results as $result) {
 			$path = substr($result, strlen(DIR_EXTENSION));


batumibiz:
Perfect!
This is exactly what is needed.

TheCartpenter:
> this is revised PR of previous #14751
> 
> Currently if controllers file is inside nested folder at more than 2 level then it is impossible to access it because of permission.
> 
> This PR will fix this issue

"I'd like to sing it on a star.". Based on the mount of stars that's in there, it looks like it's an issue that's related on your server while trying to load the PHP extension files, so far.

batumibiz:
> "I'd like to sing it on a star.". Based on the mount of stars that's in there, it looks like it's an issue that's related on your server while trying to load the PHP extension files, so far.

Why do you interfere where you do not understand?
It is better to read how the GLOB_BRACE works.

> GLOB_BRACE - Expands {a,b,c} to match 'a', 'b', or 'c'

TheCartpenter:
I'm not looking to interfere but simply to address that the main, and separate issue, is not about GLOB_BRACE but rather about the mount of stars that are indicated for the paths.

batumibiz:
> I'm not looking to interfere but simply to address that the main, and separate issue, is not about GLOB_BRACE but rather about the mount of stars that are indicated for the paths.

You need to scan several nested levels.
If you disassemble it in parts `{,*/,*/*/,*/*/*/}` (pay attention to commas)
`*/`	(nested level 1)
or
`*/*/` (nested level 2)
or
`*/*/*/` (nested level 3)

And what is unusual about it?

Or do you know a way to make it more beautiful?
Then show the working code.
If you don't know, then please don't interfere.

TheCartpenter:
While I do not encourage the integration of this code, I would rather suggest to look on StackOverFlow for more decent solutions than using levels with stars like those.

padaliyajay:
> I would rather suggest to look on StackOverFlow for more decent solutions than using levels with stars like those

It is originally level by stars. i just increased level more

batumibiz:
> While I do not encourage the integration of this code, I would rather suggest to look on StackOverFlow for more decent solutions than using levels with stars like those.

We have already tried this.
[Recursive directory scanning](https://stackoverflow.com/a/17161106) requires more code and uses more resources. In this case, it is unnecessary. The simplest and most effective solution in this PR

osworx:
Just to say it simple: **GLOB_BRACE** is sh...
Why?
Some hoster are using systems this flag is not supported (e.g.Strato).
Look at the forum - enough topics there about.

TheCartpenter:
> > While I do not encourage the integration of this code, I would rather suggest to look on StackOverFlow for more decent solutions than using levels with stars like those.
> 
> We have already tried this. [Recursive directory scanning](https://stackoverflow.com/a/17161106) requires more code and uses more resources. In this case, it is unnecessary. The simplest and most effective solution in this PR

It might be the simplest, but it's not an efficient one for everybody to use.

batumibiz:
[An alternative to this PR](https://github.com/opencart/opencart/pull/14754), without the use of GLOB_BRACE.

danielkerr:
i think the code is ok. i dont think people will go to deep so it doesn't need go infinitly deep.

batumibiz:
> i think the code is ok. i dont think people will go to deep so it doesn't need go infinitly deep.

I also like this option more.
However, they complained above that some hostings did not support GLOB_BRACE
Therefore, an alternative option was proposed.

danielkerr:
we need a list of functions to create for managing files and directories.

PR #14754, opened 2025-03-29
More in depth files from extension directory for permission

batumibiz:
An alternative to [this PR](https://github.com/opencart/opencart/pull/14753), without the use of GLOB_BRACE.
[This code](https://stackoverflow.com/questions/17160696/php-glob-scan-in-subfolders-for-a-file/17161106#17161106
) is taken as a basis.

### DIFF
--- a/upload/admin/controller/user/user_permission.php
+++ b/upload/admin/controller/user/user_permission.php
@@ -296,7 +296,7 @@ class UserPermission extends \Opencart\System\Engine\Controller {
 		$data['extensions'] = [];
 
 		// Extension permissions
-		$results = glob(DIR_EXTENSION . '*/admin/controller/*/*.php');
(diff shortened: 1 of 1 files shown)


batumibiz:
If this option does not suit you, then can you simply [roll back the changes from this comit](https://github.com/opencart/opencart/commit/6723be65bf40bbf30d7233baab24fa3cdf6e10e5)?
I checked, everything works and the problem is being solved.

TheCartpenter:
Still ends with star (*) in the ```rglob``` method as ```$flags``` has no usefulness for the platform when originally using the ```glob``` function to pull the directories and files.

padaliyajay:
I will suggest. add rglob function in helper `/system/helper/file.php`

danielkerr:
u should not need to add new methods

Issue #14755, opened 2025-03-30
[4.1.0.3] Wrong total on cart page


mhcwebdesign:
**What version of OpenCart are you reporting this for?**
4.1.0.3

**Describe the bug**
The en-gb?route=checkout/cart calculates a wrong total.

**To Reproduce**
1. Install OpenCart 4.1.0.3
2. Add the  'HTC Touch HD' to the shopping cart with a quantity of 2
3. Display the en-gb?route=checkout/cart page
4. It shows a total of 244 instead of the expected 242

**Expected behavior**
The total should be 242. The Eco Tax is always supposed to be a fixed amount of $2.00, it's not a percentage and is not supposed to be multiplied by the product quantity.

Suggested fix for **system/library/cart.php**, function **getTaxes**:
```
	public function getTaxes(): array {
		$tax_data = [];

		foreach ($this->getProducts() as $product) {
			if ($product['tax_class_id']) {
				$tax_rates = $this->tax->getRates($product['price'], $product['tax_class_id']);

				foreach ($tax_rates as $tax_rate) {
					if (!isset($tax_data[$tax_rate['tax_rate_id']])) {
						$tax_data[$tax_rate['tax_rate_id']] = ($tax_rate['amount'] * (($tax_rate['type']=='P') ? $product['quantity'] : 1));
					} else {
						$tax_data[$tax_rate['tax_rate_id']] += ($tax_rate['amount'] * (($tax_rate['type']=='P') ? $product['quantity'] : 1));
					}
				}
			}
		}

		return $tax_data;
	}
```

**Screenshots / Screen recordings**

![Image](https://github.com/user-attachments/assets/35d2d0f7-56d2-4ec9-a8a9-7bc9de0041c4)

**Server / Test environment (please complete the following information):**
 - Local development
 - Operating system: Linux Mint
 - PHP version: 8.4.4
 - Apache version: Apache/2.4.41 (Ubuntu) 
 - Browser(s) tested with: FireFox 136.0.4 (64-bit)

mhcwebdesign:
@Nathan-DC 
It still shouldn't list 2 different totals, they should be the same.

osworx:
> This may differ per country

When (I really don't know, we here have such not), then how will you currently define that???
But at the end, a final price is a final price ..

Nathan-DC:
[...]

PR #14759, opened 2025-03-31
Update cart.php

mhcwebdesign:
checkout/cart totals for fixed taxes (e.g. Eco Tax) need to be on a per product item basis, otherwise the checkout cart page lists wrong item totals.


mhcwebdesign:
See also https://github.com/opencart/opencart/issues/14755 for more details on this issue.

PR #14763, opened 2025-04-01
Enhanced Upgrade & Rollback System + Prevent Accidental Navigation

milsaware:
### What's Changed  

- Checks GitHub for the latest release and notifies the client if an update is available.  
- Displays rollback availability if applicable.  
- Formats the changelog using `formatMarkdown`.  

#### Update Process:  
1. Backs up files  
2. Backs up database  
3. Creates a database tables list  
4. Compresses backup  
5. Removes backup files  
6. Downloads update  
7. Unpacks update  
8. Installs update  
9. Removes install folder  
10. Clears cache  
11. Refreshes the page  

#### Rollback Process:  
1. Unzips backup file  
2. Rolls back database  
3. Removes database tables not part of the rolled-back version  
4. Replaces files  
5. Clears restore files from the backup folder  
6. Clears cache  
7. Refreshes the page  

### Additional Changes:  
- Included `ca.cert` for cURL commands to maintain SSL verification. 
- Implemented a window event to prevent accidental navigation during update or rollback.  
- Replaced update confirmation alert with a modal for better user experience.  

Tested rollback and update between `4.1.0.0` & `4.1.0.3` successfully. Note: Since those versions don’t contain the updated files, they were manually restored in Notepad++ during testing.


danielkerr:
looks intreasting. im guessing u need this for ur own clients.

milsaware:
I was working on a module and after a short while, I realised I was 3 releases behind (I was on 4.1.0.0, opencart was already up to 4.1.0.3). The current update process is laborious to say the least and rolling back is a bit more difficult because often there will be database changes. This way, I can now just go to the upgrade section, it'll show whether or not a new opencart version is released through Github and it takes care of the backing up, downloading and installing and if there are errors, it will just as easily roll back the system. I think it'll be useful for all devs and shop owners

TheCartpenter:
The variable conventional name structrures, ```let``` and the French street text might need a bit of revision in the meantime.

milsaware:
@TheCartpenter - sorry, none of that made any sense. Could you just give the examples of what you think needs changing, please

PR #14778, opened 2025-04-04
cache delete/unlink prevent error on file not found

hayden-t:
due to simultaneous running requests file found for deletion can be gone by time code gets to delete it, so wrapped in try, since @ suppression not coded for in error reporting.

https://github.com/opencart/opencart/pull/14096

### DIFF
--- a/upload/system/library/cache/file.php
+++ b/upload/system/library/cache/file.php
@@ -80,7 +80,7 @@ class File {
 					if (!@unlink($file)) {
 						clearstatcache(false, $file);
 					}
-				}catch(\Exception $e){}
+				}catch(\Throwable $e){}
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


ADDCreative:
Just tested and it doesn't work unfortunately. unlink doesn't generate an exemption that can be caught, only an error.

hayden-t:
Ah yes, its not an exception, but a warning. Ive changed the code, can you test again. thanks !

hayden-t:
Ive catched warnings before, but maybe with custom error handler it still gets fired like the @ suppression, thanks for feedback, will have to check it myself. But your method did work ?

danielkerr:
i'm moving away from this type of caching to generated json files within the view/data folder will take over.

hayden-t:
ok, thanks Daniel, ill close it then.

hayden-t:
i tested it, and warning is not a  throwable so modification would have to be done to the error handler or this ? https://github.com/opencart/opencart/pull/14778#issuecomment-2779900441

PR #14768, opened 2025-04-02
Version upgrade and rollback automation

milsaware:
### What's Changed  

- Checks GitHub for the latest release and notifies the client if an update is available.  
- Displays rollback availability if applicable.  
- Formats the changelog using `formatMarkdown`.  

#### Update Process:  
1. Backs up files  
2. Backs up database  
3. Creates a database tables list  
4. Compresses backup  
5. Removes backup files  
6. Downloads update  
7. Unpacks update  
8. Installs update  
9. Removes install folder  
10. Clears cache  
11. Refreshes the page  

#### Rollback Process:  
1. Unzips backup file  
2. Rolls back database  
3. Removes database tables not part of the rolled-back version  
4. Replaces files  
5. Clears restore files from the backup folder  
6. Clears cache  
7. Refreshes the page  

### Additional Changes:  
- Included `ca.cert` for cURL commands to maintain SSL verification. 
- Implemented a window event to prevent accidental navigation during update or rollback.  
- Replaced update confirmation alert with a modal for better user experience.  

Tested rollback and update between `4.1.0.0` & `4.1.0.3` successfully. Note: Since those versions don’t contain the updated files, they were manually restored in Notepad++ during testing.

Issue #14766, opened 2025-04-01
[master] Admin category filter search results in a logout


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?

At the admin in categories when using the filter function after entering some text and press enter on the keyboard you get kicked out. (When clicking the filter button it works.)

![Image](https://github.com/user-attachments/assets/8a3b44a8-0958-40a9-8331-9f7b33d9db77)

**Expected behavior**

After pressing enter to get a valid result.

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? local
 - Operating system linux
 - PHP version 8.4
 - Apache version 2.4
 - Browser(s) tested with [e.g. chrome, safari - if applicable] FF
 - Version [e.g. 22] 136.0.4

mhcwebdesign:
Indeed, I can confirm, just tested it and was able to replicate the same issue.

PR #14780, opened 2025-04-04
Admin Country Filter Improvements

milsaware:
## ✅ Summary of Changes:

- Updated the form to auto-submit on input (`keyup`, `cut`, `paste`) events.

- Removed the now-redundant submit button.

- Moved the reset/clear filter button to the top-right of the filter box for a cleaner and more accessible layout.

- Clear Filter button now clears all filter inputs and automatically submits the form for a clean reset.

- Removed history push state to improve workflow

**BEFORE**
before

**AFTER**
after

### DIFF
--- a/upload/admin/view/template/localisation/country.twig
+++ b/upload/admin/view/template/localisation/country.twig
@@ -1,87 +1,103 @@
 {{ header }}{{ column_left }}
 <div id="content">
-  <div class="page-header">
-    <div class="container-fluid">
-      <div class="float-end">
-        <button type="button" data-bs-toggle="tooltip" title="{{ button_filter }}" onclick="$('#filter-country').toggleClass('d-none');" class="btn btn-light d-lg-none"><i class="fa-solid fa-filter"></i></button>
-        <a href="{{ add }}" data-bs-toggle="tooltip" title="{{ button_add }}" class="btn btn-primary"><i class="fa-solid fa-plus"></i></a>
-        <button type="submit" form="form-country" formaction="{{ delete }}" data-bs-toggle="tooltip" title="{{ button_delete }}" onclick="return confirm('{{ text_confirm }}');" class="btn btn-danger"><i class="fa-regular fa-trash-can"></i></button>
-      </div>
-      <h1>{{ heading_title }}</h1>
-      <ol class="breadcrumb">
-        {% for breadcrumb in breadcrumbs %}
-          <li class="breadcrumb-item"><a href="{{ breadcrumb.href }}">{{ breadcrumb.text }}</a></li>
-        {% endfor %}
-      </ol>
-    </div>
-  </div>
-  <div class="container-fluid">
-    <div class="row">
-      <div id="filter-country" class="col-lg-3 col-md-12 order-lg-last d-none d-lg-block mb-3">
-        <div class="card">
-          <div class="card-header"><i class="fa-solid fa-filter"></i> {{ text_filter }}</div>
-          <div class="card-body">
-            <form id="form-filter">
-              <div class="mb-3">
-                <label class="form-label">{{ entry_name }}</label>
-                <input type="text" name="filter_name" value="{{ filter_name }}" placeholder="{{ entry_name }}" id="input-name" class="form-control"/>
-              </div>
-              <div class="mb-3">
-                <label class="form-label">{{ entry_iso_code_2 }}</label>
-                <input type="text" name="filter_iso_code_2" value="{{ filter_iso_code_2 }}" placeholder="{{ entry_iso_code_2 }}" id="input-iso-code-2" class="form-control"/>
-              </div>
-              <div class="mb-3">
-                <label class="form-label">{{ entry_iso_code_3 }}</label>
-                <input type="text" name="filter_iso_code_3" value="{{ filter_iso_code_3 }}" placeholder="{{ entry_iso_code_3 }}" id="input-iso-code-3" class="form-control"/>
-              </div>
-              <div class="text-end">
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
I don't think it's a good idea to remove submit buttons from forms.

milsaware:
@mhcwebdesign 

It auto updates itself. There's no need for the button. If there was anything that would break the auto update, then the original button wouldn't work anyway. It's far better to see the filter changes as you're typing.

I wish these things could be tested before people make comments on them.

TheCartpenter:
It looks like somebody's trying to build a new core, without the original one to be completed, in here...

TheCartpenter:
If there's one thing that's about with those changes, is that it's not about evolution.

TheCartpenter:
What it doesn't say in there is the code inconsistencies between the files you're suggesting to add about.

milsaware:
@TheCartpenter 

You're not saying it, either. 

It’s easy to criticize without considering the bigger picture, but it’s clear that what’s been done up until now hasn’t been enough to create a stable, scalable project. I'm now working on another SQL error, where have you been to deal with all this? I've only had the script 3 weeks, you’ve been doing this for years, and the lack of other devs in this community speaks volumes. It’s clear that the environment here has discouraged people from contributing. If we want real change, you need to stop gatekeeping, stop blogging, stop treating it like it's a social media platform and start encouraging a more inclusive, productive atmosphere.

mhcwebdesign:
@milsaware :
Your helping out with bugfixes is much appreciated. And yes, there are more unresolved issues, see e.g. https://github.com/opencart/opencart/issues

Let's just make sure not to implement changes until the core bugs are all fixed.

milsaware:
@mhcwebdesign 

Thank you.

The only reason I pushed this change through is because I was working on a fix for [issue #14766], where the admin panel was logging out whenever the form was submitted by pressing Enter. While working on that, I realised it could be made easier to use with some modifications. As I was modifying the one page, I wanted the theme to match throughout wherever there’s a filter box. That’s when I discovered that none of the filters were submitting properly, and two had SQL errors.

The real issue with this specific filter was that it wouldn’t submit when pressing Enter, which is the same issue across all of them. While fixing this, it was basically a copy-paste job to transfer the new functionality to all the filters, so I did that as well.

stalker780:
@milsaware 
I also think it is a bad idea of autosubmit as you type for filters.
Think about the loaded and slow servers.
Think about product list filtering when you have 200k products, if you plan to use autosubmit there too.

milsaware:
@stalker780 

🙄 

It's paginated and there are already built in autocomplete functions in the same areas which are doing the exact same thing in terms of fetching data as the input changes. You can have a million rows, it's only ever going to fetch the first 15 or 20

pemaier:
To develop further is good and get new ideas also!
Nevertheless we have to consider also the non-mainstream users and we should consider basic rules of style guides.
1. Remove the Filter button in not a good idea -> Style guides (To move the button to the top is worth to think about.)
2. Use the CR to transfer data is a good idea -> Usability improvement 
3. To implement an auto-submit of data is in my opinion not clear. 
a) Is it compatible to all browser and OS combinations of PC Tablets and Mobiles?
b) Do the users accept a uncontrolled data transfer? (... if yes it has to get labeled direct or indirect)
c) Is it a real improvement ...

pemaier:
@milsaware 
... sorry no. I do only test functions which are already within the core so called customer tests in a way the people download OpenCart install it and start to work. (Otherwise it will be a full-time job which I haven't time for.)

Right now I even can't install a new master version because the install process is not working.

milsaware:
@pemaier 

I have and currently I'm the only one with almost a fully functioning filter system across the admin panel, on top of a built in update automation function (#14768) that takes care of installing the next release, backing up the current release and rolling back to a previous one on error for me.

These changes are a good thing.

TheCartpenter:
> @pemaier
> 
> I have and currently I'm the only one with almost a fully functioning filter system across the admin panel, on top of a built in update automation function (#14768) that takes care of installing the next release, backing up the current release and rolling back to a previous one on error for me.
> 
> These changes are a good thing.

One could say that.

milsaware:
@TheCartpenter 

No need to be all salty, little buddy. It's not my fault you're so spectacularly bad that after years of working on the script, it still wound up being filled with errors, while I'm blazing through them - but hey, at least the code's pwetty. But then, given how you tried to "help" that other user with their country table error ... you really don't have a clue. The error was presented to you and you still turned around after a couple of messages claiming it wasn't enough to know what may be going on, again, after years of you piddling about with the script

You can talk all the nonsense you like about code breaking and what not, but I've been merging to OmenCart and everything works beautifully. Don't worry, kiddo. OmenCart will pick up your slack 😜

danielkerr:
ur over complicating things.

milsaware:
@danielkerr

If by "over-complicating" you mean automating core processes, fixing long-standing issues, and adding basic protections to stop the system from breaking itself—then yes, I’m doing exactly what should’ve been done years ago.

The truth is, if even a fraction of this attention had been paid to OpenCart over the past decade, there wouldn’t be so many complaints on the forums or broken features in current releases. Case in point: your latest version doesn’t even install cleanly due to missing database columns. That’s not acceptable for production-ready software.

What I’m doing now with the fork is pushing OpenCart toward being a modern, stable, and maintainable platform. These are changes you probably wouldn’t have seen until OpenCart 10, if ever. I was more than willing to do the work and contribute—openly and without ego—but instead of support, I’ve been met with resistance from people more focused on gatekeeping than improvement.

OpenCart has real potential. But it needs to be taken seriously and maintained accordingly. Otherwise, it’ll keep falling further behind.

danielkerr:
should be changed from on click to on submit to do exactly the same

PR #14800, opened 2025-04-07
More work on en-gb and fr-fr language packs

TheCartpenter:


TheCartpenter:
Don't like it. Will create a new PR.

PR #14805, opened 2025-04-10
Update category_list.twig - a discreet status view proposal

pemaier:
A proposal that is more pleasant for the eyes)
If this is accepted I'll change all lists.
image


stalker780:
imo the most pleasant and informative design was in 4.1.0.1
green/red bage on a new line after the name. What was the problem with it?
I do not understand why this awful pale row color was implemented

image

pemaier:
Sometimes is a little bit less more :)

danielkerr:
i like the current system

PR #14806, opened 2025-04-12
More en-gb and fr-fr work

TheCartpenter:


TheCartpenter:
As discovered, here: https://github.com/opencart/opencart/issues/14810 .

PR #14812, opened 2025-04-14
[cleanup] Variable already equals the assigned value

batumibiz:


batumibiz:
Variables already assigned
https://github.com/opencart/opencart/blob/master/upload/catalog/controller/mail/order.php#L184-L208

ADDCreative:
@batumibiz The replaces are different. First ones start with payment_, second with shipping_.

batumibiz:
> @batumibiz The replaces are different. First ones start with payment_, second with shipping_.

Thanks for the comment, returned Replaces

TheCartpenter:
By reverting one part of the finds, rather than re-adding the other, it means that the same find can replace both, payment and shipping, altogether.

Issue #14787, opened 2025-04-04
admin=>Customer "Login into Store" not working


chongshengdz:
admin=>Customer "Login into Store" not working
PHP Warning:  Undefined array key "customer_token" in catalog_controller_common_header.php on line 87
latest master branch

TheCartpenter:
Yes, the ```editToken()``` method seem to be missing from the account/customer model.

pemaier:
same as #14764

mhcwebdesign:
Still not working:

```
Error: <b>Notice</b>:  Undefined property: Proxy::editToken in <b>/home/juergen/htdocs/opencart-master/opencart/upload/catalog/controller/account/login.php</b> on line <b>277</b>
File: /home/juergen/htdocs/opencart-master/opencart/upload/system/engine/proxy.php
Line: 91

Backtrace: 0
File: /home/juergen/htdocs/opencart-master/opencart/upload/catalog/controller/account/login.php
Line: 277
Class: Opencart\System\Engine\Proxy
Function: __call

Backtrace: 1
File: /home/juergen/htdocs/opencart-master/opencart/upload/system/engine/action.php
Line: 96
Class: Opencart\Catalog\Controller\Account\Login
Function: token

Backtrace: 2
File: /home/juergen/htdocs/opencart-master/opencart/upload/system/framework.php
Line: 259
Class: Opencart\System\Engine\Action
Function: execute

Backtrace: 3
File: /home/juergen/htdocs/opencart-master/opencart/upload/index.php
Line: 20
Function: require_once
```

nicomintr:
I solved it by editing the file "catalog/model/account/customer.php"
At line 627, after the **public function deleteTokenByCode**
I added the method to the **Customer** class:


```
	/**
	 * editToken
	 *
	 * Updates or resets a customer's token.
	 *
	 * @param string $email Customer's email
	 * @param string $token New token (empty to reset it)
	 *
	 * @return void
	 */
	public function editToken(string $email, string $token): void {
		// Reset the empty token or update it
		$this->db->query("UPDATE `" . DB_PREFIX . "customer` SET `token` = '" . $this->db->escape($token) . "' WHERE LCASE(`email`) = '" . $this->db->escape(oc_strtolower($email)) . "'");
	}
```

Issue #14764, opened 2025-04-01
[master] Admin login as customer not working


pemaier:
**What version of OpenCart are you reporting this for?**
master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?

Go in the admin to customer list and choose a customer to login 

![Image](https://github.com/user-attachments/assets/3bd82fd4-ad3e-4543-b683-4c27137b8153)

then you get this in the log.

2025-04-01 12:01:33 - Error: <b>Notice</b>:  Undefined property: Proxy::editToken in <b>/home/localhost/public_html/opencart/opencart-4.1.0.3.company/catalog/controller/account/login.php</b> on line <b>277</b>
/home/localhost/public_html/opencart/opencart-4.1.0.3.company/system/engine/proxy.php
Line: 91

Backtrace: 0
/home/localhost/public_html/opencart/opencart-4.1.0.3.company/catalog/controller/account/login.php
Line: 277
Class: Opencart\System\Engine\Proxy
Function: __call

Backtrace: 1
/home/localhost/public_html/opencart/opencart-4.1.0.3.company/system/engine/action.php
Line: 96
Class: Opencart\Catalog\Controller\Account\Login
Function: token

Backtrace: 2
/home/localhost/public_html/opencart/opencart-4.1.0.3.company/system/framework.php
Line: 259
Class: Opencart\System\Engine\Action
Function: execute

Backtrace: 3
/home/localhost/public_html/opencart/opencart-4.1.0.3.company/index.php
Line: 20
Function: require_once

**Expected behavior**

To be logged in into the customer account.

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server? local and server
 - Operating system linux
 - PHP version 8.2 and 8.4
 - Apache version 2.4
 - Browser(s) tested with [e.g. chrome, safari - if applicable] FF
 - Version [e.g. 22] 136

mhcwebdesign:
Yes, I can confirm this issue, still not working!

wisencoder:
I am using latest OC 4.1.0.3
I do have the editToken function however, it fails with the below error, because there is no token column in the customer column where it is expected. I also don't think expecting there is good, as the token should probably be saved in the session

Error: Error: Unknown column 'token' in 'field list'
Error No: 1054
[...]

PR #14860, opened 2025-05-06
Fix for admin login to customer account from backend

pemaier:
Credits to nicomintr.
Solution for #14787 and #14764

Tested and working.


danielkerr:
ill post a fix

pemaier:
ok) thanks for the info. 
I saw that i also have to update the filter i sent. But 1st I like to check on a working system. Install is not working since ~2 weeks.

Issue #14848, opened 2025-04-30
Admin product filter brings login page


cgcmart:
**What version of OpenCart are you reporting this for?**
4.1.0.4 / master
**Describe the bug**
A clear and concise description of what the bug is
What section does it affect?
on Admin product filter, input product name, then press Filter button, that brings admin login page:

> index.php?filter_name=&filter_model=&filter_category=&filter_category_id=&filter_manufacturer=&filter_manufacturer_id=&filter_price_from=&filter_price_to=&filter_quantity_from=&filter_quantity_to=&filter_status=

![Image](https://github.com/user-attachments/assets/5cd25686-3df5-47d1-ae26-9eb7d2d3b921)

**To Reproduce**

**Expected behavior**

**Screenshots / Screen recordings**
If applicable, add screenshots or recordings to help explain your problem. Please keep these short and if you can, edit them to point out when/where the problem is.
https://www.useloom.com/ is perfect for screen recording with Chrome or with their desktop app

**Server / Test environment (please complete the following information):**
 - Local development? Deployed to a web server?
 - Operating system
 - PHP version
 - Apache version
 - Browser(s) tested with [e.g. chrome, safari - if applicable]

**Additional context**

PR #14886, opened 2025-05-18
[Master] Fix Admin Filter Submit logout

rgbworld:
- Fixes logout that occurs when submitting the filter form.
- Added to javascript:  e.preventDefault();
- Fixes opencart #14848


rgbworld:
Also fixes opencart #14766

danielkerr:
wow didnt realized i had messed up with so many filters

PR #14890, opened 2025-05-19
Update file.php

leonidukg:
To collect cache files, do it only if the random has fallen to the cleanup. Otherwise, it will slow down if there are a lot of cache files.

### DIFF
--- a/upload/system/library/cache/file.php
+++ b/upload/system/library/cache/file.php
@@ -87,9 +87,9 @@ class File {
 	 * Destructor
 	 */
 	public function __destruct() {
-		$files = glob(DIR_CACHE . 'cache.*');
-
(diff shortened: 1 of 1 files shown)


TheCartpenter:
```$files``` is still in the IF statement while the affected variable has been moved below...

**Comment by leonidukg** [comment]
My mistake.

```
	/**
	 * Destructor
	 */
	public function __destruct() {
		
		if (mt_rand(1, 100) == 1) {
		
			$files = glob(DIR_CACHE . 'cache.*');
			if ($files) {
				foreach ($files as $file) {
					$time = substr(strrchr($file, '.'), 1);

					if ($time < time()) {
						if (!@unlink($file)) {
							clearstatcache(false, $file);
						}
					}
				}
			}
		}
	}
```

PR #14897, opened 2025-05-21
Add files via upload

bettyw231:
upload/admin/view/template/common/.twig
and
upload/admin/controller/common/.php

### DIFF
--- a/common/authorize.php
+++ b/common/authorize.php
@@ -1,0 +1,311 @@
+<?php
+namespace Opencart\Admin\Controller\Common;
+/**
+ * Class Authorize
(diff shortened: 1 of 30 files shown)


TheCartpenter:
The paths do not follow the class' paths...

bettyw231:
basically the file i was uploading was too large, and i was told to load it one by one, so thats why?

TheCartpenter:
I am not referring to the split quantities that were uploaded. The previous reply is about the routes you are suggesting the changes about vs. the class paths. They don't match altogether.

Issue #14826, opened 2025-04-24
Performance Feature: Event register string action


thundor:
The memory is full of unnecessary Action objects from events even if they aren't used.
Provided Before/After code:

CURRENT:
`$this->event->register('view/*/before', new \Opencart\System\Engine\Action('extension/myextension/myextension/startup.event'));` 

TO ALSO ALLOW: 
`$this->event->register('view/*/before', 'extension/myextension/myextension/startup.event');` 

/system/engine/event.php
CURRENT:
```
	public function trigger(string $event, array $args = []) {
		foreach ($this->data as $value) {
			if (preg_match('/^' . str_replace(['\*', '\?'], ['.*', '.'], preg_quote($value['trigger'], '/')) . '/', $event)) {
				$value['action']->execute($this->registry, $args);
			}
		}

		return '';
	}
```
CHANGED TO:
```
	public function trigger(string $event, array $args = []) {
		foreach ($this->data as $value) {
			if (preg_match('/^' . str_replace(['\*', '\?'], ['.*', '.'], preg_quote($value['trigger'], '/')) . '/', $event)) {
				$value['action'] = is_string($value['action']) ? new Action($value['action']) : $value['action'];
				$this->data[$k] = $value;
				$result = $value['action']->execute($this->registry, $args);

				if (!is_null($result) && !($result instanceof \Exception)) {
					return $result;
				}
			}
		}

		return '';
	}
```

PR #14899, opened 2025-05-23
Update event.php fix issue #14826

bettyw231:
The updated code allows developers to register event actions as string paths, like: $this->event->register('some/event', 'some/controller/path'); Then, in Event::trigger, the fix checks if the action is still a string, and only then creates the Action object (lazy loading). The instantiated object is cached so it's only created once.


danielkerr:
itr does not require ann output. u can only change the arguments.

PR #14902, opened 2025-05-27
mysqli::$report_mode is deprecated

farshadfmr:
1- Error: Creation of dynamic property mysqli::$report_mode is deprecated
Fixed by replacing it with the new function.

2- Fixed a bug where having no language param in address bar would redirect to error/not_found.


danielkerr:
where does it say deprecated

https://www.php.net/manual/en/mysqli-driver.report-mode.php

farshadfmr:
NVM. both issues are resolved with new edits now.

PR #14928, opened 2025-06-09
Update IfNode.php [php 8.4.x compatibility]

osworx:


ADDCreative:
Looks like Twig need updating to 3.9 minimum for PHP 8.4 support. Twig 3.12 drops support for PHP 7. Twig 3.20 drops support for PHP 8.0.

mhcwebdesign:
> Looks like Twig need updating to 3.9 minimum for PHP 8.4 support. Twig 3.12 drops support for PHP 7. Twig 3.20 drops support for PHP 8.0.

It currently uses Twig 3.8.0, so it makes sense to upgrade that one 3.9, via composer, if that's what's needed. When I have some time I'll look into it.

PR #14956, opened 2025-06-17
google base fix

mhcwebdesign:
Added support for more currencies, fix for price tag which must include currency code


excodius:
Found a typo:
'TZS'. // Tanzanian Shilling (dot instead of comma)

PR #14971, opened 2025-06-19
Fix autocomplete in Safari

stalker780:
Safary often does not assign `click` event on dynamic, bubble or hidden elements

### DIFF
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -425,7 +425,7 @@ $(document).on('click', '[data-oc-toggle=\'image\']', function(e) {
             });
 
             // Click
-            $dropdown.on('click', 'a', function(e) {
+            $dropdown.on('mousedown', 'a', function(e) {
                 e.preventDefault();
 
                 var value = $(this).attr('href');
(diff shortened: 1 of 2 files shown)


TheCartpenter:
> @TheCartpenter Works fine in Edge, Opera, Chrome, Firefox, Safari, both desktop and mobile.
> 
> On the other hand it is just one line of code. You can easily check it with your browser before posting theory here. Tell us if it does not work.

Before posting other theories here, ```mousedown``` is also nowhere else to be found in other TWIG files.

trydalcoholic:
The click event is a sequence that includes `mousedown`. Using `mousedown` directly should be safe, as it's a fundamental part of how `click` events are generated.
@stalker780, do you have information on specific Safari versions affected? I could help test.

stalker780:
What are you arguing about? Just check autocomplete in the latest Safari desktop and mobile.

`mousedown` works in every browser. While `click` does not work in many cases in Safari. 
An element must exist, be visible, have cursor:pointer, etc. to make it work. 
This causes multiple issues with dynamic elements. Have no idea why Apple did that.

stalker780:
@TheCartpenter you are just spamming as always. What is exactly your problem with `mousedown`?

trydalcoholic:
Yes, I confirm. `click` is not working as expected in Safari Version 18.5 (20621.2.5.11.8). `mousedown` works as expected. In other browsers, both events work as expected. I believe the fix from @stalker780 is simple, effective, and safe.

PR #14973, opened 2025-06-20
fix(cli): sync installer requirements with web installer

trydalcoholic:
### Summary

This PR synchronizes the CLI installer requirements with the web installer to ensure consistency between installation methods.

### Key Changes

The CLI installer was missing some requirement checks that exist in the web installer, and had an outdated check that the web installer doesn't have. This update brings both installers into alignment:

- **Removed zlib extension check:** The web installer doesn't check for zlib, so this requirement has been removed from the CLI installer for consistency.
- **Added mbstring/iconv check:** The web installer requires either iconv or mbstring extension to be loaded. This check has been added to the CLI installer.
- **Updated database extension check:** The web installer supports multiple database drivers (mysqli, pdo, pgsql), while the CLI installer only checked for mysqli. Now both installers check for the same set of database extensions.

### DIFF
--- a/upload/install/cli_install.php
+++ b/upload/install/cli_install.php
@@ -210,8 +210,14 @@ class CliInstall extends \Opencart\System\Engine\Controller {
 			$error .= 'ERROR: OpenCart will not work with session.auto_start enabled!' . "\n";
 		}
 
-		if (!extension_loaded('mysqli')) {
-			$error .= 'ERROR: MySQLi extension needs to be loaded for OpenCart to work!' . "\n";
+		$db = [
+			'mysqli',
+			'pdo',
+			'pgsql'
+		];
+
+		if (!array_filter($db, 'extension_loaded')) {
+			$error .= 'ERROR: A database extension needs to be loaded in the php.ini for OpenCart to work!' . "\n";
 		}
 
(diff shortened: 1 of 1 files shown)


TheCartpenter:
While this line does indicate otherwise when the case occurs: https://github.com/opencart/opencart/pull/14973/commits/c47c9ba12aa14aecf2873794f8a815f90b931b37#diff-c1ef4033f8cb8b14ba7ddef1e9e2cfe8ad4dcf66f6e84ac5496c51371286fdfdR236 .

trydalcoholic:
Oh, I see what you mean. The visual diff might make it look like one module check is replacing another, but that's simply because the zlib module check was removed and the existing openssl_encrypt check remains. The iconv/mbstring module checks were missing in the CLI installer but were already present in the web installer.

For improved clarity and to precisely match the sequence in the web installer, I can reorder the checks so that the `!function_exists('openssl_encrypt')` check comes before the `!function_exists('iconv') && !extension_loaded('mbstring')` checks, exactly as implemented in the web installer here: https://github.com/opencart/opencart/blob/master/upload/install/controller/install/step_2.php#L200C7-L200C42

TheCartpenter:
So, we're no longer using ZLIB...

PR #14974, opened 2025-06-21
feat(docker): implement separate PHP-FPM and Apache containers

trydalcoholic:
### Summary

This PR refactors the Docker setup by separating the monolithic PHP-Apache container into dedicated PHP-FPM and Apache services, following Docker containerization best practices.

### Key Changes

This change implements a proper multi-container architecture that provides better separation of concerns:

- **Service Separation:** Split the single `php-apache` container into dedicated `php-fpm` and `apache` services, allowing for independent scaling and configuration.
- **Improved Architecture:** Apache now communicates with PHP-FPM via `mod_proxy_fcgi`, which is the recommended approach for production environments.
- **Future Flexibility:** This architecture enables easy migration to alternative web servers (e.g., Nginx) by simply replacing the Apache service without touching PHP-FPM configuration.
- **Enhanced Networking:** Added custom `opencart-network` for service isolation and explicit container naming with `opencart-` prefix for consistency.
- **Optimized Performance:** Reduced healthcheck intervals to minimize startup time and improve development experience.
- **Better Resource Management:** Each service can be configured, scaled, and monitored independently, providing more granular control over the application stack.

The new architecture maintains full compatibility with existing OpenCart installations while providing a more maintainable and scalable foundation.

> **⚠️ Breaking Changes:** Developers will need to rebuild containers due to the architectural changes. The OpenCart application functionality remains unchanged, but existing Docker containers must be recreated.

### Notes for developers

This PR focuses on improving the internal structure of the Docker setup. The **Docker configuration is still not considered production-ready** and further work will continue on enhancing its robustness and features.


danielkerr:
i think should be ok. i have not had time to test the docker stuff

trydalcoholic:
> i think should be ok. i have not had time to test the docker stuff

A small merge conflict did occur. The PR is now ready.

pemaier:
@trydalcoholic Is it possible to consider the main platforms of hosting companies? I guess this makes most sense.

trydalcoholic:
> @trydalcoholic Is it possible to consider the main platforms of hosting companies? I guess this makes most sense.

Regarding the `GLOB_BRACE` discussion - this might be worth exploring in a separate issue or in [GitHub Discussions](https://github.com/opencart/opencart/discussions) to avoid cluttering this PR.

You're absolutely right to focus on hosting platforms. However, I think we should look at this from the **operating system perspective** rather than individual hosting companies.

The key point is that `GLOB_BRACE` is actually part of the **operating system's libc library**, not PHP itself. This makes OpenCart **less environment-independent** when using this constant.

**Traditional hosting platforms** predominantly use Debian/Ubuntu/CentOS with glibc, where `GLOB_BRACE` works fine.

**Modern deployment platforms** increasingly use Docker containers with Alpine Linux for several reasons:
- **Size**: Alpine ~5MB vs Debian ~100MB+ 
- **Security**: Minimal attack surface
- **Performance**: Faster container startup
- **Industry standard**: Most official Docker images offer Alpine variants

Alpine uses musl libc instead of glibc, which doesn't support `GLOB_BRACE`.

From my perspective, it would be beneficial to implement broader compatibility, including support for Alpine images, which are de facto used in production Docker builds. This would make OpenCart more deployment-flexible without sacrificing functionality.

trydalcoholic:
@danielkerr Just a gentle reminder that this PR is ready for review when you have a chance! 🙂

PR #14981, opened 2025-06-26
Fix `Could not call registry key model_design_seo_url` exception

inventor96:
When editing an existing product, calling `Opencart\Admin\Model\Catalog\Product@editProduct()` would result in an exception with the message "Error: Could not call registry key model_design_seo_url!". This would happen only when `editProduct()` was the first method call after instantiation of `Opencart\Admin\Model\Catalog\Product` (via `$this->load->model('catalog/product')`). If the first method called after instantiation was one that included a call to `$this->load->model('design/seo_url')` such as `Product@deleteProduct()` or `Product@addProduct()`, the exception would not be thrown because the `design/seo_url` model was already loaded by one of the previous method calls.

### DIFF
--- a/upload/admin/model/catalog/product.php
+++ b/upload/admin/model/catalog/product.php
@@ -367,6 +367,7 @@ class Product extends \Opencart\System\Engine\Model {
 		}
 
 		// SEO
+		$this->load->model('design/seo_url');
 		$this->model_design_seo_url->deleteSeoUrlsByKeyValue('product_id', $product_id);
 
 		if (isset($data['product_seo_url'])) {


inventor96:
For additional context, I'm currently making a module extension that manages the catalog based on a 3rd party source. If the source has a change for a product before it has a new or removed product, then the `editProduct()` method gets called first, and the exception gets thrown. I've found that calling `$this->model_catalog_product->deleteProduct(-1);` before I do any processing fixes the issue, as the `design/seo_url` models gets loaded into the `catalog/product` model in that method.

```php
protected function _syncProducts() {
    $this->load->model('catalog/product');
    $this->model_catalog_product->deleteProduct(-1);

    // process the source...
}
```

PR #14983, opened 2025-06-28
Fix null product_id when joining tables without matches

inventor96:
When calling `getProducts()` with parameters that cause a `LEFT JOIN`, and the joined table does not have a match for a give row, the `product_id` field from the joined table is null. And since that version of the `product_id` field comes later in the column order, it takes precedence over the earlier columns when `mysqli_result::fetch_assoc()` is called, and so the `product_id` fields ends up being `null` in the result in PHP, which can obviously cause a number of other issues.

Specifically calling out the `product_id` field from the base table, and making it last in the order of columns, causes it to take precedence.

In my particular case, I encountered the issue when loading the list of products on the admin side, and the page I was on did not have any discounts.

### DIFF
--- a/upload/admin/model/catalog/product.php
+++ b/upload/admin/model/catalog/product.php
@@ -1021,7 +1021,7 @@ class Product extends \Opencart\System\Engine\Model {
 	 * $results = $this->model_catalog_product->getProducts($filter_data);
 	 */
 	public function getProducts(array $data = []): array {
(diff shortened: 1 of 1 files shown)


inventor96:
@danielkerr Regarding the lint test failure, I noticed that the main brach checks are currently failing, too. I looked over the errors in the lint tests here, and they seem to be the same errors you have on main. Also, I don't see anything that would be caused by my changes here... Let me know if there's something I'm missing.

trydalcoholic:
@inventor96 Don't worry, the linter isn't flagging your code. It seems like the linter has been a bit neglected during the development process. I think I'll try submitting a few PRs to clean up the linter errors.

danielkerr:
u dont need to put as product_id

PR #15012, opened 2025-07-06
Run checks for PHP 8.1 to 8.4

mhcwebdesign:
Older PHP versions 7.x or 8.0 aren't supported anymore!


mhcwebdesign:
We still to upgrade PHPStan itself on the 3.0.x.x, have asked @AJenbo to look into it.

PR #15025, opened 2025-07-08
fix(phpstan): remove unreachable statements after throw

trydalcoholic:
### PHPStan Spring Cleaning: Remove Unreachable Statements

Remove code statements that are never executed due to preceding throw statements, resolving PHPStan "deadCode.unreachable" warnings.

#### PHPStan Errors Fixed
- `Unreachable statement - code above always terminates` in system/library/db.php:66
- `Unreachable statement - code above always terminates` in system/library/db.php:76
- `Unreachable statement - code above always terminates` in install/cli_install.php:91

### DIFF
--- a/upload/install/cli_install.php
+++ b/upload/install/cli_install.php
@@ -87,8 +87,6 @@ set_error_handler(function(int $code, string $message, string $file, int $line) 
 	}
 
 	throw new \ErrorException($message, 0, $code, $file, $line);
-
-	return true;
 });
 
 /**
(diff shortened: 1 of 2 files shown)


trydalcoholic:
> we want it to terminate!

Daniel, sorry, but I’m not sure I fully understand your point.
In modern PHP, throwing an exception will automatically terminate the current script execution unless the exception is caught.
Any `exit` or `return` statements after a throw will never be executed.
Is there a specific reason for keeping the exit statements after throw?
Maybe you want to achieve some other behavior that I’m missing?

AJenbo:
Best write it like this to be sure

```
throw new \Exception;
return;
exit;
die;
break;
```

Just in case one of them stops working correctly in PHP :D

AJenbo:
> we want it to terminate!

Seriously if that's the case then you want to use exit and not throw, execution does not return to the function once you have thrown an exception, you cannot control if the application exists in case a layer higher up catches the exception.

PR #15026, opened 2025-07-08
fix(phpstan): remove unused variables from error handler closure

trydalcoholic:
### PHPStan Spring Cleaning: Remove Unused Closure Variables

Remove unused variables from closure `use` statement in error handler function, resolving PHPStan "closure.unusedUse" warnings.

#### PHPStan Errors Fixed
- `Anonymous function has an unused use $config` in system/framework.php:36
- `Anonymous function has an unused use $log` in system/framework.php:36

### DIFF
--- a/upload/system/framework.php
+++ b/upload/system/framework.php
@@ -33,7 +33,7 @@ $log = new \Opencart\System\Library\Log($config->get('error_filename'));
 $registry->set('log', $log);
 
 // Error Handler
-set_error_handler(function(int $code, string $message, string $file, int $line) use ($log, $config) {
+set_error_handler(function(int $code, string $message, string $file, int $line) {
 	// error suppressed with @
 	if (!(error_reporting() & $code)) {
 		return false;


AJenbo (approved):

danielkerr:
they are not unsed. you should check the code without just replying on phpstan. phpstan is just a tool. u need to check everything

AJenbo:
@danielkerr it's clearly not being used inside that closure. If you think it is then point out where instead of just saying so.

```php
{
	// error suppressed with @
	if (!(error_reporting() & $code)) {
		return false;
	}

	throw new \ErrorException($message, 0, $code, $file, $line);
}
```

trydalcoholic:
> they are not unsed. you should check the code without just replying on phpstan. phpstan is just a tool. u need to check everything

I understand and completely agree that everything should be checked thoroughly. We started with approximately 350 PHPStan errors and have now reduced them to around 70, primarily by fixing the most widespread and simplest ones first. Currently, I am indeed trying to examine the PHPStan logs more carefully, as I deliberately left the least clear errors for later.

Please don't be annoyed/irritated if I truly make mistakes – I'm not a robot 🙂. I gratefully rely on your personal professional experience and that of other OpenCart contributors, hoping that my unintentional errors won't be merged into the OpenCart codebase.

As for this specific PR, it seems there's no need to focus on it further, as the correction has already been added 🙂.

AJenbo:
@danielkerr I see you realize it was unused and removed it yourself as part of another big commit: https://github.com/opencart/opencart/commit/94d837830114887bebcbbcbbbd25fa9f15ccf91b

It would be good if you could have mentioned you realized this pr was correct to clear up confusion.

trydalcoholic:
> Not to be negative but it was down to 0 at one point, Daniel doesn't check error reports so someone will need to continuously clean up and deal with all the toxic comments.

Absolutely, reaching 0 PHPStan errors again would be fantastic! PHPStan is genuinely a powerful tool that absolutely deserves a place in every PHP project.

PR #15027, opened 2025-07-08
fix(phpstan): ignore missing config.php files in static analysis

trydalcoholic:
### PHPStan Spring Cleaning: Ignore Missing Config Files

Add PHPStan ignore annotations for `config.php` require statements, resolving "requireOnce.fileNotFound" warnings for environment-specific configuration files.

#### PHPStan Errors Fixed
- `Path in require_once() "config.php" is not a file or it does not exist` in admin/index.php:7
- `Path in require_once() "config.php" is not a file or it does not exist` in index.php:6
- `Path in require_once() "config.php" is not a file or it does not exist` in cron.php:8

### DIFF
--- a/upload/admin/index.php
+++ b/upload/admin/index.php
@@ -4,6 +4,7 @@ define('VERSION', '4.1.0.4');
 
 // Configuration
 if (is_file('config.php')) {
+	/** @phpstan-ignore-next-line requireOnce.fileNotFound */
 	require_once('config.php');
 }
 
--- a/upload/cron.php
+++ b/upload/cron.php
@@ -5,6 +5,7 @@ if (!is_file('config.php')) {
 }
 
 // Config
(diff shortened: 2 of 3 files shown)


AJenbo (approved):

AJenbo (approved):
Nice to see these issues being addressed

AJenbo:
> its required otherwise it would through an error before being redirected to the installer page!

That's what the comment is telling phpstan, why close it instead of merging.?

trydalcoholic:
> its required otherwise it would through an error before being redirected to the installer page!

Exactly! And that's precisely why the instruction tells PHPStan to ignore a file that isn't present. We're not changing the code itself here, but rather directly informing PHPStan that 'everything is under our control' regarding this particular file.

PR #15028, opened 2025-07-08
fix(phpstan): correct PHPDoc parameter types in task and menu models

trydalcoholic:
### PHPStan Spring Cleaning: Fix PHPDoc Parameter Types

Correct PHPDoc parameter type annotations to match actual function signatures and database schema, resolving "parameter.phpDocType" warnings.

#### PHPStan Errors Fixed
- `PHPDoc tag @param for parameter $status with type bool is incompatible with native type string` in admin/model/setting/task.php:103
- `PHPDoc tag @param for parameter $menu_id with type string is incompatible with native type int` in admin/model/tool/menu.php:76

#### Database Schema Reference
The corrections align PHPDoc with the actual database schema where:
- `task.status` is `varchar(16)` → requires `string` type

https://github.com/opencart/opencart/blob/master/upload/system/helper/db_schema.php#L7386

### DIFF
--- a/upload/admin/model/setting/task.php
+++ b/upload/admin/model/setting/task.php
@@ -89,8 +89,8 @@ class Task extends \Opencart\System\Engine\Model {
 	 *
 	 * Edit task status record in the database.
 	 *
-	 * @param int  $task_id primary key of the task record
-	 * @param bool $status
+	 * @param int    $task_id primary key of the task record
+	 * @param string $status
 	 *
 	 * @return void
 	 *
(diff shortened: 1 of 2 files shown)


trydalcoholic:
@danielkerr, I may have been hasty with this PR. After fixing the PHPDoc types, PHPStan now shows that [`admin/controller/marketplace/task.php:183`](https://github.com/opencart/opencart/blob/master/upload/admin/controller/marketplace/task.php#L183) and [`:216`](https://github.com/opencart/opencart/blob/master/upload/admin/controller/marketplace/task.php#L216) are passing `bool` values to `editStatus()`, but the method expects `string`. 

What should be the correct type for task `status` - `bool` or `string`?

danielkerr:
i have not finised with the status system!

PR #15038, opened 2025-07-09
fix(phpstan): remove redundant strict comparison checks in seo_url model

trydalcoholic:
### PHPStan Spring Cleaning: Remove Redundant Strict Comparison Checks

Remove unnecessary `!== ''` checks after `!empty()` and `isset()` conditions in SeoUrl model methods, resolving PHPStan "notIdentical.alwaysTrue" warnings.

#### PHPStan Errors Fixed
- `Strict comparison using !== between mixed~(0|0.0|''|'0'|array{}|false|null) and '' will always evaluate to true` in admin/model/design/seo_url.php:251
- `Strict comparison using !== between mixed~(0|0.0|''|'0'|array{}|false|null) and '' will always evaluate to true` in admin/model/design/seo_url.php:391  
- `Strict comparison using !== between mixed~(0|0.0|''|'0'|array{}|false|null) and '' will always evaluate to true` in admin/model/design/seo_url.php:395


danielkerr:
The current method is fine. U are now changing stuff  for the sake of it. Nothing to do with any errors.

trydalcoholic:
> The current method is fine. U are now changing stuff for the sake of it. Nothing to do with any errors.

I agree that the current code works correctly and doesn't cause any bugs.
However, these PHPStan errors point to redundant checks:
```php
// Current code
if (!empty($data['filter_language_id']) && $data['filter_language_id'] !== '') {
if (isset($data['filter_store_id']) && $data['filter_store_id'] !== '') {
```
PHPStan analyzes that after `!empty() `or `isset()`, the `!== ''` check **always returns true** and is redundant.
My goal is to resolve PHPStan errors in this file.
After these changes, the PHPStan errors are gone, so the code is now free from these static analysis warnings.
If you prefer to keep the current code, we can close the PR. Just wanted to clarify the motivation.

PR #15039, opened 2025-07-09
fix(phpstan): correct method calls in checkout order model

trydalcoholic:
### PHPStan Spring Cleaning: Fix Method Calls in Order Model

Fix undefined method calls in checkout order model by correcting model usage and method names to match actual implementations.

#### PHPStan Errors Fixed
- `Call to an undefined method Opencart\System\Engine\Proxy<Opencart\Catalog\Model\Account\Transaction>::deleteTransactionsByOrderId()` in catalog/model/checkout/order.php:320
- `Call to an undefined method Opencart\System\Engine\Proxy<Opencart\Catalog\Model\Account\Reward>::deleteRewardsByOrderId()` in catalog/model/checkout/order.php:325  
- `Call to an undefined method Opencart\System\Engine\Proxy<Opencart\Catalog\Model\Account\Customer>::getTotalTransactionsByOrderId()` in catalog/model/checkout/order.php:872
- `Call to an undefined method Opencart\System\Engine\Proxy<Opencart\Catalog\Model\Account\Customer>::addTransaction()` in catalog/model/checkout/order.php:873


AJenbo (approved):
It's nice to see some of these issues finally being addressed

trydalcoholic:
> Should be order record. Not order ID record nor order ID in PHPDocs.

@TheCartpenter Could you clarify what you mean? Should the description be `Delete customer rewards by order record from the database.`? In my opinion, and judging by the method names, we're deleting by order ID and the methods expect an `int` primitive in the arguments, not an entity like `Order`. Nevertheless, I should note that you've already done great work with the docblocks and it really helps navigate the codebase. So I'm inclined to trust your judgment.



> It's nice to see some of these issues finally being addressed

@AJenbo Thank you for implementing PHPStan! 🫡 Really great addition to the project.

trydalcoholic:
> You simply need to have the ability to compare the PHPDocs ... there's no order ID in the descriptions. Every `ID` word should be `record`.

What's with the toxic tone? What's the point of that?

trydalcoholic:
> Can you please be constructive instead of insinuating and demeaning? trydalcoholic simply asked for your help and has been nice to you so far. Comments like the ones you are making here is a large part of why I no longer actively contribute here.

@AJenbo Thanks for the support! 🙏

As I mentioned earlier, TheCartpenter wrote a lot of the docblocks and of course I corrected according to his recommendations.
I'm already afraid to write this, but we're doing:
```
DELETE FROM `" . DB_PREFIX . "customer_transaction`
```
And not `IN` the table :D
Honestly, this is getting a bit mentally exhausting. Have a good day everyone!

AJenbo:
> The mentioned `in` on the PHPDocs have nothing to do relatively with the `DELETE FROM` or any queried `FROM` statement. If you look at the rest of the PHPDocs, you will notice `in` and not `from`. Otherwise, you're using your own ways to document the process and that's simply wrong after all the work that's already been done in the core.

It seems to me that whoever wrote the original one used their own way of documenting things, since it's not grammatically correct English.

Maybe we can start with fixing this case and update the others afterwards.

TheCartpenter:
> > The mentioned `in` on the PHPDocs have nothing to do relatively with the `DELETE FROM` or any queried `FROM` statement. If you look at the rest of the PHPDocs, you will notice `in` and not `from`. Otherwise, you're using your own ways to document the process and that's simply wrong after all the work that's already been done in the core.
> 
> I was just joking. Yes, I understood you perfectly and completely agree with you. I'm also for consistency. I see how much work has been done and I sincerely respect that and consider it great work. It's just that if you had started with your last messages, it would have been a bit easier and faster. I hope we're all here to make OpenCart better and in particular to help Daniel Kerr, rather than flame or vent our bad mood. Next time I'll be more careful with my edits 🫡

It's what make the difference of analysis on the topic. While you're taking this as a joke, I rather take it seriously by following along with this project.

trydalcoholic:
> It's what make the difference of analysis on the topic. While you're taking this as a joke, I rather take it seriously by following along with this project.

Absolutely not, I'm not taking this as a joke, nor am I being ironic. I sincerely wrote exactly what I think. I saw somewhere that Daniel personally thanked you for your efforts, and that is truly valuable. I equally value your work and your feedback.

P.S. The "DELETE FROM" joke was just to lighten the mood, not to mock anyone. So, my whole point is that we can be a bit simpler if we're all here for a common good 🙂

PR #15048, opened 2025-07-12
Fix product price on subscriptions

AJenbo:


AJenbo:
I see you fixed it as part of unrelated changes after closing my pr https://github.com/opencart/opencart/commit/8e567ee73ec3f4af9eebff852169cf25c0a706fd

If you don't want to share credit with others then use the provided tools so you don't make the mistake in the first place.

PR #15053, opened 2025-07-12
[PROPOSAL] banner.php using SEO-URLs

osworx:
Changing the building of links this way (together with the proposal for class url.php) will use in future SEO-URLs.


osworx:
Okay, I could add a short check if a valid and complete URL is submitted (stored) and switch then.

PR #15054, opened 2025-07-13
refactor(docker): cleanup container structure and improve user experi…

trydalcoholic:
### Summary

This PR simplifies the Docker container structure and improves developer experience by removing unnecessary complexity:

- Removes public_html symlink to eliminate confusion between upload/ and public_html/
- Configures separate home directory for www-data user to prevent bash history cluttering project root
- Sets proper file permissions to enable OpenCart storage relocation operations
- Updates `make php` command to start directly in OpenCart directory for better developer workflow

> [!WARNING]
> ### Important: Storage Relocation
> 
> When relocating OpenCart storage via admin panel, **always choose paths within `/var/www/`** (e.g., `/var/www/storage_new/`). 
> 
> Choosing paths outside this directory (like `/var/storage/`) will result in **data loss** when containers are recreated, as those directories are not included in Docker volumes.

> [!NOTE]
> ### Notes for developers
> 
> This PR focuses on cleaning up the internal Docker structure without changing functionality. The **Docker configuration is still not considered production-ready** and further work will continue on enhancing its robustness and features.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,3 @@
 # ENV
 .env
 .env.docker
-
-# Docker-specific artifacts
-/public_html
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ apache: ## Enter the Apache container (sh)
 
 php: ## Enter the PHP container (bash)
 	$(call check_service_running,php)
-	@$(COMPOSE) exec php bash
+	@$(COMPOSE) exec --user www-data --workdir /var/www/upload php bash
 
 mysql: ## Enter the MySQL container (bash)
 	$(call check_service_running,mysql)
--- a/docker/apache/conf/httpd.conf
+++ b/docker/apache/conf/httpd.conf
@@ -23,10 +23,10 @@ Group www-data
 
 # Basic server settings
 ServerName localhost
-DocumentRoot "/var/www/public_html"
+DocumentRoot "/var/www/upload"
 
 # Main directory configuration
-<Directory "/var/www/public_html">
+<Directory "/var/www/upload">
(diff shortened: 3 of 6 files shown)


trydalcoholic:
@danielkerr Could you please review the `oc_directory_create` function? I encountered issues during storage relocation testing due to incorrect handling of absolute paths.

When relocating storage to an absolute path like `/var/www/storage_new/backup`, I get this error:
`copy(/var/www/storage_new/backup/index.html): Failed to open stream: No such file or directory in /var/www/upload/admin/controller/common/security.php on line 224`

The issue appears to be that `explode('/', '/var/www/storage/backup')` returns `['', 'var', 'www', 'storage', 'backup']` with an empty first element that breaks the path building logic in [`oc_directory_create`](https://github.com/opencart/opencart/blob/master/upload/system/helper/general.php#L219).

For my testing purposes, I made a quick fix to the function:

```php
function oc_directory_create(string $path, int $permission = 0777): bool {
    $path_new = '';
    $directories = explode('/', rtrim($path, '/'));

    foreach ($directories as $directory) {
        if ($directory === '') {
            if ($path_new === '') {
                $path_new = '/';
            }
            continue;
        }

        if (!$path_new) {
            $path_new = $directory;
        } else {
            $path_new = $path_new . '/' . $directory;
        }

        if (is_dir($path_new . '/')) {
            continue;
        }

        if (!mkdir($path_new . '/', $permission)) {
            return false;
        }
    }

    return true;
}
```

After this change, storage relocation worked successfully. Could you verify if this is expected behavior? Thank you!

PR #15069, opened 2025-07-25
[3.0.x.x] Use CDATA for image caption and title and add linebreaks

MarvinKlein1508:
This PR changes two things for the Google Sitemap extension.

1. It adds `PHP_EOL` to each line to make the file readable when you open it as a human
2. It adds the use of `CDATA` for image caption and image title. Without this the feed generation will result in an invalid feed when your caption is using html formatted special characters. For example: `&auml;`

### DIFF
--- a/upload/catalog/controller/extension/feed/google_sitemap.php
+++ b/upload/catalog/controller/extension/feed/google_sitemap.php
@@ -2,8 +2,8 @@
 class ControllerExtensionFeedGoogleSitemap extends Controller {
 	public function index() {
 		if ($this->config->get('feed_google_sitemap_status')) {
-			$output  = '<?xml version="1.0" encoding="UTF-8"?>';
-			$output .= '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1">';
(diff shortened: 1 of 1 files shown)


ADDCreative:
> Yes I know it will be escaped. In our products it escapes special characters as well. For example it makes `ä` to `&auml;` The escaped character then crashes the sitemap.

That because XML only has 5 predefined character entities, unlike HTML.

When you wrap in CDATA you are saying to ignore any markup. So once parsed the result is `&auml;` and not `ä`. The same for `&`. `&amp;` outside CDATA will correctly be parsed as `&`, but `&amp;` inside  CDATA will be parsed as `&amp;`, which would be incorrect. So you change brakes encoding for everyone else who are not writing data directly to the database.

In your case you would need to decode the entities and then just encode the ones valid in XML. Or use a numeric character reference.

mhcwebdesign:
> You shouldn't need CDATA around $product['name'] as that will already be escaped in the database. If it's not then you will have bigger problems.

Agree. Just tested it with the `ä` character in a product name, it ends up just fine as the UTF-8 Unlaut-a character in the image caption and image title. So I removed the CDATA, see https://github.com/opencart/opencart/commit/209a574817016135edbe473c5c37f9bd29887ab9

PR #15079, opened 2025-07-31
Update DB_SCHEMA - Add STATUS in OC_SSR

ppalashturov:


danielkerr:
ssr does not need status. i will remove it

PR #15082, opened 2025-08-02
[3.0.x.x] Trim posted data

mhcwebdesign:
To trim posted data once for POST requests, with subsequent validations not having to use individual trim calls. Also, currency symbols can now optionally have a blank space.


stalker780:
> Do you think that this maybe will also be integrated into the 4.x branch?

This is already done in OC4 for at a least 2 years ;)
https://github.com/opencart/opencart/blob/e2e4610983d10fa9cacd7379f68fe7d5f6ede52c/upload/system/library/request.php#L109

mhcwebdesign:
You can now set the usage of currency symbol space in the admin backend at
System > Settings > tab Local > Right Currency Symbol Space : Yes or No
System > Settings > tab Local > Left Currency Symbol Space : Yes or No

PR #15123, opened 2025-09-03
fix(docs): remove space and fix link in readme

trydalcoholic:
Removes an extra space in the `[Chat]` link to fix the broken URL.

image


danielkerr:
i thought maybe use slack but seems i need to invite people

PR #15128, opened 2025-09-09
fix(header): HTML characters(for example double quotes) break HTML

SimonLitt:
To fix this I suggest using **escape** filter
https://twig.symfony.com/doc/3.x/filters/escape.html


danielkerr:
its a meta tag. it doesn't use the description with html tags in.

PR #15137, opened 2025-09-26
Update coupon.php

hayden-t:
Default for 'uses total' should be blank, as its much less likely that this option would be needed regularly, if ever. As it limits the discount code to be used only once regardless of customer.


danielkerr:
first its needs to be an int. but its required

PR #15146, opened 2025-10-04
Removed php 7.4 from php-cs-fixer.php

condor2:


condor2:
I will make another commit to upload latest version files for phpstan & php-cs-fixer

PR #15156, opened 2025-10-07
[3.0.x.x] Trim posted data - register controller

condor2:


mhcwebdesign:
What exactly are you trying to resolve here?

If both the password and confirm are empty, then it should only warn about the password:
**Password must be between 4 and 20 characters!**
And not show this:
**Password confirmation does not match password!**
The latter should only be displayed when confirm isn't empty but isn't equal to password.

PR #15171, opened 2025-10-22
Add disallow rule for product search route

halfhope:
Disable search page indexing to prevent ISAP 
https://forum.opencart.com/viewtopic.php?p=881789


mhcwebdesign:
Done, see https://github.com/opencart/opencart/pull/15175

PR #15191, opened 2025-11-08
Fix SEO URL query to remove LIKE condition

chongshengdz:
NO NEED TO USE OR, IT WILL SLOW DOWN THE SITE


trydalcoholic (changes requested):
This change would break the current SEO URL handling in OpenCart. The use of `LIKE` here is intentional. OpenCart supports keywords with `/`, and without this condition such URLs will not be resolved correctly during route parsing. Because of that, this does not look like a safe optimization.

If you have real examples and measurements showing a noticeable performance issue with this query, it would be better to open an issue and include a reproducible scenario, table size, SQL execution plan, and benchmark results.

Issue #15194, opened 2025-11-10
Marketplace API failing with 403 error & bug in marketplace_list.twig around line 172-174


Khnaz35:
What version of OpenCart are you reporting this for?

  4.1.0.3

  Describe the bug

  The marketplace fails to load extensions due to two issues:
  1. API requests to opencart.com return 403 Forbidden errors because cURL requests lack a User-Agent header,
  causing Cloudflare to block them
  2. JavaScript error "download is not defined" appears in console when searching/filtering

  This affects the Admin Extensions > Marketplace section.

  To Reproduce

  1. Go to Admin > Extensions > Marketplace
  2. Configure API credentials (gear icon)
  3. Attempt to browse marketplace extensions
  4. See no extensions displayed and 403 errors in network tab
  5. Try searching for an extension
  6. See JavaScript error in browser console

  Expected behavior

  Marketplace should display available extensions from opencart.com API and search functionality should work without console errors.

  Screenshots / Screen recordings

  Not applicable - network tab shows HTTP 403 responses and browser console shows "Uncaught ReferenceError:
  download is not defined at line 796"

  Server / Test environment (please complete the following information):

  - Deployed to web server
  - Operating system: Linux
  - PHP version: 8.x
  - Apache version: 2.4
  - Browser(s) tested with: Chrome, Firefox
  - Version: Latest

PR #15196, opened 2025-11-10
Add CURLOPT_USERAGENT to cURL requests

Khnaz35:
issue https://github.com/opencart/opencart/issues/15194


Khnaz35:
It does perfectly work for me i will drop you a debug file on forum you can try test with it and see if that does work for you.

PR #15215, opened 2025-11-22
[php 8.5.x Compatibility] image.php

osworx:
Removed deprecated function imagedestroy() - deprecated since 8.5, has no effect since 8.0


osworx:
closed for 1 commit

PR #15228, opened 2025-11-22
[php 8.5 Compatibility] promotion.php

osworx:
Removed deprecated curl_close()

### DIFF
--- a/upload/install/controller/install/promotion.php
+++ b/upload/install/controller/install/promotion.php
@@ -32,8 +32,6 @@ class Promotion extends \Opencart\System\Engine\Controller {
 			$response = '';
 		}
 
-		curl_close($curl);
-
 		return $response;
 	}
 }


batumibiz:
OMG, it seemed to me that The Carpenter appeared again, his style is visible:)
Question to the author: why didn't you send all your improvements in the 1 PR?

osworx:
> OMG, it seemed to me that The Carpenter appeared again, his style is visible:) Question to the author: why didn't you send all your improvements in the 1 PR?

Because when done per GitHUb here, that is not possible.
And I'am using no other method.
Tell that GH to change their function.

But if you so good (or anyone else), you can show me how I can do when you work only here on a Win-PC and have to edit several files and to make then one PR of all.

Finally better x PRs then none.
Or where is your commit in this case ?

batumibiz:
> Because when done per GitHUb here, that is not possible. And I'am using no other method. Tell that GH to change their function.
> 
> But if you so good (or anyone else), you can show me how I can do when you work only here on a Win-PC and have to edit several files and to make then one PR of all.
> 
> Finally better x PRs then none. Or where is your commit in this case ?

It's a good idea to start by using the correct Git client on your PC.
And as an example, [pay attention to this PR](https://github.com/opencart/opencart/pull/15233) and how many files are affected in it.
I would like to see something like this in your PR.

mhcwebdesign:
> > Because when done per GitHUb here, that is not possible. And I'am using no other method. Tell that GH to change their function.
> > But if you so good (or anyone else), you can show me how I can do when you work only here on a Win-PC and have to edit several files and to make then one PR of all.
> > Finally better x PRs then none. Or where is your commit in this case ?
> 
> It's a good idea to start by using the correct Git client on your PC. And as an example, [pay attention to this PR](https://github.com/opencart/opencart/pull/15233) and how many files are affected in it. I would like to see something like this in your PR.

In all fairness, git itself is quite user unfriendly, always has been. Using a terminology like fork, stage, commit, push, pull, etc is quite confusing. Most git clients aren't good. I use GitKraken on a local Linux box, couldn't find a better one. I know of experienced OpenCart developers who simply stay away from github, other than reporting bugs or suggesting code improvements.

PR #15235, opened 2025-11-24
[master - php 8.5 Compatibility] 

osworx:
Removed deprecated curl_close()


osworx:
> Thank you! This is exactly what I wanted to see.

Well thank you.
Next time when you left a comment (as you did at my previous commit) , be constructive with help - that would help others to prevent stupid actions.

And we all here would like to see more of you .. until now only 1 commit - I'm sure you can more .. ;)

PR #15243, opened 2025-12-11
Non full validation PDO library

benycode:
Add addintional-missing required field avoiding error and for full validation protection

protection fully not works, because if the password is not provided, the next step will be interrupted already


trydalcoholic (changes requested):
This changes existing behavior. OpenCart currently allows an empty database password during installation and in the database drivers, so making the password required here would reject valid setups.

For that reason, we cannot merge this change.

PR #15246, opened 2025-12-16
[3.0.x.x ] Remove GLOB_BRACE for glob() to improve compatibility on musl-based systems

varyagnord:
GLOB_BRACE is not supported on some environments using musl libc (e.g., Alpine Linux, Docker images based on Alpine, FrankenPHP). 
Since the glob patterns here do not use braces (no {a,b} syntax), removing GLOB_BRACE does not affect functionality but improves cross-platform compatibility.

### DIFF
--- a/upload/admin/controller/marketplace/extension.php
+++ b/upload/admin/controller/marketplace/extension.php
@@ -29,7 +29,7 @@ class ControllerMarketplaceExtension extends Controller {
 		
 		$data['categories'] = array();
 		
-		$files = glob(DIR_APPLICATION . 'controller/extension/extension/*.php', GLOB_BRACE);
+		$files = glob(DIR_APPLICATION . 'controller/extension/extension/*.php');
 		
 		foreach ($files as $file) {
 			$extension = basename($file, '.php');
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
I think what's really needed, is something like this: 
```
/**
 * A cross-platform wrapper for glob() that simulates GLOB_BRACE 
 * on systems that do not support it natively.
 */
function safe_glob_brace($pattern, $flags = 0) {
    // 1. If GLOB_BRACE is supported and provided, use native glob
    if (defined('GLOB_BRACE') && ($flags & GLOB_BRACE)) {
        return glob($pattern, $flags);
    }

    // 2. Fallback: Manually parse {a,b,c} patterns
    if (preg_match('/\{([^}]+)\}/', $pattern, $matches)) {
        $files = [];
        $parts = explode(',', $matches[1]);
        
        foreach ($parts as $part) {
            // Replace the braced section with the current variation
            $subPattern = str_replace($matches[0], trim($part), $pattern);
            
            // Recursively call to handle multiple sets of braces
            $result = safe_glob_brace($subPattern, $flags);
            
            if (is_array($result)) {
                $files = array_merge($files, $result);
            }
        }
        
        // Remove duplicates and sort if GLOB_NOSORT isn't set
        $files = array_unique($files);
        if (!($flags & GLOB_NOSORT)) {
            sort($files);
        }
        
        return $files;
    }

    // 3. Normal glob behavior for non-braced patterns
    return glob($pattern, $flags);
}
```
There are 7 PHP files in OC 3.0.5.0 which use the GLOB_BRACE, quite often with curly expressions like {a,b,c} etc.
So in order to add support to e.g. some non-GNU systems, such as Solaris or Alpine Linux, we need a more comprehensive approach. The same issue exists for e.g. OC 4.1.0.3!

PR #15251, opened 2025-12-28
fix: Category url query

tongo-angelov:
Change category url build from array_diff_key to array_intersect_key


trydalcoholic (approved):
Looks good and consistent with the rest of the controllers.

PR #15256, opened 2026-01-06
Add country_to_store table

condor2:


upload/system/helper/db_schema.php
```diff
@@ -1491,6 +1491,28 @@ function oc_db_schema() {
 		'collate' => 'utf8mb4_unicode_ci'
 	];
 
+	$tables[] = [
+		'name'  => 'country_to_store',
+		'field' => [
+			[
+				'name'           => 'country_id',
+				'type'           => 'int(11)',
+				'auto_increment' => true
```
trydalcoholic, line 1500: auto_increment should not be used on a foreign key column in a pivot table. country_id here is a reference to country.country_id, not a primary identifier. Other *_to_store tables such as category_to_store and manufacturer_to_store do not use auto_increment, so this should be removed.

upload/system/helper/db_schema.php
```diff
@@ -1491,6 +1491,28 @@ function oc_db_schema() {
...
+			],
+			[
+				'name'    => 'store_id',
+				'type'    => 'int(11)',
+				'default' => '0'
+			]
+		],
+		'primary' => [
```
trydalcoholic, line 1508: The primary key should be composite: ['country_id', 'store_id']. With only country_id, it is possible to insert duplicate (country_id, store_id) pairs. The [upgrade_14.php](https://github.com/opencart/opencart/blob/master/upload/install/controller/upgrade/upgrade_14.php#L64C81-L64C97) script inserts multiple rows per country, one per store, so this would lead to duplicate key errors during upgrade.

trydalcoholic (changes requested):
Thanks for the contribution.

Before adding this table to `db_schema.php`, it would be good to step back and look at the bigger picture.

`country_to_store` is referenced only in one place in the codebase, in [`upgrade_14.php`](https://github.com/opencart/opencart/blob/master/upload/install/controller/upgrade/upgrade_14.php#L64C81-L64C97), where it is populated during migration. There are no controllers or models that read from this table. There is no `getCountriesByStore()` or any joins using `country_to_store`. At the moment, it looks like unused data.

Because of that, the right approach might be the opposite. Instead of adding this table to the schema, it may make more sense to remove its population from `upgrade_14.php`. Could you clarify what feature or query is expected to use this table? Without a clear use case, adding it to `db_schema.php` would create an unused table on fresh installs.

The table definition itself also has structural issues, see inline comments.

Foreign key constraints are missing as well. All `*_to_store` tables in the schema define them, so they should be added here too.

PR #15257, opened 2026-01-06
Remove duplicate text entry for 'Filter'

condor2:
Removed duplicate 'Filter' text entry.


trydalcoholic (approved):
Another duplicate key removed here as well. Good catch.

PR #15258, opened 2026-01-06
Remove duplicate text_reset entry in authorize.php

condor2:


trydalcoholic (approved):
This is indeed a duplicate key. Good catch.

PR #15259, opened 2026-01-06
Modify password length requirement in forgotten.php

condor2:
Updated password length requirement from 20 to 40 characters.


trydalcoholic (approved):
The error message was out of sync with the actual validation in `admin/controller/common/forgotten.php`, which already enforces a max length of 40 using `oc_validate_length(..., 40)`. Good catch.

PR #15264, opened 2026-01-15
[3.0.x.x] Fix admin login password encoding

ADDCreative:


nekto-kotik:
I can't believe it's finally fixed! Thank you!

PR #15268, opened 2026-01-20
Increase module table field size for html extension

hayden-t:
Due to the setting column being used for storing html for the html module/extension it would be a good idea to increase the limit on this that ive hit.

### DIFF
--- a/upload/system/helper/db_schema.php
+++ b/upload/system/helper/db_schema.php
@@ -4240,7 +4240,7 @@ function oc_db_schema() {
 			],
 			[
 				'name' => 'setting',
-				'type' => 'text'
+				'type' => 'longtext'
 			]
 		],
 		'primary' => [


trydalcoholic (commented):
Could you please share the scenario in which you reached this limit? Have you considered alternative approaches for storing HTML? It would help to understand your use case and whether such database changes should be considered.

`longtext` feels like overkill, to be honest, but `mediumtext` might be worth considering and discussing.

In any case, changes like this would require a migration in `upload/install/upgrade/*`.

hayden-t:
Its the system/default HTML extension, you can therefore store an open ended length of content. How else would the extension store the code ?

stalker780:
@trydalcoholic 
This has been a long-standing OC issue dating back to v1.5.
Fields that store JSON arrays or HTML content frequently hit the 64KB limit of the `TEXT` type, which can lead to truncated data.

These fields should use `MEDIUMTEXT` out of the box.

While this choice may have made sense 10-15 years ago, there is no practical reason today to prefer `TEXT` over `MEDIUMTEXT` in modern MySQL/MariaDB. For typical OC workloads, any differences in performance or resource usage are negligible, especially compared to the risk of losing data or settings.

I agree that `LONGTEXT` is unnecessary here. Storing more than 16MB in a single field is a very specific use case and not typical for OC, its extensions, or the default DB schema.

The fields most commonly affected are:

```
category_description.description
product_description.description
article_description.description
topic_description.description
information_description.description
manufacturer_description.description - master branch only (likely not yet ported to 4.x.x.x)
session.data
setting.value
module.setting
```

If you'd like, I can prepare PRs for both the `master` and `4.x.x.x` branches, including the necessary installer changes.

trydalcoholic:
I agree that the current 64 KB limit already feels quite restrictive without much benefit. Using `mediumtext` here makes sense.

@hayden-t would you like to update your PR?
@stalker780 it would be great if you could also open a PR. You can probably start with the 4.x.x.x branch. If @hayden-t is fine with it, @stalker780 could also open a PR for the master branch.

For consistency, OpenCart already uses `mediumtext` for `information_description.description` in master. It would make sense to do the same for:

* `product_description.description`
* `category_description.description`
* `article_description.description`
* `topic_description.description`
* `manufacturer_description.description`

`module.setting` also looks like a good candidate for `mediumtext`, and `session.data` may also be worth considering.

~~It is important to include upgrade steps for these changes. I would also suggest adding a comment in the code with a link to this PR for context.~~

trydalcoholic:
I also checked the upgrade scripts. It looks like a separate upgrade script is not needed here, OpenCart should handle updating the column automatically.

[upgrade_4.php](https://github.com/opencart/opencart/blob/master/upload/install/controller/upgrade/upgrade_4.php#L76-L104)

hayden-t:
might be best if you guys do the PR since its changed in scope and you have investigated that further, thank you, medium text is likely fine, id just chosen the biggest just coz

stalker780:
@trydalcoholic 
https://github.com/opencart/opencart/commit/f5f30e47dd3b6a45f2af7d2bc4f511109a84b486

Added a commit with larger fields types.
I have tested the upgrade process. It works fine

trydalcoholic:
Closing this PR in favor of #15373, where the field size changes have now been added.

PR #15279, opened 2026-02-02
Return inserted rows ids

stalker780:
Every Add method should return the inserted row id for usage in events

### DIFF
--- a/upload/catalog/model/checkout/order.php
+++ b/upload/catalog/model/checkout/order.php
@@ -525,8 +525,12 @@ class Order extends \Opencart\System\Engine\Model {
 	 *
 	 * $this->model_checkout_order->addOption($order_id, $order_product_id, $order_option_data);
 	 */
-	public function addOption(int $order_id, int $order_product_id, array $data): void {
+	public function addOption(int $order_id, int $order_product_id, array $data): int {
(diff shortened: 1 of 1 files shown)


upload/catalog/model/checkout/order.php
```diff
@@ -716,6 +724,10 @@ public function getTotalOrdersBySubscriptionId(int $subscription_id): int {
 	 */
 	public function addTotal(int $order_id, array $data): void {
```
trydalcoholic: `addTotal()` still has a `: void` return type but now returns `$order_total_id`. This will cause a fatal error in PHP 8. Please change the return type to `: int`.

trydalcoholic (changes requested):
Looks good overall and consistent with the existing pattern in `addOrder()` and `addProduct()`. One thing to fix: `addTotal()` should return `: int` instead of `: void`. Once this is updated, the PR should be good to merge.

upload/catalog/model/checkout/order.php
stalker780: My fault. Missed the line. Fixed it.

trydalcoholic (approved):
Thanks for the update, looks good now.

PR #15287, opened 2026-02-13
Fix missing editToken method in catalog model account customer

NanheKumar:
## Problem
Admin -> Customers -> Action -> Login into store triggered:
`Notice: Undefined property: Proxy::editToken`
from `catalog/controller/account/login.php` (token method).

## Cause
`model_account_customer->editToken()` is called in catalog login token flow, but
`catalog/model/account/customer.php` did not define `editToken()`.

## Fix
Added `editToken(string $email, string $token): void` to
`catalog/model/account/customer.php` to update `customer.token` by email.

## Repro
1. Admin panel -> Customers
2. Action -> Login into store
3. Notice is thrown

## Result after fix
Login-into-store flow works without the Proxy::editToken notice.

### DIFF
--- a/upload/catalog/model/account/customer.php
+++ b/upload/catalog/model/account/customer.php
@@ -110,6 +110,26 @@ class Customer extends \Opencart\System\Engine\Model {
 	}
 
 	/**
+	 * Edit Token
+	 *
+	 * Edit customer token record in the database.
+	 *
+	 * @param string $email
+	 * @param string $token
+	 *
+	 * @return void
+	 *
+	 * @example
+	 *
+	 * $this->load->model('account/customer');
+	 *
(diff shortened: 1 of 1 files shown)


trydalcoholic (changes requested):
The `customer.token` column exists in the schema, but it is not read or written anywhere in master. It looks like a legacy field that was replaced by the `oc_customer_token` table. Because of that, `editToken()` is not called by any current code path, and the error "Notice: Undefined property: Proxy::editToken" cannot be reproduced on master.

Could you confirm which version or branch you are testing on? If this is needed for an upcoming feature, it should include the corresponding calling code. If this is a bug in master, please share a reproduction case that works on an unmodified master.

PR #15296, opened 2026-03-01
Add missing tables and columns to db_schema

cvml11:
When the current master branch is cloned and run, it throws errors for missing columns and tables. I have filled those into the db_schema.php so others won't face the same error.


upload/system/helper/db_schema.php
```diff
@@ -7561,6 +7561,53 @@ function oc_db_schema() {
...
+				'auto_increment' => true
+			],
+			[
+				'name'    => 'store_id',
+				'type'    => 'int(11)',
+				'default' => '0'
+			],
+			[
+				'name'    => 'route',
+				'type'    => 'int(11)',
```
trydalcoholic: The `route` column is set as `int(11)` in your PR, but it stores strings like `common/home`. It should be `varchar(64)`, same as in 4.x.x.x.

upload/system/helper/db_schema.php
```diff
@@ -7561,6 +7561,53 @@ function oc_db_schema() {
...
+				'table' => 'store',
+				'field' => 'store_id'
+			]
+		],
+		'engine'  => 'InnoDB',
+		'charset' => 'utf8mb4',
+		'collate' => 'utf8mb4_unicode_ci'
+	];
+
 	$tables[] = [
```
trydalcoholic, line 7611: Missing `value` column in `translation`

The [`admin/model/design/translation.php`](https://github.com/opencart/opencart/blob/master/upload/admin/model/design/translation.php) model uses a `value` column in its INSERT and UPDATE queries, but it is not in the schema. Without it, saving translations will fail.

Please add:

```php
[
    'name' => 'value',
    'type' => 'text'
],

**Review by trydalcoholic — CHANGES_REQUESTED** [review]
Hey @cvml11, thanks for looking into this and fixing the missing schema definitions.

Before we can merge, there are two things to fix.

File: upload/system/helper/db_schema.php [hunk 3]
```diff
@@ -7561,6 +7561,53 @@ function oc_db_schema() {
...
+			],
+			[
+				'name'    => 'store_id',
+				'type'    => 'int(11)',
+				'default' => '0'
+			],
+			[
+				'name'    => 'route',
+				'type'    => 'varchar(64)',
+				'default' => '0'
```
trydalcoholic, line 7580: `route` is defined as `varchar(64)`, so the default should be an empty string `''` rather than `'0'`. Everything else looks good, thanks!

trydalcoholic (changes requested):

trydalcoholic (approved):
Looks good to me. Thanks for the updates.

PR #15299, opened 2026-03-05
modify docker config

alex93x:
configuring a user in a container to ensure compatibility of file modification rights between the local user and the user in the container


docker/php/Dockerfile
```diff
@@ -26,9 +26,17 @@ RUN install-php-extensions \
...
-# Configure www-data user home directory
-RUN usermod -d /home/www-data -m www-data
-RUN mkdir -p /home/www-data && chown www-data:www-data /home/www-data
+# Configure user home directory
+ARG PHP_UNAME
+ARG PHP_UID
+ENV PHP_UNAME=${PHP_UNAME}
+ENV PHP_UID=${PHP_UID}
+
+RUN useradd -m -G www-data,root -u $PHP_UID -d /home/$PHP_UNAME $PHP_UNAME && \
```
trydalcoholic: If `PHP_UID` or `PHP_UNAME` are empty (e.g. running `docker compose build` directly without `make`), `useradd` will fail. Please add defaults:

    ARG PHP_UNAME=www-data
    ARG PHP_UID=33

Makefile
```diff
@@ -84,7 +88,11 @@ init: ## Initialize the project (copies .env.docker)
 	fi
 
 build: ## Build images. Use 'options' for flags (e.g., --no-cache)
-	$(COMPOSE) build $(options)
+	@echo "Building images for user: $(CURRENT_USER_NAME) (ID: $(CURRENT_USER_ID))"
+	$(COMPOSE) build \
+		--build-arg PHP_UNAME="$(CURRENT_USER_NAME)" \
+        --build-arg PHP_UID="$(CURRENT_USER_ID)" \
```
trydalcoholic: This line uses spaces instead of a tab for indentation. Makefile recipes require tabs. Please fix to avoid potential issues.

trydalcoholic (changes requested):
Thanks for fixing this, UID mismatch between host and container is a real issue on Linux.

I left a few comments in the PR. Once those are addressed, this should be good to merge.

docker/php/Dockerfile
alex93x: add changes

Makefile
alex93x: add changes

trydalcoholic (changes requested):
Thanks for the update.

I re-tested the PR and there is still one issue to resolve before merge. Both build paths should work, not only `make build`, but also plain `docker compose build php`.

At the moment, the direct compose build still fails in the fallback path, so the remaining work is:

* Make sure `docker compose build php` works without going through `make`
* Keep `make build` working as well
* Verify that this actually resolves the file ownership issue on Linux, not just the image build

Once these cases are covered, this should be in good shape.

alex93x:
Added the option to use the `docker-compose` command directly; in this case, you need to manually declare the user in the `.env` file and specify the path to the file when running the command

trydalcoholic (approved):
The build error is fixed and the structure is cleaner.

I will verify that runtime file ownership works correctly on Linux with bind mounts. If any further changes are needed, I will address them in a follow-up.

Issue #14744, opened 2025-03-26
[4.1.x.x]BUG in product settlement


zhangby1234:
When there are items in the shopping cart, click "checkout", log in to the account on the checkout page, then select the delivery method and settlement gateway, and click "confirm" to jump to the payment gateway. At this time, if you find that you have purchased the wrong item and do not want to pay, and return to the shopping cart page to add or delete items and click "checkout" again to jump to the payment gateway again, the settlement price and product information displayed on the payment gateway are not the information of the product purchased again after modification, but the product information and price that were not settled last time.

danielkerr:
i would say this is an issue with ur payment gateway

danielkerr:
if u add an item in the cart, remove an item or edit the quantity opencart deletes the session data that holds for the payment and shipping. it seems ur payment gateway company is creating a transaction that can not be changed.

danielkerr:
what payment gateway is causing the issue.

danielkerr:
im going to alter the system to avoid this.

Issue #14739, opened 2025-03-25
[4.1.0.2-4.1.0.3]In the ADMIN backend upload image management window (image manager), after creating multiple folders (for example, 6 folders), the root directory images are not displayed and are hidden.


zhangby123:
In the ADMIN backend upload image management window (image manager), after creating multiple folders (for example, 6 folders), the root directory images are not displayed and are hidden.

codemake2022:
In filemanager.php, find
` foreach (array_slice($directories + $files, $start, $limit) as $path) { `
and modify it to 
`foreach (array_slice(array_merge($directories, $files), $start, $limit) as $path) {`

PR #15337, opened 2026-04-12
[4.x.x.x] More fixes from  batumibiz

mhcwebdesign:
After creating multiple folders, the root directory images are not displayed (https://github.com/opencart/opencart/issues/14739)

BUG in product settlement (https://github.com/opencart/opencart/issues/14744)

### DIFF
--- a/upload/admin/controller/common/filemanager.php
+++ b/upload/admin/controller/common/filemanager.php
@@ -130,7 +130,7 @@ class FileManager extends \Opencart\System\Engine\Controller {
 			}
 
 			// Split the array based on current page number and max number of items per page of 10
-			foreach (array_slice($directories + $files, $start, $limit) as $path) {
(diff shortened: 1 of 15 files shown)


batumibiz:
@mhcwebdesign I recommend that you constantly update [CHANGELOG](https://github.com/opencart/opencart/blob/master/CHANGELOG.md) _(which has long been abandoned)_, so it will be more clear what has already been done and there will be no confusion. Take [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) as a basis, they have a very convenient and clear structure, this is a "good practice".

Everything is already structured in my [CHANGELOG](https://github.com/batumibiz/oc4/blob/main/CHANGELOG.md), take it from there.
The update has been moved - move the item from CHANGELOG along with it.

mhcwebdesign:
Yes, you are right, we should have an updated CHANGELOG.md . The same is true for the README.md and UPGRADE.md.

Thanks for pointing it out.

PR #15350, opened 2026-04-13
[4.x.x.x] Update CHANGELOG.md

mhcwebdesign:
updated change log for 4.x.x.x


batumibiz:
You are still maintaining the CHANGELOG in a "flat" format.
Wouldn't it be better to adopt the ["Keep a Changelog"](https://keepachangelog.com/en/1.1.0/) standard instead?

It is much more practical—everything is structured and significantly easier for developers to parse at a glance.

PR #15367, opened 2026-04-20
[4.x.x.x] Fix cache

alex93x:
Added fixes for integrating various caching engines

### DIFF
--- a/upload/admin/controller/common/developer.php
+++ b/upload/admin/controller/common/developer.php
@@ -65,15 +65,7 @@ class Developer extends \Opencart\System\Engine\Controller {
 		}
 
 		if (!$json) {
-			$files = glob(DIR_CACHE . 'cache.*');
-
-			if ($files) {
-				foreach ($files as $file) {
-					if (is_file($file)) {
-						unlink($file);
-					}
-				}
-			}
+			$this->cache->clear();
 
 			$json['success'] = $this->language->get('text_cache_success');
 		}
--- a/upload/install/cli_install.php
+++ b/upload/install/cli_install.php
@@ -433,8 +433,15 @@ class CliInstall extends \Opencart\System\Engine\Controller {
 		}
 
 		if ($option['db_ssl_ca']) {
-			$output .= 'define(\'DB_SSL_CA\', \'' . addslashes($option['db_ssl_ca']) . '\');' . "\n";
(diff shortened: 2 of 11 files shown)


mhcwebdesign:
Thank you for your pull-request. Before merging, can you fix the 'system/config/default.php' in your pull-request please, as I think this line is wrong:
`$_['cache_engine']         = CACHE_ENGINE; // apc, file, mem, memcached or redis`
It should be:
`$_['cache_engine']         = 'file'; // apc, file, mem, memcached or redis`

alex93x:
> Thank you for your pull-request. Before merging, can you fix the 'system/config/default.php' in your pull-request please, as I think this line is wrong: `$_['cache_engine'] = CACHE_ENGINE; // apc, file, mem, memcached or redis` It should be: `$_['cache_engine'] = 'file'; // apc, file, mem, memcached or redis`

add changes

mhcwebdesign:
I think the cache engine, as defined in the 'system/config/default.php', doesn't need to be specified again in 'system/config/admin.php' or 'system/config/catalog.php' ! It should be sufficient to just have it in the 'system/config/default.php'.

alex93x:
> I think the cache engine, as defined in the 'system/config/default.php', doesn't need to be specified again in 'system/config/admin.php' or 'system/config/catalog.php' ! It should be sufficient to just have it in the 'system/config/default.php'.

At first, that’s exactly what I did, but following some objections, I revised it to match the way other configuration parameters (such as database settings) are defined.
The idea is that we can specify the value of `CACHE_ENGINE` in the `config.php` files, based on which the appropriate cache engine will then be enabled in `framework.php`.

mhcwebdesign:
Maybe we can use this in the system/config/admin.php and system/config/catalog.php :

```
if (defined('CACHE_ENGINE')) {
    $_['cache_engine'] = CACHE_ENGINE';
}
```
That is, only use a different cache engine when specified in the config.php or admin/config.php.

alex93x:
> Maybe we can use this in the system/config/admin.php and system/config/catalog.php :
> 
> ```
> if (defined('CACHE_ENGINE')) {
>     $_['cache_engine'] = CACHE_ENGINE';
> }
> ```
> 
> That is, only use a different cache engine when specified in the config.php or admin/config.php.

add changes

PR #15374, opened 2026-04-21
[4.x.x.x] Support multi-word product name search

stalker780:
Improve product autocomplete and name filter.
- Typing `nano ipod` will find product `ipod nano` .
- Typing `Galaxy 10` or `Samsung Tab 10` will find product `Samsung Galaxy Tab 10.1`. 

While this is impossible to find with default OC4 implementation where you need to remember the first word the product name starts with and exact words order. This is very inconvenient when you have tens or hundreds thousands of products. You cannot remember each exact product name. This is an old OC issue which lasts for decades and drives every product manager crazy :)

Also fix 'getProduct' and 'getTotalProducts' different `filter_name` SQLs.

----
Similar approach is used in frontend. But why `OR` is used there? 
Every additional word in search string must narrow the results, not expand them. 
In frontend if you search `nano ipod` you get everything with `ipod` or `nano` - you get all the ipods while you were searching for the exact model name. Imagine how this search result looks in the store with 100k products and lots of repeating words.

https://github.com/opencart/opencart/blob/f5f30e47dd3b6a45f2af7d2bc4f511109a84b486/upload/catalog/model/catalog/product.php#L163

If you make a search in frontend `ipod samsung touch canon apple macbook nikon sony` it makes no sense, but you just have selected all the store products. You can repeat the same on the store with 100k products :)

### DIFF
--- a/upload/admin/model/catalog/product.php
+++ b/upload/admin/model/catalog/product.php
@@ -1033,7 +1033,18 @@ class Product extends \Opencart\System\Engine\Model {
 		}
 
 		if (!empty($data['filter_name'])) {
-			$sql .= " AND LCASE(`pd`.`name`) LIKE '%" . $this->db->escape(oc_strtolower($data['filter_name']) . '%') . "'";
+			$implode = [];
+
+			$words = explode(' ', trim(preg_replace('/\s+/', ' ', $data['filter_name'])));
+			$words = array_filter($words);
+
+			foreach ($words as $word) {
+				$implode[] = "LCASE(`pd`.`name`) LIKE '%" . $this->db->escape(oc_strtolower($word)) . "%'";
+			}
+
+			if ($implode) {
+				$sql .= " AND (" . implode(" AND ", $implode) . ")";
+			}
 		}
 
 		if (!empty($data['filter_model'])) {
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
In a default frontend search for 'nano ipod' it comes up with these results:

iPod Classic
iPod Nano
iPod Shuffle
iPod Touch

That is, it searches for all entries containing 'nano' or 'ipod', as expected, though I understand your reasoning that for larger databases it may make sense to use a 'nano' and 'ipod' logic, instead of the or-logic. 

You can't please everyone. Perhaps it makes more sense to make these different search approaches configurable.

stalker780:
@mhcwebdesign In my experience the AND logic makes more sense as you can refine your search adding more parts of product name to the search string, which is intuitive.
But with OR logic the more words you enter in the search string the more unrelated results you get, up to the whole DB selection.

stalker780:
@mhcwebdesign 
So what do we decide with the product search?

If you wish, I can add a store setting where we can configure exact/and/or product search.
Even make this setting separate for admin and catalog.
This will add more model ifs in getProducts and getProductsTotal. 
Not sure it makes much sense.

But current search/autocomplete in admin area is definitely unusable.
My customers never liked default search/autocomplete in OC 1.5/2.x/3.x, cause it is impossible to find anything.
Literally none of my customers stores use the default product search logic, even if there are only 100 products.
And now the same problem migrated to 4.x.

What I proposed is the simplest tested approach which I use at least for the last 8 years. And it suits most customers, no matter 100k products or 10. This search logic is very convenient, intuitive and self-consistent out of the box. Of course I am speaking only about MySQL `LIKE` text search.

mhcwebdesign:
Well, could you create a pull-request for making the product search logic (AND vs OR) configurable in the admin backend? Maybe in the admin System > Stores > Settings > Options . And dependent on the setting, the product model the uses the appropriate search logic. This way we can still preserve backwards-compatibility. but also cater for stores with large number of products.

stalker780:
Added a new PR. Closing this one as the branch got too many conflicting changes

PR #15377, opened 2026-04-23
[4.x.x.x] Fix pagination icons

stalker780:
Use modern FontAwesome icons for pagination, `>|` looks outdated even for OC2, especialy in the frontend.
I know 4.1.0.4 is planned as a bugfix release. But I hardly imagine any extension this fix can break.

image


mhcwebdesign:
It already uses the correct '>|' icon.

PR #15393, opened 2026-04-27
Check code style on PHP 8.1 instead of 8.3

AJenbo:


mhcwebdesign:
Can't the code style check be removed altogether? Now PHPstan lists the messages for PHP 8.1, before it did it for PHP 8.3.

AJenbo:
That's like saying "Can't we just skip refilling the coolant? We already filled the car with a full tank of gas."

PHPStan isn't the one listing these. The remaining issues are all from PHP-CS-Fixer.

The CI setup does 3 things:
- **PHP Lint** checks if the code can be parsed.
- **PHPStan** does static analysis to check for errors.
- **PHP-CS-Fixer** does code style.

You've fixed the issues detected by PHPStan, but you still haven't addressed the issues where the code doesn't follow the declared style of the project. These are different things tracked by different tools.

We could run them in different jobs entirely, but then we'd have to pay the startup penalty for each, and there would be 11 instead of 5 jobs.

And again, this is trivial to fix. Just run `php tools/php-cs-fixer.phar fix` to have PHP-CS-Fixer correct all the style issues automatically. This is what TheCartpenter has often done manually one file at a time (simple code style fixes), but PHP-CS-Fixer automates all of that.

mhcwebdesign:
Ok, thanks for the clarification. Will try the tools/php-cs-fixer.phar next then to address these issues.

mhcwebdesign:
OK, further to this, having run the tools/php-cs-fixer.phar and merged the updated files, it now looks fine, see https://github.com/opencart/opencart/pull/15400 !

AJenbo:
I rebased the PR, you can either merge it if you prefer that the style check run on 8.1 (will require that it's updated each time you drop support for a version), or close it if you are fine with the check running on 8.3.

At least now that the status is green again it can be helpful in reviewing PRs, I'm really happy to see the status restored.

@trydalcoholic you might like this as well.

@mhcwebdesign how do you intend to sync changes from master, or is master on it's way to becoming 5.0? If so I really hope he gets rid of the awful registry that thing is making type validation impossible on a higher level.

mhcwebdesign:
> I rebased the PR, you can either merge it if you prefer that the style check run on 8.1 (will require that it's updated each time you drop support for a version), or close it if you are fine with the check running on 8.3.
> 

Let's keep it at PHP 8.3.

> @mhcwebdesign how do you intend to sync changes from master, or is master on it's way to becoming 5.0? If so I really hope he gets rid of the awful registry that thing is making type validation impossible on a higher level.

Well, the master branch really is more like becoming a 5.0 version, though Daniel treats it as a 4.2.x.x .
Either way, some of the changes applied to the 4.x.x.x should also go into the master branch, too. I'd have to go through the 4.x.x.x change log and then re-create separate needed pull-requests for the master branch.

trydalcoholic:
> @trydalcoholic you might like this as well.
Oh yeah, green pipelines 😌
Great work!

PR #15396, opened 2026-04-27
[4.x.x.x] Update order.php

billynoah:
Voiding the order here should only happen if it has already been confirmed. Since `editOrder()` is called from checkout/confirm.confirm when updating either shipping or payment method, the current logic can result in orders with a voided status that have yet to be confirmed.


mhcwebdesign:
There were some minor issues reported by PHPStan and PHP CS Fixer, which we adjusted, too, along with your changes. See https://github.com/opencart/opencart/pull/15403

PR #15407, opened 2026-05-01
[4.x.x.x] Image exists helper

alex93x:
Added a helper to check whether an image file exists


mhcwebdesign:
This is not a bug, and the check for the existence of the image file is only ever called from one file, namely the catalog/controller/product/category.php .

alex93x:
> This is not a bug, and the check for the existence of the image file is only ever called from one file, namely the catalog/controller/product/category.php .

Well, it’s not a bug fix, but it does make the code easier to read. I’ve added it to one controller as an example. So how should we improve the codebase?

PR #15410, opened 2026-05-02
[4.x.x.x] Request library changes

alex93x:
I have changed the logic of the methods so that they can accept a default value if no value is found for the specified key. This will significantly improve the code’s readability. I have added a new method, `getOnly`, to retrieve an array of values for a given set of keys.

Examples:
```
$product_id = (int)$this->request->get('product_id', 0);
```

```
$query = [
    'user_token' => $this->session->data['user_token'],
    ...$this->request->getOnly(['sort', 'order', 'page'])
];

$data['action'] = $this->url->link('extension/evasota/faq/faq.list', $query);

$this->url->link('extension/evasota/faq/faq.form', ['id' => $result['id'], ...$query])

```
```
$data['module_faq_description'] = $this->config->get('module_faq_description', []);
```

### DIFF
--- a/upload/system/engine/config.php
+++ b/upload/system/engine/config.php
@@ -47,8 +47,12 @@ class Config {
 	 *
 	 * @return mixed
 	 */
-	public function get(string $key) {
-		return $this->data[$key] ?? '';
+	public function get(string $key, $default = ''): mixed {
+		if ($this->has($key)) {
+			return $this->data[$key];
+		}
+
+		return $default;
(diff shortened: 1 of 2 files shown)


GregoryBiter (approved):

mhcwebdesign:
This is not a bugfix, but a change of the OpenCart core. The latter should only be changed as needed in 3rd party extensions, e.g. via OCmod.

PR #15413, opened 2026-05-03
Enable HttpOnly for Session Cookies by Default

uwezkhan:
This patch enables the HttpOnly attribute for OpenCart session cookies by default.

Previously, session cookies were issued with HttpOnly disabled in system/framework.php, allowing client-side JavaScript to access authentication/session cookies via document.cookie.

Setting HttpOnly to true improves session security by preventing JavaScript access to session cookies, reducing the impact of XSS vulnerabilities and aligning OpenCart with modern secure cookie best practices.

Security Impact:
- Mitigates session theft via client-side script execution
- Hardens authentication/session handling by default
- Improves secure-default posture for all OpenCart installations

Compatibility:
- No functional impact expected unless custom code explicitly relies on JavaScript access to the session cookie.


hayden-t:
sound great, let accept it asap. it should also include https://github.com/opencart/opencart/blob/88f941779e4573467b243619330fe5bead024132/upload/admin/controller/startup/session.php#L33

uwezkhan:
Good catch. Pushed the same change to admin/controller/startup/session.php so the admin session cookie gets httponly too.

Before: admin startup set 'httponly' => false, so the admin session cookie was reachable from document.cookie.
After: both the framework default and the admin startup use 'httponly' => true, so client-side JS can't read either session cookie.

Only tradeoff is any custom code that read the session cookie from JS, which shouldn't be relying on that anyway. Also dropped a stray empty headers.txt that slipped into the branch.

uwezkhan:
i am extremely sorry for replying this late . i had missed this repo.

uwezkhan:
@hayden-t  just wanted to check if this PR still matters?

hayden-t:
for sure, it hasnt been changed in master if thats what you mean and there's no reason not to adopt it for all cookies.

uwezkhan:
Makes sense. Enabled it for every cookie OpenCart sets now, not just the session ones.

Before: only the session cookie (framework.php + admin startup) got httponly from this PR. The catalog session startup, currency, tracking, language, notification, cookie-policy, and the admin/customer authorize device tokens were all still reachable from document.cookie.

After: they all set httponly=true. The two authorize cookies were still on the old 3-arg setcookie() signature, so I moved them to the options-array form to add the flag.

None of these get read from JS anywhere in the storefront or admin, so no functional change. Same tradeoff as before: custom code that read a cookie from JS would have to read it server-side instead.

uwezkhan:
any update?

PR #15414, opened 2026-05-03
Regenerate Session ID on Authentication to Prevent Session Fixation

uwezkhan:
This patch updates OpenCart to regenerate the session ID whenever a user moves from a guest session into an authenticated session.

At the moment, the same session identifier is preserved after login, registration auto-login, and other authentication flows. That creates a session fixation risk because a pre-existing session ID can remain valid after authentication.

To address this, I added session regeneration support to the core session library and introduced a helper to handle cookie/session rotation cleanly. The new session ID is now issued across all major authentication paths, including customer login, admin login, registration-based auto-login, checkout registration, and API customer authentication.

The goal is to ensure that every privilege transition from unauthenticated to authenticated state results in a fresh session identifier, bringing session handling more in line with standard security practices while keeping existing session data intact.

### DIFF
--- a/upload/admin/controller/common/login.php
+++ b/upload/admin/controller/common/login.php
@@ -95,6 +95,9 @@ class Login extends \Opencart\System\Engine\Controller {
 		}
 
 		if (!$json) {
+			// Regenerate session ID
+			oc_session_regenerate($this->registry);
+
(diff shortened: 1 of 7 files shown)


uwezkhan:
any update ?

Issue #15325, opened 2026-04-10
Adding product variant to order


epr-ds:
**What version of OpenCart are you reporting this for?**
Opencart version 4.1.0.3

**Describe the bug**
Order can't complete due to empty stock for the selected product.

What section does it affect?
Order section is affected

**To Reproduce**
1. Add a master product with quantity 1 and subtract from the stock enabled.
2. Add a variant product with its own quantity enabled, let's type 1 too.
3. Add new order that sells master product. Thus, master product will be empty in stock.
4. Try to add other order and select variant product, but system automatically add to cart the master product, and since master product is out of stock, adding new product will not proceed.

**Expected behavior**
The order should have selected the variant product and can continue to add order as variant product quantity is equal to 1.

**Screenshots / Screen recordings**

Master product and its variant (master product has zero quantity as it was sold earlier):
<img width="1207" height="215" alt="Image" src="https://github.com/user-attachments/assets/388dcac5-3d7c-4f11-8df7-723a07099b5f" />

Adding the product variant to the cart but it gets the master product selected:
<img width="1631" height="955" alt="Image" src="https://github.com/user-attachments/assets/0b4e8077-312d-4ced-aafc-db84e8c74535" />

Then click on continue, and it fails:
<img width="1630" height="422" alt="Image" src="https://github.com/user-attachments/assets/5c76fe0c-4ac7-4f04-b2a6-9caf7ccc5c79" />

**Server / Test environment (please complete the following information):**
 - In deployed web server.
 - Linux
 - PHP 8.5
 - Apache version 2.4.65
 - Browser(s) tested with Microsoft Edge
 - Version 146.0.3856.109 (Official build) (64-bit)

**Additional context**
Not sure if I misunderstood something regarding about product variants, if so, please let know how to use this feature properly.

Thanks!

mhcwebdesign:
Now fixed.

epr-ds:
> Now fixed.

Awesome!
How can I find the fix?

mhcwebdesign:
The fix is here: https://github.com/opencart/opencart/pull/15415

epr-ds:
@mhcwebdesign 
> The fix is here: [#15415](https://github.com/opencart/opencart/pull/15415)

I pulled the fix into my local and prod instances, but it is still not working as expected. Please check below images:

#### Registered products
<img width="769" height="159" alt="image" src="https://github.com/user-attachments/assets/4687876a-9cef-428f-a78c-2241d26ef763" />

[...]

PR #15415, opened 2026-05-05
[4.x.x.x] Fix add variant to cart adding master instead

tongo-angelov:
### Fixed
- <s>[#15325](https://github.com/opencart/opencart/issues/15325) - Adding product variant to order </s>

Use variant's product_id instead of master's

The issue is fixed in master branch [commit](https://github.com/opencart/opencart/commit/d68ef25d40a8603a9241c9b90745e85f965aae30)

Update: **This commit does not fix mentioned issue, but rather catalog's checkout issue**

### DIFF
--- a/upload/catalog/controller/checkout/cart.php
+++ b/upload/catalog/controller/checkout/cart.php
@@ -273,7 +273,7 @@ class Cart extends \Opencart\System\Engine\Controller {
 			}
 
 			// Validate subscription products
-			$subscriptions = $this->model_catalog_product->getSubscriptions($product_id);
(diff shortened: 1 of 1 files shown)


epr-ds:
I pulled this change into my local and prod instances, but it is still not working as expected. Please check below images:

#### Registered products
image

#### Adding new order
image

mhcwebdesign:
> I pulled this change into my local and prod instances, but it is still not working as expected. Please check below images:

All I can see from your screenshots is that you added a variant product to your shopping cart. What is the exact issue?

PR #15416, opened 2026-05-05
[4.x.x.x] More fixes of bugs and security issues from  @plemondev

mhcwebdesign:
Based on this pull request https://github.com/opencart/opencart/pull/15409 , but with further improvements especially for PHP 8.5

### DIFF
--- a/upload/system/framework.php
+++ b/upload/system/framework.php
@@ -171,7 +171,7 @@ if ($config->get('session_autostart')) {
 		'path'     => $config->get('session_path'),
 		'domain'   => $config->get('session_domain'),
 		'secure'   => $request->server['HTTPS'],
-		'httponly' => false,
+		'httponly' => true,
 		'SameSite' => $config->get('session_samesite')
 	];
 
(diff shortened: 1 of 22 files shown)


ADDCreative:
@mhcwebdesign As stated at https://github.com/opencart/opencart/pull/15409#issuecomment-4361780632. Adding stream_context_set_option to upload/system/library/mail/smtp.php is pointless. It's just setting the default options. Also stream_socket_enable_crypto is not going to use the options anyway as it's after fsockopen. Even if it did have an effect, it wouldn't make sense to set for the Explicit TLS and not Implicit TLS.

The changes to the clean in upload/system/library/request.php are also wrong. Changing ENT_COMPAT to ENT_QUOTES is going to cause mismatch with in data previously written to the database and new queries. If a word had an apostrophe in originally it would be encoded as ```'```, but now it would be in future queries ```&#039;```. It would make the original word impossible to find in a search.

mhcwebdesign:
@ADDCreative : The Query clean is changed back to using ENT_COMPAT . As regards the system/library/mail/smtp.php : When I tested it, it worked fine. But yes, you have a valid point here. Could you do a pull-request for needed changes to the smtp.php please?

Issue #15419, opened 2026-05-05
[4.x.x.x] Admin product list has wrong discount value


tongo-angelov:
**What version of OpenCart are you reporting this for?**
4.x.x.x

**Describe the bug**
If product has any active discount the "price" variable of the discount is shown under products' price in admin > catalog > products list, without checking the discount type first

**To Reproduce**
1. Go to admin panel
2. Click on Catalog then Products
3. The default `Apple Cinema 30"` product shows price $100 and discount $88, even tho the discount is set to 88%

**Expected behavior**
Discount price should be $12, on catalog page it is show correctly

**Screenshots / Screen recordings**
<img width="1170" height="236" alt="Image" src="https://github.com/user-attachments/assets/2649f98a-850f-4bab-bbfe-e2bcf6295581" />
<img width="1567" height="576" alt="Image" src="https://github.com/user-attachments/assets/a5d83f2e-2a97-40db-8de9-e1a6d2ee1c17" />
<img width="393" height="353" alt="Image" src="https://github.com/user-attachments/assets/7efbc7eb-a47e-4639-9e93-5c662e9513b9" />
(disabled tax and set 1 quantity to compare with catalog page)

**Server / Test environment:**
 - Ubuntu Server
 - PHP v8.3
 - Apache/2.4.58

**Additional context**
In catalog product model OC has $this->statement['discount'] that checks for type, but in admin OC only checks if any discount is active

tongo-angelov:
Fixed in [#15421
](https://github.com/opencart/opencart/pull/15421)

PR #15420, opened 2026-05-05
[4.x.x.x] Fix admin product list wrong discount value

tongo-angelov:
### Fixed
- [#15419](https://github.com/opencart/opencart/issues/15419) - [4.x.x.x] Admin product list has wrong discount value

Check discount type and calculate final price


mhcwebdesign:
Now merged, see https://github.com/opencart/opencart/pull/15421, with adjusted PHP CS fixer applied to it, too.

PR #15428, opened 2026-05-06
[4.x.x.x] Fix multiple issues in checkout's register component

tongo-angelov:
### Fixed
- [#15430](https://github.com/opencart/opencart/issues/15430) - Multiple issues in checkout's register component

1.Checkout index does not check if postcode is required for given country, and only on country update does UI show correctly. Added check before showing page, for better user experience.

**Before**
image

2.Fix custom fields not retaining data on guest checkout after save and refresh, as well as not saving custom field values in either account or address on register checkout:

In catalog/controller/checkout/register OC builds component with both custom fields for address and account, so it nests them under field type, after save in db customers' custom_field we have:
```
{"account":{"29":["48"]}}
```
instead of correct
```
{"29":["48"]}
```

3.Removed label for file custom field, since input is hidden and browser shows warnings
4. Fixed shipping radio custom field using "in" instead of "==" for single value check
5. Moved shipping button custom field's error outside of "div" like other custom fields
6. Fixed payment custom fields using incorrect "address_custom_field" instead of "payment_custom_field"

### DIFF
--- a/upload/catalog/controller/checkout/register.php
+++ b/upload/catalog/controller/checkout/register.php
@@ -100,6 +100,10 @@ class Register extends \Opencart\System\Engine\Controller {
 
 		$data['countries'] = $this->model_localisation_country->getCountries();
 
+		// Postcode
+		$payment_country = $this->model_localisation_country->getCountry($data['payment_country_id']);
+		$data['payment_postcode_required'] = $payment_country['postcode_required'] ?? false;
+
 		// Zones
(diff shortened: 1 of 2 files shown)


tongo-angelov:
7. address_match added in session for guest checkout. Before the value was not kept, so if user clicks addresses do not match and adds payment and shipping addresses, on refresh value defaults to 1 (match) and shipping is lost

PR #15429, opened 2026-05-06
Resolves #15143 "Argument #1 ($product_id) must be of type int, null …

### LINKED ISSUE #15143: Admin Product Filter by Model Fails with PHP Error: product_id is NULL
**What version of OpenCart are you reporting this for?**
4.1.0.3

**Describe the bug**
When attempting to filter products by 'Model' in the admin product list (`Admin -> Catalog -> Products`), the filtering process fails, and the product list either remains unchanged or does not load correctly. This is due to a fatal PHP error occurring server-side: `Opencart\Admin\Model\Catalog\Product::getOptions(): Argument #1 ($product_id) must be of type int, null given`.

This error indicates that the `getProducts` method is returning product data where the `product_id` is unexpectedly `NULL` for some records. This likely happens when a `LEFT JOIN` on the `oc_product_code` table (which is conditionally included when filtering by model) causes the `product_id` from `oc_product` to be overwritten by a `NULL` value from `oc_product_code` for products that don't have a matching entry in `oc_product_code`. Subsequent calls (like `getOptions`) then receive `NULL` instead of an integer `product_id`, leading to the PHP TypeError and breaking the product list display.

**What section does it affect?**
Admin -> Catalog -> Products

**To Reproduce**
1.  **Preparation:**
    *   Ensure you have at least two products in your catalog.
    *   For one product, ensure it *does not* have any corresponding entry in the `oc_product_code` table. Note its `product.model`.
    *   For another product, ensure it *does* have an entry in the `oc_product_code` table, and its `product_code.value` might be different from its `product.model`.
2.  Navigate to `Admin -> Catalog -> Products`.
3.  In the filter options at the top, enter a value into the "Model" field that matches the `product.model` of the product that *does not* have an `oc_product_code` entry (from step 1).
4.  Click the blue filter button.
5.  **See error:** The product list will either not update, or it will display the original unfiltered list. Upon checking the server error logs or your browser's developer console (Network tab, then look at the response for the AJAX call to load products), you will find the PHP error: `Opencart\Admin\Model\Catalog\Product::getOptions(): Argument #1 ($product_id) must be of type int, null given`. This error prevents the filtered results from being properly rendered in the admin interface.

**Expected behavior**
When filtering products by 'Model', the product list should update correctly, displaying only the matching products. No PHP Type Errors should occur, and all products should have their valid `product_id` present in the retrieved data, allowing all related functions (like `getOptions`) to operate without issue.

**Screenshots / Screen recordings**
[View Screenshot](https://image.opencart.com/original/68dc116d6f41f.jpg)

**Server / Test environment (please complete the following information):**
-   **Local development? Deployed to a web server?** Local development (XAMPP)
-   **Operating system** Windows 10
-   **PHP version** 8.1
-   **Browser(s) tested with** Google Chrome
-   **Version** 140.0.7339.128 (Official Build) (64-bit)

**Additional context**
The root cause is almost certainly the SQL query in `admin/model/catalog/product.php` (specifically in the `getProducts` method) utilizing `SELECT *` in conjunction with a `LEFT JOIN` to `oc_product_code`. When a product in `oc_product` does not have a corresponding entry in `oc_product_code`, the `LEFT JOIN` yields `NULL` values for all `oc_product_code` columns. If `oc_product_code` also has a `product_id` column (even if it's not explicitly used in the `LEFT JOIN ON` clause), `SELECT *` can cause the `NULL` `product_id` from the joined (and `NULL`-filled) `oc_product_code` row to overwrite the valid `product_id` from `oc_product` in the final result set. This leads to `NULL` `product_id`s being passed to subsequent functions that expect an integer, resulting in the TypeError.

A robust fix requires replacing `SELECT *` with an explicit column list (e.g., `SELECT p.product_id, p.model, ..., pc.value AS product_code_value`) and ensuring `GROUP BY p.product_id` is used to prevent duplicate rows and ensure `p.product_id` is always selected unambiguously.

ryanhobrock:
…given" error

### DIFF
--- a/upload/admin/model/catalog/product.php
+++ b/upload/admin/model/catalog/product.php
@@ -1025,7 +1025,13 @@ class Product extends \Opencart\System\Engine\Model {
 	 * $results = $this->model_catalog_product->getProducts($filter_data);
 	 */
 	public function getProducts(array $data = []): array {
-		$sql = "SELECT * FROM `" . DB_PREFIX . "product` `p` LEFT JOIN `" . DB_PREFIX . "product_description` `pd` ON (`p`.`product_id` = `pd`.`product_id`)";
+		$sql = "SELECT `p`.*, `pd`.*";
+
+		if (!empty($data['filter_model'])) {
+			$sql .= ", `pc`.`product_code_id`, `pc`.`code`, `pc`.`value`";
+		}
+
+		$sql .= " FROM `" . DB_PREFIX . "product` `p` LEFT JOIN `" . DB_PREFIX . "product_description` `pd` ON (`p`.`product_id` = `pd`.`product_id`)";
 
 		if (!empty($data['filter_model'])) {
 			$sql .= " LEFT JOIN `" . DB_PREFIX . "product_code` `pc` ON (`p`.`product_id` = `pc`.`product_id`)";
--- a/upload/system/helper/db_schema.php
+++ b/upload/system/helper/db_schema.php
@@ -5328,7 +5328,13 @@ function oc_db_schema() {
 				'key'  => [
 					'code'
 				]
-			]
+			],
+			[
+				'name' => 'product_id',
+				'key'  => [
+					'product_id'
+				]
+			]		
 		],
 		'engine'  => 'InnoDB',
 		'charset' => 'utf8mb4',


batumibiz:
**I have a question**: why was `product_code_id` entered into the query, since it is not used anywhere?
A full search of the entire source code turns up nothing.

ryanhobrock:
To match the previous return.

PR #15431, opened 2026-05-09
[4.x.x.x] Fix autocomplete in Safari

stalker780:
Another attempt to fix Safari autocomplete

https://github.com/opencart/opencart/pull/14971

### DIFF
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -363,7 +363,7 @@ $(document).on('click', '[data-oc-toggle=\'image\']', function(e) {
                 if (!e.relatedTarget || !$(e.relatedTarget).hasClass('dropdown-item')) {
                     this.timer = setTimeout(function(object) {
                         object.removeClass('show');
-                    }, 50, $dropdown);
+                    }, 150, $dropdown);
                 }
             });
 
@@ -373,7 +373,7 @@ $(document).on('click', '[data-oc-toggle=\'image\']', function(e) {
             });
 
             // Click
-            $dropdown.on('mousedown', 'a', function(e) {
+            $dropdown.on('click', 'a', function(e) {
                 e.preventDefault();
 
                 var value = $(this).attr('href');
--- a/upload/admin/view/javascript/common.js
+++ b/upload/admin/view/javascript/common.js
@@ -373,7 +373,7 @@ $(document).on('click', '[data-oc-toggle=\'image\']', function(e) {
             });
 
             // Click
-            $dropdown.on('click', 'a', function(e) {
+            $dropdown.on('mousedown', 'a', function(e) {
                 e.preventDefault();
 
                 var value = $(this).attr('href');
--- a/upload/catalog/view/javascript/common.js
+++ b/upload/catalog/view/javascript/common.js
@@ -282,7 +282,9 @@ var chain = new Chain();
             // Focus out
             $(element).on('focusout', function(e) {
                 if (!e.relatedTarget || !$(e.relatedTarget).hasClass('dropdown-item')) {
-                    $dropdown.removeClass('show');
+                    this.timer = setTimeout(function(object) {
+                        object.removeClass('show');
+                    }, 150, $dropdown);
                 }
             });
 
@@ -292,7 +294,7 @@ var chain = new Chain();
             });
 
             // Click
-            $dropdown.on('mousedown', 'a', function(e) {
+            $dropdown.on('click', 'a', function(e) {
                 e.preventDefault();
 
                 var value = $(this).attr('href');
@@ -309,6 +311,7 @@ var chain = new Chain();
                 clearTimeout(this.timer);
 
                 $('#autocomplete-loading').remove();
+                $dropdown.find('li').remove();
 
                 $dropdown.prepend('<li id="autocomplete-loading"><span class="dropdown-item text-center disabled"><i class="fa-solid fa-circle-notch fa-spin"></i></span></li>');
                 $dropdown.addClass('show');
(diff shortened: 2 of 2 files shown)


mhcwebdesign:
Safari's specific issue with event delegation is that it only bubbles click events to the document or parent elements if it recognizes the target as "clickable". If your $dropdown uses an element like a `<div>`, `<span>`, or `<li>` that doesn't naturally receive clicks, Safari (especially on iOS) may simply ignore the interaction. See e.g. https://stackoverflow.com/questions/9881509/jquery-on-doesnt-work-with-mobile-safari

Can you try this workaround please:

In the CSS files **admin/view/stylesheet/stylesheet.css** and **catalog/view/stylesheet/stylesheet.css** add this:

```
#dropdown-selector a {
  cursor: pointer;
}
```

This should be a common fix for both desktop and mobile as it tells Safari to treat the element as interactive.

I don't have access to Apple devices, nor a running Safari browser. But if you could test it please with above adjusted CSS, and then let us know whether it works or not, we'd be grateful.

stalker780:
@mhcwebdesign 
I have tried it back in 2025.

Tested right now:
```
.dropdown-menu a.dropdown-item {
  cursor: pointer;
}
```

Does not help.

Safari requires clickable elements not only to have `cursor: pointer`, but also to be visible when the `click` event is assigned.
While autocomplete is a dynamic element which is created hidden. I suppose this is the source of issue.

`mousedown` have no such requirements in Safari.

stalker780:
@mhcwebdesign 
I do not have Apple devices either.

I run tests on Windows 11 using this method with Playwrite and WebKit.
https://www.youtube.com/watch?v=pRpsi1Z5YY0
https://joyofcode.xyz/test-your-site-in-every-browser

At least I can reproduce all reported Safari issues with it.
It is not ideal for design, but works quite well for debugging HTML/CSS/JS issues.

mhcwebdesign:
Right, that's odd. Can you also change the **common.js** to using this:

```
// Click for Desktop, Touchstart for Mobile
$dropdown.on('mousedown touchstart', 'a', function(e) {
    // Only prevent default if it's a touch to prevent double-firing 
    // or to stop the focusout race condition
    e.preventDefault(); 

    var value = $(this).attr('href');

    if (element.items[value] !== undefined) {
        element.select(element.items[value]);
        $dropdown.removeClass('show');
    }
});
```

And maybe adjust the CSS to this:
```
[id^="autocomplete-"] .dropdown-item {
    cursor: pointer !important;
}
```

stalker780:
@mhcwebdesign 

`[id^="autocomplete-"] .dropdown-item`
CSS does not help.

----
I have found another solution. We can leave click event in it's place.

The problem is in Safari timeouts. Need to replace 50ms with at least 150ms, perhaps 250ms is even better for slower Apple devices.

Here:
```
            // Focus out
            $(element).on('focusout', function(e) {
                if (!e.relatedTarget || !$(e.relatedTarget).hasClass('dropdown-item')) {
                    this.timer = setTimeout(function(object) {
                        object.removeClass('show');
                    }, 250, $dropdown);
                }
            });
```

So the source of issue is that in Safari input `focusout` event triggers before item `click` event.
Dropdown gets hidden and it does not have `click` event anymore so it is never fired.
By setting higher timeout delay we give Safari more time to trigger `click` event.

The only downside is that when autocomplete input loses focus it takes a little longer for dropdown to disappear - 250ms instead of 50ms. I think we can live with that.

I think this will be the best fix compatible with all browsers.

mhcwebdesign:
OK, assuming this event sequence for Safari:  mousedown → focusout (blur) → mouseup → click.

So I think you are on the right track. Could you try this for the focusout :

```
$(element).on('focusout', function(e) {
    // If the new focus is inside the dropdown, do nothing
    if (e.relatedTarget && $(e.relatedTarget).hasClass('dropdown-item')) {
        return;
    }
    
    this.timer = setTimeout(function(object) {
        object.removeClass('show');
    }, 150, $dropdown); // Increased timeout slightly for Safari stability
});
```
Also, maybe iOS Safari sometimes fails to bubble events from dynamically generated <a> tags unless they have an explicit onclick attribute.Hence, in the response function, we may add this to the string builder:

`html += '<li><a href="' + json[i]['value'] + '" class="dropdown-item" onclick="void(0)">' + json[i]['label'] + '</a></li>';`

stalker780:
@mhcwebdesign 
Your `focusout` script works, but only because of the 150ms timeout.
If I reduce it to 50ms, the original issue reappears.

----

`onclick="void(0)"` does not seem to make any difference either.

----

So I assume that 50ms is often not enough for Safari's event firing order to stabilize, while increasing the delay resolves the issue on most Apple devices.

I think we should stop at increasing the timeout from 50ms to 150ms for now. At least this should cover the majority of cases.

stalker780:
@mhcwebdesign 
I have pulled a new version of this fix. Where only timeout is modified.

Also I have copied admin autocomplete module to catalog common.js. As admin version is much better tested and seems to work flawlessly now.

trydalcoholic:
I tested the behavior in Safari: increasing the timeout does indeed help. The bug reproduces consistently and disappears with a 150ms delay.

According to the logs, the event sequence is:

```
focusin
mousedown
focusout (relatedTarget = null)
mouseup
click
```

So in Safari, `focusout` fires before `click`, and the dropdown gets hidden before the item selection is processed.

From what I can see, the timeout-based fix works fine. Though long term, a more explicit approach might be a bit more reliable, for example:

* handling selection on `pointerdown` / `mousedown`
* avoiding dropdown close in `focusout` if item selection has already started
* closing the list explicitly after selection

That approach would probably depend less on Safari timing quirks and event ordering.

stalker780:
@trydalcoholic 
IMHO we are digging too deep with it.
It takes a 1-2 years to fix one single line of code.
The problem is solved with 150ms modification.
It works in all browsers, including Safari.
If it happens again we will return to fix it. As far as I understand @trydalcoholic  you are testing it on Apple device.
There were no other bug reports.
Let us just commit it and go on with other tasks rather than rewriting autocomplete module.

PR #15444, opened 2026-05-15
Document MariaDB support in install and Docker docs

robertsilen:
The Docker dev stack uses the official MariaDB image; list MariaDB alongside MySQL in install and contributor documentation.


trydalcoholic (approved):
Thanks for the docs update

PR #15445, opened 2026-05-15
[4.x.x.x] Prevent XSS in error log view

ADDCreative:


batumibiz:
It seems to me that this improvement is not needed, XSS is not possible here.
Twig itself is great at escaping HTML entities.
To check, try to write some HTML or JS code to the /logs/error.log file and view it through the admin panel.

ADDCreative:
@batumibiz OpenCart doesn't use Twig's automatic escaping, so the data will need escaping for use in HTML. I have tested and it is needed. You can always private message me on the forum if you need more information.

batumibiz:
> @batumibiz OpenCart doesn't use Twig's automatic escaping, so the data will need escaping for use in HTML. I have tested and it is needed.

About Twig you are right, I explained the situation incorrectly.
Let me explain in more detail what I meant...

The information is output not as HTML code, but as the value of the textarea field, therefore presented as plain text. You can easily check: write the line `<a href='#'>TEST-1</a><br><a href='#'>TEST-2</a>` in error.log and look at the error log from the admin panel. The code will be displayed as plain text rather than as links.
Of course, maybe I'm wrong, but I can't understand how XSS is possible here?

ADDCreative:
@batumibiz It is possible. Please private message me on the forum if you need more information.

batumibiz:
> @batumibiz Please private message me on the forum if you need more information.

Unfortunately this is not possible now.

Screenshot 2026-05-18 182802

batumibiz:
I was able to reproduce the HTML code injection.
The question is closed, **your revision is really needed**.
Thank you!

PR #15459, opened 2026-05-21
entry_article_count in setting template

chongshengdz:


mhcwebdesign:
I suggest you add it as a 'module_topic_article_count' in the **extension/opencart/admin/view/template/module/topic.twig** and then use it in the **extension/opencart/admin/controller/module/topic.php**.  See how we did it in the 4.x.x.x branch at https://github.com/opencart/opencart/pull/15486

PR #15464, opened 2026-05-21
[4.x.x.x] Add manufacturer.info route to SQL installation file

### LINKED ISSUE #15461: [4.x.x.x] Manufacturer SEO URLs append ?route=product/manufacturer.info even with SEO URLs enabled
## Bug Description

When SEO URLs are enabled in OpenCart 4.x, manufacturer pages generate URLs like:

```text
/en-gb/apple?route=product/manufacturer.info
```

instead of a fully rewritten SEO URL.

This behaviour is also present on the official OpenCart demo.

Products, categories, and information pages generate clean URLs correctly, but manufacturer pages retain the `route` query string.

---

## Cause

The SEO URL table contains a route entry for:

```text
product/manufacturer
```

but not for:

```text
product/manufacturer.info
```

The frontend manufacturer links use:

```text
route=product/manufacturer.info
```

Because no SEO route entry exists for this route, the SEO rewrite logic cannot remove the `route` parameter.

---

## Reproduction

1. Install OpenCart 4.x
2. Enable SEO URLs
3. Visit a manufacturer page (example: Apple)
4. Observe generated URL

Result:

```text
/en-gb/apple?route=product/manufacturer.info
```

Expected:

A fully rewritten SEO URL without the `route` query parameter.

Example:

```text
/en-gb/brand/apple
```

or similar.

---

## Additional Notes

Adding a SEO URL entry for:

```text
route = product/manufacturer.info
```

resolves the issue.

The SEO rewrite system already supports this correctly once the route entry exists.

Image

ddac-uk:
This adds the missing SEO URL route entries for product/manufacturer.info.

Individual manufacturer pages currently generate URLs such as:

/apple?route=product/manufacturer.info

because the installer includes an SEO URL route for product/manufacturer, but not for product/manufacturer.info.

Adding the missing route allows the existing SEO rewrite logic to generate clean manufacturer URLs such as:

/brand/apple

Fixes #15461

### DIFF
--- a/upload/install/opencart-en-gb.sql
+++ b/upload/install/opencart-en-gb.sql
@@ -2560,7 +2560,8 @@ VALUES (0, 1, 'product_id', '47', 'hp-lp3065', 1),
        (0, 1, 'route', 'information/information', 'information', -1),
        (0, 1, 'route', 'product/product', 'product', -1),
        (0, 1, 'route', 'product/category', 'catalog', -1),
-       (0, 1, 'route', 'product/manufacturer', 'brands', -1);
+       (0, 1, 'route', 'product/manufacturer', 'brands', -1),
+	   (0, 1, 'route', 'product/manufacturer.info', 'brand', -2);
 
 -----------------------------------------------------------
 


ddac-uk:
I have also tested this by applying the SQL change locally and running a clean install.

After enabling SEO URLs, the manufacturer page now generates a clean URL without the `?route=product/manufacturer.info` query string.

The brands list page still works as expected.

mhcwebdesign:
Thank you for pointing it out. I have now merged it, along with the same change for the French language, too, see https://github.com/opencart/opencart/pull/15474

PR #15465, opened 2026-05-22
Validate installer archive paths

XananasX7:
Summary
  Tighten OpenCart installer ZIP handling so archive paths cannot write outside the intended install roots.

  The marketplace extension installer and core upgrade installer manually construct output paths from ZIP entry names. Before this change, an archive entry containing traversal segments such as `../` could
  escape the expected extension, image, storage, or install directory.

  Changes
  - Add a reusable relative path validator for archive member paths.
  - Reject empty paths, NUL bytes, absolute paths, Windows drive paths, stream-wrapper style paths, repeated separators, `.`, and `..` path segments.
  - Apply the check to marketplace extension installation.
  - Apply the check to core upgrade installer extraction.
  - Add a standalone regression test for valid paths and traversal rejection.

  Validation
  - `php -l upload/system/helper/general.php`
  - `php -l upload/admin/controller/marketplace/installer.php`
  - `php -l upload/admin/controller/tool/upgrade.php`
  - `php -l upload/admin/language/en-gb/marketplace/installer.php`
  - `php -l upload/admin/language/fr-fr/marketplace/installer.php`
  - `php -l upload/admin/language/en-gb/tool/upgrade.php`
  - `php -l upload/admin/language/fr-fr/tool/upgrade.php`
  - `php tests/security/installer_path_test.php`
  - `php tools/phpstan.phar analyze --no-progress --debug upload/admin/controller/marketplace/installer.php upload/admin/controller/tool/upgrade.php upload/system/helper/general.php`
  - `git diff --check`

### DIFF
--- a/tests/security/installer_path_test.php
+++ b/tests/security/installer_path_test.php
@@ -1,0 +1,55 @@
+<?php
+require_once __DIR__ . '/../../upload/system/helper/general.php';
+
+function assert_same(mixed $expected, mixed $actual, string $message): void {
+	if ($expected !== $actual) {
+		fwrite(STDERR, $message . PHP_EOL);
+		exit(1);
+	}
+}
+
+$valid = [
+	'admin/controller/extension/demo/module/demo.php',
+	'image/catalog/demo/logo.png',
+	'system/storage/vendor/package/composer.json',
+	'install/controller/upgrade.php',
+];
+
+foreach ($valid as $path) {
+	assert_same(true, oc_validate_relative_path($path), 'Expected path to be allowed: ' . $path);
+}
+
+$invalid = [
+	'../config.php',
+	'demo/../../config.php',
+	'/tmp/config.php',
+	'C:\\tmp\\config.php',
+	'php://filter/resource=config.php',
+	'image//catalog/logo.png',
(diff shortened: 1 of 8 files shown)


XananasX7:
Summary for reviewers:

This PR adds path traversal protection to the extension installer (`admin/controller/marketplace/installer.php`).

**The vulnerability:** When extracting a ZIP extension archive, the installer builds a destination path from the archive entry's filename (`$source`) without validating it. A malicious ZIP can contain entries like `../../config.php` or `/etc/passwd` as filenames, causing writes outside the extensions directory.

**The fix:** Calls `oc_validate_relative_path($destination)` (added to `system/helper/general.php`) before using the archive entry path. This helper:
1. Rejects empty paths, null bytes, and stream wrappers (`://`)
2. Rejects absolute paths and Windows drive paths
3. Rejects `..` components in any path segment
4. Rejects double slashes and dot-segment components

A test script is included at `tests/security/installer_path_test.php` covering both valid and invalid paths.

**Attack scenario:** A malicious extension ZIP uploaded by an admin (or obtained via a compromised marketplace) could overwrite `config.php` or other sensitive files outside `DIR_EXTENSION`.

PR #15473, opened 2026-05-24
[4.x.x.x] Fix escaping in backup

ADDCreative:


mhcwebdesign:
Before merging, one note:

Order matters inside a single str_replace array. Therefore we should move the backslash `\\` replacement to the very beginning of the arrays. If you do not, PHP will replace a newline with `\n`, and then later in the same pass, it might turn that backslash into `\\n`.

So how about this one:

```
$value = str_replace(
    ['\\', "\x00", "\n", "\r", "\x1a", '\'', '"'], 
    ['\\\\', '\0', '\n', '\r', '\Z', '\\\'', '\"'], 
    $value
);
```
?

ADDCreative:
@mhcwebdesign The backslash \\ replacement is at the beginning. Your suggestion is the same as what's in the pull request.

mhcwebdesign:
> @mhcwebdesign The backslash \ replacement is at the beginning. Your suggestion is the same as what's in the pull request.

Sorry, yes, you are right, must be the heatwave here with my lack of concentration :)

PR #15493, opened 2026-05-29
Update Lint.yml checkout stage

AJenbo:
Should fix: https://github.com/opencart/opencart/pull/15332#issuecomment-4525978996

PR #15494, opened 2026-05-30
[4.x.x.x] Fixed and optimized the GitHub Actions workflow configuration

batumibiz:
In continuation of the discussion: https://github.com/opencart/opencart/pull/15493

Fixed and optimized the GitHub Actions workflow configuration.

Key Changes:
- **Updated** Actions: Upgraded to the latest actions/checkout@v6.
- **Removed Redundant Caching**: Completely removed the cache configuration. Since all OpenCart dependencies are already bundled within the repository, there is nothing to download or cache. The cache step was redundant and has been safely removed.
- **Streamlined Syntax Checks**: Removed the duplicated PHP syntax check. This task is already handled perfectly by `php-cs-fixer` and `PHPStan`, making the extra check obsolete.
- **Workflow Renaming for Badges:** Renamed the workflow file so that the `4.x.x.x` branch can have its own independent status Badge in the README. Otherwise, due to existing issues in the `master` branch, the badge would constantly remain red.


.github/workflows/Lint.yml
```diff
@@ -1,58 +0,0 @@
...
-            echo "No syntax errors were detected."
-          fi
-
-      - name: Cache dependencies
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cache/composer/files
-            ./.cache
-            ./.php-cs-fixer.cache
```
AJenbo, line 44: Caching was for the tools catch, Phpstan and php-cs-fixer only has to process changed files with caching, but without it they will have

mhcwebdesign:
@AJenbo : I am not an expert on Lint, so rely on your help for this. Where exactly would these lines
```
          path: |
            ~/.cache/composer/files
            ./.cache
            ./.php-cs-fixer.cache
```
need to be re-added into the '.github/workflows/Lint-4.yml' ?

batumibiz:
Composer caching is definitely not necessary.
Caching Tools will give a gain of about 10 seconds. There is not much PR per day and I don’t think the difference is critical.

But if necessary, I can add a cache only for Tools.

PR #15495, opened 2026-05-30
[4.x.x.x] Fix option values check and error messages

stalker780:
It is possible to add an option with no values now. This results with PHP or SQL error when adding/modifying a product.

Fix option values check and error messages:
1. Added option values check in controller
2. Added error div to options in product_form
3. Added error language strings
4. Added check in product model - ignore options without values (may be required for API calls without product controller checks)
5. Refactored bulk PHP `||` checks with `in_array()` for better code reading

### DIFF
--- a/upload/admin/controller/catalog/option.php
+++ b/upload/admin/controller/catalog/option.php
@@ -319,7 +319,7 @@ class Option extends \Opencart\System\Engine\Controller {
 			}
 		}
 
-		if (($post_info['type'] == 'select' || $post_info['type'] == 'radio' || $post_info['type'] == 'checkbox') && !isset($post_info['option_value'])) {
+		if ((in_array($post_info['type'], ['select', 'radio', 'checkbox', 'image'])) && !isset($post_info['option_value'])) {
 			$json['error']['warning'] = $this->language->get('error_type');
 		}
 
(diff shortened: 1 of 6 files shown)


stalker780:
As to option type `image`.
It was already in the code:
https://github.com/opencart/opencart/blob/5caa167bd431d6aeff3cab4752c9734270e53976/upload/admin/controller/catalog/product.php#L999
https://github.com/opencart/opencart/blob/5caa167bd431d6aeff3cab4752c9734270e53976/upload/admin/controller/catalog/option.php#L452

We all know this type of options is often used in extensions, but it is unfinished in current OC4.1 version.

Perhaps we need to make a decision either to remove this option type from the code or finish its implementation to make it usable without extensions in default OC installation.

mhcwebdesign:
The option type 'image' was last used in OpenCart 2.2.x.x versions and has since then been abandoned.
So for OpenCart 4.1.x.x versions, and the master branch, let's just remove the few occurrances of ` == 'image'` .

Of course it does make sense to check that, when a select option is used for a product, it should have at least one defined option value coming along with it, too.

stalker780:
@mhcwebdesign removed image option type

PR #15498, opened 2026-05-31
[SECURITY] Add allowed_classes => false to upgrade script unserialize() calls

XananasX7:
## Summary

Five database upgrade scripts read serialized data from legacy table columns before converting it to JSON format. All nine `unserialize()` calls are missing the `allowed_classes` option, leaving them vulnerable to PHP Object Injection (CWE-502).

## Problem

Bare `unserialize()` without `allowed_classes` can instantiate any PHP class loaded in memory when processing the input string. If an attacker can write a malicious serialized object to the database before the upgrade routine runs (via SQL injection, a compromised backup restore, or direct DB access), PHP magic methods (`__wakeup()`, `__destruct()`) can be triggered during deserialization, potentially leading to Remote Code Execution.

**Affected files and columns:**

| File | Column |
|------|--------|
| `upgrade_5.php` | `oc_setting.value` |
| `upgrade_7.php` | `oc_module.setting` |
| `upgrade_11.php` | `oc_customer.custom_field`, `oc_customer_activity.data`, `oc_address.custom_field` |
| `upgrade_12.php` | `oc_order.custom_field`, `oc_order.payment_custom_field`, `oc_order.shipping_custom_field` |
| `upgrade_17.php` | `oc_user_group.permission` |

## Fix

Pass `['allowed_classes' => false]` to each `unserialize()` call.

This is safe because all values in these columns were originally stored by OpenCart itself as serialized plain PHP arrays — no objects are ever stored. If a malformed row is encountered, `unserialize()` returns `false`, and `json_encode(false)` stores `'false'` in the column, which is a graceful failure.

## References

- https://www.php.net/manual/en/function.unserialize.php
- CWE-502: Deserialization of Untrusted Data
- OWASP: PHP Object Injection

### DIFF
--- a/upload/install/controller/upgrade/upgrade_11.php
+++ b/upload/install/controller/upgrade/upgrade_11.php
@@ -24,7 +24,7 @@ class Upgrade11 extends \Opencart\System\Engine\Controller {
 
 			foreach ($query->rows as $result) {
 				if (preg_match('/^(a:)/', $result['custom_field'])) {
-					$this->db->query("UPDATE `" . DB_PREFIX . "customer` SET `custom_field` = '" . $this->db->escape(json_encode(unserialize($result['custom_field']))) . "' WHERE `customer_id` = '" . (int)$result['customer_id'] . "'");
+					$this->db->query("UPDATE `" . DB_PREFIX . "customer` SET `custom_field` = '" . $this->db->escape(json_encode(unserialize($result['custom_field'], ['allowed_classes' => false]))) . "' WHERE `customer_id` = '" . (int)$result['customer_id'] . "'");
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


XananasX7:
Summary for reviewers:

This PR adds `['allowed_classes' => false]` to all `unserialize()` calls in the upgrade scripts (`upgrade_5.php`, `upgrade_11.php`, `upgrade_12.php`, `upgrade_17.php`).

**Why `allowed_classes => false` is safe:**
These upgrade scripts read legacy serialized data from the database and convert it to JSON. The serialized values are:
- `upgrade_5` — `setting` values: plain arrays/scalars
- `upgrade_11` — `custom_field` (customer/address): arrays
- `upgrade_12` — order custom fields: arrays
- `upgrade_17` — `user_group.permission`: arrays

All confirmed via the `preg_match('/^(a:)/', ...)` guard — only processes entries starting with `a:` (PHP array serialization prefix). No objects are stored in these fields.

**Risk mitigated:** A compromised database or SQL injection into these tables during upgrade could otherwise trigger PHP Object Injection through gadget chains. `allowed_classes => false` eliminates this risk with zero behavior change.

PR #15513, opened 2026-06-03
[4.x.x.x] Change product rating field type to float

stalker780:
Many modern stores and themes want to show product average rating as decimal 4.4, 4.9, 5.0, etc. 
But this functionality is limited by `int` field type in `oc_product`.`rating`. 
This PR allows to calculate and store decimal(3.2) ratings allowing value `4.99`, etc. 
While you still can round it to integer in your template.

### DIFF
--- a/upload/admin/model/catalog/product.php
+++ b/upload/admin/model/catalog/product.php
@@ -939,7 +939,7 @@ class Product extends \Opencart\System\Engine\Model {
 	 * Edit product rating record in the database.
 	 *
 	 * @param int $product_id primary key of the product record
-	 * @param int $rating
+	 * @param float $rating
 	 *
 	 * @return void
 	 *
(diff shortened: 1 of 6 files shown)


stalker780:
It might also make sense to remove the real-time review count calculation here:

https://github.com/opencart/opencart/blob/0be36d193a7d5aca7f442e7a4049bf2aefc0868b/upload/catalog/model/catalog/product.php#L42

The `reviews` count could be stored in the `oc_product` table and updated when a review is approved, just like the `rating` field.

Calculating review totals on every catalog request adds unnecessary `JOIN`s and `GROUP BY` operations, which can become expensive on stores with large numbers of reviews and product-heavy category pages.

While this is not a bug, it appears to be a database design issue that negatively affects catalog performance.

mhcwebdesign:
@stalker780 : While your suggestion makes sense, let's wait till after the planned release of 4.1.0.4 where backwards-compatibility to 4.1.0.3 is to be kept. 

After that, we can focus on changes like yours, or support for multi-lingual manufacturer descriptions, and some other structural changes, for a future 4.1.1.0 release.

Until then, let's keep this pull-request open.

stalker780:
@mhcwebdesign Exactly this PR does not break anything. The DB structure stays the same. Only rating field format is changed to float. So it is backward compatible with old int format.

As to my proposal on storing reviews count in a separate field, perhaps we should create a plan or a todo list for 4.1.1.0. This will help a lot OC developers and users to plan ahead.

mhcwebdesign:
> @mhcwebdesign Exactly this PR does not break anything. The DB structure stays the same. Only rating field format is changed to float. So it is backward compatible with old int format.
> 

OK, in that case it will be merged. Before doing so, can you clean it up, there are some warnings from PHPStan and PHP CS Fixer. Thanks.

> As to my proposal on storing reviews count in a separate field, perhaps we should create a plan or a todo list for 4.1.1.0. This will help a lot OC developers and users to plan ahead.

Maybe we can create a new discussion entry at https://github.com/opencart/opencart/discussions with a list of planned changes for a future 4.1.1.0 release.

stalker780:
@mhcwebdesign Lint looks fine now

PR #15527, opened 2026-06-09
[Master] reject traversal zip entries in marketplace installer

metsw24-max:
**Zip-slip in the ocmod extension installer**

Entry names from an uploaded `.ocmod.zip` get mapped onto a fixed base but are never checked for `..`, so a crafted archive can drop files outside the extension folder the extractor is meant to confine them to. Skipping entries that contain a traversal segment keeps the write within bounds and matches the containment the surrounding `image/` and `system/storage/` handling already assumes.


trydalcoholic (approved):
Looks good to me.

PR #15545, opened 2026-06-27
fix inverted .sql extension check in backup upload

nabhan06:
The backup upload handler is supposed to accept only .sql files, but the extension test in upload() is reversed: it sets the "Invalid file type" error when the filename ends with .sql and lets every other extension through, so a .php or .phar name passes validation and gets written into the storage backup directory. Negating the condition restores the intended allowlist so only .sql backups are accepted.


nabhan06:
any update?

PR #15548, opened 2026-07-01
[4.x.x.x] Help prevent long-running cron jobs from causing misses.

ryanhobrock:
Crons that took longer than 10 seconds could miss the next execution.


mhcwebdesign:
How exactly does your change solve the issue of long running cron jobs?

ryanhobrock:
It changes `date_modified` from when the job completed to when it started.

PR #15549, opened 2026-07-02
[Master] reject api requests outside the signature time window

nabhan06:
startup/api validates the request `time` against a +/-450 second window, but the two bounds are joined with && , so `$time < $time_start && $time > $time_end` can never be true and the check never rejects anything. Since `time` is folded into the HMAC-signed string only to bound replay, a captured signed api/order or api/subscription request stays valid forever instead of expiring. Switching to || rejects timestamps that fall outside the window.


nabhan06:
any update?

PR #15566, opened 2026-07-10
[4.x.x.x] Update of upgrade scripts

mhcwebdesign:
To take into account an upgrade from previous 4.1.0.x releases to 4.1.0.4

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,11 +1,11 @@
-# Upgrading from version 2.x.x
+# Upgrading from versions 4.1.0.0 - 4.1.0.3
 
 * For existing installs only
 * Don't use with previous versions
 
 ## For previous versions
 
-If you have __1.5.x__ and want to upgrade __to 3.0.x__, You may try to [contact our dedicated support](https://dedicated.opencart.com/)
+If you have an older version and want to upgrade __to 4.1.0.4__, You may try to [contact our dedicated support](https://dedicated.opencart.com/)
 
 
 ## Step by step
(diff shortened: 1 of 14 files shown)


ADDCreative:
@mhcwebdesign I don't believe you don't need to alter the tables and add the indexes in upgrade_12. That's already done in upgrade_3, which alters the the tables to match the db_schema.

Take the banner_image table link column. It will already be text as that is what is in the db_schema.
https://github.com/opencart/opencart/blob/4ba654b08d08f20930594c38e07935b6c2bff9df/upload/system/helper/db_schema.php#L475-L478

In upgrade_3 the db_schema is loaded.
https://github.com/opencart/opencart/blob/4ba654b08d08f20930594c38e07935b6c2bff9df/upload/install/controller/upgrade/upgrade_3.php#L22

And the table altered.
https://github.com/opencart/opencart/blob/4ba654b08d08f20930594c38e07935b6c2bff9df/upload/install/controller/upgrade/upgrade_3.php#L75

And the indexes added.
https://github.com/opencart/opencart/blob/4ba654b08d08f20930594c38e07935b6c2bff9df/upload/install/controller/upgrade/upgrade_3.php#L157

mhcwebdesign:
@AddCreative: In that case, could you provide an updated pull-request for the upgrade please and I'll test it again? Thanks.

ADDCreative:
@mhcwebdesign What issue were you trying to solve with the following?
https://github.com/opencart/opencart/blob/4ba654b08d08f20930594c38e07935b6c2bff9df/upload/install/controller/upgrade/upgrade_12.php#L22-L66

The only difference I see that already done is that ```mpn``` is not removed from the product table. That can be solved by adding it to the following list.
https://github.com/opencart/opencart/blob/4ba654b08d08f20930594c38e07935b6c2bff9df/upload/install/controller/upgrade/upgrade_10.php#L61-L67

mhcwebdesign:
@ADDCreative : OK, I'll adjust it, and do another upgrade test.

mhcwebdesign:
I have now simplified the **upgrade_12.php** as per your suggestions, and tested the upgrade from a 4.1.0.3 (it even had some 3rd party extensions!) to 4.1.0.4. The upgrade worked fine. 

See https://github.com/opencart/opencart/pull/15570

PR #15575, opened 2026-07-12
[4.x.x.x] Fix admin redirect and use of config_url

ADDCreative:
config_url is not available in admin, so security check always passed. Use HTTP_SERVER and HTTP_CATALOG constants instead of config_url. Also remove 'redirect' from required POST fields in the login controller.

### DIFF
--- a/upload/admin/controller/common/language.php
+++ b/upload/admin/controller/common/language.php
@@ -100,7 +100,7 @@ class Language extends \Opencart\System\Engine\Controller {
 
 			setcookie('language', $code, $option);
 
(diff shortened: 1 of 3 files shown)


mhcwebdesign:
Thanks for this, it's now merged. I am now thinking about asking Daniel to do a tagged release of 4.1.0.4 based on the 4.x.x.x.
From your perspective, are there any outstanding major issues to be addressed before doing so? Or can we go ahead with it?

ADDCreative:
@mhcwebdesign I'll flag some issues in 4.1.0.4 discussion.

PR #15577, opened 2026-07-13
[Master] check modify permission in installer upload

nabhan06:
marketplace/installer.php upload() writes the uploaded .ocmod.zip into storage and adds an extension install record, but unlike install(), vendor(), uninstall() and delete() it never checks the modify permission. The admin startup filter only enforces the access permission for a route, so a user granted access-only rights to marketplace/installer can still upload packages and register installs. This adds the same modify check the sibling methods already use.

### DIFF
--- a/upload/admin/controller/marketplace/installer.php
+++ b/upload/admin/controller/marketplace/installer.php
@@ -191,6 +191,10 @@ class Installer extends \Opencart\System\Engine\Controller {
 
 		$json = [];
 
+		if (!$this->user->hasPermission('modify', 'marketplace/installer')) {
+			$json['error'] = $this->language->get('error_permission');
+		}
+
 		// 1. Validate the file uploaded.
 		if (isset($this->request->files['file']['name'])) {
 			$filename = basename($this->request->files['file']['name']);


mhcwebdesign:
Could you test the same for the 4.x.x.x branch please, too, and then let us know whether it needs to be addressed there, too, or not? Thanks.

nabhan06:
Checked 4.x.x.x and it's affected the same way. upload() there (installer.php:222) still runs move_uploaded_file and addInstall with no permission check, while install(), xml(), uninstall() and delete() all guard on hasPermission('modify', 'marketplace/installer'). So an access-only user can upload packages and register installs there too. Same one-liner right after $json = []; closes it. Happy to send a separate PR against 4.x.x.x if you'd like.

mhcwebdesign:
I have now merged it into the 4.x.x.x branch see https://github.com/opencart/opencart/pull/15578, and it still needs to be done for the master branch. We have a lot of other changes and bugfixes already done to the 4.x.x.x branch which we will have to eventually apply to the master branch, too.

nabhan06:
Right, this PR is the master-side fix. Same one-liner as #15578, sitting right after $json = []; so the move_uploaded_file and addInstall sinks stay blocked for access-only users. It's rebased clean on master and mergeable whenever you're ready.

PR #15579, opened 2026-07-14
[Master] check return ownership before showing return history

nabhan06:
account/return.history checks the requested return_id by calling getReturns(), the customer-scoped list method whose first argument is a pagination offset, so ownership is never actually verified and getHistories()/getTotalHistories() then read the return_history rows for any return_id. A logged in customer who owns more returns than a given id can pull another customer's return status and comments through it. Swap in the owner-scoped getReturn() that info() and the order and subscription history controllers already use.


nabhan06:
any update?

PR #15580, opened 2026-07-15
Fix API auth replay/IP trust and login attempt bypass

M0nd0R:
## Summary

This change fixes two independent authentication control issues.

- reject stale signed storefront API requests by enforcing the intended plus-or-minus 450 second freshness window
- stop using client-supplied forwarding headers for security-sensitive API IP allowlist and API session checks
- preserve current forwarded-header behavior for non-security logging paths, while allowing security-sensitive callsites to require the direct peer address
- make customer login lockout checks aggregate all failed attempts for an email address so lockouts cannot be bypassed by spreading failures across multiple stored IP rows
- record customer login-attempt source IPs from the direct peer address in the failed-login path

### 1. Storefront API replay / IP trust weakness

The storefront API startup gate contained an impossible time-window comparison, so signed requests were never rejected as stale. The same gate also trusted client-controlled forwarding headers for API IP allowlist decisions, which let callers spoof an allowed address unless a trusted proxy stripped those headers first.

### 2. Customer login lockout bypass

Failed login attempts were stored per email-plus-IP, but lockout enforcement later read only by email and did not aggregate duplicate rows. A caller could therefore vary the apparent source IP and keep each stored row below the configured threshold. Accepting arbitrary forwarding headers made that bypass trivial from a single host.

## Impact

- storefront API signatures now expire as intended and can no longer be replayed indefinitely
- API IP allowlists and API token session checks now bind to the actual remote peer instead of spoofable request headers
- customer account lockouts now apply to the account as configured, even when failures originate from multiple stored IP rows

## Validation

- php -l upload/system/helper/general.php
- php -l upload/system/startup.php
- php -l upload/catalog/controller/startup/api.php
- php -l upload/catalog/model/setting/api.php
- php -l upload/catalog/model/account/customer.php
- php -l upload/admin/model/customer/customer.php

### DIFF
--- a/upload/admin/model/customer/customer.php
+++ b/upload/admin/model/customer/customer.php
@@ -1299,9 +1299,13 @@ class Customer extends \Opencart\System\Engine\Model {
 	 * $login_info = $this->model_customer_customer->getTotalLoginAttempts($email);
 	 */
 	public function getTotalLoginAttempts(string $email): array {
-		$query = $this->db->query("SELECT * FROM `" . DB_PREFIX . "customer_login` WHERE `email` = '" . $this->db->escape(oc_strtolower($email)) . "'");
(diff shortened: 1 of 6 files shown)


M0nd0R:
Closing.

PR #15590, opened 2026-07-17
[4.x.x.x] Improved exception and error handlers

mhcwebdesign:
To make them more compatible with PHP 8


ADDCreative:
Still has the issue that a Notice, Warning or Deprecated will exit() if error display is off. There should be no need to exit() with any of them.

https://github.com/opencart/opencart/blob/2640aa151605c3e5351483b5de5d6df75533be16/upload/system/framework.php#L72-L79

PR #15591, opened 2026-07-17
[4.x.x.x] Added payment instructions to order mail

mhcwebdesign:
When the payment method is Bank Transfer, then it needs the payment instructions in the order email sent to the customer.

### DIFF
--- a/upload/catalog/controller/mail/order.php
+++ b/upload/catalog/controller/mail/order.php
@@ -329,6 +329,13 @@ class Order extends \Opencart\System\Engine\Controller {
 			$data['totals'][] = ['text' => $this->currency->format($order_total['value'], $order_info['currency_code'], $order_info['currency_value'])] + $order_total;
 		}
 
+		// Payment Instructions
+		if (!empty($order_info['payment_method']['code']) && $order_info['payment_method']['code'] == 'bank_transfer.bank_transfer') {
+			if ($this->config->get('payment_bank_transfer_bank_' . $order_info['language_id'])) {
+				$data['payment_instructions'] = nl2br($this->config->get('payment_bank_transfer_bank_' . $order_info['language_id']));
+			}
+		}
+
 		// Setting
 		$this->load->model('setting/setting');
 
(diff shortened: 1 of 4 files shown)


ADDCreative:
This is the wrong way to do it. What about other payment methods, such as cheque and third-party payment extensions? You can't have a special case  for each payment that adds instructions.

The issue was this commit. https://github.com/opencart/opencart/commit/b35daf7d47e3c3d36e6387f23b432b0df5db6ac7 The comment passed into add is actually the payment instructions.

mhcwebdesign:
The order comment is correctly passed on to the order in the OpenCart database (see column oc_order.comment), And it is correctly passed on to the order confirmation email sent to the customer. The issue was with the extra field used by the bank transfer, which needed to be added to the order confirmation email.

The only other payment method I can see using an extra field is the Cheque payment, which uses 'payable' as an extra field. We may have add support for it, too. See attached screenshot.

Screenshot_2026-07-18_15-16-33

It may also make sense to change the template file **mail/order_add.twig** for this line:

old:
`<td style="font-size: 12px; border-right: 1px solid #DDDDDD; border-bottom: 1px solid #DDDDDD; background-color: #EFEFEF; font-weight: bold; text-align: left; padding: 7px; color: #222222;">{{ text_instruction }}</td>`

new:
`<td style="font-size: 12px; border-right: 1px solid #DDDDDD; border-bottom: 1px solid #DDDDDD; background-color: #EFEFEF; font-weight: bold; text-align: left; padding: 7px; color: #222222;">{{ text_comment }}</td>`

because the latter refers to the order comment, not the payment instruction.

mhcwebdesign:
Further to this, I could implement a more generic solution. 

We could have the payment methods in **extension/*/catalog/controller/payment/*.php** have an optional function named e.g. **getInstructions** included. The **catalog/controller/mail/order.php** , in its function add, would then check for the existence of the **getInstructions** function, and if there, use it to get the appropriate payment instructions included in the order email.

ADDCreative:
Try https://github.com/opencart/opencart/pull/15601.

mhcwebdesign:
Thanks, your pull request https://github.com/opencart/opencart/pull/15601 works fine.

Issue #12494, opened 2023-07-14
Opencart maintenance mode not showing the frontend when logon as admin


bpcheramie:
4.0.1.1

When you have the site set in maintenance mode and you are login as the admin and you try to view the front-end of the site it will only show you the maintenance mode screen when it should be showing the site.

1. Open admin panel and log on as admin
2. Go to system
3. Go to settings 
4. Center screen click on action to the right hand side of the screen.
5. click on the tab that has server on it.
6. turn on maintenance mode
7. click on the save button.
8. click the down arrow by your user name and select your store. 
9. it will open your website but it will only show you the maintenance page and it should be showing you the store withouy the maintenance page 

when you are logon as the admin of the store and the store is in maintenance mode you should not see the maintenance mode screen but the store.

mhcwebdesign:
OC 4.0.1.1 was never suitable for production site. Please try it again with 4.0.2.2, or the master branch.

ADDCreative:
I can confirm this affects 4.0.2.2 as well. It depends on whether the admin or catalog side creates the session cookie. See. https://forum.opencart.com/viewtopic.php?p=859249#p859203

mhcwebdesign:
@ADDCreative : I am not able to reproduce this issue in OC 4.0.2.2. Can you post the precise steps on ho0w to reproduce this error?

ADDCreative:
@mhcwebdesign This is how I could recreate it in 4.0.2.2. OpenCart was installed in a subdirectory on a domain, not local.

1. Use a totally clean browser with no cookies already set
2. Go directly to the admin URL such as `https://example.com/opencart4/admin/`
3. Login
4. Switch on Maintenance Mode if not already on
5. Click on Your Store in the Profile drop down menu. Which would take to to `https://example.com/opencart4/`

I then see the maintenance message, when I should see the store.

As I said in the forum post. If you go to in a totally clean browser to `https://example.com/opencart4/` first and then `https://example.com/opencart4/admin/` it works fine. It's because of how the path is set on the session cookie.
https://github.com/opencart/opencart/blob/f27a8e206ecb248e015a909b2fd29ea129029793/upload/system/config/default.php#L42

You end up with two session cookies /opencart4/ and /opencart4/admin/. If you go to the samin first the catalog will have a different ID as it can't read the  /opencart4/admin/ cookie.

danielkerr:
it works. just tested it.

mavelo-llc:
[...]

PR #15594, opened 2026-07-17
[4.x.x.x] Cookie changes

ADDCreative:
Set session_path to use '/' in default config, to avoid issues where admin and catalog need the same session, such as #12494.
Fix cookie name admin_authorize in user controller. Remove unused token variable retrieval in customer controller.
Set HttpOnly for all cookies. Improves security and some PCI scanners require it.
Remove hardcoded cookie paths and use configured session_path or default path.
Set SameSite to Lax for user preference cookies and session_samesite for session and security cookies.

There is still an issue in that the Secure is set based on the request. This means that if a site does not have the correct redirect from http to https, it's possible for cookies to be set without Secure. A way to prebent this would be to set Secure based on the configuration, with ```secure'   => str_starts_with(HTTP_SERVER, 'https://'),``` or ```'secure'   => str_starts_with($this->config->get('config_url'), 'https://'),``` depending on weather admin/system or catalog.


mhcwebdesign:
Now merged.

Question: On some files you are now using `'samesite' => 'Lax'` instead of the `'samesite' => $this->config->get('config_session_samesite')`, such as in the file **catalog/controller/common/cookie.php**. Is there a particular reason for this?

ADDCreative:
> Question: On some files you are now using `'samesite' => 'Lax'` instead of the `'samesite' => $this->config->get('config_session_samesite')`, such as in the file **catalog/controller/common/cookie.php**. Is there a particular reason for this?

These are user preference cookies. As the default SameSite is Stirct, it causes a few minor issues as the web browser won't send the cookie when being takin to the site via a link. If a user has already been on the site and accepted/rejected the cookie banner. The next visit from a link, the banner will appear again as the cookie can't be read, but will then disappear on the next page load as the coolie can then be read again.

Other minor issues are going to admin from a link the page can be in the wrong language initially. Marketing tracking cookie doesn't get set initially.

PR #15602, opened 2026-07-20
[4.x.x.x] Fixed order comment during checkout

mhcwebdesign:
Only allow entry into order comment after a successful selection of a payment method

### DIFF
--- a/upload/catalog/view/template/checkout/checkout.twig
+++ b/upload/catalog/view/template/checkout/checkout.twig
@@ -35,4 +35,11 @@
   </div>
   {{ column_right }}
 </div>
+
+<script type="text/javascript"><!--
+$(document).ready(function() {
+    $('#input-comment').val('').prop('disabled', true);
+});
+//--></script>
+
 {{ footer }}
--- a/upload/catalog/view/template/checkout/payment_method.twig
+++ b/upload/catalog/view/template/checkout/payment_method.twig
@@ -144,6 +144,8 @@ $(document).on('submit', '#form-payment-method', function(e) {
                     $('#input-payment-method').val($('input[name=\'payment_method\']:checked').parent().find('label').text());
                     $('#input-payment-code').val($('input[name=\'payment_method\']:checked').val());
 
+                    $('#input-comment').removeAttr('disabled');
+
                     $('#checkout-confirm').load('index.php?route=checkout/confirm.confirm&language={{ language }}');
                 }
             },


ADDCreative:
Still getting the "Order not found! Please refresh the page" message, if the comment is entered before terms & conditions have been accepted.

mhcwebdesign:
@ADDCreative : I am unable to reproduce the error. What exactly did you do? Guest checkout, Register and checkout, or checkout with existing account? With what cart items? The comment field should be disabled until the payment method has been chosen.

mhcwebdesign:
OK, try this adjusted script inside the `catalog/view/template/checkout/shipping_method.twig` , it clears and disables the comment field, because a new shipping method has to be selected, and because of that the subsequent payment method needs a new selection, too. The comment field only gets enabled after the payment selection. Enabling the comment field can't be triggered by the acceptance of terms (privacy policy) because the latter isn't always there.

```
    chain.attach(function() {
        return $.ajax({
            url: 'index.php?route=checkout/shipping_method.save&language={{ language }}',
            type: 'post',
            data: $('#form-shipping-method').serialize(),
            dataType: 'json',
            contentType: 'application/x-www-form-urlencoded',
            beforeSend: function() {
                $('#button-shipping-method').button('loading');
            },
            complete: function() {
                $('#button-shipping-method').button('reset');
                $('#input-comment').val('').prop('disabled', true);
            },
            success: function(json) {
                console.log(json);

                if (json['redirect']) {
                    location = json['redirect'];
                }

                if (json['error']) {
                    $('#alert').prepend('<div class="alert alert-danger alert-dismissible"><i class="fa-solid fa-circle-exclamation"></i> ' + json['error'] + ' <button type="button" class="btn-close" data-bs-dismiss="alert"></button></div>');
                }

                if (json['success']) {
                    $('#alert').prepend('<div class="alert alert-success alert-dismissible"><i class="fa-solid fa-circle-check"></i> ' + json['success'] + ' <button type="button" class="btn-close" data-bs-dismiss="alert"></button></div>');

                    $('#modal-shipping').modal('hide');

                    $('#input-shipping-method').val($('input[name=\'shipping_method\']:checked').parent().find('label').text());
                    $('#input-shipping-code').val($('input[name=\'shipping_method\']:checked').val());

                    $('#input-payment-method').val('');

                    $('#cart').load('index.php?route=common/cart.info&language={{ language }}');
                    $('#checkout-confirm').load('index.php?route=checkout/confirm.confirm&language={{ language }}');
                }
            },
            error: function(xhr, ajaxOptions, thrownError) {
                console.log(thrownError + "\r\n" + xhr.statusText + "\r\n" + xhr.responseText);
            }
        });
    });

```

ADDCreative:
Didn't help. By terms and conditions I mean ```I have read and agree to the Terms & Conditions``` bit just below the comment box. You need to make sure the "Checkout Terms" is set to something in the settings for it to show.

mhcwebdesign:
@ADDCreative : OK, got it. I had to enable the Terms and Conditions in the settings, and now I can reproduce the error. Will look into it.

mhcwebdesign:
I think this fix should do it: https://github.com/opencart/opencart/pull/15611

ADDCreative:
> I think this fix should do it: #15611

Thanks, that works for the terms and conditions. I think you need to remove the code that disabled the comment box now. As disabling the comment box causes seems to make the comment disappear if you do something like change the shipping method.

mhcwebdesign:
Yeah, disabling and clearing the comment box was there on purpose, but I have now removed it, see https://github.com/opencart/opencart/pull/15615 . It's now only cleared, but not disabled, upon entering the checkout page.

I just tested it with various checkout scenarios, it and seems to be working fine now.

PR #15605, opened 2026-07-21
[Master] match api routes case insensitively in the startup gate

nabhan06:
startup/api.php picks the signed-request path with substr($route, 0, 4) == 'api/' and in_array($route, $allowed), but route resolution is case insensitive, so index.php?route=API/order misses both tests and still loads catalog/controller/api/order.php with no username, IP, time window or signature check ever running, which leaves calls like history_add reachable by anyone. Lowercasing the route before the comparison puts the mixed case form back on the signed path, where it fails on the signature; the api prefix test in startup/session.php that attaches the api_token session had the same problem. 4.x.x.x carries the same code, happy to send a separate pull request for that branch.

### DIFF
--- a/upload/catalog/controller/startup/api.php
+++ b/upload/catalog/controller/startup/api.php
@@ -12,8 +12,9 @@ class Api extends \Opencart\System\Engine\Controller {
 	 * @return \Opencart\System\Engine\Action|null
 	 */
 	public function index() {
+		// Controller resolution is case insensitive, so normalise before matching
 		if (isset($this->request->get['route'])) {
-			$route = (string)$this->request->get['route'];
+			$route = strtolower((string)$this->request->get['route']);
 		} else {
 			$route = '';
 		}
--- a/upload/catalog/controller/startup/session.php
+++ b/upload/catalog/controller/startup/session.php
@@ -18,7 +18,7 @@ class Session extends \Opencart\System\Engine\Controller {
 		$this->registry->set('session', $session);
 
 		// Api
-		if (isset($this->request->get['route']) && substr((string)$this->request->get['route'], 0, 4) == 'api/' && isset($this->request->get['api_token'])) {
+		if (isset($this->request->get['route']) && substr(strtolower((string)$this->request->get['route']), 0, 4) == 'api/' && isset($this->request->get['api_token'])) {
 			$this->load->model('setting/api');
 
 			$this->model_setting_api->cleanSessions();


mhcwebdesign:
Please give us a reproducible example of how exactly it fails.

nabhan06:
Sure. The short version is that `startup/api.php` compares the raw route, but the route to controller resolution is case insensitive, so the two disagree on what "an api route" is.

Two separate effects, both reachable over plain HTTP:

**1. `index.php?route=API/cart` skips the "block direct access to other methods" branch.**
`substr('API/cart', 0, 4) == 'api/'` is false, so no `startup/api.permission`, but `Opencart\Catalog\Controller\API\Cart` still autoloads from `catalog/controller/api/cart.php` and runs.

**2. `index.php?route=API/order&call=history_add` skips the signature check.**
`in_array('API/order', $allowed)` is false, so the username lookup, IP allow list, time window and `hash_hmac` comparison never run at all, and `Api\Order::index()` dispatches `call=history_add` into `addHistory()` with whatever `order_id` / `order_status_id` you POST. No `api_token`, no signature, no credentials of any kind:

```
curl -s 'http://<host>/index.php?route=API/order&call=history_add' \
     -d 'order_id=1&order_status_id=5'
```

Here is a script that replays the gate against the real tree so you can see it without a full install. Drop it in the repo root and run `php repro.php` — it uses the actual `Engine\Autoloader` and the same class name derivation as `Engine\Factory::controller()`:

```php
<?php
define('DIR_APPLICATION', __DIR__ . '/upload/catalog/');

require DIR_APPLICATION . '../system/engine/autoloader.php';

$autoloader = new \Opencart\System\Engine\Autoloader();
$autoloader->register('Opencart\Catalog', DIR_APPLICATION);
$autoloader->register('Opencart\System', DIR_APPLICATION . '../system/');

$allowed = ['api/order', 'api/subscription'];

function gate(string $route, array $allowed, bool $patched): array {
	$route = $patched ? strtolower($route) : $route;

	return [
		'blocked'   => substr($route, 0, 4) == 'api/' && !in_array($route, $allowed),
		'signature' => in_array($route, $allowed),
	];
}

function resolves(string $route): string {
	// Engine\Factory::controller()
	$class = 'Opencart\Catalog\Controller\\' . str_replace(['_', '/'], ['', '\\'], ucwords(preg_replace('/[^a-zA-Z0-9_\/]/', '', $route), '_/'));

	return $class . ' => ' . var_export(class_exists($class), true);
}

foreach (['api/order', 'API/order', 'Api/Order', 'api/cart', 'API/cart'] as $route) {
	$before = gate($route, $allowed, false);
	$after  = gate($route, $allowed, true);

	printf(
		"route=%-10s | before: blocked=%-5s signature=%-5s | after: blocked=%-5s signature=%-5s | %s\n",
		$route,
		var_export($before['blocked'], true),
		var_export($before['signature'], true),
		var_export($after['blocked'], true),
		var_export($after['signature'], true),
		resolves($route)
	);
}
```

Output on this branch:

```
route=api/order  | before: blocked=false signature=true  | after: blocked=false signature=true  | Opencart\Catalog\Controller\Api\Order => true
route=API/order  | before: blocked=false signature=false | after: blocked=false signature=true  | Opencart\Catalog\Controller\API\Order => true
route=Api/Order  | before: blocked=false signature=false | after: blocked=false signature=true  | Opencart\Catalog\Controller\Api\Order => true
route=api/cart   | before: blocked=true  signature=false | after: blocked=true  signature=false | Opencart\Catalog\Controller\Api\Cart => true
route=API/cart   | before: blocked=false signature=false | after: blocked=true  signature=false | Opencart\Catalog\Controller\API\Cart => true
```

The `class_exists` column is the key part: every mixed case variant still loads the controller, so the route runs either way. Before the patch `API/order` is `blocked=false signature=false`, which means nothing checks it. After, it's back on the signed path and fails there, since the signature is computed over the raw route.

`startup/session.php` had the same prefix test, which is why it's in the diff.

mhcwebdesign:
Thank you for your explanations. I have now applied suggested changes to the 4.x.x.x branch, and hopefully Daniel will pick it up for the master branch, too.

PR #15608, opened 2026-07-22
[Master] reject traversal extension codes in installer upload

nabhan06:
installer upload() takes the extension code from the uploaded archive name with basename($filename, '.ocmod.zip'), and that returns '..' for a file named ...ocmod.zip, which passes both the length check and the .ocmod.zip check and gets stored by addInstall(). The code is later used as a directory component, so install() writes zip entries under DIR_EXTENSION . '../' (the existing `..` guard only looks at the entry names, not at the code prefix) and uninstall() hands DIR_EXTENSION . '../' to oc_directory_delete(), which walks the whole OpenCart root. Folded a '.'/'..' rejection into the ocmod file type check so the value never reaches the database. 4.x.x.x carries the same code, happy to open a separate PR for it.

### DIFF
--- a/upload/admin/controller/marketplace/installer.php
+++ b/upload/admin/controller/marketplace/installer.php
@@ -237,8 +237,8 @@ class Installer extends \Opencart\System\Engine\Controller {
 				$json['error'] = $this->language->get('error_filename');
 			}
 
-			// 3. Validate is ocmod file.
-			if (substr($filename, -10) != '.ocmod.zip') {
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
Thanks for pointing it out.

Yes, please also create a separate pull-request for 4.x.x.x .

nabhan06:
Done, opened #15609 for 4.x.x.x. Same derivation and same two sinks there (install() writing under DIR_EXTENSION . '../' and uninstall() handing that to oc_directory_delete()), so it's the same one-line change in upload().

PR #15612, opened 2026-07-22
{Master] Honor configured 2FA authorization expiry

carrerasdarren-cell:
## Summary

- use the configured 2FA lifetime for admin and customer trusted-device cookies
- persist that lifetime in `date_expire` instead of storing the creation time as the expiry
- reject expired authorization tokens and exclude expired records from attempt aggregation

## Why

The authorization tables and settings UI already expose a token expiry, but the setting was not applied. New records stored `date_expire = NOW()`, token lookups never checked that column, and both cookies were hard-coded to 90 days. A configured shorter lifetime therefore had no effect, and the server continued accepting the trusted-device token after that lifetime.

The minimum of one day also prevents a missing or zero setting from immediately invalidating a newly issued challenge.

## Validation

- PHP syntax checks pass for all four changed files
- focused model regression verifies both INSERT queries use the configured day interval
- focused model regression verifies both token lookups enforce expiry in the selected token and aggregate attempt count
- targeted PHP CS Fixer dry run reports no formatting changes in the modified sections
- `git diff --check` passes for the staged patch

### DIFF
--- a/upload/admin/controller/common/authorize.php
+++ b/upload/admin/controller/common/authorize.php
@@ -36,16 +36,18 @@ class Authorize extends \Opencart\System\Engine\Controller {
 		if (!$token_info) {
 			// Create a token that can be stored as a cookie and will be used to identify device is safe.
 			$token = oc_token(32);
+			$expire = max(1, (int)$this->config->get('config_2fa_expire'));
 
 			$authorize_data = [
 				'token'      => $token,
 				'ip'         => oc_get_ip(),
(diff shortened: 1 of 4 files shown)


mhcwebdesign:
Can you test the 4.x.x.x branch with it, and create a pull-request for it, too, please?

carrerasdarren-cell:
Done: I tested and ported this fix to `4.x.x.x` in #15619. That companion PR has now been merged as `5ae17a1b1831ca4ccd4000545e06edda81d41385`.

PR #15621, opened 2026-07-26
fix: correct primary key for product_tag and product_2_tag tables

WoompaLoompa:
Both `product_tag` and `product_2_tag` tables in `system/helper/db_schema.php` define `product_option_id` as the primary key, but that column doesn't exist in these tables. The correct primary key is `product_tag_id`, which is the actual auto_increment column.

This bug prevents OpenCart from being installed from scratch using the CLI installer or the web installer.


trydalcoholic:
Closing as superseded. The affected tables were removed and replaced with `product_to_tag` in the current schema. This PR also contains unrelated rendering changes that should be submitted separately and scoped to a specific issue.

PR #15625, opened 2026-07-27
[3.0.x.x] Improved mail exception handler

mhcwebdesign:
error handler that converts warnings into exceptions

### DIFF
--- a/upload/system/library/mail/mail.php
+++ b/upload/system/library/mail/mail.php
@@ -77,10 +77,26 @@ class Mail extends \stdClass {
 
 		ini_set('sendmail_from', $this->from);
 
-		if ($this->parameter) {
-			mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header, $this->parameter);
-		} else {
-			mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header);
+		// Define a temporary error handler that converts warnings into exceptions
+		set_error_handler(function ($severity, $message, $file, $line) {
+			if ($severity === E_WARNING) {
+				throw new \ErrorException($message, 0, $severity, $file, $line);
+			}
+			return false; // Let non-warnings pass to normal handling
+		});
+
+		try {
+			// Run existing mail calls
+			if ($this->parameter) {
+				mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header, $this->parameter);
+			} else {
+				mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header);
+			}
+		} catch (\ErrorException $e) {
+			// The PHP 8.5 warning is safely caught here! We ignore it.
+		} finally {
+			// Always restore the server's original error handler
+			restore_error_handler();
 		}
 	}
 }


ADDCreative:
@mhcwebdesign Out of interest, what issue are you trying to solve with this?

mhcwebdesign:
It's just to safely catch PHP Warnings and then ignore them during a mail send.

ADDCreative:
I can see it's ignoring error during mail send. Why you you want to do that and not allow them to be logged in the error log?

mhcwebdesign:
@ADDCreative : The transition was an intentional patch: 

Old code:

```
		if ($this->parameter) {
			mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header, $this->parameter);
		} else {
			mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header);
		}
```

New code:
```
		// Define a temporary error handler that converts warnings into exceptions
		set_error_handler(function ($severity, $message, $file, $line) {
			if ($severity === E_WARNING) {
				throw new \ErrorException($message, 0, $severity, $file, $line);
			}
			return false; // Let non-warnings pass to normal handling
		});

		try {
			// Run existing mail calls
			if ($this->parameter) {
				mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header, $this->parameter);
			} else {
				mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header);
			}
		} catch (\ErrorException $e) {
			// The PHP 8.5 warning is safely caught here! We ignore it.
		} finally {
			// Always restore the server's original error handler
			restore_error_handler();
		}
```

The old code was abandoned because it was unsafe for PHP 8.5. The new code was adopted to intercept the new PHP 8.5 warnings locally, acting as a buffer to keep the store online and functional even if the server's local mailer misbehaves.

In PHP 8.1 through 8.4, if the server's local sendmail background process crashed, ran out of memory, or failed to terminate correctly, the raw mail() function in the old code would swallow the error and return false silently, whereas PHP 8.5 explicitly surfaces these backend infrastructure failures by throwing a native E_WARNING.

See also https://www.php.net/manual/en/migration85.new-features.php

ADDCreative:
E_WARNING is non-fatal and execution of the script is not halted. There should be no need to supress it, even with the changes to PHP 8.5. No production server should ever have error display enable, so the warning would just be logged. I think if there was a problem with sendmail, you would wan to know about it.

Even if you did need to stop the sendmail error from being logged. There is no need to throw and catch an exception. Just return true in the error handler.

```php
if ($severity === E_WARNING) {
	return true; // Don't execute PHP internal error handler
}
return false; // Let non-warnings pass to normal handling
```

mhcwebdesign:
Well, we could just simplify the new code to something like this:

```
		$extraParam = !empty($this->parameter) ? (string)$this->parameter : '';

		$success = mail($to, $encodedSubject, $message, $headersArray, $extraParam);

		if (!$success) {
			if ($this->config->get('config_error_log')) {
				$error = error_get_last();
				$this->log->write("Mail delivery failed: " . ($error['message'] ?? 'Unknown transport error'));
			}
		}
```
In PHP 8.5: It catches actual operating system delivery errors.

In PHP 8.4: It only catches configuration errors (like a totally missing sendmail_path in php.ini). If the configuration is fine but the mail binary crashes, mail() returns false silently without setting error_get_last(). I think using above code handles this correctly via the ?? 'Unknown transport error' fallback.

What do you think ?

mhcwebdesign:
Update: Never mind, we can't even call $this->log->write() in the Mail class. I think we should just go for this one:

```
		$extraParam = !empty($this->parameter) ? (string)$this->parameter : '';

		mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header, $extraParam);

```

ADDCreative:
> Update: Never mind, we can't even call $this->log->write() in the Mail class. I think we should just go for this one:
> 
> ```
> 		$extraParam = !empty($this->parameter) ? (string)$this->parameter : '';
> 
> 		mail($to, '=?UTF-8?B?' . base64_encode($this->subject) . '?=', $message, $header, $extraParam);
> ```

That would be fine. The lack of error reporting in PHP 8.4 and lower is a PHP issue. It can be solved be using PHP 8.5.

mhcwebdesign:
> That would be fine. The lack of error reporting in PHP 8.4 and lower is a PHP issue. It can be solved be using PHP 8.5.

Agree, just tested it, and when using PHP 8.5 it now logs PHP Warnings caused by the sendmail. It's now updated, see https://github.com/opencart/opencart/pull/15638

PR #15647, opened 2026-08-07
Add product structured data

mihailovs2000:
## Summary

Adds server-rendered Schema.org JSON-LD to product pages using data already available in the product controller.

The graph includes:
- `Product` with images, brand, SKU/model/MPN and validated GTIN values
- `Offer` with currency, active price, availability and seller
- `AggregateRating` only when reviews are enabled and available
- `BreadcrumbList` based on the rendered breadcrumbs

Customer-group prices that are hidden until login are not exposed.

## Validation

- JSON output validated with products containing special prices, multiple images, GTIN, rating and out-of-stock/backorder states
- Script-closing characters are escaped and invalid UTF-8 is handled safely

### DIFF
--- a/upload/catalog/controller/product/product.php
+++ b/upload/catalog/controller/product/product.php
@@ -389,6 +389,8 @@ class ControllerProductProduct extends Controller {
 
 			$data['share'] = $this->url->link('product/product', 'product_id=' . (int)$this->request->get['product_id']);
 
+			$data['structured_data'] = $this->getStructuredData($product_info, $data);
+
 			$data['attribute_groups'] = $this->model_catalog_product->getProductAttributes($this->request->get['product_id']);
 
 			$data['products'] = array();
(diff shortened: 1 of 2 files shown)


mhcwebdesign:
I think is useful for an extension. There are a number of useful SEO extensions for structured data.

mihailovs2000:
Thanks for the feedback. I agree that extensions are the right place for advanced or merchant-specific SEO markup. However, I think a minimal, standards-compliant `Product` / `Offer` / `BreadcrumbList` graph is baseline storefront output for an e-commerce CMS and belongs in core.

Core has authoritative knowledge of the displayed price, customer-group visibility, stock state, canonical product URL, and rendered breadcrumbs. Reimplementing that baseline independently in extensions makes it easy for structured data to drift from what customers actually see, or for multiple extensions to emit conflicting graphs.

This PR intentionally keeps the scope small: it adds no dependency or configuration, emits only data already available on the product page, respects hidden customer-group prices, and omits invalid or unavailable fields. Extensions can still enrich the graph with merchant-specific features such as shipping details, return policies, additional review providers, or feed integrations.

If extensibility is the concern, I can make the core graph easy to disable or extend through the existing event/override mechanisms. But I think a fresh OpenCart store should provide correct basic product structured data by default.

PR #15649, opened 2026-08-07
Redirect legacy query URLs to SEO URLs

mihailovs2000:
## Summary

Redirects legacy GET URLs for products, categories, manufacturers and information pages to their configured SEO URLs when a rewrite is available.

For product pages, navigation-only parameters such as `path` and `manufacturer_id` are deliberately removed so all legacy variants converge on one product URL. Tracking is processed earlier in startup and is not copied into the canonical redirect.

If no SEO keyword exists, the original query URL continues to work and no redirect is issued.

## Validation

- Verified HTTP 301 behavior
- Verified that product URLs with `manufacturer_id`, `path`, or only `product_id` resolve to the same SEO URL
- Verified that missing SEO mappings do not cause redirect loops

### DIFF
--- a/upload/catalog/controller/startup/seo_url.php
+++ b/upload/catalog/controller/startup/seo_url.php
@@ -62,7 +62,7 @@ class ControllerStartupSeoUrl extends Controller {
 					$this->request->get['route'] = 'information/information';
 				}
 			}
-		} elseif ($this->request->server['REQUEST_METHOD'] == 'GET') {
+		} elseif (in_array($this->request->server['REQUEST_METHOD'], array('GET', 'HEAD'))) {
 			$this->redirectQueryUrl();
 		}
 	}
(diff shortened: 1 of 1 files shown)


mhcwebdesign:
What exact problem are you trying to resolve with this pull request? We already have link canonicals, e.g.

`<link href="https://www.mysite.com/mac" rel="canonical" />`

mihailovs2000:
The canonical tag provides a hint to search engines, but it does not normalize the request URL. The legacy product variants still return 200 and remain separate crawlable/linkable URLs, for example:

- `?route=product/product&product_id=201`
- `?route=product/product&product_id=201&path=61`
- `?route=product/product&product_id=201&manufacturer_id=16`

All three render the same product. This PR makes those variants converge with a 301 to the configured SEO URL. That also normalizes bookmarks, external links and cache entries, and consolidates link signals immediately instead of relying only on a crawler interpreting the canonical hint. The canonical link on the destination remains useful and is not replaced by this change.

The redirect is deliberately limited to supported catalog routes and only runs when a valid SEO rewrite exists; otherwise the original query URL continues to work.

PR #15651, opened 2026-08-08
Rename Google Sitemap to XML Sitemap and expose robots directive

mihailovs2000:
## Summary
- rename the visible Google Sitemap feed to the vendor-neutral XML Sitemap
- keep the existing controller route and configuration key for backward compatibility
- show a ready-to-copy `Sitemap:` directive for `robots.txt`

## Rationale
The XML Sitemap protocol is not Google-specific. A static `robots.txt` cannot include an absolute store URL in the distributed package, while the admin controller already knows `HTTP_CATALOG` and can present the exact directive without changing sitemap routing or output.

## Validation
- verified sitemap XML output remains unchanged
- verified the existing URL service continues to apply custom SEO rewrites


mhcwebdesign:
Well, we already have this directive in the '.htaccess' file:

`RewriteRule ^sitemap.xml$ index.php?route=extension/feed/google_sitemap [L]`

PR #15652, opened 2026-08-11
[4.x.x.x] Installer fix file size check for upload validation

stalker780:


stalker780:
And here we go. OC 4.0.1.4 is released with this bug.
Anyway after this release there will be a number of fixes for OC 4.0.1.5 :)

mhcwebdesign:
Well, users had many weeks to test it all out (see this dicsussion https://github.com/opencart/opencart/discussions/15426).

But this isn't the end of the road, OpenCart will continue to be refined. There might perhaps be a future 4.1.0.5, or we might include future bugfixes into a 4.1.1.0, there are other planned changes (e.g. multi-lingual manufacturer descriptions, etc, to bring it more into alignment with the master branch). 

Also, many of the recent bugfixes in the 4.x.x.x branch should eventually also be applied to the master branch.

The main goal for the 4.1.0.4 release was stability and backwards compatibility.

stalker780:
@mhcwebdesign 
I suppose we will get lots of new bug reports in the next several months, once people start using OC 4.1.0.4 and building extensions.
So I think 4.1.0.5 would be needed.

I plan to migrate to this version too. The only thing still stopping me is the new lists status design in admin area. 
Who ever decided this is convenient and informative to make hard-coded semi-transparent field rows in twigs? 🤦‍♂️

batumibiz:
> @stalker780
> Who ever decided this is convenient and informative to make hard-coded semi-transparent field rows in twigs?

The [original idea was to use a separate css class](https://github.com/opencart/opencart/pull/14518).
But Daniel decided to use the hard-wired transparency trick instead.

PR #15653, opened 2026-08-11
[4.x.x.x] Installer fix file size check for upload validation. Part2

stalker780:
Max upload file size must be taken both from `upload_max_filesize` and `post_max_size`. If any of these values is smaller than the uploaded file size, the user will not be able to upload it.


stalker780:
Also I think it should be clearly stated in the error message where these settings can be changed. 
This will reduce a great number of support requests from the customers not able to upload their extensions.

PR #15656, opened 2026-08-11
[4.x.x.x] Add end-to-end test suite and workflow using testRigor

AntonioVCruz:
## Summary

OpenCart has no automated end-to-end coverage wired into CI. This PR adds it:

- add 10 plain-English testRigor test cases and 10 shared reusable rules under `tests/e2e/`
- add `.github/workflows/testRigor_e2e_testing.yml`, which brings up OpenCart from this repository's own `docker-compose.yml` and runs the suite identified by `TESTRIGOR_SUITE_ID` against that container
- covers storefront and admin: navigation, catalogue search, required product options, cart, guest and register-at-checkout orders, the account lifecycle, admin access control, product create/edit/delete, and coupon redemption

Tests are written in plain English, so contributors can add regression coverage without writing code, and the CLI tunnels to the container so each run exercises the branch under test rather than a hosted demo store.

Store URLs and admin credentials are supplied per run. The admin defaults are the ones `docker-compose.yml` already installs, overridable with the optional `ADMIN_USERNAME` and `ADMIN_PASSWORD` secrets, and customer passwords are generated inside the tests rather than stored anywhere.

Nothing is added under `upload/`, so the distributed package is unchanged and no version increment applies. Setup, conventions and per-test coverage are documented in `tests/e2e/README.md`.

## Tests

- full suite green against 4.1.0.4: 10/10. The results are readable without a testRigor account: https://app.testrigor.com/public/tERoM8uHao9gHbZMN
- the job is a no-op unless the repository provides both `TESTRIGOR_SUITE_ID` (variable) and `TESTRIGOR_CI_TOKEN` (secret), so forks and pull requests from forks skip with a notice instead of a red check
- pull request runs execute the branch's test files without modifying the stored suite; only a push to `4.x.x.x` syncs the suite to match the repository
- latest successful workflow run: https://github.com/AntonioVCruz/opencart/actions/runs/31529569493

### First time: a testRigor account and suite

1. Sign up at [testRigor](https://www.testrigor.com/) and choose the **Public Open Source** plan.
2. Verify your email, log in, and create a test suite.
3. Add the four test data keys listed below in the suite's Test Data.

| Key | Purpose | Local value | CI value |
|---|---|---|---|
| `storeUrl` | Storefront base URL | `http://localhost` | the container just started |
| `adminUrl` | Admin panel URL | `http://localhost/admin` | the container just started, `/admin` |
| `adminUsername` | Admin account | `admin` | secret `ADMIN_USERNAME`, default `admin` |
| `adminPassword` | Admin password | `admin` | secret `ADMIN_PASSWORD`, default `admin` |

The suite ID is in its URL: for `https://app.testrigor.com/test-suites/oJvBxgjY5A8Pw48hj/test-cases` it is `oJvBxgjY5A8Pw48hj`.

### Running the suite

```bash
# Start OpenCart from this checkout (first boot runs the installer, ~2 minutes)
docker compose up -d --build opencart mysql
curl -fsS -o /dev/null -L http://localhost && echo "OpenCart is up"

# Run the suite. Without --explicit-mutations this uses your local files
# without overwriting the stored suite.
npx -y testrigor-cli test-suite run <SUITE_ID> \
  --token <CI_TOKEN> \
  --localhost \
  --url "http://localhost" \
  --test-cases-path "tests/e2e/testcases/**/*.txt" \
  --rules-path "tests/e2e/reusable-rules/**/*.txt" \
  --verbose

```

`CI_TOKEN` comes from **CI/CD Integration** in the suite sidebar. Add `--async` to start a run without waiting for it.

If the suite's test data points somewhere other than the container, override it for the run with a `--variables-path` file, as CI does. Include **all four** keys — the file replaces the variable set rather than merging into it:

```json
{
  "storeUrl": "http://localhost",
  "adminUrl": "http://localhost/admin",
  "adminUsername": "admin",
  "adminPassword": "admin"
}
```

The admin is at <http://localhost/admin> with the credentials in `docker-compose.yml`.
`upload/install.lock` is written by the container on first boot and is local state.

### Required

- Variable `TESTRIGOR_SUITE_ID`: the testRigor test suite identifier
- Secret `TESTRIGOR_CI_TOKEN`: the testRigor auth token, from the suite's CI/CD Integration
  page

Without both, the job skips with a notice; expected on forks and on pull requests from forks, which never receive secrets.

## Assistance

Claude Opus 5 assisted with consolidating the test cases, extracting the reusable rules and diagnosing failures. I reviewed and ran the changes and take responsibility for them.


batumibiz:
Interesting, but I have a question.
Can tests be run locally?
Before I create a PR, I have to check my work.

AntonioVCruz:
@batumibiz Yes, you can test your local changes using testRigor’s CLI. The tests themselves run on testRigor’s servers, but you can trigger them directly from your local environment before creating a PR. If you’d like, I can help you set it up.

mhcwebdesign:
Would it perhaps make sense to have this published as a separate OpenCart extension because it's mainly for developers if I understand it correctly?

batumibiz:
> Would it perhaps make sense to have this published as a separate OpenCart extension because it's mainly for developers if I understand it correctly?

This is a modern BDD using AI.
There is no way this will work as an extension; it has a completely different task.

BDD - This is testing the behavior of the application. That is, how a real living person would do it.
For example:
- Open admin page
- Login using the data we provided
- Next we describe the desired behavior. That is, what we manually checked every time is all done automatically, with one command.

The thing is very convenient if you write the tests correctly. Now, with AI, there shouldn’t be any problems with this, the main thing is to describe the tasks correctly.
But this is the main difficulty. Just going to the admin area and clicking through pages is of little use. The main thing is to check lists, forms, etc. using real data.

batumibiz:
> @batumibiz Yes, you can test your local changes using testRigor’s CLI. The tests themselves run on testRigor’s servers, but you can trigger them directly from your local environment before creating a PR. If you’d like, I can help you set it up.

If we use behavioral analysis, isn’t it easier to use [Behat](https://github.com/Behat/Behat), for example, it works completely locally, does not depend on external services, and the scenarios for it are similar. Writing them using AI is not a problem today.

AntonioVCruz:
@mhcwebdesign Yeah, it is mainly for developers. That's actually part of why I put it in the repo rather than making it an extension. Extensions are installed into an OpenCart store, whereas this is repository-level development tooling.

The test suite is also integrated with CI, so it can run on PRs and validate changes before they're merged. An extension wouldn't really serve that same purpose.

There's also nothing to install into OpenCart itself: no PHP or files under upload/. It's just the test scenarios and the GitHub Actions workflow, so I see it as being closer to things like phpstan.neon and Lint-4.yml that already live in the repo.

That said, I'm happy to adjust how it's structured, where it lives, or how the workflow and docs are set up. What do you think?

AntonioVCruz:
@batumibiz Fair point. Behat can run locally without relying on an external service. 

I also agree that writing the scenarios themselves isn't really the differentiator anymore. AI can generate Gherkin without much difficulty. The trade-off I see is more in the implementation and maintenance behind those scenarios. With Behat, the Gherkin steps still need PHP step definitions, and browser interaction generally needs something like Mink/WebDriver, so that implementation and the selectors behind it become part of the repo and have to be maintained as the UI changes. There’s also a bit of CI upkeep involved. You need the browser/WebDriver stack in the job, and keeping those pieces working together can be another source of CI flakes.

As a side note, I also agree with the point you made in your other comment: clicking through pages on its own isn't worth much. That's more about how the tests are written than which tool runs them, and it's the standard I tried to hold here. TC-009 creates a product in admin, verifies it on the storefront, changes the price, verifies the new price, then deletes it and confirms it's gone. TC-010 does something similar with a dynamically generated coupon code that's redeemed in the cart.

So I think the trade-off is really local and project-controlled with Behat vs. less browser-automation code to maintain in the repo with testRigor. Both approaches can cover the same behavior, and I don't think they're mutually exclusive.

batumibiz:
In general, I am glad that there was a specialist who was involved in BDD testing. If used correctly (read - competently writing scripts), it will help to avoid many mistakes when controlling what we develop.

Important nuance!
Be sure to describe how to run the script locally!
Otherwise, everything will be useless; untested code will be uploaded to the repository.

But although I'm getting ahead of myself :)
I personally would love to use something like this. But the question is: will the project’s maintainers use this?

AntonioVCruz:
Thanks, I really appreciate you taking the time to engage with this. Your comments have given me a lot to think about.

On that nuance, I see your point. The PR description does point to tests/e2e/README.md, but only as "setup and conventions". It never actually says the suite can be run locally, and that line is buried at the end of a paragraph about packaging. So in practice it was easy to miss, and I agree it should be called out explicitly. I'll fix that.

As for whether the maintainers would actually use it, I think that's the real question. The 10 cases cover what seemed like the most critical flows, but my main focus here was the CI pipeline and the overall structure. The coverage is the part that's meant to grow.

My thinking was that if the community feels this is a good fit, I'd take it to the forums and get input from more experienced members on which flows would be most useful to cover, then expand from there.

batumibiz:
By the way, I have a practical question, since you are working with testRigor, you should know the details...

What are the limitations on the free version?
How many tests can you run for free, per day, per month?

AntonioVCruz:
There’s no limit on how many tests you can run with a free account. The main trade-off is that the account is public, so other people can see the test suites, and some features are limited, such as emulators and CI/CD integrations other than the testRigor CLI.

To be more specific, the free plan runs on Linux/Chrome. Mobile emulators, real devices, Windows/macOS, and native app testing are available on the paid tiers.

batumibiz:
I tried to run tests locally.
Everything worked, but with some errors. complains about the absence of a variable `adminUrl`
Because of this, he cannot work with the admin panel.

As I saw, your `adminUrl` is set only in the CI configuration.
But this is not visible in local tests.

AntonioVCruz:
Thanks for testing it. That's a real gap, and your description pinpoints it exactly.

`adminUrl` is only used by admin-login and TC-008. Most other tests navigate from the base URL set by ` --url`, and only TC-009 / TC-010 read `storeUrl`. So when `adminUrl` is undefined, those steps go nowhere and the admin tests end up looking for the Login button on the storefront, while the rest can still pass normally.

The issue is on my side, and there are actually two parts to it. The local run command doesn't include `--variables-path`, so it silently depends on the suite's Test Data. On top of that, the setup table only showed the CI values, so there was no clear instruction for what to enter for `storeUrl `/ `adminUrl` when setting up the suite manually. I've updated the table on the PR body to list the local values explicitly. CI always passes those values, which is why the problem never showed up there.

Could you share the suite URL? I'd like to check how the Test Data is configured there to confirm that this is what’s happening.

To unblock yourself in the meantime, you can either set `adminUrl` to http://localhost/admin in the suite's Test Data, or save the JSON variables block as a file and pass `--variables-path /path/to/that.json` when running locally. All four keys need to be included in the file, since it replaces the variable set rather than merging with it.

batumibiz:
Test suite ID: jegzG8et4ZbwQoYac
But I haven’t yet figured out how to set up public access so that everyone can see it.

batumibiz:
By the way, your idea with variables.json might be useful.
You need to write a Makefile script where you immediately record all the necessary initialization steps.

Goal: not to suffer every time with command line parameters, but to enter a simple `make test` command

AntonioVCruz:
If you don't see the notification that the suite is public, your account is probably on a free trial rather than the public free plan.

To add the Test Data to the suite, just click Test Data in the left navigation menu, then click the red Add button.

I've also made my email public and added my X profile to my GitHub account. I don't really post there, but I do reply to DMs, so feel free to reach out. I'd be happy to help with the setup and give you some pointers on how to use the tool.

AntonioVCruz:
> By the way, your idea with variables.json might be useful. You need to write a Makefile script where you immediately record all the necessary initialization steps.
> 
> Goal: not to suffer every time with command line parameters, but to enter a simple `make test` command

Now that you mention it, that would have been really useful. I’ll hold off for the moment so I don’t make changes while the PR is open, but I’ll definitely do that later.

batumibiz:
I added your PR to a separate branch of the OC+ repository https://github.com/oc-plus/oc-plus/tree/testrigor
For now, I’ll try to configure the configuration for that project.
I will test it, your idea is very interesting.

PR #15672, opened 2026-08-31
[Master] Fix ISO Code 2 country lookup

zhouyang4484:
Fixes the country lookup in getCountryByIsoCode2(). The method was querying iso_code_3 instead of iso_code_2.


mhcwebdesign:
I noticed the same issue is true for the 4.x.x.x branch. Therefore, can you also create a pull-request for the 4.x.x.x branch please?

PR #15674, opened 2026-09-01
[Master] Remove the unreachable duplicate payment_methods case in the subscription API

darkdi:
`case 'payment_methods':` appears twice in the same switch with the same body. PHP takes the first arm, so the second one has never run.


darkdi:
Closing this, catalog/controller/api/subscription.php was removed on master in 470802d1 so the duplicate case is gone with it

PR #15675, opened 2026-09-01
[3.0.x.x] Remove duplicate `preg_match` executions (93-95% of all executions)

nekto-kotik:
Remove duplicate `preg_match` executions (93-95% of all executions).
There is a suprising stable effect on TTFB: 9% decrease (on average, tested on multiple installations). The effect is even visible on a fresh installation with demo data.

### DIFF
--- a/upload/system/engine/event.php
+++ b/upload/system/engine/event.php
@@ -17,6 +17,8 @@
 class Event {
 	protected $registry;
 	protected $data = array();
+	protected $processed = array();
+	protected $refresh = false;
 	
 	/**
 	 * Constructor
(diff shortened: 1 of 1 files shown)


nekto-kotik:
Hi Daniel and All,
I assume it went mostly unnoticed or not deserving attention that `event->trigger` is executed tens and sometimes hundreds of thousands of times on every page load.

Over 90% of these runs are duplicates, they compare the same pairs over and over again.
This fix eliminates duplicate `preg_match` runs (comparing the same pairs) and brings the quantity of `preg_match` executions down by over 90%: e.g. 51984 become 1248, 26462 become 1164, 18808 become 1392 (real numbers from my installations).

Even on a fresh OpenCart installation, as fast as it already is (62-78 ms TTFB on v3, depending on the page; 74-90 ms TTFB on v4), this simple change improves TTFB to 54-70 (on v3; and to 68-83 on v4) - which is not significant in milliseconds, but quite significant percentage-wise, considering how small the change is.
This average and stable ~9% TTFB improvement keeps up on real live websites, even when their TTFB is 200, 400, 800, or 1000+ ms.

`preg_match` is executed 10614-14262 times on a fresh v3 installation every time a page is loaded (13351-18025 on a fresh v4), my fix reduces it to 690-710 (1727-2013 on v4).

Real live websites usually get 30K-50K `preg_match` runs per page, reduced to 1K-2K by this fix.

mhcwebdesign:
@nekto-kotik : Interesting, thanks for pointing it out, I'll run some tests first.

nekto-kotik:
Oops, I need to make a small correction.

I wrote: "...`event->trigger` is executed tens and sometimes hundreds of thousands of times on every page load".

But I meant: "...`event->trigger` is executing `preg_match` tens and sometimes hundreds of thousands of times on every page load".

`preg_match` is executed too many times, not the `event->trigger` itself.

PR #15681, opened 2026-09-07
Fix order row delete action

huntbee-opencart:
## Description

Fixes the delete action in the admin order list.

The current implementation stores the order ID in the `href` attribute and reads it using `attr('href')`.

This change:

- stores the order ID in `data-order-id`;
- uses a dedicated `.button-delete-order` selector;
- reads the order ID from the data attribute;
- keeps the existing delete API and confirmation behaviour unchanged.

## Testing

Tested on OpenCart 3.0.5.1.

1. Open Sales > Orders.
2. Click Delete from an individual order row.
3. Confirm deletion.
4. Verify the order is deleted and the order list reloads successfully.


mhcwebdesign:
The admin order delete in the current OC 3.0.5.1 works just fine. What exact software bug or issue are you trying to solve?

huntbee-opencart:
Thanks for pointing this out. I rechecked the issue and confirmed that the delete functionality in OpenCart 3.0.5.1 is working correctly.

The problem was caused by a missing form element in the client's modified admin template, so this is not a core OpenCart bug. I'll close this PR.

PR #15682, opened 2026-09-07
[4.x.x.x] Add Brazilian Portuguese (pt-br) language

pedrohenriquers:
Adds a complete Brazilian Portuguese (`pt-br`) language pack for the storefront, the admin and the bundled extensions.

`pt-br` is not currently available: the repository ships `en-gb`, `es-es` and `fr-fr`, and Brazilian Portuguese is not among the languages translated in the Crowdin project. Brazil is one of the larger OpenCart markets, so the gap is worth closing.

## What is included

303 files across the four directories mapped by `crowdin.yml`:

| Directory | Files |
| --- | --- |
| `upload/admin/language/pt-br` | 121 |
| `upload/catalog/language/pt-br` | 87 |
| `upload/extension/opencart/admin/language/pt-br` | 62 |
| `upload/extension/opencart/catalog/language/pt-br` | 33 |

Coverage is **4,964 of 5,084 strings (97%)**. The remaining 120 are intentionally left as-is: acronyms (EAN, SKU, SMTP, ISBN, MPN), brand names, locale identifiers, HTML fragments, and words that are identical in both languages (Design, Marketing, Banner, Link).

## Where the text comes from

I think this matters for reviewing, so it is stated openly:

- **2,904 strings** were carried over from the OpenCart Brasil distribution (GPL-3.0-or-later, license-compatible), where the file path matches between versions. This is text that has been in production use in Brazilian stores.
- **1,158 strings** came from the same source by exact match of the English text, regardless of file or key, using a translation memory of 5,711 pairs. Pairs with conflicting translations were discarded rather than picking one arbitrarily.
- **1,090 strings** were translated for this contribution. Almost all belong to features that did not exist in the 3.x line: subscriptions, GDPR, blog/articles and anti-spam.

The last group has not been reviewed by a second native speaker. If a maintainer or a Brazilian contributor wants to sample anything, the checkout, shipping and payment wording is where I would look first.

## Notes on correctness

- **Format specifiers were audited key by key against `en-gb`.** Twenty-six strings inherited from the 3.x line had a different count than 4.x expects and were rewritten. One example: the footer `text_version` carried two `%s` because that distribution appended its own version number, which made the admin fail with `Error: 3 arguments are required, 2 given`. The tree now has zero mismatches.
- **`default.php` uses Brazilian conventions:** comma as decimal separator, dot for thousands, `d/m/Y` dates and 24-hour time.
- `php -l` passes on all 301 PHP files.

## Activating it

The files alone do not enable the language. It has to be registered under **System → Localisation → Languages** with code `pt-br` and locale `pt-BR,pt_BR.UTF-8,pt_BR,portuguese-brazil`, then selected in the store settings for the catalog and the admin.

Tested on a clean 4.1.0.4 install running the project's own Docker setup (Apache, PHP 8.4, MariaDB), with the storefront and admin verified in a browser.


mhcwebdesign:
@pedrohenriquers : Thank you for this. At the moment, we only have en-gb and fr-fr. If we were to add pt-br to the 4.x.x.x branch then we would need the _install/language/pt-br_ in there, too. 

I was just wondering when to add languages to the 4.x.x.x branch, as apposed to just creating a separate language pack on the OpenCart market place.

